PR #1, opened 2022-06-10
AJ-440

sagehen03:
### Running the app

1. Bring your own postgres database.  This code assumes a postgres db process is accessible on local port 5432 with a database called `wds` and a user with write access to that db called `wds`.  You can override `WDS_DB_*` environment variables if the defaults don't work for you.  See [application.properties](https://github.com/DataBiosphere/workspace-data-service/blob/3f99f2432b288a646b518b81257c42a0fe539365/src/main/resources/application.properties) for details.
2. Once the database is ready use `./gradlew bootRun` from the root of this checkout and that will run liquibase to create the schema and launch the service on port 8080.
3. Create entities `curl -v -H "Content-type: application/json" -X POST  "http://localhost:8080/api/workspaces/default/test/entities/batchUpsert" -d '[{"name": "first_entity", "entityType": "first_entity_type", "operations": [{"op": "AddUpdateAttribute", "attributeName": "col1", "addUpdateAttribute": "Spam"}]}]}]'`
4. Read entities `curl http://localhost:8080/api/workspaces/default/test/entityQuery/first_entity_type`
5. Remove entities `curl -H "Content-type: application/json" "http://localhost:8080/api/workspaces/default/test/entities/delete" -d '[{"entityName": "first_entity", "entityType": "first_entity_type"}]'`

### Database Schema
image

This schema uses a jsonb column to store the user's attribute data.  If we're going away from multitenancy, this approach doesn't make much sense.  Please chime in on db schema thoughts.

### Tests
There are no tests :cry: .  I am thinking we should follow the path of rawls and wsm and have tests that execute and validate sql statements against a real database.  The main downside is more setup required before running tests and longer test runs.  The upside would be more protection from schema changes causing regressions and a faster dev cycle for validating db code.  Please chime in with regard to tests.

### Current functionality
We can crud rows (entities) and tables (entity types) so long as they belong to the workspace default:test.  There's not currently an API to create a new workspace.  

This code includes support for references since it was largely copied from a prototype whose emphasis was replacing our existing multi-tenant solution with something more performant while still offering the same functionality.  References account for a good chunk of the code and do complicate things significantly and likely go beyond the spirit of the .1 version functionality.  However, it seems silly to remove that functionality if we're only going to add it back later.


src/main/resources/db/changesets/initial_schema.xml
```diff
@@ -0,0 +1,136 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:pro="http://www.liquibase.org/xml/ns/pro" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/pro http://www.liquibase.org/xml/ns/pro/liquibase-pro-4.6.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.6.xsd">
```
calypsomatic, line 2: Small comment!  Do we have a preference for xml vs yaml vs whatever?  I've seen yaml more often, I think, in other repos, and I find it a little easier to read.

sagehen03: I did notice the WSM uses yaml, but Rawls uses xml.  I would tend to agree that yaml is easier to read.  My preference would honestly be for just plain sql.

src/main/resources/db/changesets/add_workspace.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:pro="http://www.liquibase.org/xml/ns/pro" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/pro http://www.liquibase.org/xml/ns/pro/liquibase-pro-4.6.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.6.xsd">
+    <changeSet author="dhite" id="add_ws">
+        <sql splitStatements="false">
+            insert into workspace (id, name, namespace) values ('15f36863-30a5-4cab-91f7-52be439f1175', 'test', 'default')
+        </sql>
+    </changeSet>
```
davidangb, line 7: what is the purpose of inserting a workspace here? Is it simply to have a target for the foreign key on entity_type? See my other comment … I'm wondering if we can do away with the workspace table entirely.

src/main/resources/db/changesets/initial_schema.xml
```diff
@@ -0,0 +1,136 @@
...
+
+
+CREATE OR REPLACE FUNCTION entity_insert() RETURNS trigger AS $entity_insert$
+            BEGIN
+            with q as (select string_agg(value, ' ') as attrs, array_agg(key) as attr_names from jsonb_each_text(NEW.attributes))
+            insert into entity_attribute_info (entity_type, entity_name, entity_attribute_values, entity_attribute_names)
+            select NEW.entity_type, NEW.name, q.attrs, q.attr_names from q;
+            RETURN NEW;
+            END;
+            $entity_insert$ LANGUAGE plpgsql;
```
davidangb, line 128: nit: formatting/indentation. It doesn't matter much but since liquibase changes will fail if the content changes - including indentation - we'll have no chance to change it later!

src/main/java/org/databiosphere/workspacedataservice/shared/model/EntityType.java
```diff
@@ -0,0 +1,26 @@
...
+        this.name = name;
+    }
+
+    public List<Entity> getEntities() {
+        return entities;
+    }
+
+    public void setEntities(List<Entity> entities) {
+        this.entities = entities;
+    }
```
davidangb, line 25: why does an `EntityType` model need a list of entities? This may be answered elsewhere in the code, I will continue to read through.

src/main/java/org/databiosphere/workspacedataservice/shared/model/Entity.java
```diff
@@ -0,0 +1,104 @@
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof Entity)) return false;
+
+        Entity entity = (Entity) o;
+
+        if (!getName().equals(entity.getName())) return false;
+        return getEntityTypeId().equals(entity.getEntityTypeId());
+    }
```
davidangb, line 96: This `equals` could bite us if we ever need to compare attributes - picture a use case of an entity uploaded by a user, and we want to compare that uploaded entity to the entity in the db to see if the db requires a write. I'm fine leaving it as-is for this PR, just thinking out loud about future functionality.

src/main/java/org/databiosphere/workspacedataservice/shared/model/Entity.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.Map;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class Entity {
+
+    private String name;
+
+    private Boolean deleted;
+
+    private String entityType;
```
davidangb, line 15: My preference is for strong typing - `private EntityType entityType` instead of `private String entityType`. I think it avoids confusing signatures like `Entity(String, String …)`. I'm fine making that change as a follow-on PR, but if the team agrees, we should do it soon before we accumulate too much code.

src/main/resources/db/changesets/initial_schema.xml
```diff
@@ -0,0 +1,136 @@
...
+                <constraints nullable="false" primaryKey="true" primaryKeyName="entity_reference_pkey"/>
+            </column>
+            <column name="referenced_entity_type" type="BIGINT">
+                <constraints nullable="false" primaryKey="true" primaryKeyName="entity_reference_pkey"/>
+            </column>
+            <column name="referenced_entity_name" type="VARCHAR">
+                <constraints nullable="false" primaryKey="true" primaryKeyName="entity_reference_pkey"/>
+            </column>
+        </createTable>
+    </changeSet>
```
davidangb, line 87: I don't think we should handle entity references as hard foreign keys - eliminate this table entirely, and remove a lot of complexity around references. I think references should be simple URIs stored as strings, which can be parsed/interpreted when needed. I am working on a design for that - I'll get that out ASAP.

davidangb (commented):
I have not reviewed all the code - I skipped over a lot of the Java code - but wanted to get these first review thoughts out ASAP to start discussion. Here you go!

src/main/resources/db/changesets/initial_schema.xml
```diff
@@ -0,0 +1,136 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="name" type="VARCHAR">
+                <constraints nullable="false" primaryKey="true" primaryKeyName="workspace_pkey"/>
+            </column>
+            <column name="namespace" type="VARCHAR">
+                <constraints nullable="false" primaryKey="true" primaryKeyName="workspace_pkey"/>
+            </column>
+        </createTable>
+    </changeSet>
```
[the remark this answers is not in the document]
sagehen03: These apis were originally built to match the existing entity apis and most of those apis use workspace name and namespace in the url so I stored that (name, namespace) -> workspace id mapping in the workspace table.

src/main/resources/db/changesets/initial_schema.xml
sagehen03: I agree that there's a lot of code related to references, but I was aiming for CDC scale when I chose this implementation.  I worry about how performant parsing uri strings will be when upserting more than few thousand entities.

src/main/resources/db/changesets/add_workspace.xml
sagehen03: https://github.com/DataBiosphere/workspace-data-service/pull/1#discussion_r896181582 explains why there's a workspace table and this insertion was a temporary measure to add a workspace to the db so that the curl invocations I mention in the PR description would work.

src/main/java/org/databiosphere/workspacedataservice/shared/model/Entity.java
sagehen03: I think this is just intellij generated equals() with entity type id and entity name specified as the determinant fields.

src/main/java/org/databiosphere/workspacedataservice/shared/model/Entity.java
sagehen03: This was me being lazy.  The existing json format outputs {"entityType": "<type>"} and this was the fastest way to get there without any jackson annotations.  It may be that we need to stop making `Entity` do duty both as instructions for json serialization and as a data class within the application.

build.gradle
```diff
@@ -0,0 +1,32 @@
...
+    implementation 'org.springframework.boot:spring-boot-starter-web'
+    implementation 'org.liquibase:liquibase-core'
+    implementation 'org.apache.commons:commons-lang3'
+    implementation 'com.google.guava:guava:31.1-jre'
+    implementation 'org.postgresql:postgresql'
+    developmentOnly 'org.springframework.boot:spring-boot-devtools'
+    testImplementation 'org.springframework.boot:spring-boot-starter-test'
+}
+
+//apply plugin: "org.liquibase.gradle"
```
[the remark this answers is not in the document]
sagehen03: It was inadvertent, removed.

src/main/java/org/databiosphere/workspacedataservice/shared/model/EntityType.java
sagehen03: This is an artifact of loading a bunch of entities from a json file for the original prototype, removed.

src/main/java/org/databiosphere/workspacedataservice/shared/model/Entity.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private Map<String, Object> attributes;
```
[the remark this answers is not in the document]
sagehen03: +1 to a follow on PR and incorporating more constraints to what sort of data is allowed here.

src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
```diff
@@ -0,0 +1,303 @@
...
+    public UUID getWorkspaceId(String namespace, String name) {
+        return template.queryForObject("select id from workspace where name = ? and namespace = ?", UUID.class, name, namespace);
+    }
+
+
+    public void batchUpsert(List<Entity> entities) {
+        StopWatch watch = new StopWatch();
+        watch.start();
+        template.batchUpdate("insert into entity (name, entity_type, record_version, deleted, attributes)" +
+                "values (?, ?, ?, ?, ? :: jsonb) on conflict (name, entity_type) do update set attributes = excluded.attributes", new BatchPreparedStatementSetter() {
```
davidangb, line 99: update the `record_version` on update too (or, remove the `record_version` column and we'll add it back in later)

src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
```diff
@@ -0,0 +1,303 @@
...
+        }
+    }
+
+    public List<Entity> getNamedEntities(Long entityTypeId, Set<String> entityNames, String entityTypeName, boolean excludeDeletedEntities) {
+        String sql = "select name, attributes, entity_type, deleted from entity where entity_type = :entityType and name in (:entityNames) " + (excludeDeletedEntities ? "and deleted = false" : "");
+        return namedParameterJdbcTemplate.query(sql,
+                new MapSqlParameterSource(Map.of("entityType", entityTypeId, "entityNames", entityNames)),
+                (rs, i) -> new Entity(rs.getString("name"), new EntityType(entityTypeName), getAttributes(rs.getString("attributes")),
+                        rs.getLong("entity_type"), rs.getBoolean("deleted")));
+    }
```
davidangb, line 165: suggest adding a `boolean returnAttributes` argument to select - or not - the `attributes` column. Or maybe that needs a separate method entirely

src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
```diff
@@ -0,0 +1,303 @@
...
+    public long loadEntityType(EntityType entityType, UUID workspaceId) {
+        GeneratedKeyHolder generatedKeyHolder = new GeneratedKeyHolder();
+        template.update(con -> {
+            PreparedStatement ps = con.prepareStatement("insert into entity_type (name, workspace_id) values (?, ?)", Statement.RETURN_GENERATED_KEYS);
+            ps.setString(1, entityType.getName());
+            ps.setObject(2, workspaceId);
+            return ps;
+        }, generatedKeyHolder);
+        return (long)generatedKeyHolder.getKeys().get("id");
+    }
```
davidangb, line 191: nit: rename to `createEntityType` or similar? It wasn't clear to me what this method did just from its name; it sounds read-only.

src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
```diff
@@ -0,0 +1,303 @@
...
+
+    public Long getEntityTypeId(UUID workspaceId, String entityTypeName){
+        try {
+            LOGGER.info("Querying for {} and {}", workspaceId, entityTypeName);
+            return template.queryForObject("select id from entity_type where workspace_id = ? and name = ?",
+                    Long.class, workspaceId, entityTypeName);
+        } catch (EmptyResultDataAccessException e) {
+            return -1L;
+        }
+    }
```
davidangb, line 201: sometime - not necessarily this PR - we should have a "get or create entity type" method, I think

src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
```diff
@@ -0,0 +1,303 @@
...
+    public List<Entity> getSelectedEntities(long entityTypeId, int limit, int offset, String filter, String orderCol,
+                                            String sortOrder, List<String> fields){
+        StopWatch watch = new StopWatch();
+        watch.start();
+        StringBuilder builder = new StringBuilder("select name, attributes from entity e ");
+        MapSqlParameterSource params = new MapSqlParameterSource("entityType", entityTypeId);
+        params.addValues(Map.of("limit", limit, "offset", offset));
+        if(StringUtils.isNotBlank(filter)){
+            builder.append("join entity_attribute_info ea on ea.entity_name = e.name " +
+                    "and ea.entity_type = e.entity_type left join ft_search ft " +
```
davidangb, line 226: what is `ft_search`? That's not in the initial liquibase schema.

also, if `filter` is empty, or of `isFullTextIndexed ` is false, the join to `ft_search` is unnecessary, isn't it?

src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
```diff
@@ -0,0 +1,303 @@
...
+            builder.append("where e.entity_type = :entityType and e.deleted = false");
+        }
+        String entityTypeName = getEntityTypeName(entityTypeId);
+        builder.append(getOrderByClause(orderCol, entityTypeId, sortOrder));
+        builder.append(" limit :limit offset :offset");
+        List<Entity> result = namedParameterJdbcTemplate.query(builder.toString(), params,
+                (rs, i) -> {
+                    Map<String, Object> attributes = getAttributes(rs.getString("attributes"));
+                    if (!CollectionUtils.isEmpty(fields)) {
+                        attributes.keySet().retainAll(fields);
```
davidangb, line 240: is there a way to specify `field` selection in SQL, and let Postgres handle it?

src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
```diff
@@ -0,0 +1,303 @@
...
+                });
+        watch.stop();
+        LOGGER.info("Total time spent was {}s for {}, {}", watch.getTotalTimeSeconds(), builder, params);
+        return result;
+    }
+
+
+    public String getEntityTypeName(long entityTypeId){
+        return template.queryForObject("select name from entity_type where id = ?", String.class, entityTypeId);
+    }
```
davidangb, line 252: this is an extremely small query, but it happens so often. Not for this PR: we may want a read-through cache for this table. I expect that our data access patterns for this table will be heavily weighted to reads, with writes being heavily weighted to inserts; that sounds like a good candidate for caching.

src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
```diff
@@ -0,0 +1,303 @@
...
+
+    private Map<String, String> getEntityAttributeSortTypes(long entityTypeId) {
+        return template.query("select attr_name, sort_type from entity_type_attribute_sort where entity_type_id = ?", rs -> {
+            Map<String, String> result = new HashMap<>();
+            while(rs.next()){
+                result.put(rs.getString("attr_name"), rs.getString("sort_type"));
+            }
+            return result;
+        }, entityTypeId);
+    }
```
davidangb, line 262: is the `entity_type_attribute_sort` missing from liquibase too?

src/main/resources/db/changesets/initial_schema.xml
```diff
@@ -0,0 +1,136 @@
...
+            <column name="entity_name" type="VARCHAR">
+                <constraints nullable="false" primaryKey="true" primaryKeyName="entity_reference_pkey"/>
+            </column>
+            <column name="referenced_entity_type" type="BIGINT">
+                <constraints nullable="false" primaryKey="true" primaryKeyName="entity_reference_pkey"/>
+            </column>
+            <column name="referenced_entity_name" type="VARCHAR">
+                <constraints nullable="false" primaryKey="true" primaryKeyName="entity_reference_pkey"/>
+            </column>
+        </createTable>
```
davidangb, line 86: as commented elsewhere, this only captures references from entity A to entity B; it does not capture which _attribute_ inside entity A is doing the referencing. Given the compound primary key specified here, we'll get errors if entity A has two attributes, each of which hold a reference to entity B … that should be legal.

davidangb (commented):
a bunch more comments inline!

settings.gradle
```diff
@@ -0,0 +1 @@
+rootProject.name = 'workspace-data-service'
```
jordansafer, line 1: Naming comment :), based on the [naming doc](https://docs.google.com/document/d/1yvVlQnXDKOKSKEKIsMrWNh0LfwLuI7FWt57sa43d7Qg) & in the spirit of consistency, I propose `workspacedata` as the project name

sagehen03:
> For gradle, could you commit the .gradlew file?

Added, sorry for missing that.

src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
sagehen03: I agree https://github.com/broadinstitute/entity-service-prototypes/blob/30a793b3c8164feeb2f7dd01e05e6185e5195517/entity-postgres/src/main/java/org/broadinstitute/rdbms/dao/EntityDao.java#L191-L200 but I didn't migrate caching since it didn't seem essential for the work specified in this ticket.

src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
sagehen03: Sorry, this code references a solution I'd implemented for filtering large entity types in the prototype using full text indexing rather than `like`.  I should have looked at that code more critically when I was bringing it over, but I just did a few curls to make sure they worked, my apologies.

src/main/resources/db/changesets/initial_schema.xml
sagehen03: I don't think this is the case.  Entity A can reference multiple attributes in Entity B and the code won't try to add a new row in `entity_reference` [This line](https://github.com/DataBiosphere/workspace-data-service/blob/c41dbc9c669501d9bf8da520ae54722c2f92b65f/src/main/java/org/databiosphere/workspacedataservice/service/EntityReferenceService.java#L93) ensures that if Entity A is already referencing Entity B and we add new reference from A to B we don't try to insert a new record.

src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
```diff
@@ -0,0 +1,303 @@
...
+        this.objectMapper = objectMapper;
+    }
+
+    public List<EntityReference> getReferencesForEntities(List<Entity> entitiesForType) {
+        List<List<Entity>> chunks = Lists.partition(entitiesForType, 1_000);
+        List<EntityReference> result = new ArrayList<>();
+        for (List<Entity> chunk : chunks) {
+            result.addAll(namedParameterJdbcTemplate.query("select entity_type, entity_name, referenced_entity_type, " +
+                            "referenced_entity_name from entity_reference where entity_type = :entityType and entity_name in (:entityNames)",
+                    new MapSqlParameterSource(Map.of("entityType", chunk.get(0).getEntityTypeId(), "entityNames", chunk.stream().map(Entity::getName).collect(Collectors.toSet()))),
```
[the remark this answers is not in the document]
sagehen03: I agree that the method signature is a bit janky.  I think there's only one caller and I knew that caller would provide a list of entities belonging to the same type.  I've updated the code so that there's a separate entity id param.

src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
sagehen03: In my experience the jsonb functions for wide entities were too slow and that's probably what's going on here.

src/main/java/org/databiosphere/workspacedataservice/service/model/EntityReference.java
```diff
@@ -0,0 +1,64 @@
+package org.databiosphere.workspacedataservice.service.model;
+
+public class EntityReference {
+
+    private final String entityName;
+
+    private final long entityType;
```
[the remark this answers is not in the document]
sagehen03: See my other comment.  This table is not the only information the code uses when managing references; it also looks at the entities attributes where the reference relationship is captured as well.

src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
```diff
@@ -0,0 +1,102 @@
...
+    public ResponseEntity<String> deleteEntities(@PathVariable("workspaceNamespace") String wsNamespace,
+                                                 @PathVariable("workspaceName") String wsName, @RequestBody List<Map<String, String>> entitiesToDelete){
+        ArrayListMultimap<String, String> entitiesByType = ArrayListMultimap.create();
+        entitiesToDelete.forEach(e -> entitiesByType.put(e.get("entityType"), e.get("entityName")));
+        UUID workspaceId = getWorkspaceId(wsNamespace, wsName);
+        List<Entity> entitiesInDb = new ArrayList<>();
+        for(String entityTypeName: entitiesByType.keySet()){
+            Long entityTypeId = dao.getEntityTypeId(workspaceId, entityTypeName);
+            List<String> entities = entitiesByType.get(entityTypeName);
+            entitiesInDb.addAll(referenceService.getEntities(entityTypeId, entityTypeName, new HashSet<>(entities), true));
```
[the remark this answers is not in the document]
sagehen03: Fair enough. I've added a new method to just fetch the entity name and type and I'm passing that result to the delete method rather than re-using the method to fetch entities by name and type.  However, given that this delete code path is bound by the request size (you must specify all the entities that you want to delete in the request), I feel like it's a stretch to assert that this change will matter all that much in terms of performance.

davidangb (approved):
The only potential bug in here (that you haven't responded to) is the potentially-missing `entity_type_attribute_sort` table. Do queries that involve a sort cause errors due to this table not being present? I admit I have not tried it myself.

:+1: modulo that possible bug, with knowledge that we can refactor/edit all of this code at a later point.

calypsomatic (approved):
Approved conditional on making sure it merges with main - in particular the gradle wrapper seems necessary to make it run.

PR #3, opened 2022-06-15
Pulling in vault credentials directly via the vault-spring java library

jordansafer:
PR pulling in credential via vault-spring:
- Main vault configuration take place in `AppConfig.java`
- Additional vault properties. Mostly can be overridden easily with environment variables, except for the .vault-token path (due to annoying pathing variance which could be solved in a different way)
- Added an unnecessary call to Vault from the EntityDAO to demonstrate (see log line below)
- setup.sh walks through commands for setting up local postgres(url
Screen Shot 2022-06-15 at 5 10 40 PM
)


jordansafer:
Closing due to security complexity, since we use other tooling to populate config files in deployment-

PR #4, opened 2022-06-17
Aj 439 dev env

calypsomatic:
Add postgres script, gradle wrapper, and start to fill out the README.


README.md
```diff
@@ -1 +1,72 @@
...
+Design docs and other references may be found [here](https://broadworkbench.atlassian.net/wiki/spaces/AS/pages/2437742606/Workspace+Data+Service).
+
+## Setup
+### Prerequisites
+//TODO: Secrets
+
+Make sure you have Java 17 installed.  
+We find [jenv](https://www.jenv.be/) helpful for managing active versions.  On a Mac,jenv can be installed and used like so:
+```
+brew install jenv
```
jordansafer, line 14: Thoughts on putting sdkman vs jenv in the guide? I know we go through sdkman in the onboarding guide. Note that I'm biased because I used sdkman

README.md
```diff
@@ -1 +1,72 @@
...
+jenv add /Library/Java/JavaVirtualMachines/temurin-17.jdk/Contents/Home
+```
+
+We are using Postgres 13.1.  You do not need to have postgres installed on your system.  
+Instead use the `run_postgres.sh` script to set up a docker container running postgres (see below).
+
+## Running
+Before running WDS, you must populate secrets from Vault.  To do this run
+```
+./scripts/render-config.sh [env] [execution env]
```
jordansafer, line 33: I somewhat changed the names for these, I can update later based on what gets finalized-

local-dev/local-postgres-init.sql
```diff
@@ -0,0 +1,2 @@
+CREATE DATABASE wds;
+CREATE ROLE wds WITH LOGIN ENCRYPTED PASSWORD 'wds';
```
jordansafer, line 2: I think we also need to give the role write permissions to the database. I ran a GRANT ALL PERMISSIONS xxx command to do this, I can look up the exact command

jordansafer (approved):
Some minor comments, looks like a good improvement!

README.md
calypsomatic: Yeah I was just looking at someone else's repo when I added that.  Should we put both?

jordansafer: I think it's fine to just have one, we can just stick with jenv-

calypsomatic: Too late, I've changed it.

calypsomatic: Too late, I've changed it.

local-dev/local-postgres-init.sql
calypsomatic: Makes sense, although I don't see any such command elsewhere in the repo world.  It can also be added later.

calypsomatic: Makes sense, although I don't see any such command elsewhere in the repo world.  It can also be added later.

davidangb (approved):
proactive :+1:, with two issues to watch out for:
* merge the README.md with #2, you're already discussing this in comments
* see if the postgres user needs a `GRANT` in order to function, implement it if necessary

PR #5, opened 2022-06-21
Create build-and-test.yml

calypsomatic:
Basic workflow that runs sonarcloud.


jordansafer (approved):
Build is failing because there's no code yet, right? So ./gradlew has nothing to run :)

calypsomatic:
> Build is failing because there's no code yet, right? So ./gradlew has nothing to run :)

Yep, that's why I switched it to a draft, waiting on code appearing.

PR #7, opened 2022-06-21
Create build-and-test.yml

calypsomatic:
Adds a workflow to run sonarqube.


.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Build all projects without running tests
+        run: ./gradlew --build-cache build -x test
+
+  unit-tests-and-sonarqube:
+    needs: [ build ]
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
+        image: postgres:13
```
sagehen03, line 32: [postgres 14.4](https://www.postgresql.org/docs/14/release-14-4.html) is out.  Do we want to use that?

sagehen03 (approved):
Easy enough to change later so I won't block but 14 is the current major version of postgres.  Would it make sense to start further from obsolescence?

.github/workflows/build-and-test.yml
calypsomatic: Ah this would have been good to know for the dev env script as well.  Is there a way to have both this image and the postgres_init script point to some variable for the version so they change together?

sagehen03: We could use gh environment variables in the gh actions yml, but I'm not sure how to feed that variable value into our script.

calypsomatic: I think it's also usually possible to just put 'latest' - at least for docker containers like in the local env, but that's probably tempting fate once a breaking update comes out.

PR #8, opened 2022-06-28
API signature proposal [AJ-469]

davidangb:
Reviewers, please read in full!

This is a first-draft proposal for the "v2" API signature for WDS.

For the purposes of this description, I am calling our new preferred API the "v2" API. I call the current Rawls entity APIs the "v1" API. I believe we should focus on the v2 API for WDS and implement the v1 API as a backwards-compatibility gesture if/when we realize we need it.

These API signatures are a proposal, meant as a basis for discussion. Please offer your thoughts!

I also expect that these signatures will change as we enter implementation phases and as we define requirements. For instance, the provenance metadata is currently not well-defined. The search payloads are also not well-defined, since we have yet to nail down specifics of search features. The proposal here is NOT meant to be a big-design-up-front specification; it's more of a guide.

To visualize the APIs in this proposal, you may want to copy/paste the YAML into https://editor.swagger.io/.

What's in this PR:
* URL, request payload, and response payload definitions for what I see as the highest-priority entity APIs

What's NOT in this PR:
* signatures for the schema APIs. I envision the schema APIs will allow callers to do things like change an attribute from number to string, list the attributes for a given type, or list all types along with their attributes (this latter feature is equivalent to the "v1" entity type metadata API).
* signatures for the import and export APIs. These APIs will support PFB/TSV/TDR import as well as export to TSV, download to GCS bucket/Azure storage container, etc. Hopefully, these can replace Import Service.
* proposals for any non-success response codes. I expect we'll return 4xx and 5xx for user- and system-level errors, but I haven't attempted to list those out nor define the structure of an error response.

Note also that AJ-471 may change terminology, e.g. replace "entity" with "record". If/when that happens, these signatures will also change.

See also https://docs.google.com/spreadsheets/d/1896KTgTr10g_3cHUxxC1oF9meaW_qO9pll_AWaYiMjo/edit#gid=0

### DIFF
--- a/src/main/resources/swagger/openapi-docs.yaml
+++ b/src/main/resources/swagger/openapi-docs.yaml
@@ -10,8 +10,8 @@ info:
 paths:
   /{instanceid}/entities/{v}/{type}/{id}:
     get:
-      summary: get single entity
-      description: get single entity
+      summary: Get entity
+      description: Get entity
       parameters:
         - $ref: '#/components/parameters/instanceIdPathParam'
         - $ref: '#/components/parameters/versionPathParam'
@@ -25,8 +25,8 @@ paths:
               schema:
                 $ref: '#/components/schemas/Entity'
     put:
-      summary: create single entity
-      description: create single entity
+      summary: Create entity
+      description: Create entity
       parameters:
         - $ref: '#/components/parameters/instanceIdPathParam'
         - $ref: '#/components/parameters/versionPathParam'
@@ -40,8 +40,8 @@ paths:
               schema:
                 $ref: '#/components/schemas/Entity'
     patch:
-      summary: update single entity
-      description: update single entity
+      summary: Update entity
+      description: Update entity
       parameters:
         - $ref: '#/components/parameters/instanceIdPathParam'
         - $ref: '#/components/parameters/versionPathParam'
@@ -55,8 +55,8 @@ paths:
               schema:
                 $ref: '#/components/schemas/Entity'
     delete:
-      summary: delete single entity
-      description: delete single entity
+      summary: Delete entity
+      description: Delete entity
       parameters:
         - $ref: '#/components/parameters/instanceIdPathParam'
         - $ref: '#/components/parameters/versionPathParam'
@@ -67,10 +67,62 @@ paths:
           description: Success
   /{instanceid}/types/{v}/{type}:
     patch:
+      summary: Update entity type
+      description: |
+        Update entity type. Alpha1 supports renaming a type; future versions will support modifying schema details
+        such as changing the attribute schema for this type.
+      parameters:
+        - $ref: '#/components/parameters/instanceIdPathParam'
+        - $ref: '#/components/parameters/versionPathParam'
+        - $ref: '#/components/parameters/entityTypePathParam'
+      responses:
+        200:
+          description: Success
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/EntityTypeDefinition'
     delete:
+      summary: Delete entity type
+      description: Delete entity type. This deletes all entities of this type.
+      parameters:
+        - $ref: '#/components/parameters/instanceIdPathParam'
+        - $ref: '#/components/parameters/versionPathParam'
+        - $ref: '#/components/parameters/entityTypePathParam'
+      responses:
+        204:
+          description: Success
   /{instanceid}/types/{type}/attributes/{v}/{attribute}:
     patch:
+      summary: Update attribute
+      description: |
+        Update attribute. Alpha1 supports renaming an attribute within the specified type;
+        future versions will support modifying schema details
+        such as changing the datatype of this attribute.
+      parameters:
+        - $ref: '#/components/parameters/instanceIdPathParam'
+        - $ref: '#/components/parameters/versionPathParam'
+        - $ref: '#/components/parameters/entityTypePathParam'
+        - $ref: '#/components/parameters/attributePathParam'
+      responses:
+        200:
+          description: Success
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/EntityTypeDefinition'
     delete:
+      summary: Delete attribute
+      description: |
+        Delete attribute. This deletes all values for this attribute within the specified type.
+      parameters:
+        - $ref: '#/components/parameters/instanceIdPathParam'
+        - $ref: '#/components/parameters/versionPathParam'
+        - $ref: '#/components/parameters/entityTypePathParam'
+        - $ref: '#/components/parameters/attributePathParam'
+      responses:
+        204:
+          description: Success
   /{instanceid}/search/{v}:
     get:
     post:
@@ -78,6 +130,13 @@ paths:
     post:
 components:
   parameters:
+    attributePathParam:
+      name: attribute
+      in: path
+      description: Attribute name
+      required: true
+      schema:
+        type: string
     entityIdPathParam:
       name: id
       in: path
@@ -123,10 +182,29 @@ components:
           type: string
           description: Entity type
         attributes:
-          type: object
+          $ref: '#/components/schemas/EntityAttributes'
           description: KVPs of entity attributes
         metadata:
-          type: object
+          $ref: '#/components/schemas/EntityMetadata'
           description: Provenance and other entity metadata. May not be present in alpha1.
-
+    EntityAttributes:
+      type: object
+    EntityMetadata:
+      type: object
+    EntityAttributeDefinition:
+      required:
+        - name
+      type: object
+      properties:
+        name:
+          type: string
+          description: Attribute name
+    EntityTypeDefinition:
+      required:
+        - name
+      type: object
+      properties:
+        name:
+          type: string
+          description: Entity type name
 
--- a/src/main/resources/swagger/openapi-docs.yaml
+++ b/src/main/resources/swagger/openapi-docs.yaml
@@ -7,11 +7,18 @@ info:
   license:
     name: BSD
     url: http://opensource.org/licenses/BSD-3-Clause
+tags:
+  - name: Entities
+    description: Entity APIs
+  - name: Schema
+    description: Schema Manipulation APIs (coming soon)
 paths:
   /{instanceid}/entities/{v}/{type}/{id}:
     get:
       summary: Get entity
       description: Retrieves a single entity by its type and id
+      tags:
+        - Entities
       parameters:
         - $ref: '#/components/parameters/instanceIdPathParam'
         - $ref: '#/components/parameters/versionPathParam'
@@ -27,6 +34,8 @@ paths:
         If the entity already exists, its entire set of attributes will be overwritten by
         the attributes in the request body.
         TODO: the type and id are duplicated in the url and the body; change this? Silently ignore the type/id in the body?
+      tags:
+        - Entities
       parameters:
         - $ref: '#/components/parameters/instanceIdPathParam'
         - $ref: '#/components/parameters/versionPathParam'
@@ -44,6 +53,8 @@ paths:
         Any attributes included in the request body will be created or overwritten.
         No attributes will be deleted. To delete attributes, use the PUT api instead.
         TODO: the type and id are duplicated in the url and the body; change this? Silently ignore the type/id in the body?
+      tags:
+        - Entities
       parameters:
         - $ref: '#/components/parameters/instanceIdPathParam'
         - $ref: '#/components/parameters/versionPathParam'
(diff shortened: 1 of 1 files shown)


src/main/resources/swagger/openapi-docs.yaml
```diff
@@ -0,0 +1,482 @@
...
+        - $ref: '#/components/parameters/entityIdPathParam'
+      responses:
+        200:
+          $ref: '#/components/responses/EntityResponseBody'
+    put:
+      summary: Create or replace entity
+      description: |
+        Creates or replaces the entity using the specified type and id.
+        If the entity already exists, its entire set of attributes will be overwritten by
+        the attributes in the request body.
```
davidangb, line 35: we could also add a query param to enable/disable overwriting, i.e. throw an error if the entity already exists.

src/main/resources/swagger/openapi-docs.yaml
```diff
@@ -0,0 +1,482 @@
...
+        $ref: '#/components/requestBodies/EntityRequestBody'
+      responses:
+        201:
+          $ref: '#/components/responses/EntityResponseBody'
+    patch:
+      summary: Update entity
+      description: |
+        Updates the entity of the specified type and id.
+        Any attributes included in the request body will be created or overwritten.
+        No attributes will be deleted. To delete attributes, use the PUT api instead.
```
davidangb, line 54: so, this `PATCH` api will upsert attribute values, but has no way to delete an attribute value from an entity. Users would have to retrieve the entity, modify it locally to delete the desired attributes, and then `PUT` that modified entity in whole to perform deletes.

src/main/resources/swagger/openapi-docs.yaml
```diff
@@ -0,0 +1,482 @@
...
+        - $ref: '#/components/parameters/entityIdPathParam'
+      responses:
+        204:
+          description: Success
+  /{instanceid}/types/{v}/{type}:
+    patch:
+      summary: Update entity type
+      description: |
+        Update entity type. Returns the updated type definition.
+        Alpha1 supports renaming a type; future versions may support additional functionality.
```
davidangb, line 86: The ` /{instanceid}/types/{v}/{type}` and `/{instanceid}/types/{type}/attributes/{v}/{attribute}` APIs may not even need to exist; we might be able to roll them into the schema APIs, which are not defined in this PR.

src/main/resources/swagger/openapi-docs.yaml
```diff
@@ -0,0 +1,482 @@
...
+      responses:
+        204:
+          description: Success
+  /{instanceid}/search/{v}:
+    get:
+      summary: Query entities
+      description: |
+        Paginated list of entities matching the criteria supplied in url query parameters.
+        This API supports a basic subset of search and filtering operations. For full functionality,
+        use the POST api instead.
```
davidangb, line 152: As of this PR - the alpha1 api definitions - the `GET` and `POST` search APIs support the exact same functionality. My thinking here is that as we build more and more complicated search features, such as the ability to filter on individual attributes, we can add those features to the `POST` request body, but not necessarily add them to the `GET` query parameters. This prevents needing to represent complex structures in query params, but still allows for basic listing/pagination/filtering via a simple `GET`.

src/main/resources/swagger/openapi-docs.yaml
```diff
@@ -0,0 +1,482 @@
...
+          $ref: '#/components/schemas/EntityType'
+          description: Entity type
+        attributes:
+          $ref: '#/components/schemas/EntityAttributes'
+          description: KVPs of entity attributes
+        metadata:
+          $ref: '#/components/schemas/EntityMetadata'
+          description: Provenance and other entity metadata. May not be present in alpha1.
+    EntityAttributes:
+      type: object
```
davidangb, line 409: Not written out here: I think we should represent references as URIs, not as the bespoke format we use in Rawls. See https://docs.google.com/document/d/1kq0w8uTEiJLKRQ70fhQhZvwtuBAZ5cjGiCn0NRu7QLI/edit#heading=h.8xxqywtngs46 for info on that.

I think we should also probably represent `AttributeValueList`s as plain json arrays instead of our own bespoke format, but haven't thought about that as much.

Other than those changes, I expect we'd represent attributes exactly as we do today - mapping quite nicely to json.

src/main/resources/swagger/openapi-docs.yaml
```diff
@@ -0,0 +1,463 @@
...
+        - $ref: '#/components/parameters/versionPathParam'
+      requestBody:
+        $ref: '#/components/requestBodies/BatchRequestBody'
+      responses:
+        200:
+          description: Success
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/BatchResponse'
```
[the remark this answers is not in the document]
sagehen03: I would vote against an async only batch API.  I think for many of our user's data volumes synchronous response times will be fast enough and a synchronous API is simpler to work with.   I'd say let's start with sync and see if the need arises for async and add that alongside the sync API when we have a demonstrated need.

sagehen03 (approved):
Seems pretty reasonable.  Two high level comments that both kind of fall under the let's expose the smallest API surface that still allows our users complete their work.  Why give them two ways to do something when one suffices :shrug: ?

1. I see that we have API methods for reading and writing a single entity.  It seems like both of those are subsets of the more expansive read/query and write/batch APIs.  Working with a single entity at a time should be simpler for clients, but unless clients are dealing with a very small number of entities it's not going to be performant and it would be more API surface for us to maintain.  I'm not sure I see the value of having one entity at a time writes in particular.  Am I missing something?  Is this just for the prototype integration with the Batch team?
2. Do we need both `PATCH` and `PUT` for updates to entities?  What is the use case that the `PUT` method solves?  From what little I know about our users, it seems like they're often inserting new entities or adding new columns to existing entities.  Anecdotally, it seems less common that a user would want to blow away all the attributes for an existing entity and replace it with new ones.

src/main/resources/swagger/openapi-docs.yaml
davidangb: My thinking here is that for the most common use cases, the `GET` version would be must simpler for API consumers to call, to debug, and to understand at a glance; everything is right there in the URL. For instance, you can see the entire query in a request log; you don't have to correlate the request body.

That said, it's just a convenience/shortcut; I'd be fine launching with just the `POST` and adding the `GET` if/when we decide it's helpful to have it.

davidangb:
@sagehen03 

> Do we need both PATCH and PUT for updates to entities?

I think these do support different use cases and are both valuable.

`PATCH` supports writing a single attribute at a time - that is, the request payload can include only the attributes that are being updated, and the payload does not need to include the rest of the unchanged attributes. Use cases here are point corrections in the UI or writing batch-analysis outputs. Picture a point correction: a user changes a single cell in the UI data table and clicks save; this could be a `PATCH` to the entity in question, with a request payload of just the one attribute that changed.

The primary use case for `PUT` is to create an entity, but close behind that is the ability to delete attributes from an entity. Since `PATCH` is upsert-only, it has no way to delete attributes; but since `PUT` is a full replace, it will remove any attributes not present in the request payload.

> I see that we have API methods for reading and writing a single entity. It seems like both of those are subsets of the more expansive read/query and write/batch APIs.

You're correct; anything you can do in the single-entity APIs could also be done in the multi-entity query and batch-write APIs. If we HAD to choose one set, we'd obviously keep the batch APIs. But, I think the single-entity APIs are easier for callers to work with, easier to debug and reason about, and therefore they're worth keeping around. We could certainly *implement* the single-entity APIs at the Java layer by delegating to the multi-entity APIs, but it feels good to me to keep them around.

calypsomatic (approved):

src/main/resources/swagger/openapi-docs.yaml
KevinMarete: I agree that there is no need of duplicating functionality and maintaining multiple endpoints that do the same thing. Therefore I would stick to best practice and if that aligns with whatever is easier to use and maintain, then I would go for using `GET`

src/main/resources/swagger/openapi-docs.yaml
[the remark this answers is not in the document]
KevinMarete: Agreed, adding an extra async batch endpoint later would be ideal when a need is established.

src/main/resources/swagger/openapi-docs.yaml
KevinMarete: Will how we represent these URIs, need for the consumers of them to make any modifications?

Not sure what our bespoke format is but I agree plain JSON arrays are a good format to use unless we intend to search the json and expect large volumes hence I would advise to use JSONB format.

src/main/resources/swagger/openapi-docs.yaml
```diff
@@ -0,0 +1,482 @@
...
+        - name
+      type: object
+      properties:
+        name:
+          type: string
+          description: Attribute name
+    EntityId:
+      type: string
+    EntityMetadata:
+      type: object
```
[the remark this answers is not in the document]
KevinMarete: Makes sense, will we be using softDeletes? If yes then adding `deletedAt` would make sense!
What does the version/ETag mean? I thought version was for the endpoint definition not the entity?

src/main/resources/swagger/openapi-docs.yaml
KevinMarete: What will be the default behaviour? Enabled or Disabled

src/main/resources/swagger/openapi-docs.yaml
KevinMarete: Is this limitation for not being able to delete an attribute value by design/requirement? If not, why not have a `DELETE` endpoint that can remove (or soft delete if data exists) the attribute value. Or wrap the above stated functionality of the get entity, local modification and put for the entity into a `DELETE` endpoint!

src/main/resources/swagger/openapi-docs.yaml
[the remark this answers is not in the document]
davidangb: The implementation of soft vs hard deletes should be orthogonal to the API signatures; I think either soft or hard deletes could be represented with a `deletedAt` or other keys in the metadata.

version and/or [ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) would represent the version of the entity itself - each update to the entity would increment a version number. 

But again - all of this functionality is undefined, we are working on provenance functional requirements now. So I am trying to NOT take a stand on any of these features; I'm just making sure we have a `metadata {}` object in which we could store this info at a later point.

src/main/resources/swagger/openapi-docs.yaml
davidangb: yes, representing entity references as URIs would require changes for consumers as compared to the current Rawls format. However, ALL of the APIs defined in this PR are different from the Rawls APIs, so consumers would already be making changes across the board.

on json vs jsonb … the discussion here is about how to represent data in API request and response payloads. Are you suggesting using jsonb in a response payload?

src/main/resources/swagger/openapi-docs.yaml
davidangb: I guess it's by design. The `PATCH` endpoint allows the end user to include only those attributes that are being updated. Any attributes not present in the request payload will remain unchanged in the db. Thus, there's no way for a user to denote, using this payload, that an attribute should be deleted.

Options I see are:
1. allow deletion via `PATCH` by including an attribute but setting its value to null or some magic string like `__DELETE__`. I think we should avoid magic strings at all costs, and using null for deletes means you can't actually store a null value in the db, so I reject this option.
2. Add a `DELETE /{instanceid}/entities/{v}/{type}/{id}/attributes/{attribute}` API, as you suggest. I think this is a valid choice but might be unnecessary, if the `PUT` API suffices.
3. Use the `PUT` API to handle attribute deletes. Additionally, the `PUT` API allows a caller to both update and delete multiple attributes at once, while the `DELETE` API in the previous bullet point would only handle deletes.

davidangb: I believe the `PUT` API - option 3 - is required. Do you think we _also_ need the `DELETE` API (option 2)? Or are you suggesting that if we had the delete API, we could remove the put API?

KevinMarete: I am not saying we eliminate PUT. It is a required endpoint for updating an entire entity. I am just highlighting that we are achieving DELETE functionality without an actual DELETE endpoint. 

From your proposal it seems unnecessary to have a DELETE endpoint since removing data from the PUT request object will have the data replaced hence a similar effect to delete. Hence no need for a delete implementation.

Agreed.

PR #14, opened 2022-07-07
AJ-467 run unit tests with github actions

calypsomatic:
The sonarqube task was already named unit-and-tests-and-sonarqube plus there's a lot of duplicated action so I put the unit tests in with that task, but is there any reason to make them separate tasks?


.github/workflows/build-and-test.yml
```diff
@@ -61,6 +72,20 @@ jobs:
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
           restore-keys: ${{ runner.os }}-gradle
 
+      - name: Grant execute permission for gradlew
+        run: chmod +x gradlew
```
davidangb, line 76: is this `chmod` necessary? If it is, can we set the executable bit on `gradlew` and re-commit it, so there's no need to chmod here? (for example: https://github.com/actions/starter-workflows/issues/171)

davidangb (commented):
couple comments inline. I wouldn't block the PR on this, but my preference is to have separate workflows for unit test and sonarqube, even knowing there is duplication between them. Separate workflows would allow them to run in parallel, to run on different triggers, and to avoid any interactions between them. For instance, I think in this current workflow that if a unit test fails, sonarqube won't run.

.github/workflows/build-and-test.yml
```diff
@@ -40,6 +45,12 @@ jobs:
         with:
           fetch-depth: 0
 
+      - uses: actions/checkout@v2
```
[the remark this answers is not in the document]
calypsomatic: Ah good catch

.github/workflows/build-and-test.yml
calypsomatic: I think probably not necessary, will try it without.

sagehen03:
I think David raises a fair point that separating the two gives us future flexibility.  That being said this seems easy to change at a later date.  Do gh workflows/actions give users the ability to include common code/config so we could keep things separate while still keeping things DRY?  I honestly don't feel strongly, but if you're asking me to pick a side.  I'd say leave things as they are or investigate to see if there's solution that can make everyone happy.

sagehen03 (approved):

calypsomatic:
A quick look around makes it seem potentially plausible to avoid duplicating steps through reusable workflows, although that will take a little bit of rearrangement.  Splitting them up with duplication is pretty easy, and honestly, there's only two steps that duplicate so it's not a big deal.

jordansafer (approved):
+1 to having separate tasks, looks good

PR #17, opened 2022-07-11
Aj 485 single get

calypsomatic:
Includes code for AJ-485 - adding GET for a single entity, and AJ-486, adding PATCH for a single entity.

### DIFF
--- a/src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
+++ b/src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
@@ -36,6 +36,7 @@ public class EntityController {
         }
     }
 
+    @PatchMapping("/")
 
     @PostMapping("/api/workspaces/{workspaceNamespace}/{workspaceName}/entities/batchUpsert")
     public ResponseEntity<String> batchUpsert(@PathVariable("workspaceNamespace") String wsNamespace,
--- a/src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
+++ b/src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
@@ -44,6 +44,8 @@ public class EntityController {
                                                              @PathVariable("entityType") EntityType entityType,
                                                              @PathVariable("entityId") EntityId entityId,
                                                              @RequestBody EntityRequest entityRequest){
+        //not sure what to do with version path param just yet, thoughts?
+        Preconditions.checkArgument(version.equals("v0.2"));
         Entity singleEntity = dao.getSingleEntity(workspaceId, entityType, entityId);
         if(singleEntity == null){
             throw new ResponseStatusException(HttpStatus.NOT_FOUND, "Entity not found");
@@ -51,8 +53,11 @@ public class EntityController {
         Map<String, Object> attributesToUpdate = entityRequest.entityAttributes().attributes();
         Map<String, Object> existingAttributes = singleEntity.getAttributes();
         existingAttributes.putAll(attributesToUpdate);
-        dao.batchUpsert(List.of(singleEntity));
-        return new ResponseEntity<>(HttpStatus.CREATED);
+        EntityReferenceAction entityReferenceAction = referenceService.manageSingleEntityReference(workspaceId, singleEntity);
+        referenceService.saveReferencesAndEntities(entityReferenceAction);
+        EntityResponse response = new EntityResponse(entityId, entityType, new EntityAttributes(singleEntity.getAttributes()),
+                new EntityMetadata("TODO: SUPERFRESH"));
+        return new ResponseEntity<>(response, HttpStatus.CREATED);
     }
 
     @PostMapping("/api/workspaces/{workspaceNamespace}/{workspaceName}/entities/batchUpsert")
--- a/src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
+++ b/src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
@@ -8,6 +8,8 @@ import org.databiosphere.workspacedataservice.service.EntityReferenceService;
 import org.databiosphere.workspacedataservice.service.model.AttemptToUpsertDeletedEntity;
 import org.databiosphere.workspacedataservice.service.model.EntityReferenceAction;
 import org.databiosphere.workspacedataservice.shared.model.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.dao.EmptyResultDataAccessException;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
@@ -36,7 +38,22 @@ public class EntityController {
         }
     }
 
-    @PatchMapping("/")
+    @PatchMapping("/{workspaceId}/entities/{version}/{entityType}/{entityId}")
+    public ResponseEntity<EntityResponse> updateSingleEntity(@PathVariable("workspaceId") UUID workspaceId,
+                                                             @PathVariable("version") String version,
+                                                             @PathVariable("entityType") EntityType entityType,
+                                                             @PathVariable("entityId") EntityId entityId,
+                                                             @RequestBody EntityRequest entityRequest){
+        Entity singleEntity = dao.getSingleEntity(workspaceId, entityType, entityId);
+        if(singleEntity == null){
+            throw new ResponseStatusException(HttpStatus.NOT_FOUND, "Entity not found");
+        }
+        Map<String, Object> attributesToUpdate = entityRequest.entityAttributes().attributes();
+        Map<String, Object> existingAttributes = singleEntity.getAttributes();
+        existingAttributes.putAll(attributesToUpdate);
+        dao.batchUpsert(List.of(singleEntity));
+        return new ResponseEntity<>(HttpStatus.CREATED);
+    }
 
     @PostMapping("/api/workspaces/{workspaceNamespace}/{workspaceName}/entities/batchUpsert")
     public ResponseEntity<String> batchUpsert(@PathVariable("workspaceNamespace") String wsNamespace,
--- a/src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
+++ b/src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
@@ -121,5 +121,18 @@ public class EntityController {
                 (page-1) * pageSize, filterTerms, sortField, sortDirection, fields) : Collections.emptyList());
     }
 
+    @GetMapping("/{instanceId}/entities/{version}/{entityType}/{id}")
+    public ResponseEntity<EntityResponse> getSingleEntity(@PathVariable("instanceId") UUID instanceId,
+                                              @PathVariable("entityType") EntityType entityType,
+                                              @PathVariable("id") EntityId entityId) {
+        Entity result = dao.getSingleEntity(instanceId, entityType, entityId);
+        if (result == null){
+            throw new ResponseStatusException(HttpStatus.NOT_FOUND, "Entity not found");
+        }
+        EntityResponse response = new EntityResponse(entityId, entityType, new EntityAttributes(result.getAttributes()),
+                new EntityMetadata("TODO: ENTITYMETADATA"));
+        return new ResponseEntity<>(response, HttpStatus.OK);
+    }
+
 
 }
--- a/src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
+++ b/src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
@@ -8,6 +8,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.databiosphere.workspacedataservice.service.model.EntityReference;
 import org.databiosphere.workspacedataservice.service.model.InvalidEntityReference;
 import org.databiosphere.workspacedataservice.shared.model.Entity;
+import org.databiosphere.workspacedataservice.shared.model.EntityId;
 import org.databiosphere.workspacedataservice.shared.model.EntityToDelete;
 import org.databiosphere.workspacedataservice.shared.model.EntityType;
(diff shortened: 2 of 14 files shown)


src/main/java/org/databiosphere/workspacedataservice/shared/model/EntityId.java
```diff
@@ -0,0 +1,3 @@
+package org.databiosphere.workspacedataservice.shared.model;
+
+public record EntityId(String entityIdentifier) {}
```
calypsomatic, line 3: So `EntityId` is now a record, but `Entity` still contains a `String` for its name.  Should this be switched to `EntityId`?  Do we have both?  I have a dim notion of why having an `EntityId` is useful but maybe someone could spell it out for me.

static/swagger/openapi-docs.yaml
```diff
@@ -439,6 +457,27 @@ components:
         name:
           type: string
           description: Entity type name
+    ResponseEntity:
```
davidangb, line 460: rename to `ErrorResponse` or something like that? `ResponseEntity` imho sounds like a successful data entity response payload.

Also, this schema does not validate in https://editor.swagger.io/, though it validates locally in my IntelliJ. The error is: `Schemas with 'type: array', require a sibling 'items: ' field`

static/swagger/openapi-docs.yaml
```diff
@@ -326,6 +338,12 @@ components:
         application/json:
           schema:
             $ref: '#/components/schemas/EntityResponse'
+    ResponseEntity:
+      description: Used for error
+      content:
+        application/json:
+          schema:
+            $ref: '#/components/schemas/ResponseEntity'
```
davidangb, line 346: this seems to be currently unused … delete it or use it inside the 404s above

davidangb (commented):
Love to see this progress!

src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
```diff
@@ -36,6 +38,27 @@ private UUID getWorkspaceId(String wsNamespace, String wsName) {
...
+        }
+        Map<String, Object> attributesToUpdate = entityRequest.entityAttributes().attributes();
+        Map<String, Object> existingAttributes = singleEntity.getAttributes();
+        existingAttributes.putAll(attributesToUpdate);
+        EntityReferenceAction entityReferenceAction = referenceService.manageSingleEntityReference(workspaceId, singleEntity);
+        referenceService.saveReferencesAndEntities(entityReferenceAction);
+        EntityResponse response = new EntityResponse(entityId, entityType, new EntityAttributes(singleEntity.getAttributes()),
+                new EntityMetadata("TODO: SUPERFRESH"));
+        return new ResponseEntity<>(response, HttpStatus.CREATED);
+    }
```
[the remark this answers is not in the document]
calypsomatic: SUPERFRESH was all Drew.

src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
```diff
@@ -186,22 +201,22 @@ public void insertWorkspace(String name, String namespace, UUID id){
         template.update("insert into workspace (id, name, namespace) values (?,?,?)", id, name, namespace);
     }
 
-    public long loadEntityType(EntityType entityType, UUID workspaceId) {
+    public long loadEntityType(EntityType entityType, UUID instanceId) {
         GeneratedKeyHolder generatedKeyHolder = new GeneratedKeyHolder();
```
calypsomatic, line 205: The relationship between `EntityType` and its `id` and how each gets used is a little strange to me.  The `EntityType` class does not have an id, and there's mix and match on whether we identify a type by id or name throughout the codebase.  Willing to accept any comments or input.

src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
```diff
@@ -36,6 +38,27 @@ private UUID getWorkspaceId(String wsNamespace, String wsName) {
...
+        Preconditions.checkArgument(version.equals("v0.2"));
+        Entity singleEntity = dao.getSingleEntity(workspaceId, entityType, entityId);
+        if(singleEntity == null){
+            throw new ResponseStatusException(HttpStatus.NOT_FOUND, "Entity not found");
+        }
+        Map<String, Object> attributesToUpdate = entityRequest.entityAttributes().attributes();
+        Map<String, Object> existingAttributes = singleEntity.getAttributes();
+        existingAttributes.putAll(attributesToUpdate);
+        EntityReferenceAction entityReferenceAction = referenceService.manageSingleEntityReference(workspaceId, singleEntity);
+        referenceService.saveReferencesAndEntities(entityReferenceAction);
```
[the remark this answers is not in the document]
calypsomatic: Added a TODO in the code so we don't forget.

src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
```diff
@@ -98,5 +121,18 @@ public EntityQueryResult queryForEntities(@PathVariable("workspaceNamespace") St
...
     }
 
+    @GetMapping("/{instanceId}/entities/{version}/{entityType}/{id}")
+    public ResponseEntity<EntityResponse> getSingleEntity(@PathVariable("instanceId") UUID instanceId,
+                                              @PathVariable("entityType") EntityType entityType,
+                                              @PathVariable("id") EntityId entityId) {
+        Entity result = dao.getSingleEntity(instanceId, entityType, entityId);
+        if (result == null){
+            throw new ResponseStatusException(HttpStatus.NOT_FOUND, "Entity not found");
+        }
```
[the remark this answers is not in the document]
calypsomatic: Put a TODO in the code so we don't forget.

src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
```diff
@@ -36,13 +36,34 @@ private UUID getWorkspaceId(String wsNamespace, String wsName) {
...
+        Map<String, Object> attributesToUpdate = entityRequest.entityAttributes().attributes();
+        Map<String, Object> existingAttributes = singleEntity.getAttributes();
+        existingAttributes.putAll(attributesToUpdate);
+        //TODO: remove entityType/entityName JSON object format for references and move to URIs in the request/response payloads
+        EntityReferenceAction entityReferenceAction = referenceService.manageSingleEntityReference(instanceId, singleEntity);
+        referenceService.saveReferencesAndEntities(entityReferenceAction);
+        EntityResponse response = new EntityResponse(entityId, entityType, new EntityAttributes(singleEntity.getAttributes()),
+                new EntityMetadata("TODO: SUPERFRESH"));
+        return new ResponseEntity<>(response, HttpStatus.OK);
+    }
```
calypsomatic, line 59: Similar to using `EntityId` instead of a `String`, shouldn't we use `EntityAttributes` instead of a plain `Map` inside of `Entity`?

davidangb: yes, agreed that `EntityAttributes` is better than `Map`!

src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
```diff
@@ -102,7 +102,9 @@ public void batchUpsert(List<Entity> entities) {
             @Override
             public void setValues(PreparedStatement ps, int i) throws SQLException {
                 Entity entity = entities.get(i);
-                ps.setString(1, entity.getName());
+                //This is an example of a place where we want a string instead of an EntityId
+                //Should there be a more direct way to get entityId as a string?
+                ps.setString(1, entity.getName().entityIdentifier());
```
davidangb, line 107: sure, I could see something like:
```java
public String getNameAsString() { return name.entityIdentifier(); }
```
as a convenience method on `Entity`

src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
davidangb: I agree the `id` column on the `entity_type` table could be extraneous. We could define uniqueness for entity types by their name and the WDS instance id (which is "workspace_id" in the existing table) - and in fact we _could_ define the combination of `name` and `workspace_id` as a compound primary key. We _may_ want to keep the auto-incrementing integer id in the db as a convenience, but I don't think it should be exposed to the Java layer much/at all.

src/test/java/org/databiosphere/workspacedataservice/controller/PatchSingleEntityTest.java
```diff
@@ -26,13 +27,36 @@ public class PatchSingleEntityTest {
...
+        assertEquals(Map.of("created_at", "2022-10-01", "foo", "baz"), response.getBody().entityAttributes().attributes());
+
+        //Add a reference
+        Map<String, Object> refAttr = new HashMap<>();
+        refAttr.put("entityType", testEntityType);
+        refAttr.put("entityName", testEntityId);
+        EntityAttributes referencingAttributes = new EntityAttributes(Map.of("referencingAttr", refAttr));
+        response = controller.updateSingleEntity(UUID.randomUUID(), "v0.2", testEntityType, new EntityId("referencingEntity"),
+                new EntityRequest(testEntityId, testEntityType, referencingAttributes));
+        System.out.println(response.getBody().entityAttributes());
```
davidangb, line 57: nit: remove the `println`

src/test/java/org/databiosphere/workspacedataservice/controller/PatchSingleEntityTest.java
```diff
@@ -26,13 +27,36 @@ public class PatchSingleEntityTest {
...
+        assertTrue(response.getBody().entityAttributes().attributes().size() == 3);
+        assertEquals(Map.of("created_at", "2022-10-01", "foo", "baz", "referencingAttr", refAttr), response.getBody().entityAttributes().attributes());
```
davidangb, line 59: Could you make these three use cases - adding an attribute, overwriting an attribute, adding a reference - into separate unit tests? That way they'll always all run and we can see which pass/fail. Right now, if adding an attribute fails then the other two use cases won't run at all.

src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
calypsomatic: Done, but it's a bit of a hassle in places: whenever we manipulate the underlying map, now we have to add an extra `.attributes()` (e.g. `entity.getAttributes().attributes().put(....)`), and a slight workaround to update attributes since records are immutable.

src/test/java/org/databiosphere/workspacedataservice/controller/PatchSingleEntityTest.java
calypsomatic: whoops!  Good catch.

src/test/java/org/databiosphere/workspacedataservice/controller/PatchSingleEntityTest.java
calypsomatic: Good call.

src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
calypsomatic: Yes I was definitely considering that, it just seemed potentially about as much as a hassle to have to essentially duplicate potentially every Map method.

src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
calypsomatic: Maybe we could take Drew's suggestion and make a follow-on ticket to cover cleaning this up and making things consistent, so this PR doesn't keep ballooning out of control.

davidangb: oh absolutely. I apologize - my comment wasn't intended as a blocker for this PR. I was adding comments/input, but they're definitely for follow-on work!

calypsomatic: No, it wasn't you, I was eagerly doing all kinds of refactoring.

src/test/java/org/databiosphere/workspacedataservice/controller/EntityControllerTest.java
```diff
@@ -0,0 +1,94 @@
...
+        controller = new EntityController(referenceService, entityDao);
+    }
+
+    @Test
+    void testPatchNewSingleEntity(){
+        when(entityDao.getSingleEntity(any(), any(), any())).thenReturn(new Entity(testEntityId,
+                testEntityType, new EntityAttributes(Map.of("created_at", "2022-10-01")), 1L));
+        ResponseEntity<EntityResponse> response = controller.updateSingleEntity(UUID.randomUUID(), "v0.2", testEntityType, testEntityId,
+                new EntityRequest(testEntityId, testEntityType, new EntityAttributes(Map.of("foo", "bar"))));
+        assertTrue(response.getBody().entityAttributes().attributes().size() == 2);
```
davidangb, line 48: nit: the asserts on `size()` here are extraneous; the following line that asserts the contents of the map will inherently also check size.

davidangb (approved):
nice job getting this to the finish line!

sagehen03 (approved):
If it were up to me, I would wait to add `EntityAttributes` and other classes that just wrap a single field until we need more logic or data there.  David raises a fair point about passing args, but the IDE is pretty good (ctrl+p) there or we could use builders.  With the wrapper classes, I feel like the API is a little more clunky and compels more typing (but the IDE also helps there too).  For me it's a pretty balanced tradeoff (and a pretty minor detail) so I'm fine with more wrappers to start--perhaps that will prove helpful as the system evolves, but I wanted to register my sentiments and I'm happy to talk more about it in team topics.

PR #18, opened 2022-07-12
[AJ-472] attribute swagger details

jordansafer:
Sample json entity:

```
{
  "id": "string",
  "type": "string",
  "attributes": [
    {
      "name": "string",
      "value": "string"
    }
  ]
}
```

### DIFF
--- a/static/swagger/openapi-docs.yaml
+++ b/static/swagger/openapi-docs.yaml
@@ -344,6 +344,12 @@ components:
             items:
               $ref: '#/components/schemas/BatchOperation'
   schemas:
+    Attribute:
+      type: object
+      properties:
+        value:
+          type: string
+          description: ""
     BatchOperation:
       type: object
       required:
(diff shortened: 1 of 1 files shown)


calypsomatic (approved):

davidangb:
Looking for opinions - I was thinking a en entity representation would look like:
```json
{
  "id": "string",
  "type": "string",
  "attributes": {
    "attr1": "string",
    "attr2": true,
    "attr3": 12345,
    "attr4": [
      "array",
      "of",
      "values"
    ]
  }
}
```
which feels easier to read and for a user to understand, though it's more difficult to represent in swagger.  We could do it with:
```yaml
    EntityAttributes:
      type: object
      additionalProperties: true
```
_([doc for additionalProperties](https://swagger.io/docs/specification/data-models/dictionaries/))_

we have to make sure to handle booleans, numbers, arrays, etc in any payload, so we can't just define values as strings.

davidangb:
This was addressed in #28 (https://github.com/DataBiosphere/terra-workspace-data-service/pull/28/files#diff-da6e75dcb2752daa48de659645011da025066299c29de595d206ef62daafd417R434) - close this out?

PR #19, opened 2022-07-20
AJ-446: Create Docker Image for WDS

sagehen03:
Assuming you have postgres running on localhost on the default port, with these changes you should be to create a docker image:
`./gradlew --build-cache jibDockerBuild --image=<image-name> -Djib.console=plain`
To run:
`docker run --network=host -e WDS_DB_HOST='127.0.0.1' -p 8080:8080 <image-name>`

The docker run invocation seems to vary by platform.  The above works for me on linux with Docker v20.10, but using `host.docker.internal` is available on a Mac and should let you drop the `network=host`.   https://stackoverflow.com/questions/31249112/allow-docker-container-to-connect-to-a-local-host-postgres-database


calypsomatic (approved):

build.gradle
```diff
@@ -3,6 +3,7 @@ plugins {
     id 'io.spring.dependency-management' version '1.0.11.RELEASE'
     id 'org.liquibase.gradle' version '2.1.0'
     id 'java'
+    id 'com.google.cloud.tools.jib' version '3.2.0'
```
davidangb, line 6: very minor: could we use `3.2.1` which is latest?

davidangb (approved):
one minor comment inline, proactively giving this a 👍 

The code in here is good - I'd like to follow on with a second PR to update the README or some other developer-facing documentation explaining how to build/run the docker image, once we work through the Mac/Linux/host network/postgres access issues.

build.gradle
sagehen03: Yes! I had latest originally and changed to the version that cbas is using as a shot on the dark to see if that explained why cbas' swagger worked for docker but not ours.

PR #20, opened 2022-07-20
AJ-448 first pass at adding swagger codegen

calypsomatic:
Running `./gradlew build` will now generate code from the swagger doc and put it in `/client`.  Note this folder and thus the generated code is .gitignored.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,8 +66,8 @@ swaggerSources {
             language = 'java'
             library = "jersey2"
             outputDir = file("$projectDir/client")
-            jvmArgs = ['--add-opens=java.base/java.util=ALL-UNNAMED'] // for Swagger Codegen v3 on Java 16+
-            components = [apis: true, apiTests: false, models: true, modelTests: false]
+            jvmArgs = ['--add-opens=java.base/java.util=ALL-UNNAMED'] // needed for Java 16+
+            components = [apiTests: false]
             additionalProperties = [
                     modelPackage     : "databiosphere.workspacedataservice.model",
                     apiPackage       : "databiosphere.workspacedataservice.api"
@@ -76,6 +76,10 @@ swaggerSources {
             dependsOn validation
         }
     }
+}
+
+clean {
+    delete 'client'
 }
 
 compileJava.dependsOn tasks.generateSwaggerCode
--- a/build.gradle
+++ b/build.gradle
@@ -68,3 +68,10 @@ swaggerSources {
 
 compileJava.dependsOn tasks.generateSwaggerCode
 
+jar {
+    archiveBaseName = "swagger"
+    from "$projectDir/generated/src/main/java"
+    destinationDirectory = file("$projectDir/client")
+}
+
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -45,3 +47,18 @@ tasks.named('test') {
     }
 }
 
+swaggerSources {
+    server {
+        inputFile = file('static/swagger/openapi-docs.yaml')
+        code {
+            language = 'java'
```
calypsomatic, line 54: Do we want 'java' or 'spring' here?  'spring' gives more readable java files, and I see it used in other repos, but which form does the client want it in?

davidangb: on `java` vs `spring`: according to the listings at https://github.com/swagger-api/swagger-codegen, `java` generates a client, while `spring` generates server stubs. So, we definitely want `java`

build.gradle
```diff
@@ -45,3 +47,18 @@ tasks.named('test') {
...
+            jvmArgs = ['--add-opens=java.base/java.util=ALL-UNNAMED'] // for Swagger Codegen v3 on Java 16+
```
davidangb, line 55: what does this do? I don't immediately get an error without this, but I haven't tried using the generated files. Is it necessary?

build.gradle
```diff
@@ -45,3 +47,18 @@ tasks.named('test') {
     }
 }
 
+swaggerSources {
+    server {
```
davidangb, line 51: I think this needs to be `client`, not `server`

davidangb (commented):
Some initial config thoughts:
* I think we need config for the java package names; else it looks like the generated files are in `io.swagger.client`. See https://github.com/DataBiosphere/jade-data-repo/blob/0ae97efce6deb81a41001c4f245e217d768ce0c9/datarepo-client/build.gradle#L59-L62 for an example.
* should we add an `outputDir` config setting? I think this would make it easier for developers to be able to read the build.gradle and immediately know where the generated files are created. See https://github.com/DataBiosphere/jade-data-repo/blob/0ae97efce6deb81a41001c4f245e217d768ce0c9/datarepo-client/build.gradle#L57 for an example. If this creates a new directory, make sure to add it to `.gitignore` so we don't commit generated files.
* from reading the generated files, it looks like by default the Java client uses okhttp/gson. I see that Data Repo and WSM, and CBAS are using jersey2, should we use that? Example: https://github.com/DataBiosphere/cbas/blob/5d7528b2e2bef3a834938d3eb73051885873764e/client/swagger.gradle#L17; list of Java engine options: https://github.com/swagger-api/swagger-codegen#modifying-the-client-library-format
* from reading the generated files, it is creating tests for the apiclient. We might want to skip generating those; example https://github.com/DataBiosphere/cbas/blob/5d7528b2e2bef3a834938d3eb73051885873764e/client/swagger.gradle#L23-L26

build.gradle
calypsomatic: The docs said it was necessary for Java 16+.

PR #22, opened 2022-07-20
bugfixes for swagger ui [AJ-511]

davidangb:
* in the HTML definition, fix the path to index.css
* in the OpenAPI definition, remove "(pending)" from the GET and PATCH apis because they are now implemented.

### DIFF
--- a/static/swagger/openapi-docs.yaml
+++ b/static/swagger/openapi-docs.yaml
@@ -20,7 +20,7 @@ tags:
 paths:
   /{instanceid}/entities/{v}/{type}/{id}:
     get:
-      summary: Get entity (pending)
+      summary: Get entity
       description: Retrieves a single entity by its type and id
       tags:
         - Entities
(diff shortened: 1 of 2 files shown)


calypsomatic (approved):

sagehen03:
Not sure if you want to do this in a separate PR but swagger shows PATCH single entity accepted a payload that is just plain strings for entityType and entityName and a map for entityAttributes, but the code wants `EntityType`, `EntityId`, and `EntityAttributes` values, aka this works:
`curl -H "Content-type: application/json" -X PATCH "http://localhost:8080/15f36863-30a5-4cab-91f7-52be439f1175/entities/v0.2/sample/sample2" -d '{"entityId": {"entityIdentifier": "sample2"}, "entityType": {"name": "sample"}, "entityAttributes": {"attributes": {"date-collected": "2022-11-01"}}}'`
but what the swagger suggests:
`curl -H "Content-type: application/json" -X PATCH "http://localhost:8080/15f36863-30a5-4cab-91f7-52be439f1175/entities/v0.2/sample/sample2" -d '{"entityId": "sample2", "entityType": "sample", "entityAttributes": {"date-collected": "2022-11-01"}}'`
does not work.

sagehen03 (approved):

davidangb:
@sagehen03 ooh good catch, I would say that in this case the swagger definition is more desirable and what's actually in place is less desirable - the actual payload feels like it has too many extra layers.

PR #24, opened 2022-07-21
get/patch payloads [AJ-512]

davidangb:
* JSON request and responses for the get/patch single-entity APIs now match the swagger definition
* added unit tests for JSON de/serialization of EntityRequest/EntityResponse
* sorted the EntityAttribute keys in the JSON response; this helps with unit tests as well as being nice for the end user

In order to achieve these, I needed to turn some of the record classes into "real" classes, which allowed me to Jackson-annotate specific methods. I used the default getter/setter names, which caused some churn elsewhere in the code.

---

N.B. I still think the use of wrapper classes here is not ideal, but this PR is big enough already and I will make a separate one to illustrate how I think wrapper classes should be used.


src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
```diff
@@ -50,7 +50,7 @@ public List<EntityReference> getReferencesForEntities(List<Entity> entitiesForTy
         for (List<Entity> chunk : chunks) {
             result.addAll(namedParameterJdbcTemplate.query("select entity_type, entity_name, referenced_entity_type, " +
                             "referenced_entity_name from entity_reference where entity_type = :entityType and entity_name in (:entityNames)",
-                    new MapSqlParameterSource(Map.of("entityType", entityTypeId, "entityNames", chunk.stream().map(Entity::getName).collect(Collectors.toSet()))),
+                    new MapSqlParameterSource(Map.of("entityType", entityTypeId, "entityNames", chunk.stream().map(e -> e.getName().getEntityIdentifier()).collect(Collectors.toSet()))),
```
davidangb, line 53: see https://github.com/DataBiosphere/terra-workspace-data-service/pull/19/files#r925881849

src/main/java/org/databiosphere/workspacedataservice/shared/model/EntityRequest.java
```diff
@@ -1,3 +1,10 @@
...
 
-public record EntityRequest(EntityId entityId, EntityType entityType, EntityAttributes entityAttributes) {}
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+
+@JsonPropertyOrder({"id", "type", "attributes"})
+public record EntityRequest(
+        @JsonProperty("id") EntityId entityId,
+        @JsonProperty("type") EntityType entityType,
+        @JsonProperty("attributes") EntityAttributes entityAttributes) {}
```
davidangb, line 10: I could have just renamed the member variables, but `JsonProperty` also works; I'm not attached to either implementation.

static/swagger/openapi-docs.yaml
```diff
@@ -276,6 +276,7 @@ components:
       required: true
       schema:
         type: string
+        default: '15f36863-30a5-4cab-91f7-52be439f1175'
```
davidangb, line 279: just a convenience, not required for this PR at all

calypsomatic (approved):

jordansafer (approved):

PR #25, opened 2022-07-22
encapsulating EntityAttributes

davidangb:
Add a bunch of the `Map` methods to `EntityAttributes`, and use those to better encapsulate `EntityAttributes`' internal implementation choice of `Map<String, Object>`.

Calling code should - in almost every case - not need to know about what `EntityAttributes` uses to store its data; they can just call `EntityAttribute`'s methods.

This isn't 100% clean because the constructor still accepts a `Map<String, Object>`. Unless we also create wrapper classes for Attribute, AttributeValue, etc. we need the raw use of `Map` here. I considered making a static `EntityAttributes.fromMap(Map<String, Object>)` method to keep the naming precise, but that's not really any different; open to opinions here.

Another potential refactor is to move some methods up to `Entity`. I could imagine calling `entity.putAttribute(key, value)` instead of `entity.getAttributes().put(key, value)`, for convenience.


calypsomatic:
I like the idea of putting methods in `Entity`, although potentially it could be in addition to the methods in `EntityAttributes` instead of moving them out, there might be both use cases.

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/EntityAttributes.java
```diff
@@ -3,25 +3,51 @@
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 
-import java.util.Map;
-import java.util.Objects;
-import java.util.TreeMap;
+import java.util.*;
```
jordansafer, line 6: Nit, let's try not to use star imports

jordansafer (approved):

sagehen03 (commented):
I don't like what we've done with `EntityAttributes`, `EntityType`, and `EntityId`.  I wasn't around to review that PR.  What is the benefit of those classes when they don't add any additional validation or logic on top of their underlying java types (Map, String, and String)?  David, I think you're thinking that we may need to add validation or logic eventually so let's add the additional layer of abstraction now so that's we can seamlessly add the needed logic later?  Are there additional reasons?  Up to now, these changes have to my eyes compelled extra typing, been the source of bugs in code that didn't get migrated properly, and complicated our JSON serialization.   Probably best if we talk face to face about this.

davidangb:
closing - I still think this is worthwhile but no matter what this would need a new PR since a lot of the underlying code has changed.

PR #28, opened 2022-07-25
aj-449

calypsomatic:
AJ-448 and AJ-449, auto-generates client code from the swagger file, puts it into a jar. 
Auto-updates version - to tell it to update, simply include # + patch, minor, or major in the comment of your PR, and upon merge to main, the appropriate version update will happen, the jar built and pushed to artifactory.
Had to move the `src` directory into `service` and otherwise split up the `build.gradle`s to get it to work, but I think it works!  Please ignore the 8 billion commits made while trying to test out the github actions.


client/swagger.gradle
```diff
@@ -0,0 +1,33 @@
...
+
+	implementation 'io.swagger.core.v3:swagger-annotations'
+	swaggerCodegen 'io.swagger.codegen.v3:swagger-codegen-cli'
+}
+
+
+generateSwaggerCode {
+	inputFile = file('../service/src/main/resources/static/swagger/openapi-docs.yaml')
+	language = 'java'
+	library = 'resttemplate'
```
davidangb, line 13: oooh, `resttemplate` instead of `jersey2`. I'm not opposed - but what's the motivation to switch? What's the difference?

service/generators.gradle
```diff
@@ -0,0 +1,30 @@
...
+			springBootVersion: dependencyManagement.managedVersions['org.springframework.boot:spring-boot']
+	]
+	rawOptions = ['--import-mappings', "VersionProperties=${artifactGroup}.config.VersionProperties"]
+}
+
+String swaggerOutputSrc = "${generateSwaggerCode.outputDir}/src/main/java"
+
+idea.module.generatedSourceDirs = [file(swaggerOutputSrc)]
+sourceSets.main.java.srcDir swaggerOutputSrc
+compileJava.dependsOn generateSwaggerCode
```
davidangb, line 30: I don't see a need to generate server stubs - I think this whole `generators.gradle` file could be deleted. We only need to generate client stubs right now.

davidangb (commented):
Some comments inline! Additionally:
1. I see that when generating code, the generator warns about empty `operationId` values in the swagger definition. Not for this PR - let's track follow-on work to add those and quell the warnings.
2. I asked this in Slack too - how well does the generated client code work?

client/swagger.gradle
calypsomatic: It wasn't working with jersey2!  Maybe if I try again with this particular configuration it will work with jersey2 this time, but I'll look into the difference.

service/generators.gradle
calypsomatic: I agree with you - except that generating only the server stubs made packaging and versioning the jar not work.  I spent 2+ days on it and gave up and went with what I was seeing everywhere else - which is also generating the server stubs.  I do not understand why and if you know or can figure it out, I'd be very happy to do so.

client/swagger.gradle
davidangb: I'm not arguing for or against jersey - just want to understand the choices! If resttemplate is working, I think that's fine - and might even be easier for any Spring-based services that want to use our client.

service/generators.gradle
davidangb: If I:
* delete `service/generators.gradle`
* delete line 26 `apply from: 'generators.gradle'` from `service/build.gradle`

then I can run `./gradlew :client:artifactoryPublish` successfully (I intentionally set a bad password and get a 401 from artifactory, but hopefully that's past the point of packaging/versioning)

calypsomatic: Amazing!  😭

sagehen03:
I was able to build the client jar and then in a new project I was able to add the jar as a dependency and call my local wds with this bit of code: 
```
ApiClient client = new ApiClient();
client.setBasePath("http://localhost:8080");
EntitiesApi api = new EntitiesApi(client);

EntityResponse entityResponse = api.instanceidEntitiesVTypeIdGet("15f36863-30a5-4cab-91f7-52be439f1175", "v0.2", "sample", "sample2");
```

The two main issues I've encountered so far is EntityResponse does not seem to contain the attributes and something seems misconfigured about how this library is published.  Maybe it's not that important, but I was not able to publish the library to my local maven repo and I had to manually reference the jar in my test project.

sagehen03:
> I was able to build the client jar and then in a new project I was able to add the jar as a dependency and call my local wds with this bit of code:
> 
> ```
> ApiClient client = new ApiClient();
> client.setBasePath("http://localhost:8080");
> EntitiesApi api = new EntitiesApi(client);
> 
> EntityResponse entityResponse = api.instanceidEntitiesVTypeIdGet("15f36863-30a5-4cab-91f7-52be439f1175", "v0.2", "sample", "sample2");
> ```
> 
> The two main issues I've encountered so far are EntityResponse does not seem to contain the attributes and something seems misconfigured about how this library is published. Maybe it's not that important, but I was not able to publish the library to my local maven repo and I had to manually reference the jar in my test project.

I made a couple of changes to `open-api.yaml` and things now seem to be working reasonably well for our two methods.   With the exception that RestTemplate+PATCH seems to require some [extra config](https://stackoverflow.com/questions/29447382/resttemplate-patch-request).  It did work once I brought in apache http client and passed the apache rest template to the client lib but probably nicer not to impose that on client consumers.

client/swagger.gradle
calypsomatic: Turns out `resttemplate` does not support PATCH, whereas `jersey2` results in code that doesn't compile correctly.  SO!  I went with `jersey2` but swiped a template update from terra-workspace-manager that modified how the code is generated and seems to work.

CONTRIBUTING.md
```diff
@@ -10,6 +10,7 @@ Make sure all unit tests pass, including those you wrote for any new functionali
 ## Opening a Pull Request
 
 When you open a pull request, add the JIRA issue number (e.g. `ABC-123`) to the PR title. This will make a reference from JIRA to the GitHub issue. Describe your changes, including instructions on how to reproduce the behavior and any screenshots if applicable. 
+In the comments for one of your PRs, include '#patch', '#minor', or '#major' to increment the version.
```
davidangb, line 13: is this required? It doesn't automatically bump on merge to main?

README.md
```diff
@@ -97,4 +97,18 @@ docker run -e WDS_DB_HOST='host.docker.internal' -p 8080:8080 wdsdocker
...
+repositories {
+    maven {
+        url "https://broadinstitute.jfrog.io/broadinstitute/libs-snapshot-local/"
+    }
+}
+
+dependencies {
+    implementation(group: 'org.databiosphere', name: 'workspacedataservice-client', version: 'x.x.x')
+```
+The lastest version can be seen in our `build.gradle` file in the root directory.
```
davidangb, line 114: nit: `s/lastest/latest/`. Also, if the bumper really creates a tag, then the latest version can also be seen right in github, that might be easier to recommend than inspecting the build.gradle.

CONTRIBUTING.md
calypsomatic: Ah, good point, it automatically updates by the default bump (patch), you only comment if you want something other than that.  I'll fix it.

client/swagger.gradle
```diff
@@ -0,0 +1,49 @@
...
+	jersey_version = "2.36"
+}
+
+dependencies {
+	api 'org.springframework.boot:spring-boot-starter-json:+'
+
+	implementation 'io.swagger.core.v3:swagger-annotations'
+	swaggerCodegen 'io.swagger.codegen.v3:swagger-codegen-cli'
+	//Need to include libraries for generated code to work
+	implementation "io.swagger.core.v3:swagger-annotations:$swagger_annotations_version"
```
davidangb, line 13: `swagger-annotations` is listed twice - once at line 10 without a version, then again here at line 13 with a version

client/ApiClient.mustache
```diff
@@ -0,0 +1,845 @@
+{{!--
+  Copied from the https://github.com/swagger-api/swagger-codegen-generators tagged release 1.0.24:
+  https://github.com/swagger-api/swagger-codegen-generators/blob/v1.0.24/src/main/resources/handlebars/Java/libraries/jersey2/ApiClient.mustache
```
davidangb, line 3: We are using `swagger-codegen-cli:3.0.33`, which according to https://github.com/swagger-api/swagger-codegen/releases/tag/v3.0.33 uses [swagger-codegen-generators/releases/tag/v1.0.32](https://github.com/swagger-api/swagger-codegen-generators/releases/tag/v1.0.32). This template was copied from `1.0.24`. Could/should the template be updated to the compatible version?

Of note, looks like the latest of `swagger-codegen-cli` is 3.0.34, which uses [1.0.34](https://github.com/swagger-api/swagger-codegen-generators/releases/tag/v1.0.34). 

Is it even possible that this customization is not needed in the most recent version?

client/swagger.gradle
```diff
@@ -0,0 +1,49 @@
...
+	implementation "org.glassfish.jersey.core:jersey-client:$jersey_version"
+	implementation "org.glassfish.jersey.media:jersey-media-multipart:$jersey_version"
+	implementation "org.glassfish.jersey.media:jersey-media-json-jackson:$jersey_version"
+	implementation "com.fasterxml.jackson.core:jackson-core:$jackson_version"
+	implementation "com.fasterxml.jackson.core:jackson-annotations:$jackson_version"
+	implementation "com.fasterxml.jackson.core:jackson-databind:$jackson_version"
+	implementation "com.brsanthu:migbase64:2.2"
```
davidangb, line 20: why do we need `migbase64`? (are there any other of these dependencies we could omit?)

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -428,6 +431,7 @@ components:
           description: Provenance and other entity metadata. May not be present in v0.2.
     EntityAttributes:
       type: object
+      additionalProperties: true
```
davidangb, line 434: cool, this is exactly the short-term fix I was thinking of (see also https://github.com/DataBiosphere/terra-workspace-data-service/pull/18#issuecomment-1185529295)

davidangb (commented):
comments inline!

client/swagger.gradle
calypsomatic: All of these dependencies were adding through the painstaking process of trying to generate the code only to find it was missing some dependencies, and then adding in the dependency and letting it fail on the next missing dependency.

README.md
calypsomatic: Righto!

client/swagger.gradle
calypsomatic: Whoops

client/ApiClient.mustache
calypsomatic: Tried it out, the customization is still needed, but I did update the swagger version (and the jackson version to what they're using).

davidangb (approved):
:+1:

follow-on tickets that I see:
* investigate openapi-generator to see if it requires less customization/hacking
* review/update the `operationId` values in the yaml definition, since those become method names in the client and we want client methods to be nicely named
* gather feedback on the client and/or use it ourselves to assess model classes and other behavior
* figure out what we want for automated tests on the client

anything else you can think of?

calypsomatic:
> 👍
> 
> follow-on tickets that I see:
> 
> * investigate openapi-generator to see if it requires less customization/hacking
> * review/update the `operationId` values in the yaml definition, since those become method names in the client and we want client methods to be nicely named
> * gather feedback on the client and/or use it ourselves to assess model classes and other behavior
> * figure out what we want for automated tests on the client
> 
> anything else you can think of?

Just making a decision about the artifact group - `bio.terra` vs `org.databiosphere`

PR #39, opened 2022-08-01
AJ-482: WDS Single Tenant Relational Style

sagehen03:
Lots of code here and I'm open to suggestions on how to make this more digestible.  I implemented 3 "legacy" entity service methods since they include a full life cycle of insert, update, query and delete and demonstrate the overall feasibility of this model and together they can complement each other in tests.  

Sonar is flagging a good chunk of the SQL code due to the fact that we need to dynamically name tables and columns as per the user's writes to the system.  With a true single tenant model, it seems that SQL injection doesn't  represent a elevation in privileges (a user can use the API to read and write and there's no one else's data for them compromise) .  That being said it's good to have multiple layers of protection and we might want to use this approach for shared tenancy so it might be worthwhile to introduce an in memory cache of the user tables and their schemas and validate any generated SQL against that cache.  I'm not sure that Sonar will be smart enough not to flag in those instances where the SQL we're generating must match against some values in the cache, however.

I am not a fan of the `batchUpsert` API, but I implemented it to demonstrate we could support the existing functionality (including references).  I do think it should be replaced before too long.  

`entityQuery` does not currently stream, but the changes to do so are pretty small.  I can make that change as part of this PR, but I left it out for now since `entityQuery` pages and could easily enforce a row limit and there's already so much going on in this review.


src/main/java/org/databiosphere/workspacedataservice/DataTypeInferer.java
```diff
@@ -0,0 +1,114 @@
...
+            }
+        }
+        return result;
+    }
+
+    public DataTypeMapping selectBestType(DataTypeMapping existing, DataTypeMapping newMapping){
+        if(existing == newMapping){
+            return existing;
+        }
+        if(existing == LONG && newMapping == DOUBLE){
```
calypsomatic, line 59: Just checking - so if we already have a Double and the new mapping is a Long, we want to make it a String?

sagehen03: Good catch!  The code should keep the column type as a Double.  I'll make the change.

src/main/java/org/databiosphere/workspacedataservice/controller/SingleTenantEntityController.java
```diff
@@ -0,0 +1,197 @@
...
+            if(!singleTenantDao.entityTypeExists(workspaceId, entityType)){
+                createEntityTypeAndAddEntities(workspaceId, entityType, columnsModifiedSchema, upsertsForType);
+            } else {
+                //table exists, add columns if needed
+                List<Entity> entities = convertToEntities(upsertsForType, entityType, columnsModifiedSchema);
+                addOrUpdateColumnIfNeeded(workspaceId, entityType, columnsModifiedSchema,
+                        singleTenantDao.getExistingTableSchema(workspaceId, entityType), entities);
+                singleTenantDao.batchUpsert(workspaceId, entityType, entities, columnsModifiedSchema);            }
+        }
+        return new ResponseEntity<>(HttpStatus.NO_CONTENT);
```
calypsomatic, line 167: Do we not want to return the result or anything?

src/main/java/org/databiosphere/workspacedataservice/controller/SingleTenantEntityController.java
```diff
@@ -0,0 +1,197 @@
...
+        int entitiesInDbCount = 0;
+        for (String entityTypeName: entitiesByType.keySet()){
+            if(!singleTenantDao.entityTypeExists(workspaceId, entityTypeName)){
+                return new ResponseEntity<>("Entity type does not exist", HttpStatus.NOT_FOUND);
+            }
+            entitiesInDbCount += singleTenantDao.getEntityCount(entityTypeName,  entitiesByType.get(entityTypeName), workspaceId);
+        }
+
+        if(entitiesInDbCount != entitiesToDelete.size()){
+            return new ResponseEntity<>("Not all entities exist", HttpStatus.NOT_FOUND);
```
calypsomatic, line 184: Trying to understand this situation - it could happen if the user submits more entities to delete than are in the db, but also if they submit less.  If it's less, wouldn't we still want to delete those they asked for?

sagehen03: I believe this matches the behavior in existing entity service--if anything in the delete request doesn't exist, fail the whole request.  I'm amenable to changing things, but this API won't stick around so I don't we need to spend too much to time on this.

src/main/java/org/databiosphere/workspacedataservice/controller/SingleTenantEntityController.java
sagehen03: Yeah I think we will return more info with our eventual write APIs, but in order to have a steady target I just chose to match Rawls' existing behavior.

src/main/java/org/databiosphere/workspacedataservice/controller/SingleTenantEntityController.java
```diff
@@ -0,0 +1,197 @@
...
+        if(!singleTenantDao.entityTypeExists(workspaceId, entityType)) {
+            throw new ResponseStatusException(HttpStatus.NOT_FOUND, "Could not find entity type " + entityType);
+        }
+            Preconditions.checkArgument(Set.of("asc", "desc").contains(sortDirection.toLowerCase(Locale.ROOT)));
+        EntityQueryParameters queryParameters = new EntityQueryParameters(page, pageSize, sortField, sortDirection, filterTerms);
+        int totalEntityCount = singleTenantDao.getEntityCount(entityType, workspaceId);
+        Map<String, DataTypeMapping> schema = singleTenantDao.getExistingTableSchema(workspaceId, entityType);
+        int filteredEntityCount = StringUtils.isNotBlank(filterTerms) ? singleTenantDao.getFilteredEntityCount(workspaceId, entityType, filterTerms, schema) : totalEntityCount;
+        EntityQueryResultMetadata entityQueryResultMetadata = new EntityQueryResultMetadata(totalEntityCount, filteredEntityCount, (int) Math.ceil(filteredEntityCount / (double) pageSize));
+        return new EntityQueryResult(queryParameters, entityQueryResultMetadata, filteredEntityCount > 0 ? singleTenantDao.getSelectedEntities(entityType, pageSize,
```
calypsomatic, line 57: So the filteredEntityCount/totalEntityCount is so that we know where we are on the page?  Isn't it essentially doing the search twice then - querying for the entities to get the count, then again to actually get the result?  Three times, actually, because it also queries for the filtered count.  Are the count queries just cheap because there's no join or something?

sagehen03: It's a tired refrain at this point, but this code is meant to match Rawls' current behavior.  I don't necessarily agree with the approach it's taking, but I needed to pick a spec to implement against.  You're right that we could reduce the number of SQL queries from 3 to 2, but I think the potential savings would be cancelled by pulling so much more data across the network for large tables.

calypsomatic: I suspected it might be something like that!

src/main/java/org/databiosphere/workspacedataservice/dao/SingleTenantDao.java
```diff
@@ -0,0 +1,257 @@
...
+                Map.of("workspace", workspaceId.toString(), "tableName", tableName), (rs, rowNum) ->
+                        new SingleTenantEntityReference(rs.getString("column_name"), new EntityType(rs.getString("table_name"))));
+    }
+
+
+    private String genColUpsertUpdates(Set<String> cols) {
+        return cols.stream().filter(c -> !ENTITY_ID.getColumnName().equals(c)).map(c -> "\"" + c + "\"" + " = excluded.\"" + c + "\"").collect(Collectors.joining(", "));
+    }
+
+    private List<Object[]> getInsertBatchArgs(List<Entity> entities, Set<String> colNames) {
```
sagehen03, line 126: This code is admittedly brittle, relying on an implicit ordering of the column names.  It may be possible to convert to named parameters.

src/main/java/org/databiosphere/workspacedataservice/controller/SingleTenantEntityController.java
```diff
@@ -0,0 +1,197 @@
...
+                    throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "It looks like you're attempting to assign a reference " +
+                            "to a table, " + referencedEntityType + ", that does not exist");
+                }
+            }
+        }
+        if(!singleTenantDao.getReferenceCols(workspaceId, entityType).stream().map(SingleTenantEntityReference::getReferenceColName)
+                .collect(Collectors.toSet()).containsAll(newRefCols.keySet())){
+            throw new ResponseStatusException(HttpStatus.CONFLICT, "It looks like you're attempting to assign a reference " +
+                    "to an existing column that was not configured for references");
+        }
```
davidangb, line 131: Suggest moving this check up to the top and aborting if the request is invalid

src/main/java/org/databiosphere/workspacedataservice/dao/SingleTenantDao.java
```diff
@@ -0,0 +1,257 @@
...
+        String columnDefs = genColumnDefs(tableInfo);
+        namedTemplate.getJdbcTemplate().update("create table " + getQualifiedTableName(tableName, workspaceId) + "( " + columnDefs + ")");
+        for (SingleTenantEntityReference referencedEntityType : referencedEntityTypes) {
+            addForeignKeyForReference(tableName, referencedEntityType.getReferencedEntityType().getName(), workspaceId, referencedEntityType.getReferenceColName());
+        }
+    }
+
+    private String getQualifiedTableName(String entityType, UUID workspaceId){
+        return "\"" + workspaceId.toString() + "\".\"" + entityType + "\"";
+    }
```
davidangb, line 62: I think this effectively limits entity types to 63 chars, starting with a letter, etc, due to the restrictions on table names. We may run into issues with that.

davidangb (commented):
I've only made it through about half the PR but figured I'd submit comments so far - I will continue reviewing tomorrow!

src/main/java/org/databiosphere/workspacedataservice/DataTypeInferer.java
```diff
@@ -0,0 +1,117 @@
...
+
+
+    public DataTypeMapping inferType(Object val){
+        if(val instanceof Long || val instanceof Integer){
+            return LONG;
+        }
+
+        if(val instanceof Double || val instanceof Float){
+            return DOUBLE;
+        }
```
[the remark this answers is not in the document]
sagehen03: I'll add boolean. That will be quick.  Postgres as array data types so that should be doable too.  That will take a bit more time, but I'll familiarize myself with the existing operations on arrays and make sure we could support them fairly easily.

src/main/java/org/databiosphere/workspacedataservice/controller/SingleTenantEntityController.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    private Object convertToType(Object val, DataTypeMapping typeMapping) {
+        return switch (typeMapping){
+            case DATE -> LocalDate.parse(val.toString(), DateTimeFormatter.ISO_LOCAL_DATE);
+            case DATE_TIME -> LocalDateTime.parse(val.toString(), DateTimeFormatter.ISO_LOCAL_DATE_TIME);
+            case LONG, DOUBLE, STRING, JSON -> val;
+            case FOR_ATTRIBUTE_DEL -> throw new IllegalArgumentException("Egad, this should not happen");
+        };
+    }
```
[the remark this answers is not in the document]
sagehen03: Good call.

src/main/java/org/databiosphere/workspacedataservice/controller/SingleTenantEntityController.java
```diff
@@ -0,0 +1,197 @@
...
+                                              @RequestParam(defaultValue = "asc") String sortDirection,
+                                              @RequestParam(defaultValue = "") String filterTerms,
+                                              @RequestParam(required = false) List<String> fields) {
+        if(!singleTenantDao.entityTypeExists(workspaceId, entityType)) {
+            throw new ResponseStatusException(HttpStatus.NOT_FOUND, "Could not find entity type " + entityType);
+        }
+            Preconditions.checkArgument(Set.of("asc", "desc").contains(sortDirection.toLowerCase(Locale.ROOT)));
+        EntityQueryParameters queryParameters = new EntityQueryParameters(page, pageSize, sortField, sortDirection, filterTerms);
+        int totalEntityCount = singleTenantDao.getEntityCount(entityType, workspaceId);
+        Map<String, DataTypeMapping> schema = singleTenantDao.getExistingTableSchema(workspaceId, entityType);
```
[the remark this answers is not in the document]
sagehen03: The schema value is only used in the event that the client is using filterTerms in which case we need to construct an OR concatenation of all the fields.  So your instincts are correct that there's some unnecessary querying going on.  I'll see if it's easy to change.

src/main/java/org/databiosphere/workspacedataservice/controller/SingleTenantEntityController.java
sagehen03: I don't think that's possible all the code above this point populates variables that we need to check this condition.

src/main/java/org/databiosphere/workspacedataservice/dao/SingleTenantDao.java
sagehen03: You're right. I do see entity types that use guids in the existing system and there are some that are longer than 63 characters.  One possible solution would be moving away from information_schema immediately and having our own metadata table which could store a long name and an internal name.

src/main/java/org/databiosphere/workspacedataservice/DataTypeInferer.java
```diff
@@ -0,0 +1,117 @@
...
+            for (UpsertOperation removeAttr : removeAttrs) {
+                result.putIfAbsent(entityType, new LinkedHashMap<>());
+                Map<String, DataTypeMapping> attributeMappingForType = result.get(entityType);
+                //we need to have the attribute being removed present in the schema or else we won't generate
+                //the proper update
+                attributeMappingForType.put(removeAttr.getAttributeName(), FOR_ATTRIBUTE_DEL);
+            }
+        }
+        return result;
+    }
```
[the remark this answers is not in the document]
sagehen03: I don't think this API will stick around all that long.  I think we'd definitely want a file upload API that would be able to handle multi-gig uploads.  I experimented with something like for the first prototype and I think it's should be fairly straightforward so long as we constrain the API to just creating a populating new tables.

src/test/java/org/databiosphere/workspacedataservice/DataTypeInfererTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+    DataTypeInferer inferer = new DataTypeInferer();
+
+    @Test
+    void inferTypes() {
+        Map<String, LinkedHashMap<String, DataTypeMapping>> result = inferer.inferTypes(getSomeEntityUpserts());
+        assertThat(result.get("participants")).containsOnlyKeys("string_val", "int_val", "json_val", "date_val", "date_time_val");
+        assertThat(result.get("participants").values())
+                .containsExactlyInAnyOrder(DataTypeMapping.DATE, DataTypeMapping.JSON, DataTypeMapping.STRING, DataTypeMapping.LONG,
+                        DataTypeMapping.DATE_TIME);
```
davidangb, line 23: this test allows the possibility that the resultant data types don't match the expected input values, as long as they all exist in the result. How about:
```java
        LinkedHashMap<String, DataTypeMapping> expected = new LinkedHashMap<>();
        expected.put("string_val", DataTypeMapping.STRING);
        expected.put("int_val", DataTypeMapping.LONG);
        expected.put("json_val", DataTypeMapping.JSON);
        expected.put("date_val", DataTypeMapping.DATE);
        expected.put("date_time_val", DataTypeMapping.DATE_TIME);

        assertEquals(expected, result.get("participants"));
```

davidangb (commented):
a few more targeted comments inline.

Some higher-level thoughts:
1. I'm having a hard time knowing which code to review in this PR and which to gloss over - since it implements APIs and uses models that we know are going to go away, it gets confusing. So I may have commented on something that doesn't need it, or ignored something that should have had more attention.
2. I don't think we should change a column's datatype once it has been set in the schema, until/unless the user actively calls one of the (as yet undefined) schema-control APIs to explicitly change it. Comments welcome at https://docs.google.com/document/d/1d352ZoN5kEYWPjy0NqqWGxdf7HEu5VEdrLmiAv7dMmQ/edit#heading=h.vilx5halltu8 on this behavior, and we probably need Product-level opinion as well. Possibly we want to support multiple "schema enforcement" modes, on a per-request basis and/or as a setting for the WDS instance as a whole - we could choose between never inferring schema, only inferring schema for new columns, or changing schema for existing columns.
4. brainstorming/thinking out loud: I wonder if there's anything beneficial to inserting an entity's attributes into a temporary/staging table as jsonb, then calling `jsonb_populate_record` or similar to turn it into insertable column values …

PR #40, opened 2022-08-03
AJ-506 GH Action for publishing docker image

sagehen03:
- here we go?
- I should have secrets
- now I should definitely have secrets


calypsomatic (commented):
This looks good, just thinking about how to combine it with publishing to artifactory - first I would probably want to rename the publish.yml that pushes to artifactory to make it clear, then maybe we could have an action/workflow that does all the publishing?

calypsomatic (approved):

davidangb (commented):
two questions:
1. Should docker-image-publishing only trigger for pushes to `main`, or are you intentionally wanting to publish for a push to any branch?
2. Is it worth combining the GCR and ACR workflows so we only build and Trivy-scan the image once, but then publish it twice? Since these run on GHA runners in parallel, it's not that big a deal either way - I worry about any quantum fluctuation in which the built images are subtly different, but that _should_ never happen …

sagehen03:
> two questions:
> 
>     1. Should docker-image-publishing only trigger for pushes to `main`, or are you intentionally wanting to publish for a push to any branch?
> 
>     2. Is it worth combining the GCR and ACR workflows so we only build and Trivy-scan the image once, but then publish it twice? Since these run on GHA runners in parallel, it's not that big a deal either way - I worry about any quantum fluctuation in which the built images are subtly different, but that _should_ never happen …

I don't have strong opinions about any of these things, and I ended with actions for both Google and Azure kind of accident via the long and ridiculous troubleshooting process with devops.  What changes would you like to see?

davidangb:
if you're leaving it up to me, I suggest:
* collapse this into a single workflow that builds one and publishes to both GCR And ACR
* only run this workflow on push to `main` so that all published images are "approved" - if we find we need branch images later, we can always update

thanks!

davidangb (approved):
Thank you!

PR #44, opened 2022-08-08
BW-1346: Update swagger UI to make calls to relative paths

cjllanwarne:
Changes the swagger server definition to be relative to the hosted swagger page.

- Swagger is hosted at `.../swagger/swagger-ui.html`
- APIs are hosted at `.../<API path>

Therefore the relative server path is `../` to escape the `swagger/` URL segment.


calypsomatic (approved):

davidangb (approved):
same fix as #41!

PR #49, opened 2022-08-15
aj-496 replace reference map with string

calypsomatic:
[AJ-496](https://broadworkbench.atlassian.net/browse/AJ-496)
Instead of using a map with "entityType" and "entityName" keys, have references be accepted as simply a string beginning with an identifier, with entityType and entityName separated by a /.
Question: Presumably this convention will need to be documented somewhere.  Should that go in the openapi-docs.yaml?  In the README?  Both?  Neither?

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
@@ -49,8 +49,6 @@ public class EntityController {
     allAttrs.putAll(updatedAtts);
 
     Map<String, DataTypeMapping> typeMapping = inferer.inferTypes(updatedAtts);
-    // TODO: remove entityType/entityName JSON object format for references and move to URIs in the
-    // request/response payloads
     Map<String, DataTypeMapping> existingTableSchema =
         entityDao.getExistingTableSchema(instanceId, entityTypeName);
     singleEntity.setAttributes(new EntityAttributes(allAttrs));
@@ -59,7 +57,7 @@ public class EntityController {
         addOrUpdateColumnIfNeeded(
             instanceId, entityType.getName(), typeMapping, existingTableSchema, entities);
     try {
-      entityDao.batchUpsert(
+      convertRefsAndUpsert(
           instanceId, entityTypeName, entities, new LinkedHashMap<>(updatedSchema));
       EntityResponse response =
           new EntityResponse(
@@ -192,7 +190,7 @@ public class EntityController {
         LinkedHashMap<String, DataTypeMapping> combinedSchema =
             new LinkedHashMap<>(existingTableSchema);
         combinedSchema.putAll(requestSchema);
-        entityDao.batchUpsert(instanceId, entityTypeName, entities, combinedSchema);
+        convertRefsAndUpsert(instanceId, entityTypeName, entities, combinedSchema);
         return new ResponseEntity(response, HttpStatus.OK);
       }
     } catch (ResponseStatusException | InvalidEntityReference e) {
@@ -230,7 +228,7 @@ public class EntityController {
       List<Entity> entities = Collections.singletonList(newEntity);
       entityDao.createEntityType(
           instanceId, requestSchema, entityTypeName, RefUtils.findEntityReferences(entities));
-      entityDao.batchUpsert(
+      convertRefsAndUpsert(
           instanceId, entityTypeName, entities, new LinkedHashMap<>(requestSchema));
     } catch (MissingReferencedTableException e) {
       throw new ResponseStatusException(
@@ -239,5 +237,17 @@ public class EntityController {
               + "to a table that does not exist",
           e);
     }
+  }
+
+  private void convertRefsAndUpsert(UUID workspaceId,
+                                    String entityType,
+                                    List<Entity> entities,
+                                    LinkedHashMap<String, DataTypeMapping> schema) throws InvalidEntityReference{
+    for (Entity e: entities){
+      e.setAttributes(new EntityAttributes(e.getAttributes().getAttributes().entrySet().stream()
+              .map(entry -> RefUtils.isReferenceValue(entry.getValue()) ? new AbstractMap.SimpleEntry<>(entry.getKey(), RefUtils.getRefValue(entry.getValue())) : entry)
+              .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue))));
+    }
+    entityDao.batchUpsert(workspaceId, entityType, entities, schema);
   }
 }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
@@ -57,7 +57,7 @@ public class EntityController {
         addOrUpdateColumnIfNeeded(
             instanceId, entityType.getName(), typeMapping, existingTableSchema, entities);
     try {
-      convertRefsAndUpsert(
+      entityDao.batchUpsert(
           instanceId, entityTypeName, entities, new LinkedHashMap<>(updatedSchema));
       EntityResponse response =
           new EntityResponse(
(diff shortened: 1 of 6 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
```diff
@@ -310,16 +308,9 @@ private Map<String, Object> getAttributes(ResultSet rs) {
           if (systemCols.contains(columnName)) {
             continue;
           }
-          if (referenceColToTable.size() > 0 && referenceColToTable.containsKey(columnName)) {
```
calypsomatic, line 313: I think this can just be removed without any replacement, as `EntityController.convertRefsAndUpsert` makes sure the reference attributes are stored as the entityName only - but please correct me if I'm wrong!

service/src/main/java/org/databiosphere/workspacedataservice/service/RefUtils.java
```diff
@@ -13,11 +10,13 @@
 
 public class RefUtils {
 
+  public static final String REFERENCE_IDENTIFIER = "terra-wds:";
+
```
calypsomatic, line 14: The value of this reference identifier (and whether or not to include the : in it) is up for debate, I just put this as a default starting point.

service/src/main/java/org/databiosphere/workspacedataservice/service/RefUtils.java
```diff
@@ -33,32 +32,38 @@ public static Set<EntityReference> findEntityReferences(List<Entity> entities) {
...
+    if (obj != null){
+      String sVal = obj.toString();
+      int index = sVal.indexOf("/");
+      if (index >= 0){
+        return sVal.substring(index+1);
+      }
     }
-    throw new IllegalArgumentException("Expected {\"entityType\":<type>, \"entityName\":<name>}");
+    throw new IllegalArgumentException("Expected " + REFERENCE_IDENTIFIER + "<entityType>/<entityName>");
   }
```
sagehen03, line 54: I think your code works, but it might get tripped up on the edge case of an entity or entity type name containing a "/" (we may want to disallow that elsewhere).  Here's my suggestion which has less indentation and it throws an error if there's not exactly one "/" in the reference string.  
```
public static String getTypeValue(Object obj) {
    return splitReferenceString(obj)[0];
  }

  private static String[] splitReferenceString(Object obj) {
    String errorMessage = "Expected " + REFERENCE_IDENTIFIER + "<entityType>/<entityName>";
    Preconditions.checkNotNull(obj, errorMessage);
    String[] parts = obj.toString().split("/");
    Preconditions.checkArgument(parts.length == 2, errorMessage);
    return parts;
  }

  public static String getRefValue(Object obj) {
    return splitReferenceString(obj)[1];
  }
```

service/src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
sagehen03: It looks like it's still necessary.  The code accepts the new reference identifier format and will use to foreign keys to capture that relation, but upon read we don't show the new reference string, which I think we need to do?

service/src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
```diff
@@ -240,4 +238,16 @@ private void createEntityTypeAndInsertEntities(
...
+
+  private void convertRefsAndUpsert(UUID workspaceId,
+                                    String entityType,
+                                    List<Entity> entities,
+                                    LinkedHashMap<String, DataTypeMapping> schema) throws InvalidEntityReference{
+    for (Entity e: entities){
+      e.setAttributes(new EntityAttributes(e.getAttributes().getAttributes().entrySet().stream()
+              .map(entry -> RefUtils.isReferenceValue(entry.getValue()) ? new AbstractMap.SimpleEntry<>(entry.getKey(), RefUtils.getRefValue(entry.getValue())) : entry)
+              .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue))));
+    }
```
sagehen03, line 250: Mutating the entities with reference attributes makes me a little nervous.  I think I'm probably to blame for introducing code like this.  I think we should either make a copy of the entity list and mutate the cop or keep the logic to convert the reference representation in our serialization format in the dao and not actually change the entity attributes.  I think I'd vote for the latter.

service/src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
calypsomatic: Ah ok, I was assuming we only wanted to see the entityName string upon read.

service/src/main/java/org/databiosphere/workspacedataservice/service/RefUtils.java
calypsomatic: Yeah, I did a test in swagger and the UI there didn't accept "/" in the entityName but I didn't try with just curl or anything.

service/src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
calypsomatic: You mean, store "terra-wds:entityType/entityName" in the db, and only strip out the entityName whenever we display it anywhere?

service/src/main/java/org/databiosphere/workspacedataservice/dao/EntityDao.java
sagehen03: I'm actually not sure, but this doc seems to indicate we want the type name too.  https://docs.google.com/document/d/1kq0w8uTEiJLKRQ70fhQhZvwtuBAZ5cjGiCn0NRu7QLI/edit#heading=h.2s2k5de8ho5t

service/src/main/java/org/databiosphere/workspacedataservice/controller/EntityController.java
sagehen03: No, I mean in memory always have the attribute map always contain "terra-wds:entityType/entityName" for any reference fields and then upon write we only store entityName since the entityType information is capture in the foreign key.

calypsomatic: Ah ok well it seems like I got it doing that anyway!

sagehen03 (approved):

service/src/main/java/org/databiosphere/workspacedataservice/service/RefUtils.java
```diff
@@ -33,32 +34,32 @@ public static Set<EntityReference> findEntityReferences(List<Entity> entities) {
...
-    if (obj instanceof Map) {
-      Map map = (Map) obj;
-      return map.size() == 2 && map.keySet().containsAll(Set.of(ENTITY_TYPE_KEY, ENTITY_NAME_KEY));
-    }
-    return false;
+    return obj != null && obj.toString().startsWith(REFERENCE_IDENTIFIER);
+  }
+
+  public static String createReferenceString(String entityTypeName, String entityId){
+    return REFERENCE_IDENTIFIER + entityTypeName + "/" + entityId;
```
davidangb, line 63: for all the parsing and creation of these URIs, could you use a library instead of performing raw string parsing? I think you'd end up with something like this:
```java
  public static final String REFERENCE_IDENTIFIER = "terra-wds"; // note no colon at end

  private static String[] splitReferenceString(Object obj) {
    String errorMessage = "Expected " + REFERENCE_IDENTIFIER + ":<entityType>/<entityName>";

    // ensure we have a non-null string
    Preconditions.checkNotNull(obj, errorMessage);
    Preconditions.checkArgument(obj instanceof String, errorMessage);
    String ref = (String)obj;

    // parse the string as a uri
    UriComponents uric = UriComponentsBuilder.fromUriString(ref).build();

    // uri scheme should be "terra-wds"
    Preconditions.checkArgument(REFERENCE_IDENTIFIER.equals(uric.getScheme()), errorMessage);

    // record type is the first segment of the uri path;
    // record id is the second segment of the uri path
    List<String> pathSegments = uric.getPathSegments();
    Preconditions.checkArgument(pathSegments.size() == 2, errorMessage);

    return pathSegments.toArray(new String[0]); // or have this method return List<String> instead of array
  }

  public static String createReferenceString(String entityTypeName, String entityId){
    return UriComponentsBuilder.newInstance()
            .scheme(REFERENCE_IDENTIFIER)
            .pathSegment(entityTypeName, entityId)
            .build()
            .toUriString();
  }
```
With goals of 1) always wanting to create legal URIs, so callers can safely parse them, and 2) wanting our code to be as robust as possible, I think it's better to use a library here instead of writing our own.

davidangb (commented):
one comment …

davidangb (approved):
nice!

PR #50, opened 2022-08-16
switch google-java-format to eclipse format [AJ-470]

davidangb:
Trying to find a better spotless config, this switches from [google-java-format](https://github.com/google/google-java-format/wiki/FAQ#i-just-need-to-configure-it-a-bit-differently-how) to the eclipse predefined format.

This PR *_only_*  changes build.gradle, it does not reformat any existing code. Because of this, `spotlessCheck` will fail on this PR - the current code formatting has spots that are not legal with eclipse format. I've done this so reviewers can pull this branch, run `./gradlew spotlessApply` locally, and see the changes.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -50,7 +50,7 @@ idea.project.jdkName = '17'
 // Spotless configuration
 spotless {
     java {
-        palantirJavaFormat()
+        eclipse()
         target '**/*.java'
         targetExclude("${buildDir}/**", "**/generated/**")
     }
--- a/build.gradle
+++ b/build.gradle
@@ -50,7 +50,7 @@ idea.project.jdkName = '17'
 // Spotless configuration
 spotless {
     java {
-        googleJavaFormat()
+        palantirJavaFormat()
         target '**/*.java'
         targetExclude("${buildDir}/**", "**/generated/**")
     }


sagehen03 (commented):
It's better.  But what is the point of things like this:
```
private void createEntityTypeAndInsertEntities(
            UUID instanceId,
            EntityRequest entityRequest,
            String entityTypeName,
            Map<String, DataTypeMapping> requestSchema)
```

What's wrong with:

```
    private void createEntityTypeAndInsertEntities(UUID instanceId, EntityRequest entityRequest, String entityTypeName,
            Map<String, DataTypeMapping> requestSchema) throws InvalidEntityReference {
```
A heuristic that I often use is does the entire method fit on screen without scrolliing? If no, it might be too long.  One parameter per line makes this harder.  That being said I guess I could adjust my heuristic to look between the method's opening and closing brackets.

build.gradle
```diff
@@ -50,7 +50,7 @@ idea.project.jdkName = '17'
 // Spotless configuration
 spotless {
     java {
-        googleJavaFormat()
+        palantirJavaFormat()
```
sagehen03, line 53:
```suggestion
        eclipse()
```

sagehen03 (approved):
the eclipse formatter seems to do a better job of not creating a bunch of lines with only 10 characters of text on each.

calypsomatic (approved):

PR #52, opened 2022-08-17
Update tag.yml

calypsomatic:
Fixing residual problem from AJ-449:
My theory is that version changes are not being committed because 2 reviewers are required - [See here](https://github.com/DataBiosphere/terra-workspace-data-service/runs/7814628435?check_suite_focus=true#step:4:39) -
I believe using the BROADBOT token instead of GITHUB token should change that.  Was not apparent in testing because the reviews were not required for merging to non-main branch.

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -11,14 +11,14 @@ jobs:
       - name: Checkout current code
         uses: actions/checkout@v2
         with:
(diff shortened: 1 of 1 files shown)


sagehen03 (approved):
Let's try it!

davidangb (approved):

PR #54, opened 2022-08-17
schema API definitions [AJ-495]

davidangb:
To visualize these APIs, copy/paste the entire YAML into https://editor.swagger.io/.

Changes in this PR:
* APIs under the Schema heading are new or updated. These are the ones to review - do they look like they will serve our functional needs?
* added or changed the `operationId` for all APIs. These translate to method names in the generated Java client so please read them with your Java developer hat on.
* marked all not-yet-implemented APIs as deprecated. This is kinda funky, but it definitely helps to visualize via the swagger UI which APIs are currently working and which are not.

### DIFF
--- a/service/src/main/resources/static/swagger/openapi-docs.yaml
+++ b/service/src/main/resources/static/swagger/openapi-docs.yaml
@@ -27,7 +27,7 @@ paths:
   /{instanceid}/records/{v}/{type}/{id}:
     get:
       summary: Get record
-      operationId: getSingleRecord
+      operationId: getRecord
       description: Retrieves a single record by its type and id
       tags:
         - Records
@@ -55,6 +55,7 @@ paths:
         TODO: the type and id are duplicated in the url and the body; change this? Silently ignore the type/id in the body?
         
         TODO: add a query parameter to allow/disallow overwriting existing records?
+      operationId: createOrReplaceRecord
       tags:
         - Records
       parameters:
@@ -79,7 +80,7 @@ paths:
                 $ref: '#/components/schemas/ErrorResponse'
     patch:
       summary: Update record
-      operationId: addOrUpdateAttributes
+      operationId: updateRecord
       description: |
         Updates the record of the specified type and id.
         Any attributes included in the request body will be created or overwritten.
@@ -108,6 +109,7 @@ paths:
     delete:
       summary: Delete record (pending)
       description: Deletes the record at the specified type and id.
+      operationId: deleteRecord
       tags:
         - Records
       parameters:
@@ -115,75 +117,6 @@ paths:
         - $ref: '#/components/parameters/versionPathParam'
         - $ref: '#/components/parameters/recordTypePathParam'
         - $ref: '#/components/parameters/recordIdPathParam'
-      responses:
-        204:
-          description: Success
-  /{instanceid}/types/{v}/{type}:
-    patch:
-      summary: Update record type (pending)
-      description: |
-        Update record type. All records of the old type will be updated to the new type.
-        Returns the updated type definition.
-
-        v0.2 supports renaming a type; future versions may support additional functionality.
-      tags:
-        - Records
-      parameters:
-        - $ref: '#/components/parameters/instanceIdPathParam'
-        - $ref: '#/components/parameters/versionPathParam'
-        - $ref: '#/components/parameters/recordTypePathParam'
-      responses:
-        200:
-          description: Success
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/RecordTypeDefinition'
-    delete:
-      summary: Delete record type (pending)
-      description: Delete record type. All records of this type will be deleted.
-      tags:
-        - Records
-      parameters:
-        - $ref: '#/components/parameters/instanceIdPathParam'
-        - $ref: '#/components/parameters/versionPathParam'
-        - $ref: '#/components/parameters/recordTypePathParam'
-      responses:
-        204:
-          description: Success
-  /{instanceid}/types/{type}/attributes/{v}/{attribute}:
-    patch:
-      summary: Update attribute (pending)
-      description: |
-        Update attribute. All records containing the old attribute will now have the new attribute instead.
-        Returns the updated attribute definition.
-
-        v0.2 supports renaming an attribute within the specified type; future versions may support additional functionality.
-      tags:
-        - Records
-      parameters:
-        - $ref: '#/components/parameters/instanceIdPathParam'
-        - $ref: '#/components/parameters/versionPathParam'
-        - $ref: '#/components/parameters/recordTypePathParam'
-        - $ref: '#/components/parameters/attributePathParam'
-      responses:
-        200:
-          description: Success
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/RecordTypeDefinition'
-    delete:
-      summary: Delete attribute (pending)
-      description: |
-        Delete attribute. This deletes all values for this attribute within the specified type.
-      tags:
-        - Records
-      parameters:
-        - $ref: '#/components/parameters/instanceIdPathParam'
-        - $ref: '#/components/parameters/versionPathParam'
-        - $ref: '#/components/parameters/recordTypePathParam'
-        - $ref: '#/components/parameters/attributePathParam'
       responses:
         204:
           description: Success
@@ -196,6 +129,7 @@ paths:
         use the POST api instead.
 
         TODO: Eliminate this API in favor of POST?
+      operationId: basicQueryRecords
       tags:
         - Records
       parameters:
@@ -216,6 +150,7 @@ paths:
     post:
       summary: Query records (pending)
       description: Paginated list of records matching the criteria supplied in the request body
+      operationId: queryRecords
       tags:
         - Records
       parameters:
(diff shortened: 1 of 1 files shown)


service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -278,6 +219,134 @@ paths:
...
+        - $ref: '#/components/parameters/recordTypePathParam'
+      responses:
+        204:
+          description: Success
+  /{instanceid}/types/{type}/attributes/{v}/{attribute}:
+    get:
+      deprecated: true
+      summary: Describe attribute (pending)
+      description: |
+        Returns the schema definition for this attribute.
```
calypsomatic, line 289: This essentially just tells you the datatype?

davidangb: yup! Is there anything else you can think of that it should also return?

calypsomatic: I guess not.  Should the description make that clearer though?

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -278,6 +219,134 @@ paths:
           description: Success
         409:
           description: Conflict - schema already exists.
+  /{instanceid}/types/{v}/{type}:
+    get:
+      deprecated: true
+      summary: Describe record type (pending)
+      description: |
+        Returns the schema definition for this type, including all of its attributes.
```
sagehen03, line 227: What additional meaning does "including all of its attributes" convey on top of "Returns schema definition for this type"?  Just want to make sure I'm not missing something.

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -278,6 +219,134 @@ paths:
...
+                $ref: '#/components/schemas/RecordTypeSchema'
+    patch:
+      deprecated: true
+      summary: Update record type (pending)
+      description: |
+        Update record type. All records of the old type will be updated to the new type.
+        This API can be used to rename a record type and/or to perform batch updates to
+        this type's attributes. See also the updateAttribute API.
+        This API cannot be used to delete attributes; use deleteAttribute instead.
+        Returns the updated type definition.
```
sagehen03, line 250: So if this is analogous to PATCH for Records, then you can add a new attribute or change an existing one?  If that's correct, I think we'll need to handle errors in the event the request tries to change to a data type incompatible with the existing data (convert non-numeric text to integer e.g.).  I don't see any error responses listed yet.

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -278,6 +219,134 @@ paths:
...
+      tags:
+        - Schema
+      parameters:
+        - $ref: '#/components/parameters/instanceIdPathParam'
+        - $ref: '#/components/parameters/versionPathParam'
+        - $ref: '#/components/parameters/recordTypePathParam'
+      responses:
+        204:
+          description: Success
+  /{instanceid}/types/{type}/attributes/{v}/{attribute}:
```
sagehen03, line 284: Seems like overkill.  As a client, why not just get the types for all attributes in the record type and then pick the one you're interested in?  Less API surface seems better to me.

service/src/main/resources/static/swagger/openapi-docs.yaml
davidangb: and, added info about relations to this API as well.

I can imagine in the future this API would also return information about whether or not the attribute is nullable, or other information. I'd prefer to keep the API definition generic - that way we're not locked into a "get datatype" constraint

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -278,6 +219,134 @@ paths:
...
+            schema:
+              $ref: '#/components/schemas/AttributeSchema'
+      responses:
+        200:
+          description: Success
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AttributeSchema'
+    delete:
```
sagehen03, line 333: Do we need an error response code in the event that this attribute is referenced?

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -278,6 +219,134 @@ paths:
...
+        - $ref: '#/components/parameters/recordTypePathParam'
+        - $ref: '#/components/parameters/attributePathParam'
+      responses:
+        200:
+          description: Success
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AttributeSchema'
+    patch:
```
sagehen03, line 305: Similar to previous comment about GET, Can't we just use PATCH for the RecordType?

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -6,7 +6,7 @@ info:
   description: |
     Workspace Data Service API
 
-    This is the proposed API. Endpoints marked (pending) are not yet implemented.
+    This is the proposed API. Endpoints marked as deprecated are not yet implemented.
```
cjllanwarne, line 9: This seems like a slightly incongruous statement?

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -388,6 +457,22 @@ components:
...
+      required:
+        - name
+        - datatype
+      properties:
+        name:
+          type: string
+          description: name of this attribute.
+        datatype:
+          type: string
+          enum: [boolean, date, datetime, string, json, long, double, relation]
```
cjllanwarne, line 471: Is there a special use case for having a "json" type?

cjllanwarne, line 471: Thinking ahead warning: you may well end up wanting to make a recursive data type structure here for lists, objects, structs, etc. If you find yourselves banging your heads against `oneOf` in java's swagger codegen... I have recent experience and might be able to help.

cjllanwarne (commented):
Looks good to me as a first pass. A GET API that describes a record would be pretty convenient for providing some help when populating workflow input definitions in the UI.

I'll check in at our standup/tech talk tomorrow with the rest of the team in case they want to glance at this too

service/src/main/resources/static/swagger/openapi-docs.yaml
davidangb: That's just me being verbose and wordy and I shall strike "including all of its attributes" :)

service/src/main/resources/static/swagger/openapi-docs.yaml
davidangb: That's my thought, at this point. You're right, I haven't listed error responses! Agreed we'll have to handle lots of error cases here.

service/src/main/resources/static/swagger/openapi-docs.yaml
davidangb: I'm applying the "less surface area" argument to the calling user - if the user only wants to change a single attribute, it feels safer to call an API specific to that attribute instead of updating the parent type. But I agree that the type-schema APIs are a superset of the attribute-schema APIs, and we could release just with the type-schema APIs and only build the attribute ones when/if needed.

service/src/main/resources/static/swagger/openapi-docs.yaml
davidangb: we certainly will; I haven't attempted to list out error codes and should have made that clear in the PR description

service/src/main/resources/static/swagger/openapi-docs.yaml
davidangb: yeah, it's incongruous, but the result is a nice way for swagger-ui to visualize which APIs are active and working vs. those that are proposed but not yet implemented.

service/src/main/resources/static/swagger/openapi-docs.yaml
davidangb: Here, I'm mixing up current state and proposed state. This list of datatypes is what exists in WDS today, while the proposed list of datatypes is at https://docs.google.com/document/d/1d352ZoN5kEYWPjy0NqqWGxdf7HEu5VEdrLmiAv7dMmQ/edit#heading=h.naxag0augkgf - and the proposed list includes a "record" or "subrecord" type and array types instead of json. I should make this clearer.

davidangb: thank you! We have an investigation into `openapi-generator` in our future, and the `oneOf` is likely our primary driver.

service/src/main/resources/static/swagger/openapi-docs.yaml
sagehen03: Generally, I think it's better to wait for the client to ask.  I hear what you're saying about simplifying the API call for the caller, but the clients may have different ideas as to what's easy or preferred than we might so I'd say let's save ourselves work and wait until a client asks and then in that case we'll have pretty good assurances that our code will be used.  All that being said, I understand this all subjective and there's not a "right" answer.

sagehen03 (approved):

calypsomatic (approved):

PR #56, opened 2022-08-18
remove duplicate id and type for PUT/PATCH single record [AJ-538]

davidangb:
Removed the confusing duplication of id and type for PUT/PATCH.

Also renamed `Record.name` to `Record.id`, which caused some corresponding changes


calypsomatic (approved):
I'm especially excited about the `Record` from `RecordRequest` constructor, that's super handy

sagehen03 (approved):
Thanks getting some more of the old terminology migrated.

PR #57, opened 2022-08-22
AJ-528: don't depend on LinkedHashMap to guarantee stable iteration, use a list instead

sagehen03:
I'm dividing AJ-528 into smaller PRs.  With this PR, I originally intended to start using named sql parameters since those are usually easier to look at and reason about, but unfortunately there are a number of [constraints](https://stackoverflow.com/questions/422369/are-there-any-illegal-characters-when-using-named-parameters-in-jdbc) that spring jdbc imposes on named sql parameters that led me to abandon that effort.  I think we may want to consider limiting what we allow for record type and column names sooner rather than later or at least evaluate whether the same level of naming flexibility that we offer in Rawls is required.  

Anyhow, what was left in this PR was then switching away from using LinkedHashMap as a parameter type for the RecordDao#batchUpsert() method.  It's not exactly the most essential change, but it's good practice to use the collection interfaces rather than their implementations for methods.  I think we could eventually use List<RecordColumn> in place of Map<String, DataTypeMapping> elsewhere in the code, but for now I'm keeping things localized.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -1,6 +1,16 @@
 package org.databiosphere.workspacedataservice.dao;
 
-import static org.databiosphere.workspacedataservice.service.model.SystemColumn.RECORD_ID;
+import org.databiosphere.workspacedataservice.service.RelationUtils;
+import org.databiosphere.workspacedataservice.service.model.*;
+import org.databiosphere.workspacedataservice.shared.model.Record;
+import org.databiosphere.workspacedataservice.shared.model.*;
+import org.postgresql.util.PGobject;
+import org.springframework.dao.DataAccessException;
+import org.springframework.dao.EmptyResultDataAccessException;
+import org.springframework.jdbc.core.RowMapper;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Repository;
 
 import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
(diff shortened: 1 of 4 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -179,81 +173,74 @@ private Object getValueForSql(Object attVal) {
...
-	private class RecordRowMapper implements RowMapper<Record> {
-		private final String recordType;
-
-		private final Map<String, String> referenceColToTable;
-
-		private RecordRowMapper(String recordType, Map<String, String> referenceColToTable) {
-			this.recordType = recordType;
-			this.referenceColToTable = referenceColToTable;
-		}
+	private record RecordRowMapper(String recordType,
```
sagehen03, line 209: This was an intellij suggestion I noticed to remove a couple of lines of code.

calypsomatic (approved):
Having to make sure some things were `LinkedHashMap` was kinda a nuisance, so this is great!

davidangb:
looks like this has a failing unit test …

> we may want to consider limiting what we allow for record type and column names sooner rather than later or at least evaluate whether the same level of naming flexibility that we offer in Rawls is required.

do you know what the exact constraints are for postgres? The Rawls validation is relatively strict: https://github.com/broadinstitute/rawls/blob/40d4931dec2f82e6dba744ad200969b63a7bb7c0/model/src/main/scala/org/broadinstitute/dsde/rawls/StringValidationUtils.scala#L12-L18, though it does allow names to start with an underscore or dash, which could be a problem.

sagehen03:
> looks like this has a failing unit test …
> 
> > we may want to consider limiting what we allow for record type and column names sooner rather than later or at least evaluate whether the same level of naming flexibility that we offer in Rawls is required.
> 
> do you know what the exact constraints are for postgres? The Rawls validation is relatively strict: https://github.com/broadinstitute/rawls/blob/40d4931dec2f82e6dba744ad200969b63a7bb7c0/model/src/main/scala/org/broadinstitute/dsde/rawls/StringValidationUtils.scala#L12-L18, though it does allow names to start with an underscore or dash, which could be a problem.

I think in Postgres (though I haven't verified this systematically) that these rules apply when you don't use quotes https://www.postgresql.org/docs/7.0/syntax525.htm and table names must be 63 characters or less and column names must be 59 characters or less.  This differs from Rawls in two ways--as you note Rawls allows beginning with a number or dash and Rawls doesn't impose length restrictions.  Named parameters seem to adhere to the rules that apply when no quotes are used so dashes aren't allowed anywhere and you can't start with a number and you have shorter length restrictions.  Again, I haven't verified this systematically (the dashes were what gave me trouble in testing.  I've also noticed that when quoting there is a certain length after which postgres just substrings the first n characters of the column name.

davidangb (approved):
:+1: for the code!

Agreed about chatting with Product and maybe the rest of the team on naming restrictions. Possibly we could also scan the current prod attribute set to look for any non-compliant names … though that's an ugly and non-performant SQL investigation

sagehen03:
> +1 for the code!
> 
> Agreed about chatting with Product and maybe the rest of the team on naming restrictions. Possibly we could also scan the current prod attribute set to look for any non-compliant names … though that's an ugly and non-performant SQL investigation

I ran the queries to check for those attribute names and entity types longer than postgres' non-quoted restrictions and there were hundreds I believe, which is a comparatively small percentage of all the data.  The ones I spot checked used the workspace guid as a prefix.

PR #59, opened 2022-08-23
AJ-528: Add quoting function, Return Optional instead of null, Show All Attributes with PUT update path, New tables include FKs in create

sagehen03:
- In `RecordDao` Convert inline quoting of table and column names to method calls
- `RecordDao.getSingleRecord()` now returns an Optional
- PUT single record previously didn't include any nulled out attributes in its response payload (i.e. if PUT was used to update an existing record any existing attributes not included in the request payload would not show up in the PUT response body) This change should make the PUT response consistent with GET for the same Record, which seems correct to me, but let me know if I've misunderstood things.
- When creating a new record type, create foreign keys for references at table creation time.
- Sonar squawking about public test methods.


davidangb (approved):
lgtm, thank you!

PR #60, opened 2022-08-24
AJ-528: More consistent error handling, reserve sys_*, test improvements

sagehen03:
The big changes in this PR center around error handling.  I endeavored to ensure that all exceptions should now result in a standard error response that looks like: 
```
{
	"timestamp": "2022-08-24T15:13:56.207+00:00",
	"status": <http status error code>,
	"error": "<code description (Bad Request e.g.)>",
	"message": "<exception message>",
	"path": "<url /264dc567-d962-46a1-988a-a698d2cdff19/records/v0.21/samples/entity_3 e.g.>"
}
```
I didn't specifically pinpoint the location of this behavior in the spring code, but it seems like `ResponseStatusException` when thrown seem to work fully out of the box--they yield the expected message and `MockMvc` handles them the same as the full blown app server.  [this](https://github.com/spring-projects/spring-framework/issues/17290) seems to corroborate what I was seeing.

For the reference related exceptions in `RecordDao` where we trap and inspect Postgres level SQL inspections, it initially didn't feel right to use a `ResponseStatusException` so I kept the existing custom exceptions and annotated them `@ResponseStatus` then in order to get those annotated exceptions to work properly I needed to add `server.error.include-message=always` to `application.properties` and I needed to introduce a new test suite that uses the full web stack in order to verify behavior.  I'm not sure if all of those changes were actually worth it.  It might be better to just let `RecordDao` throw `ResponseStatusExceptions` since the dao is already coupled to spring web via the annotation so I'm not sure what I've done is an improvement, but I got curious to see if I could make it work.  Another alternative that would de-couple the dao code from web classes entirely is to go back to the way things were before this PR and have the dao throw checked exceptions and for the calling code to try catch around those calls--this leads to more indentation and uglier controller code.  It seems unlikely that we'd have a use case for using `RecordDao` apart from the web service so the coupling in exchange for cleaner controller code seems like a good tradeoff to me.

Finally, there's a change here to prohibit using the prefix "sys_" when naming attributes and some changes to parse json in tests and assert against those parsed values rather that just use string matching.


service/src/main/java/org/databiosphere/workspacedataservice/service/RelationUtils.java
```diff
@@ -75,8 +75,8 @@ public static boolean isRelationValue(Object obj) {
 		return obj != null && obj.toString().startsWith(RELATION_IDENTIFIER);
 	}
 
-	public static String createRelationString(String entityTypeName, String entityId) {
-		return UriComponentsBuilder.newInstance().scheme(RELATION_IDENTIFIER).pathSegment(entityTypeName, entityId)
+	public static String createRelationString(String recordTypeName, String recordId) {
```
sagehen03, line 78: Missed these during entity->record PR.

service/src/main/resources/application.properties
```diff
@@ -2,3 +2,6 @@ spring.datasource.driver-class-name=org.postgresql.Driver
 spring.datasource.url=jdbc:postgresql://${WDS_DB_HOST:localhost}:${WDS_DB_PORT:5432}/wds
 spring.datasource.username=${WDS_DB_USER:wds}
 spring.datasource.password=${WDS_DB_PASSWORD:wds}
+
+server.error.include-stacktrace=never
```
sagehen03, line 6: With single tenancy I think this is less of a given than with multi-tenancy (don't give prying eyes more info that necessary), but I went ahead and behaved as I would with a multi-tenant system.

sagehen03:
Just realized I need to make these error responses consistent with the swagger.  Doing that now!

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -46,22 +50,20 @@ public ResponseEntity<RecordResponse> updateSingleRecord(@PathVariable("instance
...
+		return new ResponseEntity<>(response, HttpStatus.OK);
 	}
 
 	private Map<String, DataTypeMapping> addOrUpdateColumnIfNeeded(UUID instanceId, String recordType,
 			Map<String, DataTypeMapping> schema, Map<String, DataTypeMapping> existingTableSchema,
 			List<Record> records) {
 		MapDifference<String, DataTypeMapping> difference = Maps.difference(existingTableSchema, schema);
 		Map<String, DataTypeMapping> colsToAdd = difference.entriesOnlyOnRight();
+		colsToAdd.keySet().stream().filter(s -> s.startsWith(RESERVED_NAME_PREFIX)).findAny().ifPresent(s -> {
+			throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Attribute names can't begin with " + RESERVED_NAME_PREFIX);
```
calypsomatic, line 65: I think we had a discussion somewhere about reserving "sys_" other places than attributes (or maybe I hallucinated it) - did we decide to allow it for record types and and ids?

davidangb: I think we should disallow "sys_" for types - this will allow future flexibility for WDS to create tables inside the instance schema, if we need to do so. I can imagine we might need tables for things like record version history, etc.

sagehen03: Sure thing, I'll make that change to this PR.

sagehen03:
> Just realized I need to make these error responses consistent with the swagger. Doing that now!

This is done; though admittedly I am not a swagger pro.

service/src/test/java/org/databiosphere/workspacedataservice/controller/FullStackRecordControllerTest.java
```diff
@@ -0,0 +1,110 @@
...
+        Map<String, Object> attrs = new HashMap<>();
+        attrs.put("attr_ref", RelationUtils.createRelationString("non_existent", "recordId"));
+        attrs.put("attr_ref_2", RelationUtils.createRelationString("non_existent_2", "recordId"));
+        HttpEntity<String> requestEntity = new HttpEntity<>(mapper.writeValueAsString(new RecordRequest(new RecordAttributes(attrs))), headers);
+        ResponseEntity<LinkedHashMap> response = restTemplate.exchange("/{instanceId}/records/{version}/{recordType}/{recordId}",
+                HttpMethod.PUT, requestEntity, LinkedHashMap.class, instanceId, versionId, "samples-1", "sample_1");
+        LinkedHashMap responseContent = response.getBody();
+        assertThat(responseContent).containsEntry("message", "Referenced table(s) [non_existent_2, non_existent] do(es) not exist");
+        assertThat(response.getStatusCode()).isEqualTo(HttpStatus.BAD_REQUEST);
+    }
```
davidangb, line 61: for all of these tests, it would be super cool to use the generated `ErrorResponse` class, like:
```java
ResponseEntity<ErrorResponse> response = restTemplate.exchange(
    "/{instanceId}/records/{version}/{recordType}/{recordId}", HttpMethod.PUT, requestEntity,
    ErrorResponse.class, instanceId, versionId, "samples-1", "sample_1");
ErrorResponse responseContent = response.getBody();
assertThat(responseContent.getMessage()).isEqualTo(
    "Referenced table(s) [non_existent_2, non_existent] do(es) not exist");
```
It does look like there are classpath issues preventing this and would require some build.gradle tweaking - maybe for a follow-on PR?

davidangb (approved):
looks good, one optional comment inline which is probably for a follow-on PR.

Could you run `spotlessApply` (or `build`) and commit the results? Else anytime we build locally, we'll have outstanding changes.

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -622,6 +636,9 @@ components:
         path:
           type: string
           description: requested record path
+        timestamp:
+          type: string
+          description: time of error
```
davidangb, line 641: oh one last comment - if we are not actually returning `trace` anytime, could you remove that property from this model?

calypsomatic (approved):

PR #61, opened 2022-08-24
AJ-559 don't create column for invalid relation

calypsomatic:
[AJ-559 Inserting relation to non-existent record creates relation attribute](https://broadworkbench.atlassian.net/browse/AJ-559)
If creating relation fails due to the related table not existing, the column for that relation should not be created. Included the foreign key constraint when initially creating column so we can catch the sql error at the same time.
I ran `build` so there's a number of changes that are just whitespace, hopefully it's not too disruptive.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -176,10 +176,6 @@ public class RecordDao {
 				.collect(Collectors.joining(", \n"));
 	}
 
-	public String getSingleFkSql(Relation relation) {
-		return " references " + relation.relationRecordType().getName() + "(" + RECORD_ID.getColumnName() + ")";
-	}
-
 	public List<Relation> getRelationCols(UUID instanceId, String tableName) {
 		return namedTemplate.query(
 				"SELECT kcu.column_name, ccu.table_name FROM information_schema.table_constraints tc JOIN information_schema.key_column_usage kcu "
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -88,6 +88,11 @@ public class RecordDao {
 				+ " add column " + quote(columnName) + " " + colType.getPostgresType());
 	}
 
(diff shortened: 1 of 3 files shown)


davidangb (commented):
instead of manually writing rollback `removeColumn` code for this, could the relevant statements be wrapped in a transaction, which would automatically roll back?

calypsomatic:
> instead of manually writing rollback `removeColumn` code for this, could the relevant statements be wrapped in a transaction, which would automatically roll back?

Would probably work better too!  I'll give it a try

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -212,11 +230,8 @@ private String genInsertStatement(UUID instanceId, String recordType, List<Recor
 		List<String> colNames = schema.stream().map(RecordColumn::colName).toList();
 		List<DataTypeMapping> colTypes = schema.stream().map(RecordColumn::typeMapping).toList();
 		return "insert into " + getQualifiedTableName(recordType, instanceId) + "(" + getInsertColList(colNames)
-				+ ") values (" + getInsertParamList(colTypes) + ") " + "on conflict ("
-				+ RECORD_ID.getColumnName() + ") "
-				+ (schema.size() == 1
-						? "do nothing"
-						: "do update set " + genColUpsertUpdates(colNames));
+				+ ") values (" + getInsertParamList(colTypes) + ") " + "on conflict (" + RECORD_ID.getColumnName()
+				+ ") " + (schema.size() == 1 ? "do nothing" : "do update set " + genColUpsertUpdates(colNames));
```
davidangb, line 234: thinking out loud: for these longer, multi-line strings, I wonder if it would be easier to use [Text Blocks](https://www.baeldung.com/java-text-blocks) and `String.format()` to add the variables? I don't think it matters much for the _execution_ of the code; this would purely be about whatever is more readable for WDS developers.

davidangb (approved):
lgtm, thank you - any SQL operations we can combine into a single statement instead of multiple is good!

jordansafer (approved):

gdlex4015:
Are SonarCloud failures ok to be ignored?

calypsomatic:
> Are SonarCloud failures ok to be ignored?

These are not new failures; they just showed up because I ran auto-formatting and moved some of the code around the SQL.

PR #62, opened 2022-08-29
Aj 562 delete single record

calypsomatic:
(AJ-562 Delete single record API)[https://broadworkbench.atlassian.net/browse/AJ-562]
Adds API to delete a single record.


service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -161,6 +161,20 @@ public ResponseEntity<String> createInstance(@PathVariable("instanceId") UUID in
...
+	@DeleteMapping("/{instanceId}/records/{version}/{recordType}/{recordId}")
+	public ResponseEntity deleteSingleRecord(@PathVariable("instanceId") UUID instanceId,
+			@PathVariable("version") String version, @PathVariable("recordType") RecordType recordType,
+			@PathVariable("recordId") RecordId recordId) {
+		validateVersion(version);
+		String recordTypeName = recordType.getName();
+		recordDao
+				.getSingleRecord(instanceId, recordType, recordId,
+						recordDao.getRelationCols(instanceId, recordTypeName))
+				.orElseThrow(() -> new ResponseStatusException(HttpStatus.NOT_FOUND, "Record not found"));
```
calypsomatic, line 173: Should we care if the record does not exist, or just proceed as normal since the end result is the same?

davidangb: I do think it's worth returning 404 if the record-to-be-deleted doesn't exist. That said, could you have `recordDao.deleteSingleRecord` return the count of rows it deleted, and throw 404 if it deleted zero rows?

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -144,6 +150,18 @@ public void batchUpsert(UUID instanceId, String recordType, List<Record> records
 		}
 	}
 
+	public void deleteSingleRecord(UUID instanceId, String recordType, String recordId) {
+		try {
+			namedTemplate.getJdbcTemplate().batchUpdate("delete from " + getQualifiedTableName(recordType, instanceId)
+					+ " where " + RECORD_ID + "= " + singleQuote(recordId));
```
davidangb, line 156: for the `recordId`, please use a bind parameter instead of building a string

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -172,6 +190,13 @@ private void checkForMissingRecord(SQLException sqlEx) {
 		}
 	}
 
+	private void checkForTableRelation(SQLException sqlEx) {
+		if (sqlEx != null && sqlEx.getSQLState() != null && sqlEx.getSQLState().equals("23503")) {
+			throw new IllegalDeletionException(
+					"Unable to delete this record because it has a relation with another record.");
```
davidangb, line 196: nit: I think "Unable to delete this record because another record has a relation to it." is more accurate

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -144,6 +150,18 @@ public void batchUpsert(UUID instanceId, String recordType, List<Record> records
...
+			namedTemplate.getJdbcTemplate().batchUpdate("delete from " + getQualifiedTableName(recordType, instanceId)
+					+ " where " + RECORD_ID + "= " + singleQuote(recordId));
+		} catch (DataIntegrityViolationException e) {
+			if (e.getRootCause()instanceof SQLException sqlEx) {
+				checkForTableRelation(sqlEx);
+				throw e;
+			}
+		}
+	}
+
```
sagehen03, line 164:
```suggestion
	public boolean deleteSingleRecord(UUID instanceId, String recordType, String recordId) {
		try {
			return namedTemplate.update("delete from " + getQualifiedTableName(recordType, instanceId)
					+ " where " + RECORD_ID + " = :recordId", new MapSqlParameterSource("recordId", recordId)) == 1;
		} catch (DataIntegrityViolationException e) {
			if (e.getRootCause()instanceof SQLException sqlEx) {
				checkForTableRelation(sqlEx);
			}
			throw e;
		}
	}

```
This approach switches to named SQL parameters (Sonar will be happier), eliminates the need for the singleQuote function and won't silently capture a DataIntegrityViolationException and it will return true when the record was found and deleted, false if no records were affected (aka record was not found).

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -161,6 +161,20 @@ public ResponseEntity<String> createInstance(@PathVariable("instanceId") UUID in
...
+		recordDao.deleteSingleRecord(instanceId, recordType.getName(), recordId.getRecordIdentifier());
+		return new ResponseEntity<>(HttpStatus.NO_CONTENT);
+	}
+
```
sagehen03, line 177:
```suggestion
	@DeleteMapping("/{instanceId}/records/{version}/{recordType}/{recordId}")
	public ResponseEntity deleteSingleRecord(@PathVariable("instanceId") UUID instanceId,
			@PathVariable("version") String version, @PathVariable("recordType") RecordType recordType,
			@PathVariable("recordId") RecordId recordId) {
		validateVersion(version);
		boolean recordFound = recordDao.deleteSingleRecord(instanceId, recordType.getName(), recordId.getRecordIdentifier());
		return recordFound ? new ResponseEntity<>(HttpStatus.NO_CONTENT) : new ResponseEntity<>(HttpStatus.NOT_FOUND);
	}
```
If you elect to make my suggested changes in the DAO, this is what shakes out in the controller.

sagehen03 (commented):
Looks pretty good, and I think the code works, but I think we can accomplish this with a little less code by using NamedParameterJdbcTemplate.update() Let me know what you think.

davidangb (approved):
thank you!

sagehen03 (approved):
It looks like sonar might be complaining about not using a type parameter along with ResponseEntity.  Here's what SO thinks about the issue: https://stackoverflow.com/questions/26550124/spring-returning-empty-http-responses-with-responseentityvoid-doesnt-work switching the ResponseEntity<String> or ResponseEntity<Void> will likely quiet Sonar, but probably not worth worrying about.

PR #65, opened 2022-09-06
aj-498 standard exceptions

calypsomatic:
Not much further work to be done; I added a page in the wiki and linked there from contributing.  Made one new exception class and it looks like we're already throwing `ResponseStatusException` anywhere a custom class is not used.


service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -37,6 +38,10 @@ public ResponseEntity<RecordResponse> updateSingleRecord(@PathVariable("instance
 			@PathVariable("recordId") RecordId recordId, @RequestBody RecordRequest recordRequest) {
 		validateVersion(version);
 		String recordTypeName = recordType.getName();
+		if (!recordDao.recordTypeExists(instanceId, recordTypeName)) {
+			throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
+					"Record type " + recordTypeName + " does not exist");
+		}
```
davidangb, line 44: could we do away with throwing `ResponseStatusException` and always throw a custom-defined exception?

Secondarily, we should try not to echo the user input in the error response (though we could log it), since echoing user input is a vector for XSS once we are in a UI.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -96,7 +96,7 @@ void updateWithIllegalAttributeName() throws Exception {
 						.content(mapper.writeValueAsString(new RecordRequest(new RecordAttributes(illegalAttribute)))))
 				.andExpect(status().isBadRequest())
 				.andExpect(result -> assertEquals("Attribute names can't begin with sys_",
-						result.getResponse().getErrorMessage()));
+						result.getResolvedException().getMessage()));
```
davidangb, line 99: if we use distinct exception classes for these two error conditions, then you could check the class type here instead of reading the message.

davidangb (commented):
see comments inline … what do you think about getting rid of `ResponseStatusException` everywhere?

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
calypsomatic: I was halfway to getting rid of `ResponseStatusException`, just didn't quite commit to it.  Will happily do so.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -105,12 +106,12 @@ public ResponseEntity<RecordResponse> getSingleRecord(@PathVariable("instanceId"
 		validateVersion(version);
 		if (!recordDao.instanceSchemaExists(instanceId)
 				|| !recordDao.recordTypeExists(instanceId, recordType.getName())) {
-			throw new ResponseStatusException(HttpStatus.NOT_FOUND, "Instance or table don't exist");
+			throw new MissingRecordTypeException("Instance or record doesn't exist");
```
calypsomatic, line 109: This semi-qualifies as a `MissingRecordTypeException`, not sure if we should just use that one or a new custom exception, and/or distinguish between instance and record type not existing.   Plus I see I left out "type", will get that fixed.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -155,7 +156,7 @@ public ResponseEntity<String> createInstance(@PathVariable("instanceId") UUID in
 			@PathVariable("version") String version) {
 		validateVersion(version);
 		if (recordDao.instanceSchemaExists(instanceId)) {
-			throw new ResponseStatusException(HttpStatus.CONFLICT, "This schema already exists");
+			throw new IllegalInstanceCreationException("This instance already exists");
```
calypsomatic, line 159: Changed to "instance" since "schema" means something else to the user.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -173,15 +174,14 @@ public ResponseEntity deleteSingleRecord(@PathVariable("instanceId") UUID instan
 
 	private static void validateVersion(String version) {
 		if (null == version || !version.equals("v0.2")) {
-			throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Invalid API version specified");
+			throw new InvalidAPIException("Invalid API version specified");
```
calypsomatic, line 177: For this and the previous exception (`IllegalInstanceCreationException`), not sure how much we can expect them to be needed in other circumstances, so maybe we could have a more generic exception?

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -180,8 +184,7 @@ private static void validateVersion(String version) {
 	private void createRecordTypeAndInsertRecords(UUID instanceId, Record newRecord, String recordTypeName,
 			Map<String, DataTypeMapping> requestSchema) {
 		if (recordTypeName.startsWith(RESERVED_NAME_PREFIX)) {
-			throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
-					"Record types can't start with " + RESERVED_NAME_PREFIX);
+			throw new InvalidNameException("Record types can't start with " + RESERVED_NAME_PREFIX);
```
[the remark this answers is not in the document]
sagehen03: > I would prefer to see separate `InvalidRecordTypeException` and `InvalidAttributeNameException` classes. If we have separate classes for these two error conditions, then we could move the `"Attribute names can't begin with " + RESERVED_NAME_PREFIX` and `"Record types can't start with " + RESERVED_NAME_PREFIX` verbiage into the exception classes themselves and out of this controller.
> 
> I expect that as we build out the batch and schema APIs, we'll find more uses for these exceptions.

If it was up to me, I would wait until we found those uses for these new exceptions before adding them, but I agree that having a different exception to match differing conditions as opposed to just distinct message strings is a win.

sagehen03 (approved):
I think we're adding code without commensurate benefit to our users or ourselves, but it's pretty subjective and I'll go along with the majority will.

davidangb (commented):
I do like the approach of having distinct exceptions thrown for our various use cases.

On passing the error message from the callier into the exception vs. coding the message into the exception class: imho the only reason to pass the message from caller is if the message needs to be dynamic. In all cases where the message is the same every time we throw it, we should code it into the exception class itself - and can do so right in the `@ResponseStatus` annotation, like so:
```java
@ResponseStatus(code = HttpStatus.BAD_REQUEST, reason = "Record types can't start with "
		+ ReservedNames.RESERVED_NAME_PREFIX)
public class InvalidRecordTypeException extends RuntimeException {
}
```

Then, we have a centralized place for when Product (or others) requests verbiage changes for error messages - and in our unit tests we can check for `instanceof` instead of the string message, which means if we change the error message we don't need to care about changing the unit tests.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -51,7 +53,7 @@ void createInstanceAndTryToCreateAgain() throws Exception {
 		UUID uuid = UUID.randomUUID();
 		mockMvc.perform(post("/{instanceId}/{version}/", uuid, versionId)).andExpect(status().isCreated());
 		mockMvc.perform(post("/{instanceId}/{version}/", uuid, versionId)).andExpect(status().isConflict()).andExpect(
-				result -> assertEquals("This schema already exists", result.getResponse().getErrorMessage()));
+				result -> assertEquals("This instance already exists", result.getResolvedException().getMessage()));
```
davidangb, line 56: use `instanceof` instead of looking at the message:
```suggestion
result -> assertTrue(result.getResolvedException() instanceof IllegalInstanceCreationException));
```

davidangb (commented):
could you keep going and bring `IllegalDeletionException` and `IllegalInstanceCreationException` in line with the other exceptions, by moving the error message into the exception? 

That only leaves two exceptions which are inconsistent:
* `InvalidRelationException`, has two different error messages passed in from the caller. Should this be two separate exceptions?
* `MissingReferencedTableException`, which outputs the offending table name. Should this _not_ echo the user input in the content of the error message?

I'd be fine tackling `InvalidRelationException` and `MissingReferencedTableException` in a follow-on PR because they may need more discussion/thought, unless you'd rather tackle them as part of this one.

calypsomatic:
> could you keep going and bring `IllegalDeletionException` and `IllegalInstanceCreationException` in line with the other exceptions, by moving the error message into the exception?

I left `IllegalDeletionException` and `IllegalInstanceCreationException` the way they are on the theory that there might potentially be other reasons in the future why it's illegal to delete a record or create an instance.  But we can deal with those situations when or if they arise.

davidangb (approved):
thanks for all the iteration on this!

sagehen03:
> could you keep going and bring `IllegalDeletionException` and `IllegalInstanceCreationException` in line with the other exceptions, by moving the error message into the exception?
> 
> That only leaves two exceptions which are inconsistent:
> 
>     * `InvalidRelationException`, has two different error messages passed in from the caller. Should this be two separate exceptions?
> 
>     * `MissingReferencedTableException`, which outputs the offending table name. Should this _not_ echo the user input in the content of the error message?
> 
> 
> I'd be fine tackling `InvalidRelationException` and `MissingReferencedTableException` in a follow-on PR because they may need more discussion/thought, unless you'd rather tackle them as part of this one.

It seems like we're moving towards having one type of message per exception.  I would argue that's not the intention of most Exceptions otherwise they wouldn't expose the `message` field and would instead expose only the part of the message that changes (record type mostly in our case).  I do think it we were to ever use this data access code outside of a controller this work is useful, but my philosophy is generally we're bad at anticipating the future so let's not make things more complicated right now so as to make something that may come to pass easier if it does in fact come to pass.

service/src/main/java/org/databiosphere/workspacedataservice/service/model/exception/IllegalDeletionException.java
```diff
@@ -3,11 +3,6 @@
 import org.springframework.http.HttpStatus;
 import org.springframework.web.bind.annotation.ResponseStatus;
 
-@ResponseStatus(HttpStatus.BAD_REQUEST)
+@ResponseStatus(code = HttpStatus.BAD_REQUEST, reason = "Unable to delete this record because another record has a relation to it.")
```
jordansafer, line 6: I think this status should be HttpStatus.FORBIDDEN, since the operation is prevented because of the relation: https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_client_errors

jordansafer (approved):
After reviewing the new exception types, I agree, for the purposes of cleaner testing that we could add a few exception types and remove the ResponseStatusException()

jordansafer (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -167,15 +166,15 @@ public void addForeignKeyForReference(String recordType, String referencedRecord
...
 			}
 			throw e;
 		}
 	}
 
-	private void checkForMissingTable(SQLException sqlEx, String... missingTableName) {
+	private void checkForMissingTable(SQLException sqlEx) {
 		if (sqlEx != null && sqlEx.getSQLState() != null && sqlEx.getSQLState().equals("42P01")) {
-			throw new MissingReferencedTableException(missingTableName);
+			throw new MissingObjectException("Referenced record type");
```
davidangb, line 177: nit: for potentially user-facing messages, let's use "relation" instead of "reference"

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -83,9 +87,8 @@ private Map<String, DataTypeMapping> addOrUpdateColumnIfNeeded(UUID instanceId,
 		}
 		if (!recordDao.getRelationCols(instanceId, recordType).stream().map(Relation::relationColName)
 				.collect(Collectors.toSet()).containsAll(newRefCols.keySet())) {
-			throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
-					"It looks like you're attempting to assign a relation "
-							+ "to an existing column that was not configured for relations");
+			throw new InvalidRelationException("It looks like you're attempting to assign a relation "
+					+ "to an existing column that was not configured for relations");
```
davidangb, line 91: nit: use "attribute" instead of "column" for user-facing messages (I know this verbiage is not part of your PR)

davidangb (commented):
two verbiage nits inline. I also notice that `MissingReferencedTableException` is now unused - delete that class if we don't need it?

davidangb (approved):
thanks for sticking with this!

gdlex4015:
Are SonarCloud warnings bogus?

PR #66, opened 2022-09-07
Aj 566 check relation attributes when updating columns

calypsomatic:
AJ-566
Compares new and old relation attributes to make sure different record types are not being assigned to one attribute. Previously only checked new relations.


service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -67,12 +67,11 @@ private Map<String, DataTypeMapping> addOrUpdateColumnIfNeeded(UUID instanceId,
 					"Attribute names can't begin with " + RESERVED_NAME_PREFIX);
 		});
 		Set<Relation> relations = RelationUtils.findRelations(records);
```
davidangb, line 69: re: not calling `recordDao.getRelationCols()` twice … could you get it once at the beginning and then modify its result during the phase of adding new columns? Like so:

```suggestion
		List<Relation> existingRelationCols = recordDao.getRelationCols(instanceId, recordType);
		Set<Relation> relations = RelationUtils.findRelations(records);
		relations.addAll(existingRelationCols);
		Map<String, List<Relation>> newRefCols = relations.stream()
				.collect(Collectors.groupingBy(Relation::relationColName));
		Preconditions.checkArgument(newRefCols.values().stream().filter(l -> l.size() > 1).findAny().isEmpty(),
				"Relation attribute can only be assigned to one record type");
		for (String col : colsToAdd.keySet()) {
			String referencedRecordType = null;
			if (newRefCols.containsKey(col)) {
				referencedRecordType = newRefCols.get(col).get(0).relationRecordType().getName();
				existingRelationCols.add(new Relation(col, new RecordType(referencedRecordType)));
			}
			recordDao.addColumn(instanceId, recordType, col, colsToAdd.get(col), referencedRecordType);
			schema.put(col, colsToAdd.get(col));
		}
		if (!existingRelationCols.stream().map(Relation::relationColName)
				.collect(Collectors.toSet()).containsAll(newRefCols.keySet())) {

```

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -67,12 +67,11 @@ private Map<String, DataTypeMapping> addOrUpdateColumnIfNeeded(UUID instanceId,
...
 				.collect(Collectors.groupingBy(Relation::relationColName));
-		// TODO: better communicate to the user that they're trying to assign multiple
-		// record types to a
-		// single column
-		Preconditions.checkArgument(newRefCols.values().stream().filter(l -> l.size() > 1).findAny().isEmpty());
+		Preconditions.checkArgument(newRefCols.values().stream().filter(l -> l.size() > 1).findAny().isEmpty(),
+				"Relation attribute can only be assigned to one record type");
 		for (String col : colsToAdd.keySet()) {
 			String referencedRecordType = null;
 			if (newRefCols.containsKey(col)) {
```
davidangb, line 77: re: not calling `recordDao.getRelationCols()` twice … could you get it once at the beginning and then modify its result during the phase of adding new columns? Like so:

```suggestion
		List<Relation> existingRelationCols = recordDao.getRelationCols(instanceId, recordType);
		Set<Relation> relations = RelationUtils.findRelations(records);
		relations.addAll(existingRelationCols);
		Map<String, List<Relation>> newRefCols = relations.stream()
				.collect(Collectors.groupingBy(Relation::relationColName));
		Preconditions.checkArgument(newRefCols.values().stream().filter(l -> l.size() > 1).findAny().isEmpty(),
				"Relation attribute can only be assigned to one record type");
		for (String col : colsToAdd.keySet()) {
			String referencedRecordType = null;
			if (newRefCols.containsKey(col)) {
				referencedRecordType = newRefCols.get(col).get(0).relationRecordType().getName();
				existingRelationCols.add(new Relation(col, new RecordType(referencedRecordType)));
			}
			recordDao.addColumn(instanceId, recordType, col, colsToAdd.get(col), referencedRecordType);
			schema.put(col, colsToAdd.get(col));
		}
		if (!existingRelationCols.stream().map(Relation::relationColName)
				.collect(Collectors.toSet()).containsAll(newRefCols.keySet())) {

```

davidangb: whoops, sorry - I couldn't make this a code suggestion against the full set of lines it changes - hopefully you can interpret what I meant though!

davidangb (commented):
could you add a unit test for this condition? Apologies for not asking for this in the initial review.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -231,6 +226,36 @@ void putRecordWithMissingTableReference() throws Exception {
...
+		String invalid_ref = RelationUtils.createRelationString("missing", recordId);
+		new_attributes.put("ref-attr", invalid_ref);
+
+		mockMvc.perform(put("/{instanceId}/records/{version}/{recordType}/{recordId}", instanceId, versionId,
+				referringType, "new_record")
+						.content(mapper.writeValueAsString(new RecordRequest(new RecordAttributes(new_attributes))))
+						.contentType(MediaType.APPLICATION_JSON))
+				.andExpect(status().isBadRequest());
+	}
+
```
davidangb, line 258: Thinking out loud: this is interesting - once the relation is configured from `referring_Type` to `referenced_Type`, this test proves that you can't add an attribute with a relation to anything other than `referenced_Type`, _*even if the new target type is nonexistent*_. If we ever added extra validation logic to ensure that all relation targets exist, the new logic might tweak this test. 

Nothing to change, going to 👍 this PR!

davidangb (approved):

PR #68, opened 2022-09-13
AJ-567: Sanitize user input for table and column names to remediate SQL injection risk

sagehen03:
I threw in a few small sonar fixes in addition to disallowing characters for record type names and attribute names.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -66,12 +66,13 @@ public class RecordController {
 		MapDifference<String, DataTypeMapping> difference = Maps.difference(existingTableSchema, schema);
 		Map<String, DataTypeMapping> colsToAdd = difference.entriesOnlyOnRight();
 		colsToAdd.keySet().stream().filter(s -> s.startsWith(RESERVED_NAME_PREFIX)).findAny().ifPresent(s -> {
-			throw new InvalidNameException("Attribute");
+			throw new InvalidNameException(InvalidNameException.NameType.ATTRIBUTE);
 		});
 		validateRelationsAndAddColumns(instanceId, recordType, schema, records, colsToAdd, existingTableSchema);
 		Map<String, MapDifference.ValueDifference<DataTypeMapping>> differenceMap = difference.entriesDiffering();
-		for (String column : differenceMap.keySet()) {
-			MapDifference.ValueDifference<DataTypeMapping> valueDifference = differenceMap.get(column);
+		for (Map.Entry<String, MapDifference.ValueDifference<DataTypeMapping>> entry : differenceMap.entrySet()) {
+			String column = entry.getKey();
+			MapDifference.ValueDifference<DataTypeMapping> valueDifference = entry.getValue();
 			DataTypeMapping updatedColType = inferer.selectBestType(valueDifference.leftValue(),
 					valueDifference.rightValue());
 			recordDao.changeColumn(instanceId, recordType, column, updatedColType);
@@ -100,13 +101,15 @@ public class RecordController {
 			throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
 					"Relation attribute can only be assigned to one record type");
 		}
-		for (String col : colsToAdd.keySet()) {
+		for (Map.Entry<String, DataTypeMapping> entry : colsToAdd.entrySet()) {
 			String referencedRecordType = null;
+			String col = entry.getKey();
+			DataTypeMapping dataType = entry.getValue();
 			if (allRefCols.containsKey(col)) {
 				referencedRecordType = allRefCols.get(col).get(0).relationRecordType().getName();
 			}
-			recordDao.addColumn(instanceId, recordType, col, colsToAdd.get(col), referencedRecordType);
-			requestSchema.put(col, colsToAdd.get(col));
+			recordDao.addColumn(instanceId, recordType, col, dataType, referencedRecordType);
+			requestSchema.put(col, dataType);
 		}
 	}
 
@@ -152,8 +155,8 @@ public class RecordController {
 					recordTypeName);
 			// null out any attributes that already exist but aren't in the request
 			existingTableSchema.keySet().forEach(attr -> attributesInRequest.putIfAbsent(attr, null));
-			Record record = new Record(recordId, recordType, recordRequest.recordAttributes());
-			List<Record> records = Collections.singletonList(record);
+			Record newRecord = new Record(recordId, recordType, recordRequest.recordAttributes());
+			List<Record> records = Collections.singletonList(newRecord);
 			addOrUpdateColumnIfNeeded(instanceId, recordType.getName(), requestSchema, existingTableSchema, records);
 			Map<String, DataTypeMapping> combinedSchema = new HashMap<>(existingTableSchema);
 			combinedSchema.putAll(requestSchema);
@@ -176,7 +179,7 @@ public class RecordController {
 	}
 
 	@DeleteMapping("/{instanceId}/records/{version}/{recordType}/{recordId}")
-	public ResponseEntity deleteSingleRecord(@PathVariable("instanceId") UUID instanceId,
+	public ResponseEntity<Void> deleteSingleRecord(@PathVariable("instanceId") UUID instanceId,
 			@PathVariable("version") String version, @PathVariable("recordType") RecordType recordType,
 			@PathVariable("recordId") RecordId recordId) {
 		validateVersion(version);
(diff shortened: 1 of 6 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/service/model/exception/InvalidNameException.java
```diff
@@ -6,7 +6,18 @@
...
+
+	public enum NameType {
+		ATTRIBUTE("Attribute"), RECORD_TYPE("Record Type");
+
+		private final String name;
+
+		NameType(String name) {
+			this.name = name;
+		}
+	}
```
sagehen03, line 18: I wanted to avoid having a proliferation of `throw new IvalidNameException("Record Type"|"Attribute")` and run the risk of small typos, but I'm not all that sure that this is proper place to address that concern.  Thoughts?

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -100,13 +101,15 @@ private void validateRelationsAndAddColumns(UUID instanceId, String recordType,
 			throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
 					"Relation attribute can only be assigned to one record type");
 		}
-		for (String col : colsToAdd.keySet()) {
+		for (Map.Entry<String, DataTypeMapping> entry : colsToAdd.entrySet()) {
```
sagehen03, line 104: This is a change to make sonar happier, iterating over the entryset rather than keyset if both values are used.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -66,12 +66,13 @@ private Map<String, DataTypeMapping> addOrUpdateColumnIfNeeded(UUID instanceId,
...
 		Map<String, DataTypeMapping> colsToAdd = difference.entriesOnlyOnRight();
 		colsToAdd.keySet().stream().filter(s -> s.startsWith(RESERVED_NAME_PREFIX)).findAny().ifPresent(s -> {
-			throw new InvalidNameException("Attribute");
+			throw new InvalidNameException(InvalidNameException.NameType.ATTRIBUTE);
 		});
 		validateRelationsAndAddColumns(instanceId, recordType, schema, records, colsToAdd, existingTableSchema);
 		Map<String, MapDifference.ValueDifference<DataTypeMapping>> differenceMap = difference.entriesDiffering();
-		for (String column : differenceMap.keySet()) {
-			MapDifference.ValueDifference<DataTypeMapping> valueDifference = differenceMap.get(column);
+		for (Map.Entry<String, MapDifference.ValueDifference<DataTypeMapping>> entry : differenceMap.entrySet()) {
```
sagehen03, line 73: Change for sonar.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -152,8 +155,8 @@ public ResponseEntity<RecordResponse> upsertSingleRecord(@PathVariable("instance
 					recordTypeName);
 			// null out any attributes that already exist but aren't in the request
 			existingTableSchema.keySet().forEach(attr -> attributesInRequest.putIfAbsent(attr, null));
-			Record record = new Record(recordId, recordType, recordRequest.recordAttributes());
-			List<Record> records = Collections.singletonList(record);
+			Record newRecord = new Record(recordId, recordType, recordRequest.recordAttributes());
```
sagehen03, line 158: Sonar change

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -176,7 +179,7 @@ public ResponseEntity<String> createInstance(@PathVariable("instanceId") UUID in
 	}
 
 	@DeleteMapping("/{instanceId}/records/{version}/{recordType}/{recordId}")
-	public ResponseEntity deleteSingleRecord(@PathVariable("instanceId") UUID instanceId,
+	public ResponseEntity<Void> deleteSingleRecord(@PathVariable("instanceId") UUID instanceId,
```
sagehen03, line 182: Sonar change

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -292,7 +309,7 @@ private Map<String, Object> getAttributes(ResultSet rs) {
 					} else {
 						Object object = rs.getObject(columnName);
 						attributes.put(columnName,
-								object instanceof PGobject ? ((PGobject) object).getValue() : object);
+								object instanceof PGobject pGobject? pGobject.getValue() : object);
```
sagehen03, line 312: Sonar change.

service/src/main/java/org/databiosphere/workspacedataservice/service/model/exception/InvalidNameException.java
calypsomatic: I wasn't real pleased with passing the type of thing in that way, I just wanted to avoid duplication, so I think this is a step in the right direction at least.

calypsomatic (approved):
I'm very excited to stop having sonar yelling at me for code I didn't write!  Make sure to put a note somewhere in the readme and/or wiki about what types of names are allowed.

sagehen03:
> I'm very excited to stop having sonar yelling at me for code I didn't write! Make sure to put a note somewhere in the readme and/or wiki about what types of names are allowed.

I've made a change to the swagger and the readme.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -32,6 +34,7 @@
 public class RecordDao {
 
 	private final NamedParameterJdbcTemplate namedTemplate;
+	private static final Pattern DISALLOWED_CHARS_PATTERN = Pattern.compile("[^a-z0-9\\-_\\s]", Pattern.CASE_INSENSITIVE);
```
davidangb, line 37: hm,  I think we probably don't want to support ALL whitespace characters - vertical tab, newline, carriage return, form feed all seem like they could cause weird bugs especially when rendered in a UI. I'd defer to product if we have reasons we DO want to support those - by my inclination is to just allow spaces, not the full `\s` character class.

EDIT: from your swagger & error message changes it seems like you also intended this to be spaces only!

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -66,10 +70,23 @@ public void createRecordType(UUID instanceId, Map<String, DataTypeMapping> table
...
-		return quote(instanceId.toString()) + "." + quote(recordType);
+		return quote(instanceId.toString()) + "." + quote(validateName(recordType, InvalidNameException.NameType.RECORD_TYPE));
+	}
+
+	private String validateName(String name, InvalidNameException.NameType nameType){
+		if(containsDisallowedSqlCharacter(name)){
+			throw new InvalidNameException(nameType);
+		} else {
+			return name;
+		}
```
davidangb, line 85: nit: the `else` is extraneous here:
```suggestion
		if(containsDisallowedSqlCharacter(name)){
			throw new InvalidNameException(nameType);
		}
		return name;
```
no big deal if you have a reason to keep it as-is.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
davidangb: TIL! JEP 305, huh?

davidangb (commented):
comments inline!

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
sagehen03: Good catch! Thank you!

davidangb (approved):
:+1:-ing, but one other thought that came up as I was working on my PR: we validate that the record type doesn't start with "sys_" inside `createRecordTypeAndInsertRecords`, but we validate its sql legality elsewhere in RecordDao. Could/should these be consolidated? Looking for opinions here, I'm ok to merge the PR as-is.

sagehen03:
> +1-ing, but one other thought that came up as I was working on my PR: we validate that the record type doesn't start with "sys_" inside `createRecordTypeAndInsertRecords`, but we validate its sql legality elsewhere in RecordDao. Could/should these be consolidated? Looking for opinions here, I'm ok to merge the PR as-is.

Yeah, the inconsistency is a little troubling, but I see this one as pretty grey, the existing validation for the names that begin with sys_ is only the write code paths since it's not necessary on the read paths.  Putting all the validation in one place has that nice property of reducing lines of code, but results in extra validation on reads and does extra work before failing the operation.  I can see both sides, but I prefer to leave it as is since that gives the flexibility to potentially deal with these scenarios (someone trying to SQL inject vs. someone who chooses a namespace we want to reserve) differently in the future.

PR #69, opened 2022-09-13
AJ-517 first pass describeRecordType

calypsomatic:
Adds API to describe the schema of a recordType.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -201,7 +201,6 @@ public class RecordController {
 
 	private RecordTypeSchema getSchemaDescription(UUID instanceId, String recordType) {
 		Map<String, DataTypeMapping> schema = recordDao.getExistingTableSchema(instanceId, recordType);
-		schema.remove(RECORD_ID);
 		Map<String, RecordType> relations = recordDao.getRelationCols(instanceId, recordType).stream()
 				.collect(Collectors.toMap(Relation::relationColName, Relation::relationRecordType));
 		List<AttributeSchema> attrSchema = schema.entrySet().stream().map(entry -> createAttributeSchema(entry.getKey(),
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -205,8 +205,9 @@ public class RecordController {
 		Map<String, RecordType> relations = recordDao.getRelationCols(instanceId, recordType).stream()
 				.collect(Collectors.toMap(Relation::relationColName, Relation::relationRecordType));
 		List<AttributeSchema> attrSchema = schema.entrySet().stream().map(entry -> createAttributeSchema(entry.getKey(),
-				entry.getValue(), relations.getOrDefault(entry.getKey(), null))).collect(Collectors.toList());
-		return new RecordTypeSchema(recordType, attrSchema);
+				entry.getValue(), relations.getOrDefault(entry.getKey(), null))).toList();
+		int recordCount = recordDao.countRecords(instanceId, recordType);
+		return new RecordTypeSchema(recordType, attrSchema, recordCount);
 	}
 
 	private AttributeSchema createAttributeSchema(String name, DataTypeMapping datatype, RecordType relation) {
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -5,8 +5,7 @@ import com.google.common.collect.Maps;
 import org.databiosphere.workspacedataservice.dao.RecordDao;
 import org.databiosphere.workspacedataservice.service.DataTypeInferer;
 import org.databiosphere.workspacedataservice.service.RelationUtils;
-import org.databiosphere.workspacedataservice.service.model.DataTypeMapping;
-import org.databiosphere.workspacedataservice.service.model.Relation;
+import org.databiosphere.workspacedataservice.service.model.*;
 import org.databiosphere.workspacedataservice.service.model.exception.*;
 import org.databiosphere.workspacedataservice.shared.model.Record;
 import org.databiosphere.workspacedataservice.shared.model.*;
@@ -18,6 +17,7 @@ import org.springframework.web.server.ResponseStatusException;
 import java.util.*;
 import java.util.stream.Collectors;
 
+import static org.databiosphere.workspacedataservice.service.model.ReservedNames.RECORD_ID;
 import static org.databiosphere.workspacedataservice.service.model.ReservedNames.RESERVED_NAME_PREFIX;
 
 @RestController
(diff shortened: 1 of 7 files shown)


sagehen03 (commented):
Do we need to add a count for the number of records in a record type?

calypsomatic:
> Do we need to add a count for the number of records in a record type?

Yes, I didn't do it at first because it's not explicitly required in the ticket, but I've added it in the describe-all-types ticket which I'm also working on.  I can port it over.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -306,6 +305,46 @@ void deleteReferencedRecord() throws Exception {
...
+				.andExpect(content().string(containsString("attr1\",\"datatype\":\"STRING\"")))
+				.andExpect(content().string(containsString("attr2\",\"datatype\":\"DOUBLE\"")))
+				.andExpect(content().string(containsString("attr3\",\"datatype\":\"DATE\"")))
+				.andExpect(content().string(containsString("attr4\",\"datatype\":\"STRING\"")))
+				.andExpect(content().string(containsString("attr5\",\"datatype\":\"LONG\"")))
+				.andExpect(content().string(containsString("attr-boolean\",\"datatype\":\"BOOLEAN\"")))
+				.andExpect(content().string(
+						containsString("attr-ref\",\"datatype\":\"RELATION\",\"relatesTo\":\"" + referencedType)))
+				.andExpect(content().string(not(containsString("sys_name"))))
+				.andExpect(content().string(containsString("name\":\"" + type)));
```
sagehen03, line 338: Not a huge deal, but you can use jsonPath matchers to more precisely validate the json.  Here's an example from elsewhere in `RecordControllerMockMvcTest`: `mockMvc.perform(get("/{instanceId}/records/{version}/{recordType}/{recordId}", instanceId, versionId,
				recordType, "record_0")).andExpect(status().isOk()).andExpect(jsonPath("$.id", is("record_0")));`

calypsomatic: Oh that would be very nice!  Thank you!

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -185,6 +188,35 @@ public ResponseEntity deleteSingleRecord(@PathVariable("instanceId") UUID instan
...
+		if (!recordDao.recordTypeExists(instanceId, recordType.getName())) {
+			throw new MissingObjectException("Record type");
+		}
+		RecordTypeSchema result = getSchemaDescription(instanceId, recordType.getName());
+		return new ResponseEntity<>(result, HttpStatus.OK);
+	}
+
+	private RecordTypeSchema getSchemaDescription(UUID instanceId, String recordType) {
+		Map<String, DataTypeMapping> schema = recordDao.getExistingTableSchema(instanceId, recordType);
+		schema.remove(RECORD_ID);
```
sagehen03, line 204: Suggestion: I think we can modify `getExistingTableSchema` to not return the `sys_name` column and then we wouldn't have to remove the column from result.  
```
public Map<String, DataTypeMapping> getExistingTableSchema(UUID instanceId, String tableName) {
		MapSqlParameterSource params = new MapSqlParameterSource("instanceId", instanceId.toString());
		params.addValue("tableName", tableName);
		params.addValue("recordName", RECORD_ID);
		return namedTemplate
				.query("select column_name, data_type from INFORMATION_SCHEMA.COLUMNS where table_schema = :instanceId "
						+ "and table_name = :tableName and column_name != :recordName", params, rs -> {
							Map<String, DataTypeMapping> result = new HashMap<>();
							while (rs.next()) {
								result.put(rs.getString("column_name"),
										DataTypeMapping.fromPostgresType(rs.getString("data_type")));
							}
							return result;
						});
	}
```
I don't think the other callers need that column, but we should double check.

sagehen03 (approved):
This seems like it will work.  I made a few suggestions, but I don't see any bugs.  With this change we're one step closer to going forward with users with this data type model.  So if we want to make changes, such as just a single number data type (that's postgres numeric behind the scenes) or if we don't want to include json for now, we should address that in the relative near term.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -210,6 +210,11 @@ public List<Relation> getRelationCols(UUID instanceId, String tableName) {
 				(rs, rowNum) -> new Relation(rs.getString("column_name"), new RecordType(rs.getString("table_name"))));
 	}
 
+	public int countRecords(UUID instanceId, String recordTypeName) {
```
sagehen03, line 213: Thinking out loud, at some point I think we may want to leverage spring's caching around this method and the put method.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
calypsomatic: I think it would only be a problem if we were relying on it to potentially remove columns, which we don't currently do.  But we'd need to remember to take it into account if we're updating the schema and deleting columns.

service/src/test/java/org/databiosphere/workspacedataservice/dao/RecordDaoTest.java
```diff
@@ -157,4 +156,26 @@ void testDeleteRelatedRecord() {
...
+		assertEquals(1, count);
+
+		List<Record> batch = new ArrayList<>();
+		for (int i = 0; i < 10; i++) {
+			batch.add(new Record(new RecordId("record" + i), recordType, new RecordAttributes(new HashMap<>())));
+		}
+		recordDao.batchUpsert(instanceId, recordType.getName(), batch, new HashMap<>());
+		int newcount = recordDao.countRecords(instanceId, recordType.getName());
+		assertEquals(11, newcount);
+	}
```
davidangb, line 179: suggestion, will not block the PR on this: you could make this test shorter and more robust by checking inside the loop, like so:

```java
	void testCountRecords() {
		// ensure we start with a count of 0 records
		assertEquals(0, recordDao.countRecords(instanceId, recordType.getName()));
		// insert records and test the count after each insert
		for (int i = 0; i < 10; i++) {
			Record testRecord = new Record(new RecordId("record" + i), recordType, new RecordAttributes(new HashMap<>()));
			recordDao.batchUpsert(instanceId, recordType.getName(), Collections.singletonList(testRecord), new HashMap<>());
			assertEquals(i+1, recordDao.countRecords(instanceId, recordType.getName()),
					"after inserting " + (i+1) + " records");
		}
	}
```

davidangb (commented):
see comments inline, this is working well for me in practice!

service/src/main/java/org/databiosphere/workspacedataservice/service/model/RecordTypeSchema.java
```diff
@@ -0,0 +1,9 @@
+package org.databiosphere.workspacedataservice.service.model;
+
+import java.util.List;
+
+public record RecordTypeSchema(String name, List<AttributeSchema> attributes, int count) {
+	public RecordTypeSchema(String name, List<AttributeSchema> attributes) {
+		this(name, attributes, 0);
+	}
```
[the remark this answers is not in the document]
calypsomatic: Ah right I was using it in an earlier iteration

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -185,6 +188,34 @@ public ResponseEntity deleteSingleRecord(@PathVariable("instanceId") UUID instan
...
+		}
+		RecordTypeSchema result = getSchemaDescription(instanceId, recordType.getName());
+		return new ResponseEntity<>(result, HttpStatus.OK);
+	}
+
+	private RecordTypeSchema getSchemaDescription(UUID instanceId, String recordType) {
+		Map<String, DataTypeMapping> schema = recordDao.getExistingTableSchema(instanceId, recordType);
+		Map<String, RecordType> relations = recordDao.getRelationCols(instanceId, recordType).stream()
+				.collect(Collectors.toMap(Relation::relationColName, Relation::relationRecordType));
+		List<AttributeSchema> attrSchema = schema.entrySet().stream().map(entry -> createAttributeSchema(entry.getKey(),
```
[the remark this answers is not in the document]
calypsomatic: Wasn't there some discussion about whether to sort because users have an expected order for their attributes and alphabetizing them isn't what they want?  Or is it just because it's preferable to the default, which is random?

service/src/test/java/org/databiosphere/workspacedataservice/dao/RecordDaoTest.java
calypsomatic: Huh not sure why I did it that weird way

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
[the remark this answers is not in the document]
davidangb: There's definitely a discussion about allowing users to define and control the sort order for their attributes, especially in the UI. But, the current state of this PR doesn't allow that. So in order of preference, I think we have:
* least preferable: Java map key (hash) order, current state of this PR
* better: alphabetical order, my code comment
* best: user-defined order, a bunch of work required and exact use cases undefined

davidangb (approved):
lgtm, thank you!

calypsomatic:
All the sonarcloud failures were previously existing/getting fixed in a concurrent PR.

PR #70, opened 2022-09-14
RecordType wrapper class refactor

davidangb:
Highlights of this refactor:
* `RecordType` class is now immutable
* we perform validation in the RecordType constructor (starts with "sys_", is sql-legal)
* all methods that accept a record type as argument now accept RecordType instead of String, and we pass around RecordType instead of repeatedly transforming it to and from String
* we only translate RecordType to a String value where absolutely necessary, such as when using in a sql statement
* `toString` for RecordType now returns just the name value, making it easy to use a RecordType directly in url templates (such as inside unit tests)

There's a lot of lines of code that changed due to the pervasiveness of RecordType

After this PR merges, I'll make follow-on PRs:
- [ ] eliminate `RecordId` and replace with a plain `String`
- [ ] proposal to simplify `RecordAttributes`

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -129,9 +129,9 @@ public class RecordController {
 
 	@PostMapping("/{instanceid}/search/{version}/{recordType}")
 	public RecordQueryResponse queryForEntities(@PathVariable("instanceid") UUID instanceId,
-			@PathVariable("recordType") String recordTypeName,
+			@PathVariable("recordType") RecordType recordType,
 			@RequestBody(required = false) SearchRequest searchRequest) {
-		validateRecordType(instanceId, recordTypeName);
+		checkRecordTypeExists(instanceId, recordType);
 		if (null == searchRequest) {
 			searchRequest = new SearchRequest();
 		}
@@ -139,11 +139,11 @@ public class RecordController {
 			throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
 					"Limit must be more than 0 and can't exceed " + MAX_RECORDS + ", and offset must be positive.");
 		}
-		int totalRecords = recordDao.countRecords(instanceId, recordTypeName);
+		int totalRecords = recordDao.countRecords(instanceId, recordType);
 		if (searchRequest.getOffset() > totalRecords) {
 			return new RecordQueryResponse(searchRequest, Collections.emptyList(), totalRecords);
 		}
-		List<Record> records = recordDao.queryForRecords(recordTypeName, searchRequest.getLimit(),
+		List<Record> records = recordDao.queryForRecords(recordType, searchRequest.getLimit(),
 				searchRequest.getOffset(), searchRequest.getSort().name().toLowerCase(), instanceId);
 		List<RecordResponse> recordList = records.stream().map(
 				r -> new RecordResponse(r.getId(), r.getRecordType(), r.getAttributes(), new RecordMetadata("UNUSED")))
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -131,7 +131,6 @@ public class RecordController {
 			@PathVariable("version") String version, @PathVariable("recordType") RecordType recordType,
 			@PathVariable("recordId") RecordId recordId, @RequestBody RecordRequest recordRequest) {
 		validateVersion(version);
-		// String recordTypeName = recordType;
 		Map<String, Object> attributesInRequest = recordRequest.recordAttributes().getAttributes();
 		Map<String, DataTypeMapping> requestSchema = inferer.inferTypes(attributesInRequest);
 		HttpStatus status = HttpStatus.CREATED;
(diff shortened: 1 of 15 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -134,33 +131,32 @@ public ResponseEntity<RecordResponse> upsertSingleRecord(@PathVariable("instance
 			@PathVariable("version") String version, @PathVariable("recordType") RecordType recordType,
 			@PathVariable("recordId") RecordId recordId, @RequestBody RecordRequest recordRequest) {
 		validateVersion(version);
-		String recordTypeName = recordType.getName();
+		// String recordTypeName = recordType;
```
calypsomatic, line 134: A commented out line of code!  Unacceptable!

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -78,13 +80,13 @@ void tryCreatingIllegallyNamedRecordType() throws Exception {
 						.content(mapper.writeValueAsString(new RecordRequest(new RecordAttributes(attributes))))
 						.contentType(MediaType.APPLICATION_JSON))
 				.andExpect(status().isBadRequest())
-				.andExpect(result -> assertTrue(result.getResolvedException() instanceof InvalidNameException));
+				.andExpect(result -> assertTrue(result.getResolvedException() instanceof MethodArgumentTypeMismatchException));
```
calypsomatic, line 83: Just curious - where did this `MethodArgumentTypeMismatchException` come from?

calypsomatic (approved):
👍 Although you might want to throw some `squid`s in there to quiet down sonar.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -254,13 +248,11 @@ private void validateInstance(UUID instanceId) {
 		}
 	}
 
-	private void createRecordTypeAndInsertRecords(UUID instanceId, Record newRecord, String recordTypeName,
+	private void createRecordTypeAndInsertRecords(UUID instanceId, Record newRecord, RecordType recordType,
 			Map<String, DataTypeMapping> requestSchema) {
-		if (recordTypeName.startsWith(RESERVED_NAME_PREFIX)) {
-			throw new InvalidNameException(InvalidNameException.NameType.RECORD_TYPE);
-		}
+		recordType.validate();
```
sagehen03, line 253: Do we need this call?  I think valiidate() is already called in the RecordType constructor.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -73,25 +74,25 @@ public void createRecordType(UUID instanceId, Map<String, DataTypeMapping> table
 
 	}
 
-	private String getQualifiedTableName(String recordType, UUID instanceId) {
+	private String getQualifiedTableName(RecordType recordType, UUID instanceId) {
 		return quote(instanceId.toString()) + "."
-				+ quote(validateName(recordType, InvalidNameException.NameType.RECORD_TYPE));
+				+ quote(validateName(recordType.getName(), InvalidNameException.NameType.RECORD_TYPE));
```
sagehen03, line 79: I think we can get rid of the validateName() call in `getQualifiedTable` now since we do the validation upon creating RecordTypes?

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -73,25 +74,25 @@ public void createRecordType(UUID instanceId, Map<String, DataTypeMapping> table
...
 	}
 
-	private String validateName(String name, InvalidNameException.NameType nameType) {
+	public static String validateName(String name, InvalidNameException.NameType nameType) {
```
sagehen03, line 82: I feel like we should move this method out of RecordDao.  Maybe to a utility class and we should probably rename it `sanitizeSqlName` or something more descriptive than validateName.

sagehen03 (approved):
I don't see any bugs.  It looks like there might be an unnecessary RecordType#valiidate() call and I'd like to see validation move out of RecordDao eventually with RecordType and perhaps a new class, RecordAttributeName all doing validation upon object creation.  I'm not sure if that's your long term vision.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
davidangb: good question; I should have explained this. Because we've added validation into the `RecordType` constructor, Spring now gets an error when attempting to create a `RecordType` from the URL path params, and Spring throws a `MethodArgumentTypeMismatchException` error before any of our code runs. 

I have a similar comment on @sagehen03 's PR: https://github.com/DataBiosphere/terra-workspace-data-service/pull/73#discussion_r973204964 and I think we'll need a follow-on story (or a few) to improve error messages for invalid URL/request body parameters.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
davidangb: good catch, will fix!

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
davidangb: done!

calypsomatic (approved):

PR #71, opened 2022-09-14
Aj 572 describe all types

calypsomatic:
[AJ-572](https://broadworkbench.atlassian.net/browse/AJ-572)
Adds API to describe all record types in an instance.


service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -352,6 +352,43 @@ void describeNonexistentType() throws Exception {
 				.andExpect(status().isNotFound());
 	}
 
+	@Test
+	@Transactional
+	void describeAllTypes() throws Exception {
+		// replace instanceId for this test so only these records are found
+		instanceId = UUID.randomUUID();
```
calypsomatic, line 359: Would be happy for recommendations for a better way to make sure I know all the record types to expect; `@Transactional` doesn't clear out the database between tests and a few other annotations I tried (`@Order`, `@DirtiesContext`) didn't seem to help.

jordansafer: I haven't found any clean solution. Looks like you'd have to make a cleanup task which deletes DB entries between tasks. Other options like @DirtiesContext can apparently be somewhat expensive if they are working-

jordansafer (approved):

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
davidangb: hm - I think that mutating the `instanceId` class variable is not a good idea - that could cause problems for future tests. My suggestion is to allow `createSomeRecords` to accept an instance argument, like:
```java
	private void createSomeRecords(String recordType, int numRecords) throws Exception {
		createSomeRecords(recordType, numRecords, instanceId);
	}
	private void createSomeRecords(String recordType, int numRecords, UUID instanceId) throws Exception {
          …
        }
```
and then in this test you can create a test-local random UUID and use it just for these tests.

davidangb: and to clarify my previous comment: I think that simply using a different instanceId is fine for this test, no need to wipe/recreate the database at all. That's what multiple instances are for!

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -322,4 +322,9 @@ public Optional<Record> getSingleRecord(UUID instanceId, RecordType recordType,
 			return Optional.empty();
 		}
 	}
+
+	public List<String> getAllRecordTypes(UUID instanceId) {
+		return namedTemplate.queryForList("select tablename from pg_tables WHERE schemaname = :workspaceSchema",
+				new MapSqlParameterSource("workspaceSchema", instanceId.toString()), String.class);
```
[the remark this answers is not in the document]
calypsomatic: And coincidentally, my test tables - "firstType", "secondType" and "thirdType" are already in alphabetical order!

davidangb (approved):
looks good, thank you!

PR #72, opened 2022-09-15
Aj 574 delete record type

calypsomatic:
[AJ-574](https://broadworkbench.atlassian.net/browse/AJ-574)
Adds an API to delete record type

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -187,9 +187,15 @@ public class RecordDao {
 	}
 
 	private void checkForTableRelation(SQLException sqlEx) {
-		if (sqlEx != null && sqlEx.getSQLState() != null && sqlEx.getSQLState().equals("23503")) {
-			throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
-					"Unable to delete this record because another record has a relation to it.");
+		if (sqlEx != null && sqlEx.getSQLState() != null) {
+			if (sqlEx.getSQLState().equals("23503")) {
+				throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
+						"Unable to delete this record because another record has a relation to it.");
+			}
+			if (sqlEx.getSQLState().equals("2BP01")) {
+				throw new ResponseStatusException(HttpStatus.FORBIDDEN,
+						"Unable to delete this record type because another record type has a relation to it.");
+			}
 		}
 	}
 
@@ -327,5 +333,16 @@ public class RecordDao {
 		return namedTemplate.queryForList(
 				"select tablename from pg_tables WHERE schemaname = :workspaceSchema order by tablename",
 				new MapSqlParameterSource("workspaceSchema", instanceId.toString()), String.class);
+	}
+
+	public void deleteRecordType(UUID instanceId, String recordType) {
+		try {
+			namedTemplate.getJdbcTemplate().update("drop table " + getQualifiedTableName(recordType, instanceId));
+		} catch (DataAccessException e) {
+			if (e.getRootCause()instanceof SQLException sqlEx) {
+				checkForTableRelation(sqlEx);
+			}
+			throw e;
+		}
 	}
 }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -74,11 +74,12 @@ public class RecordDao {
 	}
 
 	private String getQualifiedTableName(String recordType, UUID instanceId) {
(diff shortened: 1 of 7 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -187,9 +187,15 @@ private void checkForMissingRecord(SQLException sqlEx) {
 	}
 
 	private void checkForTableRelation(SQLException sqlEx) {
-		if (sqlEx != null && sqlEx.getSQLState() != null && sqlEx.getSQLState().equals("23503")) {
-			throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
-					"Unable to delete this record because another record has a relation to it.");
+		if (sqlEx != null && sqlEx.getSQLState() != null) {
```
calypsomatic, line 190: This is kind of a placeholder to get the job done while we think/discuss about how best to deal with this.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -203,9 +203,15 @@ private void checkForMissingRecord(SQLException sqlEx) {
...
+			if (sqlEx.getSQLState().equals("23503")) {
+				throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
+						"Unable to delete this record because another record has a relation to it.");
+			}
+			if (sqlEx.getSQLState().equals("2BP01")) {
+				throw new ResponseStatusException(HttpStatus.FORBIDDEN,
+						"Unable to delete this record type because another record type has a relation to it.");
+			}
```
davidangb, line 214: I'm fine with this implementation - thinking out loud for the future: if we find we need to detect a third or a fourth error code, we should probably move to switch/case:
```java
	private void checkForTableRelation(SQLException sqlEx) {
		if (sqlEx != null && sqlEx.getSQLState() != null) {
			switch (sqlEx.getSQLState()) {
				case "23503" -> throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
						"Unable to delete this record because another record has a relation to it.");
				case "2BP01" -> throw new ResponseStatusException(HttpStatus.FORBIDDEN,
						"Unable to delete this record type because another record type has a relation to it.");
				default -> throw new ResponseStatusException(HttpStatus.INTERNAL_SERVER_ERROR,
						"Unexpected SQL error: " + sqlEx.getMessage());
                                                  // or, skip throwing here and log something, or just noop
			}
		}
	}
```

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -203,9 +203,15 @@ private void checkForMissingRecord(SQLException sqlEx) {
...
-		if (sqlEx != null && sqlEx.getSQLState() != null && sqlEx.getSQLState().equals("23503")) {
-			throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
-					"Unable to delete this record because another record has a relation to it.");
+		if (sqlEx != null && sqlEx.getSQLState() != null) {
+			if (sqlEx.getSQLState().equals("23503")) {
+				throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
+						"Unable to delete this record because another record has a relation to it.");
+			}
+			if (sqlEx.getSQLState().equals("2BP01")) {
+				throw new ResponseStatusException(HttpStatus.FORBIDDEN,
```
davidangb, line 212: swagger defines that this case will return a 409 conflict. I think it should either be 409 or 400 - forbidden should be reserved for auth issues. Could you update this to throw 409, or update both the swagger and here to throw 400? (or speak up about why it should be something other than 400/409!)

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -344,4 +354,15 @@ public List<String> getAllRecordTypes(UUID instanceId) {
 				"select tablename from pg_tables WHERE schemaname = :workspaceSchema order by tablename",
 				new MapSqlParameterSource("workspaceSchema", instanceId.toString()), String.class);
 	}
+
+	public void deleteRecordType(UUID instanceId, String recordType) {
+		try {
+			namedTemplate.getJdbcTemplate().update("drop table " + getQualifiedTableName(recordType, instanceId));
```
calypsomatic, line 360: Also welcome any input on a better way to do this to make Sonar happy.

davidangb (approved):
lgtm, modulo decision on fixing or ignoring sonar errors. I did not test this myself manually, I assume you did!

sagehen03 (approved):
Gah, to Sonar!  It looks like this is a separate yet related from SQLi warning about formatted SQL.  II think we're just going to have to disable that warning.

calypsomatic:
> Gah, to Sonar! It looks like this is a separate yet related from SQLi warning about formatted SQL. II think we're just going to have to disable that warning.

How do we disable the warning?

PR #74, opened 2022-09-15
Aj 555 put created

calypsomatic:
[AJ-555](https://broadworkbench.atlassian.net/browse/AJ-555)
Distinguish between PUT creating or updating a record in the HTTPStatus returned.

I looked around to see if I could find whether or not a record already existed without making an extra database call, but nothing popped out at me.  I figured at least avoiding figuring out the relation columns and getting all the extraneous data was unnecessary and so created a new method, but I'm not sure how thrilled I am with the overall solution, so I'm happy to take suggestions.

### DIFF
--- a/service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
@@ -99,13 +99,18 @@ public class RecordControllerMockMvcTest {
 	@Transactional
 	void putNewRecord() throws Exception {
 		String newRecordType = "newRecordType";
-		RecordAttributes attributes = new RecordAttributes(Map.of("foo","bar","num",123));
+		RecordAttributes attributes = new RecordAttributes(Map.of("foo", "bar", "num", 123));
+		// create new record with new record type
 		mockMvc.perform(put("/{instanceId}/records/{version}/{recordType}/{recordId}", instanceId, versionId,
-						newRecordType, "newRecordId")
-						.content(mapper.writeValueAsString(new RecordRequest(attributes)))
+				newRecordType, "newRecordId").content(mapper.writeValueAsString(new RecordRequest(attributes)))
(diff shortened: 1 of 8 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -94,6 +94,24 @@ void updateWithIllegalAttributeName() throws Exception {
...
+		mockMvc.perform(put("/{instanceId}/records/{version}/{recordType}/{recordId}", instanceId, versionId,
+				newRecordType, "newRecordId").content(mapper.writeValueAsString(new RecordRequest(attributes)))
+						.contentType(MediaType.APPLICATION_JSON))
+				.andExpect(status().isCreated());
+
+		RecordAttributes attributes2 = new RecordAttributes(Map.of("foo", "baz", "num", 888));
+		mockMvc.perform(put("/{instanceId}/records/{version}/{recordType}/{recordId}", instanceId, versionId,
+				newRecordType, "newRecordId2").content(mapper.writeValueAsString(new RecordRequest(attributes2)))
+						.contentType(MediaType.APPLICATION_JSON))
+				.andExpect(status().isCreated());
```
davidangb, line 113: for good measure, could you add an update and check that it returns OK?
```java
		// now update the second new record
		RecordAttributes attributes3 = new RecordAttributes(Map.of("foo", "updated", "num", 999));
		mockMvc.perform(put("/{instanceId}/records/{version}/{recordType}/{recordId}", instanceId, versionId,
						newRecordType, "newRecordId2").content(mapper.writeValueAsString(new RecordRequest(attributes3)))
						.contentType(MediaType.APPLICATION_JSON))
				.andExpect(status().isOk());
```

davidangb (approved):
lgtm, I think the solution you have here is a good one, it's what I would have done too!

One suggestion inline, proactive :+1:. If you change stuff _other_ than the test tweak, I will re-review.

sagehen03 (approved):

PR #75, opened 2022-09-20
remove RecordId model [AJ-507]

davidangb:
remove the `RecordId` class and use a plain `String` instead.


sagehen03 (approved):
Seems pretty straightforward!

calypsomatic (approved):

PR #76, opened 2022-09-21
RecordAttributes model [AJ-507]

davidangb:
Philosophically, I am realizing that my motivations for refactoring classes like this are:
* where I'm less confident about the underlying representations - like using `Map<String, Object>` to represent attributes - I want to abstract callers from knowing about those representations
* providing domain objects that include some encapsulated logic, as opposed to pure record classes

---

Highlights of this PR:

1. Calling code is able to call methods directly on `Record` or `RecordAttributes`, instead of retrieving the underlying `Map` of attributes.
[Example](https://github.com/DataBiosphere/terra-workspace-data-service/pull/76/files#diff-6fdffee8355da583e8db9f8b0995789baa890a7fc0b2ba3da10e381f60afee25R284
), using `Record toInsert`:
```java
toInsert.getAttributes().getAttributes().get(colName)
```
becomes
```java
toInsert.getAttributeValue(colName)
```

2. Calling code can add `RecordAttributes` to a `Record` or to a `RecordAttributes` without retrieving the underlying `Map`:
[Example](https://github.com/DataBiosphere/terra-workspace-data-service/pull/76/files#diff-494957a4a881ff0f5307171889be87fc44dd3c1a06618aafcbe18998113cbfb1R43):
```java
Map<String, Object> updatedAtts = recordRequest.recordAttributes().getAttributes();
Map<String, Object> allAttrs = new HashMap<>(singleRecord.getAttributes().getAttributes());
allAttrs.putAll(updatedAtts);
```
becomes:
```java
RecordAttributes incomingAtts = recordRequest.recordAttributes();
RecordAttributes allAttrs = singleRecord.putAllAttributes(incomingAtts).getAttributes();
```

3. Reduce occurrences of getting the underlying attribute `Map`, then constructing a new `RecordAttributes` object from that map

4. The underlying implementation of `Map<String, Object>` for `RecordAttributes` is now mostly abstracted (there are a few non-abstracted places where we iterate over the map entries). I suspect the underlying representation will change in the future, so it's nice to be abstracted from it

5. By moving the mutator methods onto `RecordAttributes`, we can chain them, whereas `Map`'s methods return void and are not changeable. Thus we can change ([example](https://github.com/DataBiosphere/terra-workspace-data-service/pull/76/files#diff-ba2227f2e1675ac7d3e2a73ef90f4babd21833fb5166ea2e399e37c8dbc0dd8cL15)):
```java
	public static Map<String, Object> generateRandomAttributes() {
		Map<String, Object> attributes = new HashMap<>();
		attributes.put("attr1", RandomStringUtils.randomAlphabetic(6));
		attributes.put("attr2", RandomUtils.nextFloat());
		attributes.put("attr3", "2022-11-01");
		attributes.put("attr4", RandomStringUtils.randomNumeric(5));
		attributes.put("attr5", RandomUtils.nextLong());
		attributes.put("attr-dt", "2022-03-01T12:00:03");
		attributes.put("attr-json", "{\"foo\":\"bar\"}");
		attributes.put("attr-boolean", true);
		return attributes;
	}
```
to
```java
	public static RecordAttributes generateRandomAttributes() {
		return RecordAttributes.empty()
				.putAttribute("attr1", RandomStringUtils.randomAlphabetic(6))
				.putAttribute("attr2", RandomUtils.nextFloat())
				.putAttribute("attr3", "2022-11-01")
				.putAttribute("attr4", RandomStringUtils.randomNumeric(5))
				.putAttribute("attr5", RandomUtils.nextLong())
				.putAttribute("attr-dt", "2022-03-01T12:00:03")
				.putAttribute("attr-json", "{\"foo\":\"bar\"}")
				.putAttribute("attr-boolean", true);
	}
```

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -41,7 +41,7 @@ public class RecordController {
 				.getSingleRecord(instanceId, recordType, recordId, recordDao.getRelationCols(instanceId, recordType))
 				.orElseThrow(() -> new MissingObjectException("Record"));
 		RecordAttributes incomingAtts = recordRequest.recordAttributes();
-		RecordAttributes allAttrs = singleRecord.getAttributes().putAll(incomingAtts);
+		RecordAttributes allAttrs = singleRecord.putAllAttributes(incomingAtts).getAttributes();
 		Map<String, DataTypeMapping> typeMapping = inferer.inferTypes(incomingAtts);
 		Map<String, DataTypeMapping> existingTableSchema = recordDao.getExistingTableSchema(instanceId, recordType);
 		singleRecord.setAttributes(allAttrs);
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -40,13 +40,11 @@ public class RecordController {
 		Record singleRecord = recordDao
 				.getSingleRecord(instanceId, recordType, recordId, recordDao.getRelationCols(instanceId, recordType))
 				.orElseThrow(() -> new MissingObjectException("Record"));
-		Map<String, Object> updatedAtts = recordRequest.recordAttributes().getAttributes();
-		Map<String, Object> allAttrs = new HashMap<>(singleRecord.getAttributes().getAttributes());
-		allAttrs.putAll(updatedAtts);
-
-		Map<String, DataTypeMapping> typeMapping = inferer.inferTypes(updatedAtts);
+		RecordAttributes incomingAtts = recordRequest.recordAttributes();
+		RecordAttributes allAttrs = singleRecord.getAttributes().putAll(incomingAtts);
+		Map<String, DataTypeMapping> typeMapping = inferer.inferTypes(incomingAtts);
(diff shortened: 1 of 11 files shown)


sagehen03 (commented):
I agree this is better than the current state of affairs.  I am still skeptical as to whether this in an improvement over just using a Java Map.  In your view what advantages does this offer vs. just using a Map?  It does give us a layer of abstraction above the Map which could make it easier in the future to change how we store and access the attribute data in memory.  It also allows for chaining while adding attributes to a record.  Is there anything else?  If no, it doesn't feel like enough to me.  It's less code to just use a Map and the Map's API is something that all Java devs should know fairly well.  I am also pretty skeptical that we'll change our underlying use of a Map.

calypsomatic:
> I agree this is better than the current state of affairs. I am still skeptical as to whether this in an improvement over just using a Java Map. In your view what advantages does this offer vs. just using a Map? It does give us a layer of abstraction above the Map which could make it easier in the future to change how we store and access the attribute data in memory. It also allows for chaining while adding attributes to a record. Is there anything else? If no, it doesn't feel like enough to me. It's less code to just use a Map and the Map's API is something that all Java devs should know fairly well. I am also pretty skeptical that we'll change our underlying use of a Map.

I really like the chaining!

calypsomatic (approved):

sagehen03 (approved):

PR #78, opened 2022-09-27
run build/test on both push and pull_request events

davidangb:


calypsomatic (approved):
Works for me!

sagehen03 (approved):
I'm still a little confused as to the behavior I saw previously, but somethings stay mysterious I suppose.

PR #80, opened 2022-09-30
enable CORS [AJ-585]

davidangb:
in my attempts to get Terra UI talking to WDS, I realized CORS is not set up. This PR enables Ajax requests to WDS.


calypsomatic (approved):
I'll trust you that we don't need any restrictions!

jordansafer (approved):

davidangb:
@calypsomatic I updated the PR description; this is the same `Access-Control-Allow-Origin` setting that Terra uses today, so I'm keeping it in place.

PR #81, opened 2022-10-03
AJ-596: TSV Streaming Download

sagehen03:
I've tested this with a record type larger than 1gb and confirmed that it works even on a 100mb heap.  Changes are in place to make sure we stream from db to java.sql.Resultset, from Resultset to intermediate domain object, and from domain object to http response.  I can share a large json file you can use as input to the batch write API if you'd like to test for yourself.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
@@ -15,6 +15,9 @@ import org.springframework.context.annotation.Primary;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 
 import javax.sql.DataSource;
+
+import org.springframework.lang.NonNull;
+import org.springframework.web.servlet.config.annotation.AsyncSupportConfigurer;
 import org.springframework.web.servlet.config.annotation.CorsRegistry;
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
 
@@ -66,8 +69,13 @@ public class WorkspaceDataServiceApplication {
 	public WebMvcConfigurer corsConfigurer() {
 		return new WebMvcConfigurer() {
 			@Override
-			public void addCorsMappings(CorsRegistry registry) {
+			public void addCorsMappings(@NonNull CorsRegistry registry) {
 				registry.addMapping("/**").allowedOrigins("*");
+			}
+
+			@Override
+			public void configureAsyncSupport(@NonNull AsyncSupportConfigurer configurer) {
+				configurer.setDefaultTimeout(-1);
 			}
 		};
 	}
--- a/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
@@ -11,6 +11,7 @@ import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.jdbc.DataSourceBuilder;
 import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Primary;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 
 import javax.sql.DataSource;
@@ -34,6 +35,19 @@ public class WorkspaceDataServiceApplication {
 		//https://jdbc.postgresql.org/documentation/head/query.html#query-with-cursor
 		hikariDataSource.setAutoCommit(false);
 		return hikariDataSource;
+	}
+
+	@Bean
+	@Primary
+	@ConfigurationProperties("spring.datasource")
+	public DataSource mainDb(){
+		return DataSourceBuilder.create().build();
+	}
+
+	@Bean
+	@Primary
+	public NamedParameterJdbcTemplate namedParameterJdbcTemplate(){
+		return new NamedParameterJdbcTemplate(mainDb());
 	}
 
 	@Bean
--- a/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
@@ -35,7 +35,7 @@ public class WorkspaceDataServiceApplication {
 	public DataSource streamingDs(){
 		DataSource ds = DataSourceBuilder.create().build();
 		HikariDataSource hikariDataSource = (HikariDataSource) ds;
-		//https://jdbc.postgresql.org/documentation/head/query.html#query-with-cursor
+		//https://jdbc.postgresql.org/documentation/query/#getting-results-based-on-a-cursor
 		hikariDataSource.setAutoCommit(false);
 		return hikariDataSource;
 	}
@@ -58,7 +58,7 @@ public class WorkspaceDataServiceApplication {
 	public NamedParameterJdbcTemplate templateForStreaming(@Qualifier("streamingDs") DataSource ds,
 														   @Value("${twds.streaming.fetch.size:50}") int fetchSize){
 		NamedParameterJdbcTemplate jdbcTemplate = new NamedParameterJdbcTemplate(ds);
-		//https://jdbc.postgresql.org/documentation/head/query.html#query-with-cursor
+		//https://jdbc.postgresql.org/documentation/query/#getting-results-based-on-a-cursor
 		jdbcTemplate.getJdbcTemplate().setFetchSize(fetchSize);
 		return jdbcTemplate;
 	}
--- a/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
@@ -3,9 +3,17 @@ package org.databiosphere.workspacedataservice;
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.json.JsonMapper;
+import com.zaxxer.hikari.HikariDataSource;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.jdbc.DataSourceBuilder;
 import org.springframework.context.annotation.Bean;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+
+import javax.sql.DataSource;
 
 @SpringBootApplication
 public class WorkspaceDataServiceApplication {
@@ -14,6 +22,28 @@ public class WorkspaceDataServiceApplication {
 	public ObjectMapper objectMapper() {
 		return JsonMapper.builder().enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS).build();
 	}
+
+	@Bean
+	@Qualifier("streamingDs")
+	@ConfigurationProperties("streaming.query")
+	public DataSource streamingDs(){
+		DataSource ds = DataSourceBuilder.create().build();
+		HikariDataSource hikariDataSource = (HikariDataSource) ds;
+		//https://jdbc.postgresql.org/documentation/head/query.html#query-with-cursor
+		hikariDataSource.setAutoCommit(false);
+		return hikariDataSource;
+	}
+
+	@Bean
+	@Qualifier("streamingDs")
+	public NamedParameterJdbcTemplate templateForStreaming(@Qualifier("streamingDs") DataSource ds,
+														   @Value("${twds.streaming.fetch.size:50}") int fetchSize){
+		NamedParameterJdbcTemplate jdbcTemplate = new NamedParameterJdbcTemplate(ds);
+		//https://jdbc.postgresql.org/documentation/head/query.html#query-with-cursor
+		jdbcTemplate.getJdbcTemplate().setFetchSize(fetchSize);
+		return jdbcTemplate;
+	}
+
 	public static void main(String[] args) {
 		SpringApplication.run(WorkspaceDataServiceApplication.class, args);
 	}
--- a/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
@@ -56,7 +56,7 @@ public class WorkspaceDataServiceApplication {
 	@Bean
 	@Qualifier("streamingDs")
 	public NamedParameterJdbcTemplate templateForStreaming(@Qualifier("streamingDs") DataSource ds,
-			@Value("${twds.streaming.fetch.size:50}") int fetchSize) {
+			@Value("${twds.streaming.fetch.size:5000}") int fetchSize) {
 		NamedParameterJdbcTemplate jdbcTemplate = new NamedParameterJdbcTemplate(ds);
 		// https://jdbc.postgresql.org/documentation/query/#getting-results-based-on-a-cursor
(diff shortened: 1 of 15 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
```diff
@@ -17,14 +29,54 @@ public ObjectMapper objectMapper() {
 		return JsonMapper.builder().enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS).build();
 	}
 
+	@Bean
```
sagehen03, line 32: There's a set of cascading changes here.  We need connections with `autoCommit=false` in order to not load the whole sql resultset in memory, but the rest of our db code relies on `autoCommit=true` so we need a new connection pool.  As soon as we introduce a declared connection pool, Spring uses that one and no longer creates a default one so we need to also set up another DataSource to replace the old default one.  A similar thing is at work with the NamedParameterJdbcTemplate beans.  It might be possible to do all this by temporarily changing the settings on a sql connection and sql statement when we're using those particular instances, but that seems error prone and likely to introduce hard to repro bugs.

service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
```diff
@@ -17,14 +29,54 @@ public ObjectMapper objectMapper() {
...
+	}
+
+
 	// CORS: allow Ajax requests from anywhere for all endpoints.
 	@Bean
 	public WebMvcConfigurer corsConfigurer() {
 		return new WebMvcConfigurer() {
 			@Override
-			public void addCorsMappings(CorsRegistry registry) {
+			public void addCorsMappings(@NonNull CorsRegistry registry) {
```
sagehen03, line 72: Intellij was giving me a warning here.

service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
```diff
@@ -17,14 +29,54 @@ public ObjectMapper objectMapper() {
...
 				registry.addMapping("/**").allowedOrigins("*");
 			}
+
+			@Override
+			public void configureAsyncSupport(@NonNull AsyncSupportConfigurer configurer) {
```
sagehen03, line 77: No timeout now for async requests.  I don't think this is an issue since it's single tenant--a user will be more forgiving if they've managed to DOS themselves.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -72,6 +75,27 @@ public ResponseEntity<RecordResponse> getSingleRecord(@PathVariable("instanceId"
...
 
+	@GetMapping("{instanceId}/tsv/{version}/{recordType}")
+	public ResponseEntity<StreamingResponseBody> streamAllEntities(@PathVariable("instanceId") UUID instanceId,
+																   @PathVariable("version") String version,
+																   @PathVariable("recordType") RecordType recordType) {
+		validateVersion(version);
+		validateInstance(instanceId);
+		checkRecordTypeExists(instanceId, recordType);
+		List<String> headers = new ArrayList<>(Collections.singletonList(ReservedNames.RECORD_ID));
+		headers.addAll(recordDao.getExistingTableSchema(instanceId, recordType).keySet());
```
sagehen03, line 86: Not sure if we need to do some column ordering?  The `headers` list will determined the left to right ordering of the columns in the tsv output.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -166,8 +171,8 @@ public void batchUpsert(UUID instanceId, RecordType recordType, List<Record> rec
 		} catch (DataAccessException e) {
 			if (e.getRootCause()instanceof SQLException sqlEx) {
 				checkForMissingRecord(sqlEx);
-				throw e;
 			}
+			throw e;
```
sagehen03, line 175: I noticed while testing that we were swallowing exceptions here.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -190,6 +195,7 @@ public void batchUpsertWithErrorCapture(UUID instanceId, RecordType recordType,
 					throw new BatchWriteException(rowErrors);
 				}
 			}
+			throw e;
```
sagehen03, line 198: Also swallowing exceptions here.

service/src/main/java/org/databiosphere/workspacedataservice/service/TsvSupport.java
```diff
@@ -0,0 +1,47 @@
...
+
+public class TsvSupport {
+
+    private TsvSupport() {
+    }
+
+    public static CSVFormat getOutputFormat(List<String> headers) {
+         return CSVFormat.DEFAULT.builder().setDelimiter('\t')
+                .setQuoteMode(QuoteMode.MINIMAL)
+                 .setRecordSeparator("\n")
```
sagehen03, line 20: I could see that we may want to use the line separator for the client browser's OS rather than the Linux new line, but I'm not sure if it matters?

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -163,6 +163,31 @@ paths:
...
+        - $ref: '#/components/parameters/versionPathParam'
+        - $ref: '#/components/parameters/recordTypePathParam'
+      responses:
+        200:
+          description: Records in tsv format
+          content:
+            text/plain:
+              schema:
+                type: string
+                format: binary
```
sagehen03, line 184: Cribbed from Orch, is this right?  I don't think of TSV as binary.

service/src/test/java/org/databiosphere/workspacedataservice/controller/TsvDownloadTest.java
```diff
@@ -0,0 +1,62 @@
...
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+class TsvDownloadTest {
+
+    @Autowired
+    private TestRestTemplate restTemplate;
+
+    @Autowired
+    private RecordController recordController;
+
+    @Test
```
sagehen03, line 35: Marking this test as `@Transactional` causes it to fail.  I think that's because we're using two connection pools so we can't share a transaction.  Thus, the code to read the tsv won't see the tsv records.  Downside of all this means that this test code will leave data in the test db.  I don't think that's a huge deal since our test db servers are ephemeral?

service/src/main/resources/application.properties
```diff
@@ -1,8 +1,11 @@
-spring.datasource.driver-class-name=org.postgresql.Driver
-spring.datasource.url=jdbc:postgresql://${WDS_DB_HOST:localhost}:${WDS_DB_PORT:5432}/wds?reWriteBatchedInserts=true
+spring.datasource.jdbc-url=jdbc:postgresql://${WDS_DB_HOST:localhost}:${WDS_DB_PORT:5432}/wds?reWriteBatchedInserts=true
 spring.datasource.username=${WDS_DB_USER:wds}
 spring.datasource.password=${WDS_DB_PASSWORD:wds}
 
+streaming.query.jdbc-url=jdbc:postgresql://${WDS_DB_HOST:localhost}:${WDS_DB_PORT:5432}/wds
```
sagehen03, line 5: We could reuse the exact url from `spring.datasource.jdbc-url` but then you get an IntelliJ warning where we reference that property.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -72,6 +75,27 @@ public ResponseEntity<RecordResponse> getSingleRecord(@PathVariable("instanceId"
...
+		Stream<Record> allRecords = recordDao.streamAllRecordsForType(instanceId, recordType);
+
+		StreamingResponseBody responseBody = httpResponseOutputStream -> {
+			try (CSVPrinter writer = TsvSupport.getOutputFormat(headers).print(new OutputStreamWriter(httpResponseOutputStream))) {
+				TsvSupport.RecordEmitter recordEmitter = new TsvSupport.RecordEmitter(writer, headers.subList(1, headers.size()));
+				allRecords.forEach(recordEmitter);
+			}
+		};
+		return ResponseEntity.status(HttpStatus.OK).contentType(new MediaType("text", "csv")).body(responseBody);
+	}
```
calypsomatic, line 96: There's some mix of CSV and TSV here, with apparently a CSV at the end.  How important is it that we do a TSV instead of a CSV?

sagehen03: I tried it with tsv and my Firefox didn't open my spreadsheet program when I a made a GET request to this controller through my browser so I switched to "csv".  We should experiment with a few different browsers and choose the mime type that does the optimal thing in the highest percentage of cases.

service/src/test/java/org/databiosphere/workspacedataservice/controller/TsvDownloadTest.java
```diff
@@ -0,0 +1,62 @@
...
+    void batchWriteFollowedByTsvDownload() throws IOException {
+        RecordType recordType = RecordType.valueOf("tsv-test");
+        String version = "v0.2";
+        UUID instanceId = UUID.randomUUID();
+        recordController.createInstance(instanceId, version);
+        InputStream is = FullStackRecordControllerTest.class.getResourceAsStream("/batch_write_tsv_data.json");
```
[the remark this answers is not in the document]
sagehen03: That's a mistake, though it doesn't really matter, updated.

service/src/main/java/org/databiosphere/workspacedataservice/service/TsvSupport.java
davidangb: meh, I say stick with just `\n`

calypsomatic (approved):

service/src/main/resources/static/swagger/openapi-docs.yaml
davidangb: I _think_ this should be:
```yaml
content:
  text/tab-separated-values:
    schema:
      type: string
      format: binary
```

and therefore Orchestration is wrong in its specification of `text/plain` but correct in its specification of `format: binary`. The `format: binary` looks required by OpenAPI; search for "Response That Returns a File" at https://swagger.io/docs/specification/describing-responses/

service/src/test/java/org/databiosphere/workspacedataservice/controller/TsvDownloadTest.java
```diff
@@ -0,0 +1,62 @@
...
+        String version = "v0.2";
+        UUID instanceId = UUID.randomUUID();
+        recordController.createInstance(instanceId, version);
+        InputStream is = TsvDownloadTest.class.getResourceAsStream("/batch_write_tsv_data.json");
+        ResponseEntity<BatchResponse> response = recordController.streamingWrite(instanceId, version, recordType, is);
+        assertThat(response.getStatusCodeValue()).isEqualTo(200);
+        HttpHeaders headers = new HttpHeaders();
+        headers.setContentType(MediaType.APPLICATION_JSON);
+        ResponseEntity<Resource> stream = restTemplate.exchange("/{instanceId}/tsv/{version}/{recordType}",
+                HttpMethod.GET, new HttpEntity<>(headers), Resource.class, instanceId, version, recordType);
```
davidangb, line 47: for my edification: does this actually require setting `MediaType.APPLICATION_JSON)` on the request?

davidangb (commented):
See comments inline!

Also - there's some funky indentation here, could you format the changed files with Spotless? Obviously the Spotless integration isn't working seamlessly, we should make a follow-on ticket to make spotless more automated.

service/src/test/java/org/databiosphere/workspacedataservice/controller/TsvDownloadTest.java
davidangb: Maybe a good compromise - as part of this PR or as a follow-on - is to create a "deleteInstance" method that drops the schema. Since this test creates a new instance at a random UUID, it could drop that schema when its done.

service/src/test/java/org/databiosphere/workspacedataservice/controller/TsvDownloadTest.java
sagehen03: It does not.  It was copy pasta.  Good eye, removed.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -275,6 +281,11 @@ private void checkForTableRelation(SQLException sqlEx) {
 		}
 	}
 
+	public Stream<Record> streamAllRecordsForType(UUID instanceId, RecordType recordType){
+		return templateForStreaming.getJdbcTemplate().queryForStream("select * from " + getQualifiedTableName(recordType, instanceId),
+				new RecordRowMapper(recordType, getRelationColumnsByName(getRelationCols(instanceId, recordType))));
+	}
```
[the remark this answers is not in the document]
sagehen03: I've added an order by sys_name.  I think the important thing here is we return all records. My guess is users would sort or filter in a spreadsheet program afterward or just immediately upload to some other place so I'm skeptical that a variation in the order of the rows in the tsv output would cause confusion, but it's also very easy to add the order by so I did that.

service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
```diff
@@ -17,14 +29,54 @@ public ObjectMapper objectMapper() {
...
+
+	@Bean
+	@Qualifier("streamingDs")
+	public NamedParameterJdbcTemplate templateForStreaming(@Qualifier("streamingDs") DataSource ds,
+														   @Value("${twds.streaming.fetch.size:50}") int fetchSize){
+		NamedParameterJdbcTemplate jdbcTemplate = new NamedParameterJdbcTemplate(ds);
+		//https://jdbc.postgresql.org/documentation/query/#getting-results-based-on-a-cursor
+		jdbcTemplate.getJdbcTemplate().setFetchSize(fetchSize);
+		return jdbcTemplate;
+	}
```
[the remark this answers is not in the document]
sagehen03: Updated, I had it low because I didn't see much of a performance improvement when I tested higher numbers and thought the increased memory usage wasn't worth the tradeoff, but a 15% perf improvement is probably a better tradeoff than the memory economy.

sagehen03:
> See comments inline!
> 
> Also - there's some funky indentation here, could you format the changed files with Spotless? Obviously the Spotless integration isn't working seamlessly, we should make a follow-on ticket to make spotless more automated.

I see that in github, but it looks fine my IDE.  I've run spotless.

service/src/test/java/org/databiosphere/workspacedataservice/controller/TsvDownloadTest.java
sagehen03: I vote for later :grimacing:

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -72,6 +75,28 @@ public ResponseEntity<RecordResponse> getSingleRecord(@PathVariable("instanceId"
...
+		StreamingResponseBody responseBody = httpResponseOutputStream -> {
+			try (CSVPrinter writer = TsvSupport.getOutputFormat(headers)
+					.print(new OutputStreamWriter(httpResponseOutputStream))) {
+				TsvSupport.RecordEmitter recordEmitter = new TsvSupport.RecordEmitter(writer,
+						headers.subList(1, headers.size()));
+				allRecords.forEach(recordEmitter);
+			}
+		};
+		return ResponseEntity.status(HttpStatus.OK).contentType(new MediaType("text", "tab-separated-values"))
+				.body(responseBody);
```
davidangb, line 97: I've been trying to figure out why this didn't behave like the Orch TSV download. It's not the swagger definition; I believe this needs:
```java
.header(HttpHeaders.CONTENT_DISPOSITION, "attachment;filename=" + recordType.getName() + ".tsv")
```
to tell it to download the file.

davidangb (approved):
Approving, but note one more comment about adding a `Content-Disposition` header to better indicate that this is a file download.

PR #83, opened 2022-10-04
AJ-611: don't require trailing slash for create-instance API

davidangb:
title.

This was tripping up CBAS integration because us humans were confused by it


calypsomatic (approved):
Who put that there?! (I think it was me)

sagehen03 (approved):

PR #86, opened 2022-10-05
aj-615 paginated sort by column

calypsomatic:
[AJ-615](https://broadworkbench.atlassian.net/browse/AJ-615)
Allows sorting by a given attribute.  Naive implementation of however postgres interprets the sort.  Defaults to RECORD_ID if none given.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -86,14 +86,14 @@ public class RecordController {
 		}
 
 		if(searchRequest.getSortAttribute() != null && !recordDao.getExistingTableSchema(instanceId, recordType).keySet().contains(searchRequest.getSortAttribute())){
-			throw new MissingObjectException("Record attribute");
+			throw new MissingObjectException("Requested sort attribute");
 		}
 		int totalRecords = recordDao.countRecords(instanceId, recordType);
 		if (searchRequest.getOffset() > totalRecords) {
 			return new RecordQueryResponse(searchRequest, Collections.emptyList(), totalRecords);
 		}
 		List<Record> records = recordDao.queryForRecords(recordType, searchRequest.getLimit(),
-				searchRequest.getOffset(), searchRequest.getSort().name().toLowerCase(), instanceId, searchRequest.getSortAttribute());
+				searchRequest.getOffset(), searchRequest.getSort().name().toLowerCase(), searchRequest.getSortAttribute(), instanceId);
 		List<RecordResponse> recordList = records.stream().map(
 				r -> new RecordResponse(r.getId(), r.getRecordType(), r.getAttributes(), new RecordMetadata("UNUSED")))
 				.toList();
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -84,12 +84,16 @@ public class RecordController {
 			throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
 					"Limit must be more than 0 and can't exceed " + MAX_RECORDS + ", and offset must be positive.");
 		}
+
+		if(searchRequest.getSortAttribute() != null && !recordDao.getExistingTableSchema(instanceId, recordType).keySet().contains(searchRequest.getSortAttribute())){
+			throw new MissingObjectException("Record attribute");
+		}
 		int totalRecords = recordDao.countRecords(instanceId, recordType);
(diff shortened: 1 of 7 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/controller/FullStackRecordControllerTest.java
```diff
@@ -94,6 +96,43 @@ void testQuerySuccess() throws Exception {
...
+		body = executeQuery(recordType, RecordQueryResponse.class, sortByInt)
+				.getBody();
+		assertThat(body.records().get(0).recordAttributes().getAttributeValue("attr3"))
+				.as("Record with attr3 1 should be first record in ascending order").isEqualTo(1);
+		assertThat(body.records().get(4).recordAttributes().getAttributeValue("attr3")).isEqualTo(5);
+		SearchRequest sortByDate = new SearchRequest(limit, offset, SortDirection.DESC, "attr-dt");
+		body = executeQuery(recordType, RecordQueryResponse.class, sortByDate)
+				.getBody();
+		assertThat(body.records()).hasSize(limit);
+		assertThat(Instant.ofEpochMilli((Long)body.records().get(0).recordAttributes().getAttributeValue("attr-dt")).atZone(ZoneId.systemDefault()).toLocalDate()).as("Record with attr-dt 2022-04-25 should be first record in descending order")
```
calypsomatic, line 130: Somehow really struggled with translating between the Long value of the dates and the String value that I could understand.  Will welcome a better way to deal with this, and/or will have to be updated if/when we change how dates are returned.

sagehen03: Sorry, I'm likely at fault here. I thought this [line](https://github.com/DataBiosphere/terra-workspace-data-service/blob/2f4874af8675e0d81a5c687cb0d94c90373b4470/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java#L430) would put a LocalDate or a LocalDateTime into the attribute map, but it looks like it defaults to Date and TimeStamp types.  We can get the code to do the right thing if include the schema info in the RowMapper so that the rs call look like:  

```
Object object = rs.getObject(columnName, schema.get(columnName).getJavaType());
```

This might help with your issue, but it imposes another db call to fetch the schema, and at the moment we don't do anything with the attributes other than serialize them to json so David's [approach](https://github.com/DataBiosphere/terra-workspace-data-service/pull/87) might be better, but I'm reluctant to lose that typed info lest we need it in the future.

sagehen03 (approved):
I didn't look very closely at the test code (it seems like you have a lot there), but controller and dao code looks reasonable to me.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -87,9 +87,9 @@ private String getQualifiedTableName(RecordType recordType, UUID instanceId) {
 
 	@SuppressWarnings("squid:S2077")
 	public List<Record> queryForRecords(RecordType recordType, int pageSize, int offset, String sortDirection,
-			UUID instanceId) {
+			UUID instanceId, String sortAttribute) {
```
davidangb, line 90: nit: I think the method signature would be cleaner if `sortDirection` and `sortAttribute` are next to each other, instead of having `instanceId` sitting between them. I believe there's only one caller so this shouldn't incur too much refactoring (and IntelliJ can reorder those anyway)

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -84,12 +84,16 @@ public RecordQueryResponse queryForEntities(@PathVariable("instanceid") UUID ins
 			throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
 					"Limit must be more than 0 and can't exceed " + MAX_RECORDS + ", and offset must be positive.");
 		}
+
+		if(searchRequest.getSortAttribute() != null && !recordDao.getExistingTableSchema(instanceId, recordType).keySet().contains(searchRequest.getSortAttribute())){
+			throw new MissingObjectException("Record attribute");
```
davidangb, line 89: nit: change to `MissingObjectException("Sort attribute")` or `MissingObjectException("Requested sort attribute")` or similar, to help our users

davidangb (approved):
Minor comments inline, hopefully not very big.

On the request payload: this will allow sorting by one attribute only, we could not do a compound sort e.g. `order by col1 asc, col2 desc`. I realize this is out of scope - and I'm not suggesting a change in this PR, just thinking out loud: if we need to support compound sorts down the line, we'll need to change that signature.

Proactively approving - if you change anything other than addressing the comments, I'll re-review!

PR #87, opened 2022-10-05
dates and timestamps should be human-readable [AJ-620]

davidangb:
* Datetime attributes now serialize to e.g. `2022-04-25T12:00:01`
* Date attributes now serialize to e.g. `2022-04-25`
* The `RecordAttributes` underlying `Map` now contains `LocalDate` and `LocalDateTime` values

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
@@ -2,6 +2,7 @@ package org.databiosphere.workspacedataservice;
 
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.json.JsonMapper;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
@@ -14,7 +15,8 @@ public class WorkspaceDataServiceApplication {
 
 	@Bean
 	public ObjectMapper objectMapper() {
-		return JsonMapper.builder().enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS).build();
+		return JsonMapper.builder().enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS)
+				.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false).findAndAddModules().build();
 	}
 
 	// CORS: allow Ajax requests from anywhere for all endpoints.
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -419,9 +419,9 @@ public class RecordDao {
 						Object object = rs.getObject(columnName);
 
 						if (object instanceof java.sql.Date date) {
-							object = date.toString();
+							object = date.toLocalDate();
 						} else if (object instanceof java.sql.Timestamp ts) {
-							object = ts.toString();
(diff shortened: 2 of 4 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -417,8 +417,16 @@ private RecordAttributes getAttributes(ResultSet rs) {
...
+
+						if (object instanceof java.sql.Date date) {
+							object = date.toString();
+						} else if (object instanceof java.sql.Timestamp ts) {
+							object = ts.toString();
+						} else if (object instanceof PGobject pGobject) {
+							object = pGobject.getValue();
+						}
+
+						attributes.putAttribute(columnName, object);
```
sagehen03, line 429: This approach is one step closer to Map<String, String> to keep the attributes in memory.  The current functionality doesn't require anything more, but we might need it at some point.  We can use something like `Object object = rs.getObject(columnName, schema.get(columnName).getJavaType());` to get the properly typed java objects into the attribute map, but it does require the schema info in the RowMapper.

davidangb: An approach that would keep the real java types in the attribute map:
```java
						if (object instanceof java.sql.Date date) {
							object = date.toLocalDate();
						} else if (object instanceof java.sql.Timestamp ts) {
							object = ts.toLocalDateTime();
						} else if (object instanceof PGobject pGobject) {
							object = pGobject.getValue();
						}
```

If we do that, then we just need to configure the `ObjectMapper` to serialize `LocalDate` and `LocalDateTime`

service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
```diff
@@ -14,7 +15,8 @@ public class WorkspaceDataServiceApplication {
 
 	@Bean
 	public ObjectMapper objectMapper() {
-		return JsonMapper.builder().enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS).build();
+		return JsonMapper.builder().enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS)
+				.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false).findAndAddModules().build();
```
davidangb, line 19: the `.findAndAddModules()` could be reduced to just `mapper.registerModule(new JavaTimeModule());` but I believe the former is the default Spring behavior, so I'm choosing that approach.

service/src/test/java/org/databiosphere/workspacedataservice/controller/FullStackRecordControllerTest.java
```diff
@@ -126,10 +127,12 @@ void testSortedQuerySuccess() throws Exception {
...
-		assertThat(Instant.ofEpochMilli((Long)body.records().get(0).recordAttributes().getAttributeValue("attr-dt")).atZone(ZoneId.systemDefault()).toLocalDate()).as("Record with attr-dt 2022-04-25 should be first record in descending order")
-				.isEqualTo("2022-04-25");
-		assertThat(Instant.ofEpochMilli((Long)body.records().get(4).recordAttributes().getAttributeValue("attr-dt")).atZone(ZoneId.systemDefault()).toLocalDate())
-				.isEqualTo("2002-03-23");
+
+		assertThat(body.records().get(0).recordAttributes().getAttributeValue("attr-dt"))
+				.as("Record with attr-dt 2022-04-25 should be first record in descending order")
+				.isEqualTo("2022-04-25T12:00:01");
+		assertThat(body.records().get(4).recordAttributes().getAttributeValue("attr-dt"))
+				.isEqualTo("2002-03-23T12:00:01");
```
davidangb, line 135: this test expected datetimes to be returned as epoch millis; they are now returned as human-readable strings.

sagehen03 (approved):

PR #88, opened 2022-10-06
AJ-596: download TSV as attachment

davidangb:
follow-on to #81 : when calling the TSV endpoint, download the result as a file by setting `Content-Disposition: attachment`

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -10,6 +10,7 @@ import org.databiosphere.workspacedataservice.service.model.*;
 import org.databiosphere.workspacedataservice.service.model.exception.MissingObjectException;
 import org.databiosphere.workspacedataservice.shared.model.Record;
 import org.databiosphere.workspacedataservice.shared.model.*;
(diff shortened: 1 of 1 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -110,15 +112,17 @@ public RecordQueryResponse queryForEntities(@PathVariable("instanceid") UUID ins
 					"Limit must be more than 0 and can't exceed " + MAX_RECORDS + ", and offset must be positive.");
 		}
 
-		if(searchRequest.getSortAttribute() != null && !recordDao.getExistingTableSchema(instanceId, recordType).keySet().contains(searchRequest.getSortAttribute())){
+		if (searchRequest.getSortAttribute() != null && !recordDao.getExistingTableSchema(instanceId, recordType)
+				.keySet().contains(searchRequest.getSortAttribute())) {
```
davidangb, line 116: spotless reformat

jordansafer (approved):

sagehen03 (approved):
FWIW in my testing it downloaded as a file already so I wasn't aware of this issue, but I know there's variation between browsers.

PR #89, opened 2022-10-07
MDC logging [AJ-444]

davidangb:
In this PR:
* ~bring in [Lombok](https://projectlombok.org/) for its super-easy `@Slf4j` annotation~
* ~annotate `RecordController` and `RecordDao` with `@Slf4j`~
* add example logging statements to `RecordController.queryForEntities()` and `RecordDao.queryForRecords()`
* create a `MDCFilter` servlet filter that adds a request id to MDC logging
  * look for a request id in the incoming request; use it if found
  * else, generate a new UUID and use that value
* configure a logback-spring.xml file based on https://www.baeldung.com/spring-boot-logging but that also adds the MDC RequestId


service/src/main/java/org/databiosphere/workspacedataservice/service/MDCFilter.java
```diff
@@ -0,0 +1,63 @@
...
+@Component
+public class MDCFilter implements Filter {
+
+    // where the unique request id is stored in the MDC context
+    static final String MDC_KEY = "requestId";
+    // the response header containing the unique request id
+    public static final String RESPONSE_HEADER = "x-b3-traceid";
+    // the list of request headers, in order, we will search for an incoming unique request id
+    // we could add other headers here, such as: x-vcap-request-id, X─B3─ParentSpanId, X─B3─SpanId, X─B3─Sampled
+    static final List<String> INCOMING_HEADERS = Arrays.asList("x-b3-traceid", "x-request-id", "trace-id", "x-amzn-trace-id");
```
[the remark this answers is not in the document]
jordansafer: Do we need x-amzn-trace headers? I’m wondering what the context is where we’d want to use other types of request headers. I know x-amzn-trace is used within aws

jordansafer (approved):

sagehen03 (approved):
I think this will work for our current set up and we can always modify later to adjust according to our deployment environment so I'm approving, but I have some questions. 

1. Did you consider using a ServletRequestListener rather than a filter?  I don't know all the details, but since everything goes through one servlet with Spring using a filter is fine, but if we (an unlikely scenario I know) added another servlet that we wanted MDC request logging we'd need to add this filter to that servlet while a listener would just work with not changes.  The listener approach also seems to have the advantage of the requestDestroyed() callback which would be an ideal place for...
2. You don't call MDC.clear() or MDC.remove() in this implementation.  I think you'll always end up updating the map to the unique id for the request but belt and suspenders might be nice so that we don't under unforeseen circumstances get log lines erroneously attached to unrelated requests.
3. Do we need to add all the logback config now?  I'm not sure what the best or existing practice is for Azure logging, but my preference would be to wait until we know more details there.  If what you've added here will plug in play with Azure, then ignore this comment.
4. I don't want to be a party pooper but it looks lombok substitutes typing something like `private static final Logger LOGGER = LoggerFactory.getLogger(DataTypeInferer.class);` at the top of a file in exchange for `@Slf4j`.  My feeling is if we want to save typing why not do so directly with the IDE rather than through whatever less direct means that Lombok uses.  I just type `sl+TAB` in IntelliJ to get the same result with a live template that I'm happy to share.  However, I know part of the fun of software is to use and learn new stuff so I don't want to stand in the way of Lombok adoption if that's what folks want.  I've just had problems with inscrutable version compatibility issues with Lombok, and my core argument is if we're using it to save typing why not just save typing more directly by using the IDE to do the typing for us?
5. I don't know the exact details but I'm not sure we'll get the desired behavior for async request processing which we use to stream all records because I think most MDC implementations use ThreadLocal variables and assumes the same thread will process the request all the way through?

service/src/main/java/org/databiosphere/workspacedataservice/service/MDCFilter.java
```diff
@@ -0,0 +1,63 @@
...
+        MDC.put(MDC_KEY, uniqueId);
+
+        log.trace("Request IP address is {}", servletRequest.getRemoteAddr());
+        log.trace("Request content type is {}", servletRequest.getContentType());
+
+        // add id to the response
+        if (servletResponse instanceof HttpServletResponse httpServletResponse) {
+            httpServletResponse.setHeader(RESPONSE_HEADER, uniqueId);
+            log.trace("Response header is set with uuid {}", uniqueId);
+        }
```
sagehen03, line 59: What is the use case for adding the uniqueId for the response headers?

service/src/main/java/org/databiosphere/workspacedataservice/service/MDCFilter.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse, FilterChain filterChain) throws IOException, ServletException {
+        String uniqueId = null;
+        // look for a non-blank unique request id in the incoming request headers
+        if (servletRequest instanceof HttpServletRequest httpServletRequest) {
+            for (String hdr : INCOMING_HEADERS) {
+                String foundIncoming = httpServletRequest.getHeader(hdr);
+                if (foundIncoming != null && !foundIncoming.isBlank()) {
+                    uniqueId = foundIncoming;
```
sagehen03, line 38: Should we verify that the header passed to us is a UUID or add some other validation?

davidangb:
@jordansafer I removed support for `x-amzn-trace`; there's no need to support it now, and we can always add it in later if we need it.

@sagehen03 we discussed a lot of this in person, but here goes:
1. I split the logic between a ServletRequestListener (look in request headers, add id to MDC, clear MDC on requestDestroyed) and a Filter (if request id in MDC, add it to response headers). We need the Filter since ServletRequestListener can't write to response headers.
2. we don't need all the logback config now, but it was mostly a big copy/paste; if we find problems with it in Azure or elsewhere we can change it at that point - I'd prefer not to do the work to figure out which parts we want and which we don't.
3. I removed Lombok and am grabbing Loggers directly from SLF4J now
4. yes - the async responses jump threads and therefore lose the request id in the logs. This is what I see:

```
2022-10-11 17:22:43,484 DEBUG [ebd7f518-04e6-4ee4-b41a-f38afe1311ef] org.springframework.core.log.LogFormatUtils: GET "/e558c1e0-4d09-4c2f-90bc-2ac9877dcaee/tsv/v0.2/thing", parameters={}
2022-10-11 17:22:43,485 DEBUG [ebd7f518-04e6-4ee4-b41a-f38afe1311ef] org.springframework.web.servlet.handler.AbstractHandlerMapping: Mapped to org.databiosphere.workspacedataservice.controller.RecordController#streamAllEntities(UUID, String, RecordType)
2022-10-11 17:22:43,499 DEBUG [ebd7f518-04e6-4ee4-b41a-f38afe1311ef] org.springframework.web.context.request.async.WebAsyncManager: Started async request
2022-10-11 17:22:43,499 DEBUG [ebd7f518-04e6-4ee4-b41a-f38afe1311ef] org.springframework.web.servlet.FrameworkServlet: Exiting but response remains open for further handling
2022-10-11 17:22:43,501 DEBUG [] org.springframework.web.context.request.async.WebAsyncManager: Async result set, dispatch to /e558c1e0-4d09-4c2f-90bc-2ac9877dcaee/tsv/v0.2/thing
2022-10-11 17:22:43,501 DEBUG [] org.springframework.core.log.LogFormatUtils: "ASYNC" dispatch for GET "/e558c1e0-4d09-4c2f-90bc-2ac9877dcaee/tsv/v0.2/thing", parameters={}
2022-10-11 17:22:43,502 DEBUG [] org.springframework.core.log.LogFormatUtils: Resume with async result []
2022-10-11 17:22:43,502 DEBUG [] org.springframework.web.servlet.FrameworkServlet: Exiting from "ASYNC" dispatch, status 200
```

I was wrong about the async result having a second uuid … I definitely saw that at one point but now I'm not sure how I made that happen :(

sagehen03 (approved):

PR #91, opened 2022-10-14
AJ-581: Add arrays as data type

sagehen03:
This PR adds the ability to upload a JSON representation of an array and store that data in a postgres array column.  Boolean, String, Number, LocalDate, and LocalDateTime are the supported array types. For the date types it was necessary to convert to String[] and then cast in the SQL insert in order for db writes to succeed. Handling both the TSV and the JSON representations of the data increases the complexity.


service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
```diff
@@ -27,8 +32,19 @@ public class WorkspaceDataServiceApplication {
 
 	@Bean
 	public ObjectMapper objectMapper() {
-		return JsonMapper.builder().enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS)
-				.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false).findAndAddModules().build();
+		JsonMapper mapper = JsonMapper.builder().enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS)
+				.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false)
+				.configure(DeserializationFeature.ACCEPT_FLOAT_AS_INT, false)
+				.findAndAddModules().build();
+		mapper.coercionConfigFor(LogicalType.Boolean).setCoercion(CoercionInputShape.Integer, CoercionAction.Fail);
```
sagehen03, line 39: Don't coerce 0 to false and everything else to true.

service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
```diff
@@ -27,8 +32,19 @@ public class WorkspaceDataServiceApplication {
...
+		mapper.coercionConfigFor(LogicalType.Float).setCoercion(CoercionInputShape.String, CoercionAction.Fail);
+		mapper.coercionConfigFor(LogicalType.Integer).setCoercion(CoercionInputShape.String, CoercionAction.Fail);
```
sagehen03, line 41: Quoted numbers should stay strings, aka don't make ["1", "2"] an array of long.

service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
```diff
@@ -27,8 +32,19 @@ public class WorkspaceDataServiceApplication {
 
 	@Bean
 	public ObjectMapper objectMapper() {
-		return JsonMapper.builder().enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS)
-				.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false).findAndAddModules().build();
+		JsonMapper mapper = JsonMapper.builder().enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS)
+				.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false)
+				.configure(DeserializationFeature.ACCEPT_FLOAT_AS_INT, false)
```
sagehen03, line 37: Don't convert floats to ints.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -127,7 +128,7 @@ public Map<String, DataTypeMapping> getExistingTableSchema(UUID instanceId, Reco
 		params.addValue("tableName", recordType.getName());
 		params.addValue("recordName", RECORD_ID);
 		return namedTemplate
-				.query("select column_name, data_type from INFORMATION_SCHEMA.COLUMNS where table_schema = :instanceId "
+				.query("select column_name, udt_name::regtype as data_type from INFORMATION_SCHEMA.COLUMNS where table_schema = :instanceId "
```
sagehen03, line 131: This change was necessary to get the array type aka text[] as opposed to just returning ARRAY.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -368,9 +369,43 @@ private Object getValueForSql(Object attVal, DataTypeMapping typeMapping) {
...
+				return attVal.toArray(new Boolean[0]);
+			case ARRAY_OF_DOUBLE:
+				List<Object> doubleList = attVal.stream().map(i -> {
+					if (i instanceof Integer){
+						return (double) ((Integer) i).intValue();
+					}
+					else {
+						return i;
+					}
+				}).toList();
```
sagehen03, line 399: I hate this, but I haven't been able to find a better way.  If Spring serializes from JSON something like `attrFoo: [1, 2.2]`, that will reach this part of the code as an ArrayList containing an Integer and Double and Java won't let us just call `attVal.toArray(new Double[0])` https://stackoverflow.com/questions/5566336/classcastexception-casting-integer-to-double

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -82,6 +96,8 @@ public DataTypeMapping inferTypeForTsvSource(Object val) {
 		if (RelationUtils.isRelationValue(val)) {
 			return STRING;
 		}
+		//libreoffice at least uses left and right quotes which cause problems when we try to parse as JSON
+		sVal = sVal.replaceAll("[“”]", "\"");
```
sagehen03, line 100: When I created a TSV in my spreadsheet program it used left and right quotes, which made Jackson puke.

service/src/main/java/org/databiosphere/workspacedataservice/service/model/DataTypeMapping.java
```diff
@@ -2,33 +2,41 @@
...
 
 	static {
-		MAPPING_BY_PG_TYPE.put("date", DATE);
-		MAPPING_BY_PG_TYPE.put("timestamp with time zone", DATE_TIME);
-		MAPPING_BY_PG_TYPE.put("text", STRING);
-		MAPPING_BY_PG_TYPE.put("jsonb", JSON);
-		MAPPING_BY_PG_TYPE.put("bigint", LONG);
-		MAPPING_BY_PG_TYPE.put("numeric", DOUBLE);
-		MAPPING_BY_PG_TYPE.put("boolean", BOOLEAN);
+		Arrays.stream(DataTypeMapping.values()).forEach(e -> MAPPING_BY_PG_TYPE.put(e.getPostgresType(), e));
```
sagehen03, line 33: Oops I realized there was a more concise way to do this and keep things DRYer

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
sagehen03: It's also worth mentioning that arrays with very large numbers aren't supported.  The json serialization will convert those numbers to BigDecimal of BigInteger, but postgres' supported numeric array types can't handle numbers that size.  I think the best we can do in that case is store the data as text or json.  It also looks like very big numbers for plain (non-array) numeric attributes will overflow which is not unexpected since we're coercing to Double and Long.  All of this opens up the question of what scale of numeric values do we need to support?

service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
```diff
@@ -26,9 +30,18 @@ public class WorkspaceDataServiceApplication {
...
 				.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false)
 				.findAndAddModules().build();
+		mapper.coercionConfigFor(LogicalType.Boolean).setCoercion(CoercionInputShape.Integer, CoercionAction.Fail);
+		mapper.coercionConfigFor(LogicalType.Float).setCoercion(CoercionInputShape.String, CoercionAction.Fail);
+		mapper.coercionConfigFor(LogicalType.Integer).setCoercion(CoercionInputShape.String, CoercionAction.Fail);
+		return mapper;
+	}
+
+	@Bean
+	public DataTypeInferer inferer(ObjectMapper mapper){
```
sagehen03, line 43: We've changed a number of defaults on the Object mapper, there are a few changes like this to share a single instance across the app so that we get consistent JSON (de)serialization behavior.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -57,6 +64,12 @@ public DataTypeMapping selectBestType(DataTypeMapping existing, DataTypeMapping
 		if (newMapping == NULL || existing == NULL) {
 			return newMapping != NULL ? newMapping : existing;
 		}
+		if (existing.isArrayType() && newMapping.isArrayType() && Set.of(existing, newMapping).contains(EMPTY_ARRAY)) {
+			return newMapping != EMPTY_ARRAY ? newMapping : existing;
+		}
+		if(newMapping.isArrayType() && existing.isArrayType()){
+			return ARRAY_OF_STRING;
```
calypsomatic, line 71: I'm having a hard time following where something is for sql or tsv or whatnot - so this looks like `ARRAY_OF_STRING` is the only array type we have, so is this method just for the database?

service/src/test/resources/small-test.tsv
```diff
@@ -1,5 +1,5 @@
-sys_name	greeting	unicode	json	date	embedded characters	double	long
+sys_name	greeting	unicode	json	date	embedded characters	double	long	z_array_of_string	z_long_array	z_double_array	z_z_boolean_array	zz_array_of_date	zz_array_of_datetime
```
calypsomatic, line 1: Are these z's just to keep it alphabetical order?

sagehen03: yes!

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
calypsomatic: Aha!  Gotcha.  Thanks

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -469,6 +497,13 @@ private RecordAttributes getAttributes(ResultSet rs) {
 							object = ts.toLocalDateTime();
 						} else if (object instanceof PGobject pGobject) {
 							object = pGobject.getValue();
+						} else if (object instanceof PgArray pgArray) {
+							object = pgArray.getArray();
+							if(object.getClass() == Timestamp[].class) {
+								object = convertToLocalDateTime(object);
+							} else if(object.getClass() == Date[].class) {
+								object = convertToLocalDate(object);
```
davidangb, line 505: I suggest using `object instanceof Date[] dtArray` to cast the object up here, instead of inside `convertToLocalDate` - and changing `convertToLocalDate` to accept a `Date[]` instead of an `Object`. This is minor, but it prevents a future coder from a bad call to `convertToLocalDate` and then running into a ClassCastException. Same for `convertToLocalDateTime`.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -479,6 +514,24 @@ private RecordAttributes getAttributes(ResultSet rs) {
...
+			}
+			return result;
+		}
+		private LocalDate[] convertToLocalDate(Object object) {
+			Date[] tzArray = (Date[]) object;
+			LocalDate[] result = new LocalDate[tzArray.length];
+			for (int i = 0; i < tzArray.length; i++) {
+				result[i] = tzArray[i].toLocalDate();
+			}
+			return result;
```
davidangb, line 532: both `convertToLocalDateTime` and `convertToLocalDate` could use streams, e.g.
```java
return Arrays.stream(tzArray).map(Date::toLocalDate).toArray(LocalDate[]::new);
```

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -469,6 +497,13 @@ private RecordAttributes getAttributes(ResultSet rs) {
 							object = ts.toLocalDateTime();
 						} else if (object instanceof PGobject pGobject) {
 							object = pGobject.getValue();
+						} else if (object instanceof PgArray pgArray) {
+							object = pgArray.getArray();
+							if(object.getClass() == Timestamp[].class) {
+								object = convertToLocalDateTime(object);
+							} else if(object.getClass() == Date[].class) {
+								object = convertToLocalDate(object);
+							}
```
davidangb, line 506: also - are there unit tests for these date/timestamp arrays? I was messing around with the code here and couldn't get tests to fail, no matter what I changed.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -57,6 +64,12 @@ public DataTypeMapping selectBestType(DataTypeMapping existing, DataTypeMapping
...
+		}
```
davidangb, line 72: I think we should discuss requirements with PM first before making any additional changes here … but for future thought: if existing is DATE and newMapping is DATETIME, we could change to DATETIME instead of STRING

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -84,7 +97,12 @@ public DataTypeMapping inferTypeForTsvSource(Object val) {
 		if (isNumericValue(sVal)) {
 			return NUMBER;
 		}
-		return getTypeMappingFromString(sVal);
+		return getTypeMappingFromString(replaceLeftRightQuotes(sVal));
+	}
+
+	//libreoffice at least uses left and right quotes which cause problems when we try to parse as JSON
+	private String replaceLeftRightQuotes(String val){
+		return val.replaceAll("[“”]", "\"");
```
davidangb, line 105: my opinion is we shouldn't do this at all - but if we do, we should only replace leading/trailing smart quotes; we shouldn't replace anything in the middle of the string

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -162,13 +186,64 @@ public boolean isNumericValue(String sVal) {
...
+		DataTypeMapping bestMapping = null;
+		for (Object o : list) {
+			if(bestMapping == null){
+				bestMapping = inferType(o, InBoundDataSource.JSON);
+			} else {
+				bestMapping = selectBestType(bestMapping, inferType(o, InBoundDataSource.JSON));
+			}
+		}
+		return DataTypeMapping.getArrayTypeForBase(bestMapping);
+	}
```
davidangb, line 217: what I was picturing when we spoke in person was something like this:
```java
	private <T> DataTypeMapping findArrayType(List<T> list){
		DataTypeMapping bestMapping = null;
		// find the distinct inferred types from the input
		List<DataTypeMapping> inferredTypes = list.stream()
				.map(item -> inferType(item, InBoundDataSource.JSON))
				.distinct()
				.toList();
		if (inferredTypes.size() == 1) {
			bestMapping = inferredTypes.get(0);
		} else {
			for (DataTypeMapping type : inferredTypes) {
				bestMapping = selectBestType(bestMapping, type);
			}
		}
		return DataTypeMapping.getArrayTypeForBase(bestMapping);
	}
```

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -162,13 +186,64 @@ public boolean isNumericValue(String sVal) {
...
+
+	public <T> T[] getArrayOfType(String val, Class<T[]> clazz) {
 		try {
-			JsonNode jsonNode = objectMapper.readTree(val);
-			return jsonNode.isArray() || jsonNode.isObject();
+			return objectMapper.readValue(replaceLeftRightQuotes(val), clazz);
 		} catch (JsonProcessingException e) {
-			return false;
+			return null;
```
davidangb, line 246: thinking out loud … 

`findArrayTypeFromJson` and `getArrayOfType` seem pretty heavyweight, but I can't realistically think of anything better to do ... maybe we just keep an eye here for any potential problems?

davidangb (commented):
see inline! This is a big complex PR

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -518,7 +518,7 @@ components:
           description: name of this attribute.
         datatype:
           type: string
-          enum: [boolean, date, datetime, string, json, long, double, relation]
+          enum: [boolean, date, datetime, string, json, long, double, relation, array_of_boolean, array_of_string, array_of_number, array_of_date, array_of_datetime]
```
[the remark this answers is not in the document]
sagehen03: Your suggestion feels more complicated to me.

service/src/test/java/org/databiosphere/workspacedataservice/DataTypeInfererTest.java
```diff
@@ -101,18 +105,34 @@ void inferSomeTypes() {
...
+	void ambiguousConversions() {
+		assertThat(inferer.inferType("[true, \"false\", \"True\"]", InBoundDataSource.JSON)).isEqualTo(DataTypeMapping.ARRAY_OF_BOOLEAN);
+		assertThat(inferer.inferType("[true, \"false\", True]", InBoundDataSource.JSON)).isEqualTo(DataTypeMapping.STRING);
+		assertThat(inferer.inferType("[\"11\", \"99\"]", InBoundDataSource.JSON)).isEqualTo(DataTypeMapping.ARRAY_OF_STRING);
+		assertThat(inferer.inferType("[\"11\", 99, \"foo\"]", InBoundDataSource.JSON)).isEqualTo(DataTypeMapping.ARRAY_OF_STRING);
+		assertThat(inferer.inferType("", InBoundDataSource.TSV)).isEqualTo(DataTypeMapping.NULL);
+		assertThat(inferer.inferType("", InBoundDataSource.JSON)).isEqualTo(DataTypeMapping.STRING);
+		assertThat(inferer.inferType("[11, 99, -3.14]", InBoundDataSource.JSON)).isEqualTo(DataTypeMapping.ARRAY_OF_NUMBER);
+		assertThat(inferer.inferType("[11, 99, -3.14, 09]", InBoundDataSource.JSON)).isEqualTo(DataTypeMapping.STRING);
+		assertThat(inferer.inferType("[]", InBoundDataSource.JSON)).isEqualTo(DataTypeMapping.EMPTY_ARRAY);
```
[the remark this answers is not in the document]
sagehen03: I agree with you on the inconsistency of quoted coercible boolean values vs. numbers.  I would say if we want to prioritize consistency I would say let's switch to making "true" remain strings.  The existing non array boolean behavior was that we convert them to booleans so I wanted to make array booleans consistent with that, but if we want to re-visit the incosistency between boolean vs. numbers let's do that.  I'll piing Beth and Alessandro.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -365,9 +368,33 @@ private Object getValueForSql(Object attVal, DataTypeMapping typeMapping) {
...
+		}
 		return attVal;
 	}
 
+	private Object getArrayValues(Object attVal, DataTypeMapping typeMapping) {
+		if (attVal instanceof List<?> valAsList) {
+			return getListAsArray(valAsList, typeMapping);
+		}
+		return inferer.getArrayOfType(attVal.toString(), typeMapping.getJavaArrayTypeForDbWrites());
+	}
```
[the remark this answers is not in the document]
sagehen03: Correct or if the user submits quoted json as an attribute value via the json code paths.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
sagehen03: Good call!

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
sagehen03: What did you change?  There are definitely tests that write to these array date types, but I think we could add some tests for reading and verifying the proper value.  I will do add those.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
sagehen03: That seems reasonable.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
sagehen03: I will test and see what jackson does with non-leading left right quotes.  If it doesn't puke, I will adjust the regex to look only for leading and trailing quotes.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
sagehen03: That's usually my preference!

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
davidangb: this still passes tests:
```java
						} else if (object instanceof PgArray pgArray) {
							object = pgArray.getArray();
							if(object.getClass() == Timestamp[].class) {
								object = "failme";
							} else if(object.getClass() == Date[].class) {
								object = 1234;
							}
						}
```

sagehen03: Ah that gibes with my understanding of the tests.  I was trying to keep this PR from ever expanding and was going to add a follow on ticket to clean up the tests as we discussed on Tuesday and include in that more validation that we're returning the proper values, but I'll add some date read validation now.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -131,7 +132,57 @@ void simpleTsvUploadWithBatchingShouldSucceed(@Value("${twds.write.batch.size}")
...
+		//empty array
+		for (int i = 0; i < 10; i++) {
+			tsvContent.append(i + "empty\t[]\n");
+		}
+		//array of long
+		for (int i = 0; i < 10; i++) {
+			tsvContent.append(i + "valid\t[12]\n");
+		}
+		MockMultipartFile file = new MockMultipartFile("records", "simple.tsv", MediaType.TEXT_PLAIN_VALUE,
+				tsvContent.toString().getBytes());
```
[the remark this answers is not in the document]
sagehen03: For a narrow tsv with only one or two columns this doesn't seem unwieldy to me.  It's also nice not to have to look at two editor panes to understand the test

sagehen03:
@davidangb I think I've addressed your PR feedback except for two cases: 

string coercion consistency--we're waiting on some guidance in slack I'd like to open a separate ticket for this and address it once we have consensus 

replacing directional quotes--the quotes that trip jackson aren't at the edges of the string so I think it should stay, but if I need to remove them for your approval I can do that and we can see what happens once we start getting usage.

davidangb (approved):
:+1: thank you!

On the smart-quote issue … could you document what you have run into, in a unit test or in a jira ticket or whatever form makes sense to you? Based on your comment about interior quotes being a problem, I tried a very simple unit test:
```java
	@Test
	void testSmartQuotes() throws JsonProcessingException {
		String input = "\"f“o“o\"";
		String actual = jacksonObjectMapper.readValue(input, String.class);
		assertEquals("f“o“o", actual);
	}
```
and the test passed … so I think I am not understanding the issue correctly and want to make sure we follow up on it.

sagehen03:
> ```java
> String input = "\"f“o“o\"";
> 		String actual = jacksonObjectMapper.readValue(input, String.class);
> 		assertEquals("f“o“o", actual);
> ```
This illustrates the issue with smart quotes.  It might be possible to craft a complicated regex to ferret those element delimiting smart quotes but I feel like complicated regexes never end well.
```
@Test
void quoting() throws JsonProcessingException {
String input = "[“foo”, “bazz”, “bar”]";
String[] actual = mapper.readValue(input, String[].class);
assertArrayEquals(new String[]{"foo", "baz", "bar"}, actual);
}
```

PR #93, opened 2022-10-17
AJ-638: consolidate to single numeric data type

sagehen03:
Remove the Long and Double data types and replace them with Number.


service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -131,12 +125,8 @@ public DataTypeMapping inferTypeForJsonSource(Object val) {
 			return NULL;
 		}
 
-		if (val instanceof Long || val instanceof Integer) {
-			return LONG;
-		}
-
-		if (val instanceof Double || val instanceof Float) {
-			return DOUBLE;
+		if (val instanceof Integer || val instanceof Double || val instanceof Long || val instanceof BigInteger) {
```
sagehen03, line 128: Jackson will convert integers to either Integer, Long, or BigInteger depending upon the size of the integer.

calypsomatic (approved):
Thank you Drew!

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -352,11 +352,8 @@ private Object getValueForSql(Object attVal, DataTypeMapping typeMapping) {
 			return RelationUtils.getRelationValue(attVal);
 		}
 		if (attVal instanceof String sVal) {
-			if (stringIsCompatibleWithType(typeMapping == DataTypeMapping.LONG, inferer::isLongValue, sVal)) {
-				return Long.parseLong(sVal);
-			}
-			if (stringIsCompatibleWithType(typeMapping == DataTypeMapping.DOUBLE, inferer::isDoubleValue, sVal)) {
-				return Double.parseDouble(sVal);
+			if (stringIsCompatibleWithType(typeMapping == DataTypeMapping.NUMBER, inferer::isNumericValue, sVal)) {
+				return Double.valueOf(sVal);
```
davidangb, line 356: TOL: I wonder if this change to return a `Double` instead of a `double` has a similar performance impact as using `BigDecimal` - that is, I wonder if moving from a primitive to an Object is the big deal. No need to test or answer this, I think the changes to the code are the right thing to do, and it's not required to figure this out.

davidangb (approved):
thank you!

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
sagehen03: I think if we used `parseDouble` it would get auto boxed to a Double so my logic was to just go straight to Double if that's where it was going to end up anyway, but I have not tested or looked deeply into what happens.  Hopefully, we can switch to BigDecimal before too long anyhow.

PR #109, opened 2022-10-19
Aj 628 python client

calypsomatic:
This PR will generate a python client and push it to PYPI.
The `python` yaml is called by the `publish` yaml so that it only occurs after the latest tag has been generated and uses that one.

### DIFF
--- a/.github/workflows/release-python-client.yaml
+++ b/.github/workflows/release-python-client.yaml
@@ -3,8 +3,6 @@ name: Publish Python client to PyPI
 on:
   workflow_call:
     secrets:
-      GITHUB_TOKEN:
-        required: true
       PYPI_API_TOKEN:
         required: true
 jobs:
--- a/.github/workflows/release-python-client.yaml
+++ b/.github/workflows/release-python-client.yaml
@@ -26,9 +26,9 @@ jobs:
       - name: Use Node.js ${{matrix.node-version }}
         uses: actions/setup-node@v2
         with:
-          node-version: 14
+          node-version: 18
       - name: Install openapi-generator-cli
-        run: npm install openapitools/openapi-generator-cli@1.0.18-4.3.1 -g
+        run: npm install @openapitools/openapi-generator-cli@1.0.18-4.3.1 -g
       - name: Generate Python client
         run: |
           openapi-generator-cli generate \
--- a/.github/workflows/release-python-client.yaml
+++ b/.github/workflows/release-python-client.yaml
@@ -11,7 +11,10 @@ jobs:
     name: Build and publish Python client to PyPI
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       - name: Set up Python 3.7
         uses: actions/setup-python@v1
         with:
--- a/.github/workflows/release-python-client.yaml
+++ b/.github/workflows/release-python-client.yaml
@@ -16,7 +16,7 @@ jobs:
         uses: actions/setup-python@v1
         with:
           python-version: 3.7
-      - name: 'Get Previous tag'
+      - name: Get previous tag
         id: apiprevioustag
         uses: "broadinstitute/github-action-get-previous-tag@master"
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/release-python-client.yaml
```diff
@@ -0,0 +1,55 @@
...
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
+      - name: Get previous tag
+        id: apiprevioustag
+        uses: "broadinstitute/github-action-get-previous-tag@master"
+        env:
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+      - name: Use Node.js ${{matrix.node-version }}
+        uses: actions/setup-node@v2
```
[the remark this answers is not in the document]
calypsomatic: That's what it said in the [workflow I was copying from](https://github.com/DataBiosphere/jade-data-repo/blob/3e736149c46b676847e9cfec7d203fe4384354a5/.github/workflows/release-python-client.yaml#L24) :p

.github/workflows/release-python-client.yaml
```diff
@@ -0,0 +1,55 @@
...
+        with:
+          node-version: 14
+      - name: Install openapi-generator-cli
+        run: npm install @openapitools/openapi-generator-cli -g
```
aaronkanzer, line 31: `I think you could just do npm install @openapitools/openapi-generator-cli@1.0.18-4.3.1 -g and remove the set version to 4.3.1 step below?

.github/workflows/release-python-client.yaml
```diff
@@ -0,0 +1,55 @@
...
+      - name: set version to 4.3.1
+        run: openapi-generator-cli version-manager set 4.3.1
+      - name: Generate Python client
+        run: |
+          openapi-generator-cli generate \
+          -i service/src/main/resources/static/swagger/openapi-docs.yaml \
+          -g python \
+          -o wds-client \
+          --additional-properties=projectName=wds-client,packageName=wds_client,packageVersion=${{ steps.apiprevioustag.outputs.tag }} \
```
aaronkanzer, line 40: This might be nitpicky (curious to get thoughts), but would it be beneficial to append `terra-` or some Terra-related naming convention since the package is coupled with Terra?

calypsomatic: Yep!  I was just throwing something out there to prove this works, making it presentable is for future us to worry about.

.github/workflows/release-python-client.yaml
calypsomatic: I can give it try!

.github/workflows/release-python-client.yaml
```diff
@@ -0,0 +1,53 @@
...
+          python-version: 3.7
+      - name: Get previous tag
+        id: apiprevioustag
+        uses: "broadinstitute/github-action-get-previous-tag@master"
+        env:
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+      - name: Use Node.js ${{matrix.node-version }}
+        uses: actions/setup-node@v2
+        with:
+          node-version: 14
```
jordansafer, line 29: could we use node 18? Node 14 hits end of life in april 2023 (pretty soon) and Node 18 won't reach it til april 2025 (also somewhat soon, but it's the latest available) https://endoflife.date/nodejs

jordansafer (approved):

.github/workflows/release-python-client.yaml
calypsomatic: Well it didn't work; it works fine just on my local terminal but not in the GHA.  Maybe I just did something wrong, but I'm going back to the two-step version and maybe it can be investigated and cleaned up into one step in some future PR.

.github/workflows/release-python-client.yaml
calypsomatic: Good call!

.github/workflows/release-python-client.yaml
aaronkanzer: Did it work after changing the version of `node` to `18` you are referencing in the GHA here?

calypsomatic: Nope!

sagehen03 (approved):
I pip installed and made some wds api calls from python.

PR #114, opened 2022-10-21
AJ-639, Support Arbitrary Precision

sagehen03:
The changes here go into three buckets--1. configure jackson to use bigdecimal and biginteger for JSON deserialization of numbers. 2. share the same jackson config everywhere so that item 1 applies everywhere. 3. change tests to account for change from step 1.


service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -67,6 +68,9 @@ public DataTypeMapping selectBestType(DataTypeMapping existing, DataTypeMapping
 		if (existing.isArrayType() && newMapping.isArrayType() && Set.of(existing, newMapping).contains(EMPTY_ARRAY)) {
 			return newMapping != EMPTY_ARRAY ? newMapping : existing;
 		}
+		if (existing.isArrayType() && newMapping.isArrayType() && Set.of(existing, newMapping).contains(EMPTY_ARRAY)) {
```
calypsomatic, line 71: This looks like the same code block twice in a row, or am I missing something?

sagehen03: You're right, looks like I messed up that part of the merge.

jordansafer (approved):

calypsomatic (approved):

PR #115, opened 2022-10-24
AJ-640: Fix for writing JSON objects to jsonb

sagehen03:
In addition to what's referenced in the jira ticket, I took the opportunity to improve some of our tests and to ensure that we're reading and writing data as expected.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -377,7 +377,7 @@ public class RecordDao {
 				return LocalDateTime.parse(sVal, DateTimeFormatter.ISO_LOCAL_DATE_TIME);
 			}
 		}
-		if(attVal instanceof Map<?,?> || attVal instanceof List<?>){
+		if(attVal instanceof Map<?,?>){
 			try {
 				return objectMapper.writeValueAsString(attVal);
 			} catch (JsonProcessingException e) {
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -76,8 +76,7 @@ public class RecordDao {
 	private final ObjectMapper objectMapper;
 
 	public RecordDao(NamedParameterJdbcTemplate namedTemplate,
-			@Qualifier("streamingDs") NamedParameterJdbcTemplate templateForStreaming, DataTypeInferer inf) {
-					 @Qualifier("streamingDs") NamedParameterJdbcTemplate templateForStreaming, ObjectMapper objectMapper) {
+			@Qualifier("streamingDs") NamedParameterJdbcTemplate templateForStreaming, DataTypeInferer inf, ObjectMapper objectMapper) {
 		this.namedTemplate = namedTemplate;
 		this.templateForStreaming = templateForStreaming;
 		this.inferer = inf;
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -398,7 +398,7 @@ public class RecordDao {
 
 	private Object[] getListAsArray(List<?> attVal, DataTypeMapping typeMapping) {
 		switch (typeMapping){
-			case ARRAY_OF_STRING, ARRAY_OF_DATE, ARRAY_OF_DATE_TIME, ARRAY_OF_NUMBER:
+			case ARRAY_OF_STRING, ARRAY_OF_DATE, ARRAY_OF_DATE_TIME, ARRAY_OF_NUMBER, EMPTY_ARRAY:
 				return attVal.stream().map(Object::toString).toList().toArray(new String[0]);
 			case ARRAY_OF_BOOLEAN:
 				//accept all casings of True and False if they're strings
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -129,7 +129,7 @@ public class RecordDao {
 				"select * from " + getQualifiedTableName(recordType, instanceId) + " order by "
 						+ (sortAttribute == null ? RECORD_ID : quote(sortAttribute)) + " " + sortDirection + " limit "
 						+ pageSize + " offset " + offset,
-				new RecordRowMapper(recordType, getRelationColumnsByName(getRelationCols(instanceId, recordType)), objectMapper, getExistingTableSchema(instanceId, recordType)));
+				new RecordRowMapper(recordType, objectMapper, instanceId));
 	}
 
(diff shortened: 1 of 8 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -68,11 +70,15 @@ public class RecordDao {
 	private static final Logger LOGGER = LoggerFactory.getLogger(RecordDao.class);
 
 	private final DataTypeInferer inferer;
+
+	private final ObjectMapper objectMapper;
+
 	public RecordDao(NamedParameterJdbcTemplate namedTemplate,
-			@Qualifier("streamingDs") NamedParameterJdbcTemplate templateForStreaming, DataTypeInferer inf) {
+			@Qualifier("streamingDs") NamedParameterJdbcTemplate templateForStreaming, DataTypeInferer inf, ObjectMapper objectMapper) {
```
sagehen03, line 77: Jackson will pass us a map that we have to serialize to a json string in order to get it into the jsonb column so we need an ObjectMapper where we're writing to the db.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -468,14 +482,18 @@ public int getBatchSize() {
 	}
 
 	private record RecordRowMapper(RecordType recordType,
-			Map<String, RecordType> referenceColToTable) implements RowMapper<Record> {
+			Map<String, RecordType> referenceColToTable, ObjectMapper objectMapper, Map<String, DataTypeMapping> schema) implements RowMapper<Record> {
```
sagehen03, line 485: I added the schema parameter because there wasn't a surefire way to determine that column is jsonb just using the SQL resultset's metadata.  This does impose another query on read operations to first read the schema for the record type and then fetch the data.  Not great, I know.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -490,24 +508,7 @@ private RecordAttributes getAttributes(ResultSet rs) {
...
-							object = pGobject.getValue();
-						} else if (object instanceof PgArray pgArray) {
-							object = pgArray.getArray();
-							if(object.getClass() == Timestamp[].class) {
-								object = convertToLocalDateTime(object);
-							} else if(object.getClass() == Date[].class) {
-								object = convertToLocalDate(object);
-							}
-						}
-
```
sagehen03, line 509: Adding the json reading functionality here got sonar pretty upset about the cyclomatic complexity which I kind of agreed with so I moved some of this code to two new methods.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -212,15 +217,15 @@ public boolean isArray(String val){
...
+		}
 		List<DataTypeMapping> inferredTypes = list.stream()
 				.map(item -> inferType(item, InBoundDataSource.JSON))
 				.distinct()
 				.toList();
-		if (inferredTypes.size() == 1) {
-			bestMapping = inferredTypes.get(0);
-		} else {
+		DataTypeMapping bestMapping = inferredTypes.get(0);
+		if (inferredTypes.size() > 1) {
```
sagehen03, line 228: There was a bug here that we didn't catch because DataTypeInfererTest was using stringified arrays rather than the java objects that jackson gives us.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -81,12 +86,6 @@ void tsvWithNoRowsShouldReturn400() throws Exception {
 				.file(file)).andExpect(status().isBadRequest());
 	}
 
-	/**
-	 * Slightly strange test to show loss of precision as a result of storing a large
-	 * integer in WDS.  If we switch to using BigDecimal throughout we should be able to
-	 * to change the test to start asserting equality
-	 * @throws Exception
-	 */
```
sagehen03, line 89: Forgot to delete this comment when I changed the test to assert equality.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -468,14 +482,18 @@ public int getBatchSize() {
...
 	private record RecordRowMapper(RecordType recordType,
-			Map<String, RecordType> referenceColToTable) implements RowMapper<Record> {
+			Map<String, RecordType> referenceColToTable, ObjectMapper objectMapper, Map<String, DataTypeMapping> schema) implements RowMapper<Record> {
 
 		@Override
 		public Record mapRow(ResultSet rs, int rowNum) throws SQLException {
-			return new Record(rs.getString(RECORD_ID), recordType, getAttributes(rs));
+			try {
+				return new Record(rs.getString(RECORD_ID), recordType, getAttributes(rs));
+			} catch (JsonProcessingException e) {
```
calypsomatic, line 491: Are we catching this exception only to throw a new one just to make the compiler happy about expecting a `JsonProcessingException` and/or because we definitely want to throw a `RuntimeException` or some other reason?

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -123,7 +129,7 @@ public List<Record> queryForRecords(RecordType recordType, int pageSize, int off
 				"select * from " + getQualifiedTableName(recordType, instanceId) + " order by "
 						+ (sortAttribute == null ? RECORD_ID : quote(sortAttribute)) + " " + sortDirection + " limit "
 						+ pageSize + " offset " + offset,
-				new RecordRowMapper(recordType, getRelationColumnsByName(getRelationCols(instanceId, recordType))));
+				new RecordRowMapper(recordType, getRelationColumnsByName(getRelationCols(instanceId, recordType)), objectMapper, getExistingTableSchema(instanceId, recordType)));
```
davidangb, line 132: it looks like all three call sites that create a `RecordRowMapper` first call `getRelationColumnsByName()` and `getExistingTableSchema()`. How about moving those calls inside of `RecordRowMapper` for consistency? This would cascade a minor refactor of `getSingleRecord()`, which has a slightly different implementation of `getRelationColumnsByName`.

This is optional - I wouldn't block the PR on it - but I think it would be a good encapsulation.

davidangb (approved):
One optional comment inline about a potential cleanup/refactor; otherwise lgtm!

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
sagehen03: I tried this refactoring and it does mean having to pipe the instanceid to the RecordRowMapper and couples the RecordRowMapper more tightly to RecordDao, but it does reduce the number of args for RecordRowMapper by 1, reduces call sites, and also reduces the number of args for getSingleRecord() by 1 so overall it seems like a close win over the status quo so I've added it :)

PR #133, opened 2022-10-26
Aj 641 versions

calypsomatic:
This PR creates a new tag upon push to main first, and then, using the new tag, publishes all the clients and images to ensure they are in sync.  Docker images are given both the git hash and the tag.

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -78,7 +78,7 @@ jobs:
         run: docker push --all-tags ${{ steps.gcr-image-name.outputs.name }}
 
       - name: Re-tag image for ACR
-        run: docker tag ${{ steps.gcr-image-name.outputs.name }} \ 
+        run: docker tag ${{ steps.gcr-image-name.outputs.name }}:${{ steps.setHash.outputs.git_short_sha }} \ 
           ${{ steps.acr-image-name.outputs.name }}:${{ steps.setHash.outputs.git_short_sha }}
 
       - name: Add version tag to ACR
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -90,18 +90,18 @@ jobs:
           token: ${{ secrets.BROADBOT_TOKEN }} # Has to be set at checkout AND later when pushing to work
           path: cromwhelm
           
-      - name: Update WDS image tag for CBAS Helm Chart in cromwhelm
-        env:
-          BROADBOT_TOKEN: ${{ secrets.BROADBOT_TOKEN }}
-        run: |
-          set -e
-          cd cromwhelm
-          git checkout main
-          HELM_CUR_TAG=$(grep "/terra-workspace-data-service:" terra-batch-libchart/values.yaml | sed "s,.*/terra-workspace-data-service:,,")
-          HELM_NEW_TAG=${{ steps.setHash.outputs.git_short_sha }}
-          [[ -n "$HELM_CUR_TAG" && -n "$HELM_NEW_TAG" ]]
-          sed -i "s/$HELM_CUR_TAG/$HELM_NEW_TAG/" terra-batch-libchart/values.yaml
-          git config --global user.name "broadbot"
-          git config --global user.email "broadbot@broadinstitute.org"
-          git commit -am "Auto update WDS version to $HELM_NEW_TAG"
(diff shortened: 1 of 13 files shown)


.github/workflows/publish-docker.yml
```diff
@@ -33,22 +40,26 @@ jobs:
           echo "::set-output name=git_short_sha::${git_short_sha}"
       - name: Construct GCR docker image name and tag
         id: gcr-image-name
-        run: echo ::set-output name=name::${GCR_REGISTRY}/${GOOGLE_PROJECT}/${SERVICE_NAME}:${{ steps.setHash.outputs.git_short_sha }}
+        run: echo ::set-output name=name::${GCR_REGISTRY}/${GOOGLE_PROJECT}/${SERVICE_NAME}
```
aaronkanzer, line 43: Should we be referencing the style prescribed here as it seems `::set-output` will soon be deprecated: 

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/#examples

.github/workflows/publish-docker.yml
```diff
@@ -64,17 +75,20 @@ jobs:
...
+      - name: Add version tag to ACR
+        run: docker image tag ${{ steps.acr-image-name.outputs.name }}:${{ steps.setHash.outputs.git_short_sha }} ${{ steps.acr-image-name.outputs.name }}:${{ inputs.new-tag }}
 
       - name: Push image to Azure
         run: |
           echo ${{ secrets.ACR_SP_PASSWORD }} | docker login ${ACR_REGISTRY} \
           --username ${{ secrets.ACR_SP_USER }} \
           --password-stdin
-          docker push ${{ steps.acr-image-name.outputs.name }}
+          docker push --all-tags ${{ steps.acr-image-name.outputs.name }}
```
aaronkanzer, line 91: What is the reason to push multiple tags to the registry?

.github/workflows/publish-docker.yml
```diff
@@ -33,22 +40,26 @@ jobs:
...
         run: |
           ./gradlew --build-cache jibDockerBuild \
-          --image=${{ steps.gcr-image-name.outputs.name }} \
+          --image=${{ steps.gcr-image-name.outputs.name }}:${{ steps.setHash.outputs.git_short_sha }} \
           -Djib.console=plain
+          
+      - name: Add version tag to GCR
+        run: docker image tag ${{ steps.gcr-image-name.outputs.name }}:${{ steps.setHash.outputs.git_short_sha }} ${{ steps.gcr-image-name.outputs.name }}:${{ inputs.new-tag }}
+          
       - name: Run Trivy vulnerability scanner
```
aaronkanzer, line 58: Just out of curiousity, why do we not run Trivy on images pushed to ACR?

.github/workflows/publish-docker.yml
```diff
@@ -33,22 +40,26 @@ jobs:
           echo "::set-output name=git_short_sha::${git_short_sha}"
       - name: Construct GCR docker image name and tag
         id: gcr-image-name
-        run: echo ::set-output name=name::${GCR_REGISTRY}/${GOOGLE_PROJECT}/${SERVICE_NAME}:${{ steps.setHash.outputs.git_short_sha }}
+        run: echo ::set-output name=name::${GCR_REGISTRY}/${GOOGLE_PROJECT}/${SERVICE_NAME}
 
       - name: Construct ACR docker image name and tag
```
davidangb, line 45: same here, remove "and tag"?

davidangb (commented):
I like this setup! I also don't have ACR access yet so can't verify what's there, let's remember to do that once we get access.

How about making a manual change to `build.gradle` to set `version = 0.2.0-SNAPSHOT` in this PR, to do a soft reset of our versioning?

.github/workflows/publish-docker.yml
davidangb: some systems like our helm charts use the short git hash, but we'd like that hash to be associated with the semantic version number we're using for our clients and github tags. Having both tags on the image makes that nice and easy:
Screenshot 10-26-2022 at 05 02 PM

.github/workflows/publish-docker.yml
calypsomatic: I guess I was assuming it's the exact same image, just with a different name/tag

aaronkanzer (approved):

sagehen03 (approved):

PR #134, opened 2022-10-26
AJ-642: Test the generated client

sagehen03:


service/src/test/java/org/databiosphere/workspacedataservice/GeneratedClientTests.java
```diff
@@ -0,0 +1,120 @@
...
+
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+class GeneratedClientTests {
+
+    private ApiClient apiClient;
+    @LocalServerPort
+    int port;
+
+    private final UUID instanceId = UUID.randomUUID();
+    private final String version = "v0.2";
```
aaronkanzer, line 36: Do we want to make this a constant somewhere in our test suite?

It seems that a few other places are using it. I'm not quite sure how often/when we will bump the value though...

service/src/test/java/org/databiosphere/workspacedataservice/GeneratedClientTests.java
```diff
@@ -0,0 +1,120 @@
...
+
+    @BeforeEach
+    void init(){
+        apiClient = new ApiClient();
+        apiClient.setBasePath("http://localhost:" + port);
+    }
+
+    @Test
+    void createInstance() throws ApiException {
```
aaronkanzer, line 45: Should there be some type of assertion/test in here since it is decorated as `@Test`?

calypsomatic (commented):
Do we want to do more testing of different attribute types?

service/src/test/java/org/databiosphere/workspacedataservice/GeneratedClientTests.java
```diff
@@ -0,0 +1,120 @@
...
+    int port;
+
+    private final UUID instanceId = UUID.randomUUID();
+    private final String version = "v0.2";
+
+    @BeforeEach
+    void init(){
+        apiClient = new ApiClient();
+        apiClient.setBasePath("http://localhost:" + port);
+    }
```
aaronkanzer, line 42: I think you could throw `createNewInstance(instanceId)` in  here as well since you are calling it in the beginning of each unit test?

(I think you might need to just add `throws ApiException` to the function signature as well)

service/src/test/java/org/databiosphere/workspacedataservice/GeneratedClientTests.java
sagehen03: In theory yes, but if an ApiException is thrown the test will fail.  That seems like enough to me.  Let me know if you have better ideas.  The createInstance method is void so I think all we want to do here is ensure that no exception is thrown.  Again, I'm open to ideas.

service/src/test/java/org/databiosphere/workspacedataservice/GeneratedClientTests.java
sagehen03: I originally wanted to write a test for all the client methods that CBAS is using so I had a separate test for `createInstance` but if the consensus is we should move that to the init method, I can do that.  I think the result should be the same if `createInstance` throws an exception in the init method and this would resolve your other question around no assertion being present in the test.

sagehen03:
> Do we want to do more testing of different attribute types?

I think that's already covered by our existing tests.  These tests are basically just to make sure that the swagger client is not broken due to a bad yaml file.  Ensuring the correctness of the logic elsewhere in the codebase falls to our existing tests in my view.

service/src/test/java/org/databiosphere/workspacedataservice/GeneratedClientTests.java
sagehen03: In theory yes, but it's a pretty quick change with the IDE so I'm going to say not worth worrying about at the moment.

service/src/test/java/org/databiosphere/workspacedataservice/GeneratedClientTests.java
aaronkanzer: Yeah I assume if the `init` fails, then that would be enough to reveal a regression in the code?

I'm not as familiar with CBAS yet, so if the current code is more regression-proof, fine by me.

sagehen03: I've adopted this suggestion and removed the specific test for createInstance.

PR #135, opened 2022-10-27
AJ-613: Add Spring Boot Actuator to provide endpoints for WDS info, health

aaronkanzer:
Add Spring Boot Actuator to provide endpoints for WDS info, health

Sample payload from `/version`:

```
{"git":{"branch":"dh-spring-boot-actuator","commit":{"id":"7e9eabe","time":"2022-10-28T19:39:01Z"}},"build":{"artifact":"service","name":"service","time":"2022-10-28T20:09:09.844Z","version":"0.1.56-SNAPSHOT","group":"org.databiosphere"}}
```

Sample payload from `/status`:

```
{
  "status": "UP"
}
```


service/VERSION.txt
```diff
@@ -0,0 +1,2 @@
+1.2.3
+sha123
```
calypsomatic, line 2: My instinct would be to make this file a `.json` since it's going to end up in essentially a json-like format anyway.  Not sure I have a real strong argument there, but it might be slightly easier to read it with a json library than parsing it.

calypsomatic:
Not sure if this belongs as part of this ticket, but just mentioning for discussion/so that we all remember to address it: right now the ['version' that is populated by default in `openapi-docs.yaml`](https://github.com/DataBiosphere/terra-workspace-data-service/blob/2056c7c0ed1ee2c69a58d70cf1791b5e4466ed15/service/src/main/resources/static/swagger/openapi-docs.yaml#L5) is hard-coded as 'v0.2'.  We'll want that to be automatically updated as well.

service/VERSION.txt
aaronkanzer: That's a good point -- I refactored here to JSON.

Seems `jq` is available in the Github Actions runners, so that should make altering the `VERSION.json` file simple.

davidangb:
I'm wondering if we can extract all the info we need at gradle build time, instead of relying on github actions to calculate it for us. At a high level, I think it would look something like this:
1. Use https://plugins.gradle.org/plugin/com.gorylenko.gradle-git-properties or similar to store the git info. That plugin generates a `git.properties` file on the classpath
2. Add some gradle magic to write the [version](https://github.com/DataBiosphere/terra-workspace-data-service/blob/2056c7c0ed1ee2c69a58d70cf1791b5e4466ed15/build.gradle#L19) to a properties file during build, OR move the version to `gradle.properties` … the latter would require tweaking our existing github actions
3. Write a new `@Configuration`-annotated class which specifies `@PropertySource`s of the properties files from the previous two bullets, and produces a smiple `@Bean` holding the hash and semantic version number
4. Bring the bean into `RecordController` via `@Autowired`, and return its values in the `/version` API response

service/build.gradle
```diff
@@ -5,6 +5,10 @@ plugins {
 	id 'org.sonarqube'
```
sagehen03, line 5: I think you need `id 'com.gorylenko.gradle-git-properties` here and need to include the version info for that plugin for the project's build.gradle

aaronkanzer: Added back in!

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -26,7 +26,25 @@ tags:
     description: Instance APIs
   - name: Schema
     description: Schema Manipulation APIs (coming soon)
+  - name: General WDS Information
+    description: Information regarding versioning, health, info, etc. for WDS
 paths:
+  /actuator/info:
```
calypsomatic, line 32: I think `actuator` doesn't make sense to anybody but us.  Is there a more general, public-friendly URL we could use?

davidangb (commented):
I agree with Bria's comment on URLs. Terra has historically used `/version` and `/status` for these, and the status APIs typically expose information about the subcomponents they check.

I think we could achieve that with some actuator config, e.g.:
```
management.endpoints.web.base-path=/
management.endpoints.web.path-mapping.info=version
```

which would drive some changes to the swagger definition, too

davidangb:
( sorry, hit enter too soon ) … the other thing I want to do before merging this is check with infosec & other dev teams to see if anyone has any hesitations about using actuator for these. I know we have turned off everything but these two actuator endpoints - but some of the other ones would be dangerous if they got exposed.

davidangb:
@calypsomatic I waffle on automatically updating the version in openapi-docs. On one hand, that version could be seen as the most up-to-date patch version, which means we want auto updates. On the other hand, it could be viewed as the version that users should enter in the path param for API endpoints - which means we want it to stay stable until/unless we change those implementations.

sagehen03 (approved):

aaronkanzer:
> ( sorry, hit enter too soon ) … the other thing I want to do before merging this is check with infosec & other dev teams to see if anyone has any hesitations about using actuator for these. I know we have turned off everything but these two actuator endpoints - but some of the other ones would be dangerous if they got exposed.

Just linking for any future reference to a Slack thread with infosec related to this

service/src/main/resources/static/swagger/openapi-docs.yaml
aaronkanzer: Updated to use `/version` and `\status` instead!

calypsomatic (approved):

PR #141, opened 2022-11-15
dont autocreate instance in put-record; add delete-instance API [AJ-614, AJ-674]

davidangb:
there's some yak-shaving in here.

The ultimate goal is for the `createOrReplaceRecord` PUT single-record API to stop auto-creating instances if the supplied instanceId does not exist. That runtime change was pretty easy.

However, it caused some issues in unit tests. So I updated unit tests to create the relevant instance before each test ran. In `FullStackRecordControllerTest`, this ran into even more issues because the database is persistent, and each create-instance call past the first would fail with a 409 conflict.

So, I also added a delete-instance API (with a couple tests of its own and swagger doc). This isn't the only solution to the testing situation, but it felt like a beneficial addition for debugging and such. If reviewers think the delete-instance API is a bad idea, I can find a different solution for tests.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -183,12 +183,10 @@ public class RecordController {
 			@PathVariable("version") String version, @PathVariable("recordType") RecordType recordType,
 			@PathVariable("recordId") String recordId, @RequestBody RecordRequest recordRequest) {
 		validateVersion(version);
+		validateInstance(instanceId);
 		RecordAttributes attributesInRequest = recordRequest.recordAttributes();
 		Map<String, DataTypeMapping> requestSchema = inferer.inferTypes(attributesInRequest, InBoundDataSource.JSON);
 		HttpStatus status = HttpStatus.CREATED;
-		if (!recordDao.instanceSchemaExists(instanceId)) {
-			recordDao.createSchema(instanceId);
-		}
 		if (!recordDao.recordTypeExists(instanceId, recordType)) {
 			RecordResponse response = new RecordResponse(recordId, recordType, recordRequest.recordAttributes());
 			Record newRecord = new Record(recordId, recordType, recordRequest);
@@ -222,6 +220,17 @@ public class RecordController {
 		}
 		recordDao.createSchema(instanceId);
 		return new ResponseEntity<>(HttpStatus.CREATED);
+	}
+
+	@DeleteMapping("/instances/{version}/{instanceId}")
+	public ResponseEntity<String> deleteInstance(@PathVariable("instanceId") UUID instanceId,
+												 @PathVariable("version") String version) {
+		validateVersion(version);
+		if (!recordDao.instanceSchemaExists(instanceId)) {
+			throw new ResponseStatusException(HttpStatus.NOT_FOUND, "This instance does not exist");
+		}
+		recordDao.dropSchema(instanceId);
+		return new ResponseEntity<>(HttpStatus.OK);
 	}
 
 	@DeleteMapping("/{instanceId}/records/{version}/{recordType}/{recordId}")
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -79,6 +79,7 @@ public class RecordController {
 			@PathVariable("version") String version, @PathVariable("recordType") RecordType recordType,
 			@PathVariable("recordId") String recordId, @RequestBody RecordRequest recordRequest) {
 		validateVersion(version);
+		validateInstance(instanceId);
 		checkRecordTypeExists(instanceId, recordType);
 		Record singleRecord = recordDao
 				.getSingleRecord(instanceId, recordType, recordId)
@@ -114,8 +115,8 @@ public class RecordController {
 	public ResponseEntity<TsvUploadResponse> tsvUpload(@PathVariable("instanceId") UUID instanceId,
 			@PathVariable("version") String version, @PathVariable("recordType") RecordType recordType,
 			@RequestParam("records") MultipartFile records) throws IOException {
-		validateInstance(instanceId);
 		validateVersion(version);
+		validateInstance(instanceId);
 		int recordsModified;
 		try (InputStreamReader inputStreamReader = new InputStreamReader(records.getInputStream())) {
 			recordsModified = batchWriteService.uploadTsvStream(inputStreamReader, instanceId, recordType);
@@ -150,7 +151,10 @@ public class RecordController {
 	@PostMapping("/{instanceid}/search/{version}/{recordType}")
 	public RecordQueryResponse queryForEntities(@PathVariable("instanceid") UUID instanceId,
 			@PathVariable("recordType") RecordType recordType,
+			@PathVariable("version") String version,
 			@RequestBody(required = false) SearchRequest searchRequest) {
+		validateVersion(version);
+		validateInstance(instanceId);
 		checkRecordTypeExists(instanceId, recordType);
 		if (null == searchRequest) {
 			searchRequest = new SearchRequest();
@@ -226,9 +230,7 @@ public class RecordController {
 	public ResponseEntity<String> deleteInstance(@PathVariable("instanceId") UUID instanceId,
 												 @PathVariable("version") String version) {
 		validateVersion(version);
-		if (!recordDao.instanceSchemaExists(instanceId)) {
-			throw new ResponseStatusException(HttpStatus.NOT_FOUND, "This instance does not exist");
-		}
+		validateInstance(instanceId);
 		recordDao.dropSchema(instanceId);
 		return new ResponseEntity<>(HttpStatus.OK);
 	}
@@ -238,6 +240,7 @@ public class RecordController {
 			@PathVariable("version") String version, @PathVariable("recordType") RecordType recordType,
 			@PathVariable("recordId") String recordId) {
 		validateVersion(version);
+		validateInstance(instanceId);
 		boolean recordFound = recordDao.deleteSingleRecord(instanceId, recordType, recordId);
 		return recordFound ? new ResponseEntity<>(HttpStatus.NO_CONTENT) : new ResponseEntity<>(HttpStatus.NOT_FOUND);
 	}
@@ -256,6 +259,7 @@ public class RecordController {
 	public ResponseEntity<RecordTypeSchema> describeRecordType(@PathVariable("instanceId") UUID instanceId,
 			@PathVariable("v") String version, @PathVariable("type") RecordType recordType) {
 		validateVersion(version);
+		validateInstance(instanceId);
 		checkRecordTypeExists(instanceId, recordType);
 		RecordTypeSchema result = getSchemaDescription(instanceId, recordType);
 		return new ResponseEntity<>(result, HttpStatus.OK);
@@ -306,6 +310,7 @@ public class RecordController {
 	public ResponseEntity<BatchResponse> streamingWrite(@PathVariable("instanceid") UUID instanceId,
 			@PathVariable("v") String version, @PathVariable("type") RecordType recordType, InputStream is) {
 		validateVersion(version);
+		validateInstance(instanceId);
 		int recordsModified = batchWriteService.consumeWriteStream(is, instanceId, recordType);
 		return new ResponseEntity<>(new BatchResponse(recordsModified, "Huzzah"), HttpStatus.OK);
 	}
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -91,6 +91,7 @@ public class RecordDao {
 		namedTemplate.getJdbcTemplate().update("create schema " + quote(instanceId.toString()));
 	}
 
+	@SuppressWarnings("squid:S2077") // since instanceId must be a UUID, it is safe to use inline
 	public void dropSchema(UUID instanceId) {
 		namedTemplate.getJdbcTemplate().update("drop schema " + quote(instanceId.toString()) + " cascade");
 	}
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -91,6 +91,10 @@ public class RecordDao {
 		namedTemplate.getJdbcTemplate().update("create schema " + quote(instanceId.toString()));
 	}
 
+	public void dropSchema(UUID instanceId) {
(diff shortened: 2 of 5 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -62,9 +64,20 @@ class RecordControllerMockMvcTest {
 
 	private static String versionId = "v0.2";
 
-	@BeforeAll
-	private static void createWorkspace() {
+	@BeforeEach
+	void beforeEach() throws Exception {
 		instanceId = UUID.randomUUID();
+		createInstance(instanceId);
+	}
```
davidangb, line 71: note the change here: moving `UUID.randomUUID();` from `BeforeAll` to `BeforeEach` means that tests will now each use a *different* instance, instead of most tests using the same instance.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -62,9 +64,20 @@ class RecordControllerMockMvcTest {
...
+
+	@AfterEach
+	void afterEach() throws Exception {
+		try {
+			mockMvc.perform(delete("/instances/{v}/{instanceid}",
+					versionId, instanceId).content("")).andExpect(status().isOk());
+		} catch (Throwable t)  {
+			 // noop - if we fail to delete the instance, don't fail the test
+		}
```
davidangb, line 80: the noop catch here feels yucky, but 9 tests that throw errors cause the delete-instance call to fail due to an aborted transaction. Removing `@Transactional` from each test makes this better and allows delete-instance to succeed. I think that's the right solution but I wanted to stop shaving yaks somewhere … I could make this change in a follow-on PR.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -75,13 +85,50 @@ void createInstanceAndTryToCreateAgain() throws Exception {
...
+		mockMvc.perform(delete("/instances/{version}/{instanceId}", versionId, uuid)).andExpect(status().isOk());
+	}
+
 	@Test
 	@Transactional
 	void tsvWithNoRowsShouldReturn400() throws Exception {
 		MockMultipartFile file = new MockMultipartFile("records", "no_data.tsv", MediaType.TEXT_PLAIN_VALUE,
 				"col1\tcol2\n".getBytes());
 
-		mockMvc.perform(post("/instances/{version}/{instanceId}", versionId, instanceId));
```
davidangb, line 84: here and in a handful of other tests, I deleted an explicit call to create-instance in favor of the create-instance in the `@BeforeEach`.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -713,14 +751,12 @@ void describeNonexistentType() throws Exception {
 	@Test
 	@Transactional
 	void describeAllTypes() throws Exception {
-		// replace instanceId for this test so only these records are found
-		UUID instId = UUID.randomUUID();
```
davidangb, line 717: here and in 2 other tests, I removed the code that used a brand new instanceId, since the `@BeforeEach` now creates new instanceIds automatically.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -224,6 +222,17 @@ public ResponseEntity<String> createInstance(@PathVariable("instanceId") UUID in
 		return new ResponseEntity<>(HttpStatus.CREATED);
 	}
 
+	@DeleteMapping("/instances/{version}/{instanceId}")
+	public ResponseEntity<String> deleteInstance(@PathVariable("instanceId") UUID instanceId,
+												 @PathVariable("version") String version) {
+		validateVersion(version);
+		if (!recordDao.instanceSchemaExists(instanceId)) {
```
aaronkanzer, line 229: Could you call `validateInstance(instanceId)` here as well instead of the conditional?

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -91,6 +91,11 @@ public void createSchema(UUID instanceId) {
 		namedTemplate.getJdbcTemplate().update("create schema " + quote(instanceId.toString()));
 	}
 
+	@SuppressWarnings("squid:S2077") // since instanceId must be a UUID, it is safe to use inline
+	public void dropSchema(UUID instanceId) {
+		namedTemplate.getJdbcTemplate().update("drop schema " + quote(instanceId.toString()) + " cascade");
```
aaronkanzer, line 96: Just wanted to double-check -- are there any use-cases where we would not want to cascade-delete?

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -224,6 +222,17 @@ public ResponseEntity<String> createInstance(@PathVariable("instanceId") UUID in
 		return new ResponseEntity<>(HttpStatus.CREATED);
 	}
 
+	@DeleteMapping("/instances/{version}/{instanceId}")
```
aaronkanzer, line 225: This might be the WDS-newbie in me...but what kind of authentication do we have wrapped around these endpoints? (especially those with `delete` scare me slightly....)

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -287,6 +287,27 @@ paths:
             'application/json':
               schema:
                 $ref: '#/components/schemas/ErrorResponse'
+    delete:
+      summary: Delete an instance
+      description: |
+        Delete the instance with the given UUID. This API is liable to change.
```
aaronkanzer, line 293: What does `This API is liable to change.` mean in this case?

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -75,13 +85,50 @@ void createInstanceAndTryToCreateAgain() throws Exception {
...
+		// creating again should 201
+		mockMvc.perform(post("/instances/{version}/{instanceId}", versionId, uuid)).andExpect(status().isCreated());
+	}
+
+	@Test
+	@Transactional
+	void deleteInstanceContainingData() throws Exception {
+		UUID uuid = UUID.randomUUID();
+		// creating the instance should 201
+		mockMvc.perform(post("/instances/{version}/{instanceId}", versionId, uuid)).andExpect(status().isCreated());
```
aaronkanzer, line 109: Per this comment: https://github.com/DataBiosphere/terra-workspace-data-service/pull/141/files#r1023212693 -- is there a reason that you have the random UUI and mock called here that is normally handled in your `@BeforeEach` setup?

service/src/test/java/org/databiosphere/workspacedataservice/controller/FullStackRecordControllerTest.java
```diff
@@ -69,6 +72,22 @@ static void setUp() {
...
+	}
+
+	@AfterEach
+	void afterEach() {
+		ResponseEntity<String> response = restTemplate.exchange("/instances/{v}/{instanceid}", HttpMethod.DELETE,
+				new HttpEntity<>("", headers), String.class,
+				versionId, instanceId);
+		assertEquals(HttpStatus.OK, response.getStatusCode());
+	}
+
```
sagehen03, line 90: These changes do ensure that each test has a clean schema, but they also result in significantly more API calls.  It's anecdotal but the few times I ran the tests in Intellij, the main branch was generally 100-200ms faster when running all the tests in FullStackRecordControllerTest than your PR branch.  Are the any advantages to this change besides ensuring that no test will step on another?

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
sagehen03: Similar to my comment on FullStackRecordController, my preference would be to make fewer API calls and to preserve test speed as much as possible.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
davidangb: good call, changed.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
davidangb: not that I can think of! Do you have one?

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
davidangb: for now - when WDS is single-user - no auth built into WDS itself. Leo and the Azure Relay that sit in front of WDS auth the user for us. Once WDS becomes multi-user, we'll need to build authorization into WDS; see AJ-667 and others.

service/src/main/resources/static/swagger/openapi-docs.yaml
davidangb: when we support multiple WDS instances per workspace, we'll need a change to this API to specify the parent workspace id. Beyond that, I'm not sure. We could also just make a new version of the API and leave this one in place. For now, this seems like a safe statement to make.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
davidangb: good call, I missed this one. Fixed!

sagehen03 (approved):
I'd rather not switch to @BeforeEach mostly for performance reasons but I understand that it does ensure a clean slate for each test and eliminates the chance of the tests interacting with each other.  If there's another reason for the switch please share.  On balance seems like an even-ish trade off.

davidangb:
@sagehen03 I ran some test timings of my own after seeing your comments. Full results are here: https://docs.google.com/spreadsheets/d/1fj4Bz63E11k2tP8k10ll3jDAXussQzYlX1QvzblVkm8/edit#gid=0

I compared this PR branch to the main branch, and ran each of the following 5 times. The averaged results are:
* `FullStackRecordControllerTest`, as measured by IntelliJ: PR branch is 220.2ms slower
* `RecordControllerMockMvcTest `, as measured by IntelliJ: PR branch is 212.6ms slower (but note that the branch also adds two test cases)
* `service:test`, as measured by `gradle --profile` at a command line: PR branch is 489ms _faster_

I honestly don't know why the branch is faster for `service:test` overall (and I triple-checked I didn't mix up the branches).

My overall impression is that ~200ms is a more than acceptable tradeoff to ensure test correctness. In the dev process for this PR, as I started to make changes to `FullStackRecordControllerTest`, I was dismayed that changing one test affected the others, because they were all using the same instance.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
aaronkanzer: Seems there might be one more place here too? https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java#L220

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -150,7 +151,10 @@ public ResponseEntity<StreamingResponseBody> streamAllEntities(@PathVariable("in
 	@PostMapping("/{instanceid}/search/{version}/{recordType}")
 	public RecordQueryResponse queryForEntities(@PathVariable("instanceid") UUID instanceId,
 			@PathVariable("recordType") RecordType recordType,
+			@PathVariable("version") String version,
 			@RequestBody(required = false) SearchRequest searchRequest) {
+		validateVersion(version);
+		validateInstance(instanceId);
 		checkRecordTypeExists(instanceId, recordType);
```
aaronkanzer, line 158: Just a random thought/curious to get others thoughts...

It seems we are calling `validateInstance` separately in each of these endpoints, prior to another call to the DB to do whatever else the endpoint prescribes. Just wonder if perhaps it would be worthwhile to append the query in `validateInstance` with whatever the subsequent query would be in the endpoint?

This perhaps might make the code more complex than necessary -- just curious about performance here as well.

sagehen03:
> @sagehen03 I ran some test timings of my own after seeing your comments. Full results are here: https://docs.google.com/spreadsheets/d/1fj4Bz63E11k2tP8k10ll3jDAXussQzYlX1QvzblVkm8/edit#gid=0
> 
> I compared this PR branch to the main branch, and ran each of the following 5 times. The averaged results are:
> 
>     * `FullStackRecordControllerTest`, as measured by IntelliJ: PR branch is 220.2ms slower
> 
>     * `RecordControllerMockMvcTest `, as measured by IntelliJ: PR branch is 212.6ms slower (but note that the branch also adds two test cases)
> 
>     * `service:test`, as measured by `gradle --profile` at a command line: PR branch is 489ms _faster_
> 
> 
> I honestly don't know why the branch is faster for `service:test` overall (and I triple-checked I didn't mix up the branches).
> 
> My overall impression is that ~200ms is a more than acceptable tradeoff to ensure test correctness. In the dev process for this PR, as I started to make changes to `FullStackRecordControllerTest`, I was dismayed that changing one test affected the others, because they were all using the same instance.

Fair enough, but I don't think test correctness was compromised in previous set up.  It was a choice I made to do set up one time before all tests in the suite and to let the tests share an instance and impose upon test writers the burden of ensuring they choose unique record type names for their test amongst all the record type names used in test suite.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
davidangb: That link is the opposite. Everywhere else we call `validateInstance`, we are ensuring the instance already exists and will throw a 404 if it does not. In your link - inside of the `createInstance` method - we ensure the instance does NOT already exist and will throw a 409 if it does.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
davidangb: @aaronkanzer I would prefer to keep it as a separate call for now - I don't think combining makes sense. Honestly, sometime soon I'd like to break `RecordController` into smaller files, potentially leaving all the validation in place but the implementations in another file, and I think combining the SQL calls would get messy.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
aaronkanzer: I don't -- just wanted to double-check :D

aaronkanzer (approved):

PR #142, opened 2022-11-16
AJ-695: Let user specify primary key column for TSV upload

sagehen03:
- setting this aside for a bit
- getting closer, need to restore name validation stuff
- tests all passing
- reducing changes
- adding caching
- switch to optional for tsv
- naming cleanup

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
@@ -17,6 +17,7 @@ import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.jdbc.DataSourceBuilder;
+import org.springframework.cache.annotation.EnableCaching;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Primary;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
@@ -28,6 +29,7 @@ import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
 import javax.sql.DataSource;
 
 @SpringBootApplication
+@EnableCaching
 public class WorkspaceDataServiceApplication {
 
 	@Bean
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -32,6 +32,7 @@ import org.springframework.jdbc.core.RowMapper;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.stereotype.Repository;
+import org.springframework.transaction.annotation.Transactional;
 import org.springframework.web.server.ResponseStatusException;
 
 import java.math.BigDecimal;
@@ -116,6 +117,7 @@ public class RecordDao {
 	}
 
 	@SuppressWarnings("squid:S2077")
+	@Transactional
 	public void createRecordType(UUID instanceId, Map<String, DataTypeMapping> tableInfo, RecordType recordType,
 								 Set<Relation> relations, String recordTypePrimaryKey) {
 		setRecordIdColumn(recordType, instanceId, recordTypePrimaryKey);
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -5,7 +5,6 @@ import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.MapDifference;
 import com.google.common.collect.Maps;
-import com.google.common.collect.Streams;
 import org.databiosphere.workspacedataservice.service.DataTypeInferer;
 import org.databiosphere.workspacedataservice.service.InBoundDataSource;
 import org.databiosphere.workspacedataservice.service.RelationUtils;
@@ -46,7 +45,6 @@ import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -58,7 +56,6 @@ import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import static org.databiosphere.workspacedataservice.service.model.ReservedNames.RESERVED_NAME_PREFIX;
 import static org.databiosphere.workspacedataservice.service.model.exception.InvalidNameException.NameType.ATTRIBUTE;
 import static org.databiosphere.workspacedataservice.service.model.exception.InvalidNameException.NameType.RECORD_TYPE;
 
@@ -91,22 +88,17 @@ public class RecordDao {
 
 	public void createSchema(UUID instanceId) {
 		namedTemplate.getJdbcTemplate().update("create schema " + quote(instanceId.toString()));
-		namedTemplate.getJdbcTemplate().update("create table " + quote(instanceId.toString()) + ".sys_record_type_identifiers ( record_type varchar not null, record_id varchar not null, primary key (record_type))");
 	}
 
 	public String getRecordIdColumn(RecordType recordType, UUID instanceId){
-		return namedTemplate.queryForObject("select record_id from " + getQualifiedTableName("sys_record_type_identifiers", instanceId) + " where record_type = :recordType",
-				new MapSqlParameterSource("recordType", recordType.getName()), String.class);
-	}
-
-	public void setRecordIdColumn(RecordType recordType, UUID instanceId, String recordTypeRowIdentifier){
 		MapSqlParameterSource params = new MapSqlParameterSource("recordType", recordType.getName());
-		params.addValue("recordTypeRowIdentifier", recordTypeRowIdentifier);
-		namedTemplate.update("insert into " + getQualifiedTableName("sys_record_type_identifiers", instanceId)
-				+ "(record_type, record_id) values (:recordType, :recordTypeRowIdentifier)", params);
+		params.addValue("instanceId", instanceId.toString());
+		return namedTemplate.queryForObject("select kcu.column_name FROM information_schema.table_constraints tc " +
+						"JOIN information_schema.key_column_usage kcu ON tc.constraint_name = kcu.constraint_name AND tc.table_schema = kcu.table_schema " +
+						"JOIN information_schema.constraint_column_usage ccu ON ccu.constraint_name = tc.constraint_name AND ccu.table_schema = tc.table_schema " +
+						"WHERE tc.constraint_type = 'PRIMARY KEY' AND tc.table_schema = :instanceId AND tc.table_name= :recordType",
+				params, String.class);
 	}
-
-
 
 	public boolean recordTypeExists(UUID instanceId, RecordType recordType) {
 		return Boolean.TRUE.equals(namedTemplate.queryForObject(
@@ -120,7 +112,6 @@ public class RecordDao {
 	@Transactional
 	public void createRecordType(UUID instanceId, Map<String, DataTypeMapping> tableInfo, RecordType recordType,
 								 Set<Relation> relations, String recordTypePrimaryKey) {
-		setRecordIdColumn(recordType, instanceId, recordTypePrimaryKey);
 		String columnDefs = genColumnDefs(tableInfo, recordTypePrimaryKey);
 		try {
 			namedTemplate.getJdbcTemplate().update("create table " + getQualifiedTableName(recordType, instanceId)
@@ -140,11 +131,6 @@ public class RecordDao {
 		// N.B. recordType is sql-validated in its constructor, so we don't need it here
 		return quote(instanceId.toString()) + "."
 				+ quote(SqlUtils.validateSqlString(recordType.getName(), RECORD_TYPE));
-	}
-
-	private String getQualifiedTableName(String tableName, UUID instanceId) {
-		// N.B. recordType is sql-validated in its constructor, so we don't need it here
-		return quote(instanceId.toString()) + "." + tableName;
 	}
 
 	@SuppressWarnings("squid:S2077")
@@ -236,7 +222,7 @@ public class RecordDao {
 	}
 
 	private List<RecordColumn> getSchemaWithRowId(Map<String, DataTypeMapping> schema, String recordIdColumn) {
-		return Stream.concat(Stream.of(new RecordColumn(recordIdColumn, DataTypeMapping.STRING)), schema.entrySet().stream().map(e -> new RecordColumn(e.getKey(), e.getValue()))).collect(Collectors.toSet()).stream().toList();
+		return Stream.concat(Stream.of(new RecordColumn(recordIdColumn, DataTypeMapping.STRING)), schema.entrySet().stream().map(e -> new RecordColumn(e.getKey(), e.getValue()))).toList();
 	}
 
 	public void batchUpsertWithErrorCapture(UUID instanceId, RecordType recordType, List<Record> records,
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -22,6 +22,8 @@ import org.postgresql.jdbc.PgArray;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.cache.annotation.CacheEvict;
+import org.springframework.cache.annotation.Cacheable;
 import org.springframework.dao.DataAccessException;
 import org.springframework.dao.DataIntegrityViolationException;
 import org.springframework.dao.EmptyResultDataAccessException;
@@ -73,7 +75,7 @@ public class RecordDao {
 	private final ObjectMapper objectMapper;
 
 	public RecordDao(NamedParameterJdbcTemplate namedTemplate,
-			@Qualifier("streamingDs") NamedParameterJdbcTemplate templateForStreaming, DataTypeInferer inf, ObjectMapper objectMapper) {
+					 @Qualifier("streamingDs") NamedParameterJdbcTemplate templateForStreaming, DataTypeInferer inf, ObjectMapper objectMapper) {
 		this.namedTemplate = namedTemplate;
 		this.templateForStreaming = templateForStreaming;
 		this.inferer = inf;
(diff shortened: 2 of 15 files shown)


service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -206,6 +206,7 @@ paths:
             'application/json':
               schema:
                 $ref: '#/components/schemas/ErrorResponse'
+  /{instanceid}/tsv/{v}/{type}/{recordTypePrimaryKey}:
```
sagehen03, line 209: The underlying API end point in RecordController will work if the last path parameter (recordTypePrimaryKey) is absent and assume sys_name in that case, but I could not figure out how to make swagger and the swagger client support that invocation mode.  If reviewers know offhand how to make this work, please share.

davidangb: How about making it an optional query parameter, instead of a path parameter?

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -132,18 +150,18 @@ public List<Record> queryForRecords(RecordType recordType, int pageSize, int off
 		LOGGER.info("queryForRecords: {}", recordType.getName());
 		return namedTemplate.getJdbcTemplate().query(
 				"select * from " + getQualifiedTableName(recordType, instanceId) + " order by "
-						+ (sortAttribute == null ? RECORD_ID : quote(sortAttribute)) + " " + sortDirection + " limit "
+						+ (sortAttribute == null ? getPrimaryKeyColumn(recordType, instanceId) : quote(sortAttribute)) + " " + sortDirection + " limit "
 						+ pageSize + " offset " + offset,
 				new RecordRowMapper(recordType, objectMapper, instanceId));
 	}
 
 	public Map<String, DataTypeMapping> getExistingTableSchema(UUID instanceId, RecordType recordType) {
```
davidangb, line 158: thinking out loud, no changes requested for this PR: the name `getExistingTableSchema` feels misleading to me now; personally, I now consider the user-specified primary key column to be part of the schema instead of a system-managed column. But that's up for debate.

I also wonder if there's any chance to simplify/refactor any of the code that introspects or loops over columns, if we use the assumption that the user-specified primary key is a "real" column.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -91,6 +94,17 @@ public void createSchema(UUID instanceId) {
...
+	@Cacheable(value = PRIMARY_KEY_COLUMN_CACHE, key = "{ #recordType.name, #instanceId.toString()}")
+	public String getPrimaryKeyColumn(RecordType recordType, UUID instanceId){
+		MapSqlParameterSource params = new MapSqlParameterSource("recordType", recordType.getName());
+		params.addValue(INSTANCE_ID, instanceId.toString());
+		return namedTemplate.queryForObject("select kcu.column_name FROM information_schema.table_constraints tc " +
+						"JOIN information_schema.key_column_usage kcu ON tc.constraint_name = kcu.constraint_name AND tc.table_schema = kcu.table_schema " +
+						"JOIN information_schema.constraint_column_usage ccu ON ccu.constraint_name = tc.constraint_name AND ccu.table_schema = tc.table_schema " +
+						"WHERE tc.constraint_type = 'PRIMARY KEY' AND tc.table_schema = :instanceId AND tc.table_name = :recordType",
+				params, String.class);
+	}
```
davidangb, line 106: I *think* this is safe, but I want us to triple-check if, under a Leo/k8s deployment, there's any chance we have multiple replicas of WDS running simultaneously, offering a chance for cache to get out of sync.

service/src/test/java/org/databiosphere/workspacedataservice/controller/TsvDownloadTest.java
```diff
@@ -30,13 +36,45 @@ class TsvDownloadTest {
...
+		InputStreamReader reader = new InputStreamReader(inputStream, StandardCharsets.UTF_8);
+		Iterable<CSVRecord> records = new CSVParser(reader, csvFormat);
+		Iterator<CSVRecord> iterator = records.iterator();
+		CSVRecord rcd = iterator.next();
+		assertThat(rcd.get("sample_id")).isEqualTo("bazinga");
+		assertThat(rcd.get("col_1")).isEqualTo("Fido");
+		assertThat(rcd.get("col_2")).isEqualTo("Jerry");
+		assertThat(iterator.hasNext()).isFalse();
+		reader.close();
+	}
```
davidangb, line 72: Could you also add a test that verifies we'll use the first column as the primary key if an explicit PK isn't specified? (I don't know if it should go in this file or not, I'm just adding my comment here)

davidangb (commented):
I have a few, hopefully minor, questions/comments on the syntax of this PR - see inline!

I also have some broader questions about what functionality we want to have in the end. I will jot those down and distribute too.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -100,22 +114,26 @@ public boolean recordTypeExists(UUID instanceId, RecordType recordType) {
 		return Boolean.TRUE.equals(namedTemplate.queryForObject(
 				"select exists(select from pg_tables where schemaname = :instanceId AND tablename  = :recordType)",
 				new MapSqlParameterSource(
-						Map.of("instanceId", instanceId.toString(), "recordType", recordType.getName())),
+						Map.of(INSTANCE_ID, instanceId.toString(), "recordType", recordType.getName())),
 				Boolean.class));
 	}
 
 	@SuppressWarnings("squid:S2077")
+	@CacheEvict(value = PRIMARY_KEY_COLUMN_CACHE, key = "{ #recordType.name, #instanceId.toString()}")
```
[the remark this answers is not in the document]
sagehen03: Yes.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -319,7 +342,7 @@ private void checkForTableRelation(SQLException sqlEx) {
 
 	public Stream<Record> streamAllRecordsForType(UUID instanceId, RecordType recordType) {
 		return templateForStreaming.getJdbcTemplate().queryForStream(
-				"select * from " + getQualifiedTableName(recordType, instanceId) + " order by " + RECORD_ID,
+				"select * from " + getQualifiedTableName(recordType, instanceId) + " order by " + getPrimaryKeyColumn(recordType, instanceId),
```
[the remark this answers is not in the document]
sagehen03: This is good point.  I don't think I had a particular criteria.  I will look to see if I can move all invocations inside RecordDao.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
sagehen03: If there were multiple WDS instances sharing a db and you created a record type and then deleted that record type and then created a new record type using that same name as the deleted record type but used a different primary then you would have a problem if the WDS instance(s) not doing the second record type creation had previously cached the old primary key value.  I wrote this caching code with the understanding that we would not be deploying multiple instances of WDS.

service/src/test/java/org/databiosphere/workspacedataservice/controller/TsvDownloadTest.java
sagehen03: All the tsv upload tests that pre-date this PR exercise that code path `storeLargeIntegerValue` is one example.

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -116,28 +117,30 @@ private void validateRelationsAndAddColumns(UUID instanceId, RecordType recordTy
 	}
 
 	@Transactional
-	public int uploadTsvStream(InputStreamReader is, UUID instanceId, RecordType recordType) throws IOException {
+	public int uploadTsvStream(InputStreamReader is, UUID instanceId, RecordType recordType, Optional<String> uniqueRowIdentifierColumn) throws IOException {
 		CSVFormat csvFormat = TsvSupport.getUploadFormat();
 		CSVParser rows = csvFormat.parse(is);
+		String leftMostColumn = rows.getHeaderMap().keySet().iterator().next();
```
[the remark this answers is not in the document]
sagehen03: Yes, it's a LinkedHashMap behind the scenes. Though I did switch to your suggestion since it has fewer dots.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
davidangb: > the understanding that we would not be deploying multiple instances of WDS on top of the same db

agreed - and I am pretty certain that's the case, just want to be absolutely doubly triply sure

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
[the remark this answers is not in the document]
sagehen03: I've combined two of your comments into one changeset.  I renamed `getExistingTableSchema` to the descriptive and awkward `getExistingTableLessPrimaryKey` then I added `getAllAttributeNames` which I then used to replace the one `getPrimaryKeyColumn` call site that was required to be outside of `RecordDao`.  All call sites for getPrimaryKeyColumn are now inside RecordDao, but that also has implications for caching which you can see from the updated code.

service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
```diff
@@ -29,6 +30,7 @@
 import javax.sql.DataSource;
 
 @SpringBootApplication
+@EnableCaching
```
aaronkanzer, line 33: Does a `CacheManager` need to be registered here in tandem with this decorator?

Just noticed in the docs here: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/cache/annotation/EnableCaching.html

```
A bean of type [CacheManager](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/cache/CacheManager.html) must be registered, as there is no reasonable default that the framework can use as a convention. And whereas the <cache:annotation-driven> element assumes a bean named "cacheManager", @EnableCaching searches for a cache manager bean by type. Therefore, naming of the cache manager bean method is not significant.
```

service/src/main/java/org/databiosphere/workspacedataservice/dao/CachedQueryDao.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private final NamedParameterJdbcTemplate namedTemplate;
+
+    public CachedQueryDao(NamedParameterJdbcTemplate namedTemplate) {
+        this.namedTemplate = namedTemplate;
+    }
+
+    @Cacheable(value = PRIMARY_KEY_COLUMN_CACHE, key = "{ #recordType.name, #instanceId.toString()}")
+    public String getPrimaryKeyColumn(RecordType recordType, UUID instanceId){
+        LOGGER.warn("Calling the db to retrieve primary key for {}", recordType.getName());
```
aaronkanzer, line 35: Why a `LOGGER.warn` here?

service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
sagehen03: Since we're using Spring Boot in addition to the Spring Framework, that step is not necessary.  https://stackoverflow.com/questions/41739581/default-cache-manager-with-spring-boot-using-enablecaching

service/src/main/java/org/databiosphere/workspacedataservice/dao/CachedQueryDao.java
sagehen03: I thought it would be useful to be sure we track how often we're hitting the db.  Multiple invocations for the same record type and instance would signify caching is not working.  I used these log lines to identify the fact that caching wasn't working in my initial implementation.  Do you have opinions about this?  It seems like a pretty anodyne detail to me.

calypsomatic (approved):
Great work!

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -100,22 +105,25 @@ public boolean recordTypeExists(UUID instanceId, RecordType recordType) {
...
 		try {
 			namedTemplate.getJdbcTemplate().update("create table " + getQualifiedTableName(recordType, instanceId)
 					+ "( " + columnDefs + (!relations.isEmpty() ? ", " + getFkSql(relations, instanceId) : "") + ")");
 		} catch (DataAccessException e) {
 			if (e.getRootCause()instanceof SQLException sqlEx) {
 				checkForMissingTable(sqlEx);
 			}
+			//this exception is thrown from getFkSql if the referenced relation doesn't exist
+			if(e instanceof EmptyResultDataAccessException){
+				throw new MissingObjectException("Record type for relation");
```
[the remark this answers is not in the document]
sagehen03: This code is consistent with the existing error handling for the case of a missing record type in a relation https://github.com/DataBiosphere/terra-workspace-data-service/blob/6d9192c6a185bbfc843552aa400114da24d0dc0f/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java#L317 and unfortunately there's not an easy way to pull out what was the name of the record type that's missing.

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -194,7 +197,7 @@ public int consumeWriteStream(InputStream is, UUID instanceId, RecordType record
 				List<Record> records = info.getRecords();
 				if (firstUpsertBatch && info.getOperationType() == OperationType.UPSERT) {
 					schema = inferer.inferTypes(records, InBoundDataSource.JSON);
-					createOrModifyRecordType(instanceId, recordType, schema, records);
+					createOrModifyRecordType(instanceId, recordType, schema, records, RECORD_ID);
```
aaronkanzer, line 200: Just for clarification, this is where we are converting the user's primary key to be `sys_*` on our side of things?

service/src/main/java/org/databiosphere/workspacedataservice/dao/CachedQueryDao.java
aaronkanzer: No strong opinions -- was just curious if the warning was problematic in any manner

service/src/test/resources/small-test-no-sys.tsv
```diff
@@ -0,0 +1,5 @@
+greeting	unicode	json	date	embedded characters	double	long	z_array_of_string	z_long_array	z_double_array	z_z_boolean_array	zz_array_of_date	zz_array_of_datetime
+hello	🚀	"{""foo"":""bar""}"	2021-01-10	"test
+Foo
+Bar"	-2.287	-99	[“test1”, “test2”]		[-11.2221, -11]	"[true, “false""]"	[“2021-10-03”, “2022-10-03”]	[“2021-10-03T19:00:00”, “2022-10-03T23:01:00”]
+goodbye	😈		2022-11-01	"Two Beers	Or	Not Two Beers?"	3.14	1147417414841	[“98”, “99”]	[199, 12]			[]	[“2021-10-03T19:00:00”, “2022-10-03T23:01:00”]
```
aaronkanzer, line 5: "Two Beers	Or	Not Two Beers?" 

_That is the question...._

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
sagehen03: In this code path and in the PUT single record code path, the inbound JSON data contains a list of attributes and a record id for each record.  We route the record id value to a sys_name column for storage in the db.  This line of code ensures that the target table for the record type gets created with a sys_name column with a primary key index. The TSV payload is different because we don't have record ids, it's just a list of attributes per record with one of the attributes designated as the row unique identifier.  Does that help answer your question?

service/src/test/resources/small-test-no-sys.tsv
sagehen03: It's almost the weekend!

davidangb:
For the sake of complete comment threads, noting here that https://docs.google.com/document/d/1Mm0ugEgtjF2dD5Uu49CqeRK8pDKvxvEym8T12kjUANw/edit# has outstanding questions about this PR (and for the team - this is on the agenda for Monday to discuss in person)

service/src/main/java/org/databiosphere/workspacedataservice/dao/CachedQueryDao.java
davidangb, line 35: nit: since the cache key considers both recordType and instanceId, could you add the instanceId to the log statement here?

davidangb, line 35: also, thinking out loud for a future PR - I wonder if we can safely expose the cache metrics from actuator to watch how these caches perform. I would expect an extremely high hit rate.

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -116,28 +117,30 @@ private void validateRelationsAndAddColumns(UUID instanceId, RecordType recordTy
...
+		String uniqueIdentifierAsString = uniqueRowIdentifierColumn.orElse(leftMostColumn);
 		int recordsProcessed = 0;
 		for (CSVRecord row : rows) {
 			Map<String, Object> m = (Map) row.toMap();
-			m.remove(ROW_ID_COLUMN_NAME);
+			m.remove(uniqueIdentifierAsString);
 			changeEmptyStringsToNulls(m);
 			try {
-				batch.add(new Record(row.get(ROW_ID_COLUMN_NAME), recordType, new RecordAttributes(m)));
+				batch.add(new Record(row.get(uniqueIdentifierAsString), recordType, new RecordAttributes(m)));
```
davidangb, line 134: Have you tested what happens here if the row has the empty string for its id (i.e. `row.get(uniqueIdentifierAsString)` returns ""? (I have not tested this, I don't know the answer). Even if it all works and we insert a row with a pk of "", I think that record might end up being unretrievable by some of our APIs. Depending on the behavior in practice, we might need to change something here.

davidangb (commented):
one minor nit, one question, one thought for later. Beyond those I am :+1: for the code in this PR. I do want to discuss the issues in the separate google doc before merging, though, so we can coordinate all the follow-on work and understand what needs to be in place before this merges.

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
aaronkanzer: Yes, this is great -- thanks!

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
sagehen03: Your intuition was correct.  The code did store an empty string in the primary key in that case.  I've made the change to throw an illegal argument exception now, which fails the tsv upload in its entirety since I'm pretty sure the user would never want to set "" as their row identifier.

service/src/main/java/org/databiosphere/workspacedataservice/dao/CachedQueryDao.java
sagehen03: I looked into the actuator support for our cache and what I was able to activate was pretty underwhelming.  I couldn't figure out how to expose hit rate or cache size.  I think it might be necessary to use a more fully featured cache implementation but I'm not sure.

davidangb: just from reading the docs, the `caches` endpoint for actuator definitely looks boring. It might be hidden under metrics: https://docs.spring.io/spring-boot/docs/1.5.22.RELEASE/reference/html/production-ready-metrics.html#production-ready-datasource-cache … but this is all for some later time anyway

davidangb (approved):
thank you!

PR #143, opened 2022-11-17
disable CORS headers for non-local development [AJ-696]

davidangb:
When deployed behind Terra's Azure Relay, WDS's CORS responses and the Relay's CORS responses conflict with each other, resulting in CORS errors in the browser. Relay configuration [here](https://github.com/DataBiosphere/terra-azure-relay-listeners/blob/a9576c888e934d51ef9bad486aa69e8c2aa92e4a/service/src/main/java/org/broadinstitute/listener/relay/Utils.java#L34).

This PR changes WDS to _not_ issue CORS responses, unless WDS is running with the "local" Spring profile. That profile is not enabled by default, but engineers can toggle it on if they need to connect to a local WDS from a browser.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
@@ -19,6 +19,7 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.jdbc.DataSourceBuilder;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Primary;
+import org.springframework.context.annotation.Profile;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.lang.NonNull;
 import org.springframework.web.servlet.config.annotation.AsyncSupportConfigurer;
@@ -83,8 +84,18 @@ public class WorkspaceDataServiceApplication {
 		return jdbcTemplate;
 	}
 
-	// CORS: allow Ajax requests from anywhere for all endpoints.
+	/**
+	 * Configure CORS response headers.
+	 *
+	 * When running behind Terra's Azure Relay, the Relay handles CORS response headers, so WDS should not;
+	 * the two CORS configurations will conflict.
+	 *
+	 * When running WDS locally for development - i.e. not behind a Relay - you may need to enable headers.
+	 * To do so, activate the "local" Spring profile by setting spring.profiles.active=local in
+	 * application.properties (or other Spring techniques for activating a profile)
+	 */
 	@Bean
+	@Profile("local")
 	public WebMvcConfigurer corsConfigurer() {
 		return new WebMvcConfigurer() {
 			@Override
@@ -93,7 +104,15 @@ public class WorkspaceDataServiceApplication {
 						.allowedOrigins("*");
 
 			}
+		};
+	}
 
+	/**
+	 * Configure the app for asynchronous request processing.
+	 */
+	@Bean
+	public WebMvcConfigurer asyncConfigurer() {
+		return new WebMvcConfigurer() {
 			@Override
 			public void configureAsyncSupport(@NonNull AsyncSupportConfigurer configurer) {
 				configurer.setDefaultTimeout(-1);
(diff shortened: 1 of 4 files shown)


service/src/main/resources/application.properties
```diff
@@ -20,3 +20,7 @@ spring.servlet.multipart.max-file-size=5GB
 
 twds.write.batch.size=5000
 twds.streaming.fetch.size=5000
+
+# activate the "local" profile to turn on CORS response headers,
```
aaronkanzer, line 24: Don't want to increase scope here too much, but wondering if we should slowly transition WDS to use an environment-specific `application.properties` file? (Feel free to disregard for now, unless this is something for production-readiness that we could tackle sooner than later)

e.g. https://docs.spring.io/spring-boot/docs/2.1.9.RELEASE/reference/html/boot-features-external-config.html#boot-features-external-config-application-property-files

service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
```diff
@@ -83,8 +84,18 @@ public NamedParameterJdbcTemplate templateForStreaming(@Qualifier("streamingDs")
...
+	 *
+	 * When running behind Terra's Azure Relay, the Relay handles CORS response headers, so WDS should not;
+	 * the two CORS configurations will conflict.
+	 *
+	 * When running WDS locally for development - i.e. not behind a Relay - you may need to enable headers.
+	 * To do so, activate the "local" Spring profile by setting spring.profiles.active=local in
+	 * application.properties (or other Spring techniques for activating a profile)
+	 */
 	@Bean
+	@Profile("local")
```
sagehen03, line 98: I'd rather use something like `@Profile("!azure")` that way any deployment scripts can include azure among the active profiles, and this bean won't be active in those environments, but for local dev it will.  The code in this PR will require devs to set  `-Dspring.profiles.active=local` in their jvm args or to update their application.properties and remember to set those values whenever they set up a new dev environment.  I'd rather impose that burden upon the deploy script than the local environments.

service/src/main/resources/application.properties
sagehen03, line 24: If we go this route (which I'd vote against), I'd rather update the readme and advise devs to set `-Dspring.profiles.active=local `

davidangb: I'm totally open to using an env-specific properties file. I don't see a need to do so as part of this ticket - the CORS on/off is just one feature. Do you think that's something we need to do now?

aaronkanzer: Yeah no need to do it for this ticket -- I figured I'd mention in case this is a pattern that starts to arise as we transition from dev-only to production.

Seems @sagehen03 's approach for the jvm arg seems simplistic enough for accommodating this though

service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
davidangb: I disagree pretty strongly with what we're optimizing for here. This local development issue currently affects less than 5 people, who are all internal - and even then, it only affects those developers if they are running a UI against their local WDS; simply developing WDS on its own is unaffected by CORS. I'd much rather optimize for seamless production and BEE deployments, without imposing requirements on what those deployments need to specify and running the risk of anything going wrong there.

As a middle ground, what if I added a new gradle task named "runLocal" or something, which automatically enables the profile, and we changed the dev instructions to say `./gradlew runLocal` instead of `./gradlew bootRun`?

sagehen03: I think (admittedly I don't know what it's going to look like) specifying the Azure profile would be done one time by our deployment mechanism.  That doesn't feel like a big imposition.  We have 5 developers plus any developers on adjacent teams that want to run WDS locally, plus coops, plus new machines, plus new checkout.. in all those cases the developer would have to remember that step.

sagehen03: Thanks for chatting over meet and making me feel more heard, @davidangb I agree to disagree with you on this.

sagehen03 (approved):

aaronkanzer (approved):

PR #144, opened 2022-11-21
AJ-723: Make it possible to turn off logback file logging.

sagehen03:
- add dev profile and don't log when it's active
- change profile name to local


davidangb (approved):
looks good, did not test myself.

calypsomatic (approved):

PR #146, opened 2022-11-22
AJ-726: Add Primary Key to JSON response

sagehen03:
This PR adds the primary key to the attribute map in the JSON response.    I also took the liberty to add code to order the JSON attributes with PK first and all other attributes following in alphabetical order and to order the TSV columns by that same ordering scheme.  When I tested with terra UI, [this code](https://github.com/DataBiosphere/terra-ui/blob/7976a29f882b3814ff7d65a08cbfbcf949257ed0/src/components/data/DataTable.js#L154-L156) led to problems, my fix was to make those lists match on the server side and update the schema API to include the primary key column too as part of this PR.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -278,7 +278,7 @@ public class RecordController {
 	}
 
 	private RecordTypeSchema getSchemaDescription(UUID instanceId, RecordType recordType) {
-		Map<String, DataTypeMapping> schema = recordDao.getExistingTableSchemaLessPrimaryKey(instanceId, recordType);
+		Map<String, DataTypeMapping> schema = recordDao.getExistingTableSchema(instanceId, recordType);
 		Map<String, RecordType> relations = recordDao.getRelationCols(instanceId, recordType).stream()
 				.collect(Collectors.toMap(Relation::relationColName, Relation::relationRecordType));
 		List<AttributeSchema> attrSchema = schema.entrySet().stream().sorted(Map.Entry.comparingByKey())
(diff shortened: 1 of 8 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/shared/model/AttributeComparator.java
```diff
@@ -0,0 +1,19 @@
...
+    @Override
+    public int compare(String o1, String o2) {
+        if (o1.equals(primaryKey)) {
+            return -1;
+        }
+        if (o2.equals(primaryKey)) {
+            return 1;
+        }
+        Collator collator = Collator.getInstance();
+        return collator.compare(o1, o2);
```
davidangb, line 17: What's the difference between this and doing `o1.compareTo(o2)`?

davidangb (approved):
:+1:, modulo the question about `collator.compare` vs `String.compareTo`. I don't know if there are any subtle differences based on collations/codepoints here.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -278,7 +278,7 @@ public ResponseEntity<List<RecordTypeSchema>> describeAllRecordTypes(@PathVariab
 	}
 
 	private RecordTypeSchema getSchemaDescription(UUID instanceId, RecordType recordType) {
-		Map<String, DataTypeMapping> schema = recordDao.getExistingTableSchemaLessPrimaryKey(instanceId, recordType);
+		Map<String, DataTypeMapping> schema = recordDao.getExistingTableSchema(instanceId, recordType);
```
sagehen03, line 281: This change was necessary to accommodate this [code](https://github.com/DataBiosphere/terra-ui/blob/7976a29f882b3814ff7d65a08cbfbcf949257ed0/src/components/data/DataTable.js#L154-L156)

calypsomatic (approved):

PR #147, opened 2022-11-22
DRAFT: corner case TSV inputs

davidangb:
Creating as a draft. This contains two failing unit tests so can't be merged as-is. It also likely duplicates tests elsewhere in the code; we should hunt down and resolve dupilcations.

This PR adds a parameterized unit test that checks various TSV syntaxes and asserts on the data created by those inputs.


davidangb:
This has been incorporated into #152. Closing.

PR #148, opened 2022-11-22
Aj 653 relation arrays

calypsomatic:
See [AJ-653 Support arrays-of-relations](https://broadworkbench.atlassian.net/browse/AJ-653)
This PR allows arrays of relations as an attribute column in a record.  The column is stored in the record type table along with a join table to ensure foreign-key consistency.  
This PR does not implement update or deletion of relation-array columns.  These features will be addressed in a follow-on PR.


service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
```diff
@@ -0,0 +1,123 @@
...
+        this.recordDao = recordDao;
+        this.inferer = inferer;
+    }
+
+    public void prepareAndUpsert(UUID instanceId, RecordType recordType, List<Record> records,
+                                 Map<String, DataTypeMapping> requestSchema) {
+        prepareAndUpsert(instanceId, recordType, records, requestSchema, RECORD_ID);
+    }
+
+        public void prepareAndUpsert(UUID instanceId, RecordType recordType, List<Record> records,
```
sagehen03, line 40: There's a lot going on in this method.  As I see things, these are the steps to handling arrays of relations: 1. Identify which attributes in the request are arrays of relations. 2. Verify that those attributes that are array of relations point to the same record type within the array and across records for the same attribute 3. Create join table with the proper foreign key(s) for the data identified in the previous step. 4. "Flatten" the array of relations data into a list of (referenced record_id, parent_record_id) so that the data can be inserted into the join tables. 5. Insert into the join table and upon error roll back and report foreign key violations as best we can.  6. If step 5 succeeds, insert the rest of the attribute data (the non array of realations).

It seems like in this implementation you are doing steps 2, 4, and 5.  I think there's an advantage to putting each step into its own method and naming it accordingly.  Also I think there's a small advantage to doing the input validation before creating the join table as you can do that step outside of a transaction and fail the operation before you've made any writes if there are invalid inputs.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
```diff
@@ -0,0 +1,123 @@
...
+                    if (!RelationUtils.getTypeValue(r).equals(relDef.relationRecordType())){
+                        throw new InvalidRelationException("It looks like you're attempting to assign a relation "
+                                + "to multiple record types");
+                    }
+                    relList.add(new RelationValue(rec, new Record(RelationUtils.getRelationValue(r), RelationUtils.getTypeValue(r), new RecordAttributes(Collections.emptyMap()))));
+                }
+                relationArrayValues.put(relDef, relList);
+            }
+        }
+        recordDao.batchUpsert(instanceId, recordType, records, relationArrays.get(false), primaryKey);
```
sagehen03, line 73: As we discussed last week, I would not remove the array of relation data from what gets written with the batchUpsert call.  Doing so will make reads work with no changes, while making writes more complicated but I think it's worth the tradeoff.

sagehen03 (commented):
I left a couple of high level comments.  Let's discuss synchoronously if I'm not being especially clear.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -182,6 +183,25 @@ public RecordQueryResponse queryForEntities(@PathVariable("instanceid") UUID ins
 		return new RecordQueryResponse(searchRequest, recordList, totalRecords);
 	}
 
+	private List<Record> queryForRecords(RecordType recordType, int pageSize, int offset, String sortDirection,
```
sagehen03, line 186: I think most of this code can go away if we store the AoR data twice?

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -326,6 +346,16 @@ private void createRecordTypeAndInsertRecords(UUID instanceId, Record newRecord,
 			Map<String, DataTypeMapping> requestSchema) {
 		List<Record> records = Collections.singletonList(newRecord);
 		recordDao.createRecordType(instanceId, requestSchema, recordType, RelationUtils.findRelations(records), ReservedNames.RECORD_ID);
-		recordDao.batchUpsert(instanceId, recordType, records, requestSchema);
+		recordService.prepareAndUpsert(instanceId, recordType, records, requestSchema);
 	}
+
+	private Map<String, DataTypeMapping> getFullTableSchema(UUID instanceId, RecordType recordType){
+		List<Relation> relationArrays = recordDao.getRelationArrayCols(instanceId, recordType);
+		Map<String, DataTypeMapping> schema = recordDao.getExistingTableSchemaLessPrimaryKey(instanceId, recordType);
```
sagehen03, line 354: I think this call should already return the AoR attributes since you're using domaiins?

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -110,12 +114,61 @@ public boolean recordTypeExists(UUID instanceId, RecordType recordType) {
...
+	public void createRelationJoinTable(UUID instanceId, String tableName, RecordType referringRecordType,
+								 RecordType referencedRecordType) {
+		String fromCol = getFromColumnName(referringRecordType);
+		String toCol = getToColumnName(referencedRecordType);
+		String columnDefs =  quote(fromCol) + " text, " + quote(toCol) + " text";
+		//Possibly temporary fake relations to make existing methods work for this situation
+		Set<Relation> relations = Set.of(new Relation(fromCol, referringRecordType), new Relation(toCol, referencedRecordType));
+		try {
+			namedTemplate.getJdbcTemplate().update("create table " + getQualifiedJoinTableName(instanceId, tableName, referringRecordType) +
+					"( " + columnDefs +  (!relations.isEmpty() ? ", " + getFkSql(relations, instanceId) : "") + ")");
```
sagehen03, line 171: I think relations will always be non-empty.

service/src/main/java/org/databiosphere/workspacedataservice/service/RelationUtils.java
```diff
@@ -23,16 +24,20 @@ public class RelationUtils {
 	 *            - all records whose references to check
 	 * @return Set of Relation for all referencing attributes
 	 */
-	public static Set<Relation> findRelations(List<Record> records) {
-		Set<Relation> result = new HashSet<>();
+	public static RelationCollection findRelations(List<Record> records) {
+		Set<Relation> relations = new HashSet<>();
+		Set<Relation> relationArrays = new HashSet<>();
 		for (Record record : records) {
 			for (Map.Entry<String, Object> entry : record.attributeSet()) {
```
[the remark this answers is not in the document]
calypsomatic: Yes that makes my life a lot easier!  But I still need to distinguish between tsv and json sources, and haven't got a nice way to do so without passing the source around, which I think i can't do here because of methods getting reused upstream.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
calypsomatic: I pulled out a number of actions into separate methods; could probably still use some cleanup but it's at least better now.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
calypsomatic: It wasn't when I was only storing AoR once, but now it is!

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -173,7 +173,7 @@ void writeAndReadJson() throws Exception {
 	void writeAndReadAllDataTypesJson() throws Exception {
 		String rt = "all-types";
 		RecordAttributes attributes = TestUtils.getAllTypesAttributesForJson();
-		assertEquals(attributes.attributeSet().size(), DataTypeMapping.values().length);
+//		assertEquals(attributes.attributeSet().size(), DataTypeMapping.values().length);
```
calypsomatic, line 176: RELATION and ARRAY_OF_RELATION are now included in the list of DataTypeMapping types; however adding in default values for them is more complicated since it would require existing records to refer to.  Debating whether to remove this line, edit it to expect the `allTypesAttributes` to contain everything BUT relations, or include relations in the `allTypes`.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
```diff
@@ -0,0 +1,132 @@
...
+        return new RelationValue(fromRecord, new Record(RelationUtils.getRelationValue(toString), RelationUtils.getTypeValue(toString), new RecordAttributes(Collections.emptyMap())));
+    }
+
+
+    private void checkAllRefsMatch(List<String> rels, RecordType typeToMatch){
+        if (!rels.stream().map(RelationUtils::getTypeValue).allMatch(r -> r.equals(typeToMatch))){
+            throw new InvalidRelationException("It looks like you're attempting to assign a relation "
+                    + "to multiple record types");
+        }
+    }
```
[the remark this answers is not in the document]
davidangb: Update: I realize that my suggestion essentially reimplements `RelationUtils.getTypeValueForList`. Could you do away with `checkAllRefsMatch` entirely and use `RelationUtils.getTypeValueForList` instead?

service/src/main/java/org/databiosphere/workspacedataservice/service/model/RelationValue.java
```diff
@@ -0,0 +1,4 @@
+package org.databiosphere.workspacedataservice.service.model;
+
+public record RelationValue(org.databiosphere.workspacedataservice.shared.model.Record fromRecord, org.databiosphere.workspacedataservice.shared.model.Record toRecord) {
```
davidangb, line 3: nit: add `import org.databiosphere.workspacedataservice.shared.model.Record` and make this just:
```java
public record RelationValue(Record fromRecord, Record toRecord)
```

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
davidangb: `getAllTypesAttributesForJson()` is only called from here, so it seems like any approach you take would be fine.

I'd lean towards more test coverage, which would mean:
* insert one or two dummy target records
* add a scalar relation and array of relations to `getAllTypesAttributesForJson()`, targeting the dummy records you just inserted
* also update the `getExpectedAllAttributesJsonText()` value

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -250,6 +250,36 @@ void simpleTsvUploadWithBatchingShouldSucceed(@Value("${twds.write.batch.size}")
...
+	@Test
+	@Transactional
+	void tsvUploadWithRelationsShouldSucceed() throws Exception {
+		RecordType refType = RecordType.valueOf("refType");
+		createSomeRecords(refType, 3);
+
+		StringBuilder tsvContent = new StringBuilder("sys_name\trel\trelArr\n");
+		String singleRel = RelationUtils.createRelationString(refType, "record_0");
+		String relArr = "[\"" + RelationUtils.createRelationString(refType, "record_1") + "\", \"" + RelationUtils.createRelationString(refType, "record_2") + "\"]";
+		for (int i = 0; i < 5 + 1; i++) {
```
davidangb, line 262: nit: `i < 5 + 1` could be `i < 6`, yeah?

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -110,12 +114,58 @@ public boolean recordTypeExists(UUID instanceId, RecordType recordType) {
...
+			}
+			//this exception is thrown from getFkSql if the referenced relation doesn't exist
+			if(e instanceof EmptyResultDataAccessException){
+				throw new MissingObjectException("Record type for relation");
+			}
+			throw e;
+		}
+	}
+
+	private String getJoinTableName(String tableName, RecordType referringRecordType){
```
davidangb, line 139: nit: for super clarity, give both the arguments a descriptive name? Maybe "toTable" and "fromTable", or whatever works for you? I find just "tableName" to be confusing here.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -110,12 +114,58 @@ public boolean recordTypeExists(UUID instanceId, RecordType recordType) {
...
+	}
+
+	private String getToColumnName(RecordType referencedRecordType){
+		return "to_" + referencedRecordType.getName() + "_key";
+
+	}
+
+	@SuppressWarnings("squid:S2077")
+	public void createRelationJoinTable(UUID instanceId, String tableName, RecordType referringRecordType,
+								 RecordType referencedRecordType) {
```
davidangb, line 160: "tableName" here again is pretty confusing (to me). Could this be "referringColumn" or something like that?

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -110,12 +114,58 @@ public boolean recordTypeExists(UUID instanceId, RecordType recordType) {
...
+		String fromCol = getFromColumnName(referringRecordType);
+		String toCol = getToColumnName(referencedRecordType);
+		String columnDefs =  quote(fromCol) + " text, " + quote(toCol) + " text";
```
davidangb, line 163: this creates a join table with dynamically-named columns, e.g. `from_sample_key` and `to_participant_key`. To simplify the implementation, could every join table have the same column names, e.g. `from` and `to`?

If the table name has the relevant info in it, like `sys_sample_participant`, it should still be understandable to a human even if the column names are generic.

davidangb (commented):
here's a first-pass review. There's a lot here, and this is complex! Nice job getting it this far.

I want to spend a little time trying this out in practice and looking at the database, I will definitely report back anything else I find.

service/src/main/java/org/databiosphere/workspacedataservice/service/model/RelationValue.java
calypsomatic: Yeah this was just Intellij being difficult for unknown reasons and I didn't get back to fixing it yet.  Thanks for the reminder!

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
calypsomatic: I.... don't know what happened here

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -281,4 +284,31 @@ public boolean isValidDate(String val) {
...
+	 * @return Set of Relation for all referencing attributes
+	 */
+	public RelationCollection findRelations(List<Record> records, Map<String, DataTypeMapping> schema) {
+		Set<Relation> relations = new HashSet<>();
+		Set<Relation> relationArrays = new HashSet<>();
+		for (Record rec : records) {
+			for (Map.Entry<String, Object> entry : rec.attributeSet()) {
+				if (schema.get(entry.getKey()) == RELATION){
+					relations.add(new Relation(entry.getKey(), RelationUtils.getTypeValue(entry.getValue())));
+					//TODO deal with tsv vs json source a bit smarter
```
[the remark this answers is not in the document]
calypsomatic: I would love to standarize inputs early on, but that would obviate a lot of the tsv vs json methods we have and is probably out of scope of this PR.   I could maybe pass a datasource along, like we do in a couple other places, although I'm not thrilled with that idea.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
```diff
@@ -0,0 +1,132 @@
...
+        for (Record rec : records) {
+            for (Map.Entry<String, Object> attribute : rec.attributeSet()){
+                if (relationArrays.containsKey(attribute.getKey()) && attribute.getValue() != null){
+                    //How to read relation list depends on its source, which we don't know here so we have to check
+                    List<String> rels;
+                    if (attribute.getValue() instanceof List<?>){
+                        rels = (List<String>) attribute.getValue();
+                    } else {
+                        rels = Arrays.asList(inferer.getArrayOfType(attribute.getValue().toString(), String[].class));
+                    }
```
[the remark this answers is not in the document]
calypsomatic: I was kinda ignoring that type of thing, banishing it to the realm of "update" instead of "create" and therefore a later ticket.  But upon a quick test, it looks like the column is change to the new type (which is most likely going to always be String) before it reaches this point.  So not the worst behavior, although the join table is now floating free.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
calypsomatic: I think I started out that way, realized the join table also needed to have the relation-array-column name in it in case a single record type has two different relation-array-columns relating to the same record type, and thought putting the names in the columns seemed easier than trying to parse out the names of two tables AND a column from a table name - we'd have to know 2 out of the 3 to be sure.  I do like the idea of a simpler implementation though, maybe I can manage, I'll look into it.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -95,6 +96,9 @@ public boolean instanceSchemaExists(UUID instanceId) {
 
 	public void createSchema(UUID instanceId) {
 		namedTemplate.getJdbcTemplate().update("create schema " + quote(instanceId.toString()));
+		//Add domains to distinguish relations and arrays of relations from strings and arrays of strings
+		namedTemplate.getJdbcTemplate().update("create domain " + quote(instanceId.toString()) + ".relation as text");
+		namedTemplate.getJdbcTemplate().update("create domain " + quote(instanceId.toString()) + ".array_of_relation as text[]");
```
sagehen03, line 101: I wonder if want to create these domains in the public schema upon app startup?  We could do a create if not exists so it wouldn't be a problem to have it re-run.  Then if we did that, we wouldn't need [this code](https://github.com/DataBiosphere/terra-workspace-data-service/blob/d0ebfbb14fc33519e13adc874bc2000a65bcfa8b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java#L260-L262)

sagehen03:
The code allows for the same record to referenced repeatedly in the same array of refs.  Is that what we want?  I think it makes sense when we use the array language, but the existing entity service calls this feature set tables which implies no dupes.  Do we need to dedupe or throw an error if the same record is specified more than once in a single array of relation?

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
```diff
@@ -0,0 +1,132 @@
...
+        //Identify relation arrays
+        Map<String, DataTypeMapping> relationArrays = requestSchema.entrySet().stream()
+                .filter(entry -> entry.getValue() == DataTypeMapping.ARRAY_OF_RELATION)
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+        Map<Relation, List<RelationValue>> relationArrayValues = getAllRelationArrayValues(records, relationArrays);
+        recordDao.batchUpsert(instanceId, recordType, records, requestSchema, primaryKey);
+        for (Map.Entry<Relation, List<RelationValue>> rel : relationArrayValues.entrySet()) {
+            recordDao.insertIntoJoin(instanceId, rel.getKey(), recordType, rel.getValue());
+        }
+    }
```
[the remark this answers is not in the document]
calypsomatic: A transaction would be good! Unfortunately, if I just throw `@Transactional` on there, apparently there's [all](https://sonarcloud.io/organizations/broad-databiosphere/rules?open=java%3AS2229&rule_key=java%3AS2229) [kinds](https://stackoverflow.com/questions/5109343/transactional-method-called-from-another-method-doesnt-obtain-a-transaction) of trouble because it's called from other methods.  It's not clear to me if I can fix it by adding `@Transactional` to the other calling methods as well.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
calypsomatic: I was wondering if it was really necessary to have the domains per-schema.  Seems like a good idea to move it to public.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -290,4 +293,31 @@ public boolean isValidDate(String val) {
...
+	 *
+	 * @param records
+	 *            - all records whose references to check
+	 * @return Set of Relation for all referencing attributes
+	 */
+	public RelationCollection findRelations(List<Record> records, Map<String, DataTypeMapping> schema) {
+		Set<Relation> relations = new HashSet<>();
+		Set<Relation> relationArrays = new HashSet<>();
+		for (Record rec : records) {
+			for (Map.Entry<String, Object> entry : rec.attributeSet()) {
```
sagehen03, line 308: It seems like this code is going to go over all attribute data even if we don't have any relations?  Can we either not call this method or return an empty RelationCollection if no relation or array of relation columns are present in the schema map?  In a similar vein, even if there are relations present it will go over all attributes when it should only need to iterate over the relation attributes.  This second request is admittedly a little more debatable since filtering out those non relation attributes is likely an iterative action itself.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -649,7 +716,7 @@ public boolean recordExists(UUID instanceId, RecordType recordType, String recor
 
 	public List<RecordType> getAllRecordTypes(UUID instanceId) {
 		return namedTemplate.queryForList(
-				"select tablename from pg_tables WHERE schemaname = :workspaceSchema order by tablename",
+				"select tablename from pg_tables WHERE schemaname = :workspaceSchema and hasindexes is true order by tablename",
```
sagehen03, line 719: Tell me more about this change to the where clause.  Is this just to exclude the join tables from the record type listing?  If so, why not do something more direct like exclude tables starting with sys?

calypsomatic: Yes it's just to exclude join tables - and I think I made this change before I had named the tables with sys_, so I think I should be able to update it to be less opaque now!

service/src/main/resources/application.properties
```diff
@@ -8,6 +8,9 @@ spring.datasource.jdbc-url=jdbc:postgresql://${WDS_DB_HOST:localhost}:${WDS_DB_P
 spring.datasource.username=${WDS_DB_USER:wds}
 spring.datasource.password=${WDS_DB_PASSWORD:wds}
 
+#To ensure data.sql is run on startup
+spring.sql.init.mode=always
+
```
davidangb, line 13: TOL, not for this PR: we should evaluate this approach vs. using Liquibase - this is just fine for now!

service/src/test/java/org/databiosphere/workspacedataservice/dao/RecordDaoTest.java
```diff
@@ -302,4 +322,94 @@ void testDeleteRecordTypeWithRelation() {
...
+		recordDao.batchUpsert(instanceId, relationArrayType, Collections.singletonList(recordWithRelationArray), schema);
+
+		Map<String, DataTypeMapping> createdSchema = recordDao.getExistingTableSchemaLessPrimaryKey(instanceId, relationArrayType);
+		assertEquals(3, createdSchema.size());
+		List<Relation> relationArrayCols = recordDao.getRelationArrayCols(instanceId, relationArrayType);
+		assertEquals(List.of(arrayRelation), relationArrayCols);
+		Record record = recordDao.getSingleRecord(instanceId, relationArrayType, relArrId).get();
+		assertNotNull(record);
+		//Inspecting the values shows they match but the test still fails at this line
+//		assertEquals(relArr, Arrays.asList(record.getAttributeValue("relArrAttr")));
```
davidangb, line 390: I think you need to cast the attribute value to `String[]` before calling `Arrays.asList()` on it:
```java
		String[] actualAttrValue = assertInstanceOf(String[].class, record.getAttributeValue("relArrAttr"));
		assertIterableEquals(relArr, Arrays.asList(actualAttrValue));
```

davidangb (commented):
more comments inline!

service/src/test/java/org/databiosphere/workspacedataservice/dao/RecordDaoTest.java
calypsomatic: Ah I missed this comment!  It saved my life, thank you!

davidangb (approved):

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -47,12 +39,7 @@
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStreamWriter;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.UUID;
+import java.util.*;
```
jordansafer, line 42: I'm pretty strongly against star imports (just encountered a bug the other day due to a mis-used type from the wrong namespace). Can we open a ticket to add a linting rule and remove current star imports? I think I commented this before on one of Drew's PRs and didn't feel strongly about it, but after dealing with a bug from it, I think there's a good argument not to use them.

jordansafer (approved):

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
calypsomatic: I have no opinion about star imports - pretty sure this is all automatically done by Intellij.  Lint all you want, modulo all the heartache we've had from linting before.

PR #149, opened 2022-11-23
AJ-730: Fix for strange primary key names

sagehen03:


calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -610,7 +610,7 @@ private LocalDate[] convertToLocalDate(Object object) {
 	public Optional<Record> getSingleRecord(UUID instanceId, RecordType recordType, String recordId) {
 		try {
 			return Optional.ofNullable(namedTemplate.queryForObject(
-					"select * from " + getQualifiedTableName(recordType, instanceId) + " where " + cachedQueryDao.getPrimaryKeyColumn(recordType, instanceId)
+					"select * from " + getQualifiedTableName(recordType, instanceId) + " where " + quote(cachedQueryDao.getPrimaryKeyColumn(recordType, instanceId))
```
aaronkanzer, line 613: Could we add a unit test for here?

Seems that nothing broke when removing the `quote()` call.

aaronkanzer:
Just a question for clarification -- why cast some of these values with `quote()` that reference a key, and not all?

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
sagehen03: Removing the quote call will only cause things to break on reads if the record type was created with a user specified primary key which can currently only be specified through tsv upload.  I've added code to TsvDownloadTest to retrieve by id after uploading and you can see those tests fail now if you remove the quote call you're referencing.  Overall, the postgres behavior around this is frustrating as postgres won't fail when it encounters a "weird" primary key name, instead it will just rename it to something less weird--ButtHead->butthead for example.

aaronkanzer:
> > Just a question for clarification -- why cast some of these values with `quote()` that reference a key, and not all?
> 
> Can you give me an example of what you're referring to?

Perhaps something like `params.addValue("recordTypeRowIdentifier", cachedQueryDao.getPrimaryKeyColumn(recordType, instanceId));`

at Line 158ish in `RecordDao.java`

sagehen03:
> > > Just a question for clarification -- why cast some of these values with `quote()` that reference a key, and not all?
> > 
> > 
> > Can you give me an example of what you're referring to?
> 
> Perhaps something like `params.addValue("recordTypeRowIdentifier", cachedQueryDao.getPrimaryKeyColumn(recordType, instanceId));`
> 
> at Line 158ish in `RecordDao.java`

TBH it was trial and error and just tweaking the code until the tests passed, but with the benefit of experience I think I can say that in cases where we're using the value to specify a column name in a SQL statement, we need to quote in the event that the user chose a column name that includes upper case letters, spaces, dashes among other reserved characters.  In cases where we we're just pulling a string out of table data like the example you cited, we don't need to quote.  We may one day regret not making users comply with postgres' restrictions on column and table names, but for now it doesn't see too unmanageable, especially now that we're adding more tests :crossed_fingers:

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -149,7 +149,7 @@ public ResponseEntity<StreamingResponseBody> streamAllEntities(@PathVariable("in
 	}
 
 	@PostMapping("/{instanceid}/search/{version}/{recordType}")
-	public RecordQueryResponse queryForEntities(@PathVariable("instanceid") UUID instanceId,
+	public RecordQueryResponse queryForRecords(@PathVariable("instanceid") UUID instanceId,
```
sagehen03, line 152: This method was missed when we refactored from entity to records.

aaronkanzer:
Just flagging two other occurrences of `cachedQueryDao.getPrimaryKeyColumn` that are still not casted with `quote()`

https://github.com/DataBiosphere/terra-workspace-data-service/blob/73286a48248274e6a49d762d34bc5da22873a9ec/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java#L526

https://github.com/DataBiosphere/terra-workspace-data-service/blob/dh-aj-730-1/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java#L234

(Perhaps a mini-thought, but is there any downside of always casting this value just to regression-proof every usage?)

davidangb (approved):
thank you!

sagehen03:
> Just flagging two other occurrences of `cachedQueryDao.getPrimaryKeyColumn` that are still not casted with `quote()`
> 
> https://github.com/DataBiosphere/terra-workspace-data-service/blob/dh-aj-730-1/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java#L526
> 
> and
> 
> https://github.com/DataBiosphere/terra-workspace-data-service/blob/dh-aj-730-1/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java#L234
> 
> (Perhaps a mini-thought, but is there any downside of always casting this value just to regression-proof every usage?)

In the first case, we use the value to access [java.sql.Resultset data by column name ](https://github.com/DataBiosphere/terra-workspace-data-service/blob/10c5ee9e4100c51e6a9c79bea74b0e9bcab00954/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java#L533) if you quote the value in that case you'll get a SQLException.  The second case you cite eventually get quoted.  We can't quote in all cases because there are instances like the one I mentioned a second ago that require the non-quoted version.  Basically, I believe it works like this.. Don't quote when it's used with a jdbc resultset, just being included in the response body, or a where clause value.  Do quote when you need to reference that column name in a sql statement.

PR #150, opened 2022-11-28
AJ-729: Let callers specify primary key for JSON methods that create record types

sagehen03:
This PR adds an optional query param to let the caller specify the primary key column for the `streamingWrite` and `upsertSingleRecord` endpoints.  It also replaces the variable name `uniqueRowIdentifier` with `primaryKey` in RecordController and RecordDao.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -290,7 +290,7 @@ public class RecordController {
 	}
 
 	private RecordTypeSchema getSchemaDescription(UUID instanceId, RecordType recordType) {
-		Map<String, DataTypeMapping> schema = recordDao.getExistingTableSchemaLessPrimaryKey(instanceId, recordType);
+		Map<String, DataTypeMapping> schema = recordDao.getExistingTableSchema(instanceId, recordType);
 		Map<String, RecordType> relations = recordDao.getRelationCols(instanceId, recordType).stream()
 				.collect(Collectors.toMap(Relation::relationColName, Relation::relationRecordType));
 		List<AttributeSchema> attrSchema = schema.entrySet().stream().sorted(Map.Entry.comparingByKey())
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -111,10 +111,10 @@ public class RecordController {
 		return new ResponseEntity<>(response, HttpStatus.OK);
 	}
 
-	@PostMapping({"/{instanceId}/tsv/{version}/{recordType}/{uniqueRowIdentifierColumn}", "/{instanceId}/tsv/{version}/{recordType}"} )
+	@PostMapping( "/{instanceId}/tsv/{version}/{recordType}")
 	public ResponseEntity<TsvUploadResponse> tsvUpload(@PathVariable("instanceId") UUID instanceId,
 			   @PathVariable("version") String version, @PathVariable("recordType") RecordType recordType,
-			   @PathVariable(name= "uniqueRowIdentifierColumn", required = false) Optional<String> uniqueRowIdentifierColumn,
+			   @RequestParam(name= "uniqueRowIdentifierColumn", required = false) Optional<String> uniqueRowIdentifierColumn,
                @RequestParam("records") MultipartFile records) throws IOException {
 		validateVersion(version);
 		validateInstance(instanceId);
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -132,7 +132,7 @@ public class RecordController {
 		validateVersion(version);
 		validateInstance(instanceId);
 		checkRecordTypeExists(instanceId, recordType);
-		List<String> headers = new ArrayList<>(Collections.singletonList(ReservedNames.RECORD_ID));
+		List<String> headers = new ArrayList<>(Collections.singletonList(recordDao.getPrimaryKeyColumn(recordType, instanceId)));
 		headers.addAll(recordDao.getExistingTableSchema(instanceId, recordType).keySet());
 		Stream<Record> allRecords = recordDao.streamAllRecordsForType(instanceId, recordType);
 
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -110,15 +110,15 @@ public class RecordController {
 		return new ResponseEntity<>(response, HttpStatus.OK);
 	}
 
-	@PostMapping("/{instanceId}/tsv/{version}/{recordType}")
+	@PostMapping("/{instanceId}/tsv/{version}/{recordType}/{recordId}")
 	public ResponseEntity<TsvUploadResponse> tsvUpload(@PathVariable("instanceId") UUID instanceId,
 			@PathVariable("version") String version, @PathVariable("recordType") RecordType recordType,
-			@RequestParam("records") MultipartFile records) throws IOException {
+			@PathVariable("recordId") String recordId, @RequestParam("records") MultipartFile records) throws IOException {
 		validateInstance(instanceId);
 		validateVersion(version);
 		int recordsModified;
 		try (InputStreamReader inputStreamReader = new InputStreamReader(records.getInputStream())) {
-			recordsModified = batchWriteService.uploadTsvStream(inputStreamReader, instanceId, recordType);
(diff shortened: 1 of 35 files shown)


davidangb (commented):
code/syntax in here looks good, thank you!

Functionally, I have two concerns, which may need product input:

If I create a record in a nonexistent type and specify a primaryKey of "foo", it will create a table and set the table's pk to "foo", as expected. If I then create a second record of the same type but specify a primaryKey of "bar" … it creates the second record, and still uses the pk of "foo". Should we throw an error instead, if the user specifies primaryKey in the query param, but the value they specified is incorrect?

If I specify a primaryKey for a non-existent type, and I include an attribute of that name in the request payload, I get a SQL error. Similarly, if the type already exists and I include the primary key in the request payload - whether or not I also specify it in the query param - I get a SQL error. In my opinion, it's quite likely that users will do this. My preference would be to handle this case - validate that the pk value in the payload is the same as the pk value in the URL (throw error if not), and allow the pk to be present in attributes. (we might want, for a future revision of the APIs, to not even require the record id in the URL). If product feels this isn't important to handle, I'll defer to that - though I'd want to improve the error message, at least.

sagehen03:
> code/syntax in here looks good, thank you!
> 
> Functionally, I have two concerns, which may need product input:
> 
> If I create a record in a nonexistent type and specify a primaryKey of "foo", it will create a table and set the table's pk to "foo", as expected. If I then create a second record of the same type but specify a primaryKey of "bar" … it creates the second record, and still uses the pk of "foo". Should we throw an error instead, if the user specifies primaryKey in the query param, but the value they specified is incorrect?
> 
> If I specify a primaryKey for a non-existent type, and I include an attribute of that name in the request payload, I get a SQL error. Similarly, if the type already exists and I include the primary key in the request payload - whether or not I also specify it in the query param - I get a SQL error. In my opinion, it's quite likely that users will do this. My preference would be to handle this case - validate that the pk value in the payload is the same as the pk value in the URL (throw error if not), and allow the pk to be present in attributes. (we might want, for a future revision of the APIs, to not even require the record id in the URL). If product feels this isn't important to handle, I'll defer to that - though I'd want to improve the error message, at least.

Good catches!  I think we should consider the primaryKey static once the record type has been created and throw an error if the user specifies a different primary key for subsequent writes to an existing record type. 

The second issue feels more ambiguous; and I think I'd prefer to adjust the code so that it doesn't throw an error and as usual just takes the recordId as the pk value.  I feel like the swagger is pretty clear on the recordId meaning.  Let me know if you feel strongly that we need to check that pk attribute in the attribute matches the recordId.

service/src/test/java/org/databiosphere/workspacedataservice/controller/TsvDownloadTest.java
```diff
@@ -80,7 +80,7 @@ void batchWriteFollowedByTsvDownload() throws IOException {
 		RecordType recordType = RecordType.valueOf("tsv-test");
 
 		InputStream is = TsvDownloadTest.class.getResourceAsStream("/batch_write_tsv_data.json");
-		ResponseEntity<BatchResponse> response = recordController.streamingWrite(instanceId, version, recordType, is);
+		ResponseEntity<BatchResponse> response = recordController.streamingWrite(instanceId, version, recordType, Optional.empty(), is);
```
aaronkanzer, line 83: Would it be useful to include a unit test where the `primaryKey` is provided?

sagehen03: I've added some more tests around the cases that David identified and those tests specify a primary key.   `GeneratedClientTests#putRecordWithSpecifiedPk` also previously exercised that code.

sagehen03:
> > code/syntax in here looks good, thank you!
> > Functionally, I have two concerns, which may need product input:
> > If I create a record in a nonexistent type and specify a primaryKey of "foo", it will create a table and set the table's pk to "foo", as expected. If I then create a second record of the same type but specify a primaryKey of "bar" … it creates the second record, and still uses the pk of "foo". Should we throw an error instead, if the user specifies primaryKey in the query param, but the value they specified is incorrect?
> > If I specify a primaryKey for a non-existent type, and I include an attribute of that name in the request payload, I get a SQL error. Similarly, if the type already exists and I include the primary key in the request payload - whether or not I also specify it in the query param - I get a SQL error. In my opinion, it's quite likely that users will do this. My preference would be to handle this case - validate that the pk value in the payload is the same as the pk value in the URL (throw error if not), and allow the pk to be present in attributes. (we might want, for a future revision of the APIs, to not even require the record id in the URL). If product feels this isn't important to handle, I'll defer to that - though I'd want to improve the error message, at least.
> 
> Good catches! I think we should consider the primaryKey static once the record type has been created and throw an error if the user specifies a different primary key for subsequent writes to an existing record type.
> 
> The second issue feels more ambiguous; and I think I'd prefer to adjust the code so that it doesn't throw an error and as usual just takes the recordId as the pk value. I feel like the swagger is pretty clear on the recordId meaning. Let me know if you feel strongly that we need to check that pk attribute in the attribute map matches the recordId value.

@davidangb I've made the changes to reject a write to an existing record type where the primary key specified is not the same as the existing primary key.  I've also made it so that we no longer throw a SQL error in the event that the primary key data is specified in the attributes for JSON writes and we just ignore the pk data specified in the attributes.  Let me know if you strongly object to this approach, but I feel like we make it fairly clear as to what the recordId values mean in RecordRequest payloads.

service/src/main/java/org/databiosphere/workspacedataservice/service/model/exception/NewPrimaryKeyException.java
```diff
@@ -0,0 +1,15 @@
...
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(code = HttpStatus.BAD_REQUEST)
+public class NewPrimaryKeyException extends RuntimeException {
+
+
+	public NewPrimaryKeyException(String oldPk, String newPk, RecordType recordType) {
+		super("The primary key for " + recordType + " is already set to " + oldPk
+				+ " if you wish to change it to " + newPk + " you'll need to create a new record type.");
```
davidangb, line 13: nit: leave `newPk` out of the error message, have it just be:
```
super("The primary key for " + recordType + " is already set to " + oldPk
				+ ". If you wish to change it, you'll need to create a new record type.");
```
this prevents any kind of XSS injection if/when this message is echoed to the browser

davidangb (approved):
one nit on the error message inline, else :+1: and thanks for the updates! Proactively approving, but expecting you'll change the error message.

I do think we should throw an error if the id in the url and the id in attributes (if specified) do not match - but I'm totally fine filing that as a follow-on change and prioritizing it with PM and the team for some later time.

service/src/main/java/org/databiosphere/workspacedataservice/service/model/exception/NewPrimaryKeyException.java
sagehen03: I've made the change you suggested and your suggestion reminded me that we weren't "sql validating" the user specified primary key so I've corrected that too.  I was trying to give the user (in my head a programmatic API user) as useful of an error message as possible.  I reluctantly agree that it's probably too much to ask web client authors to consider the contents of the error message before putting it on the screen, but if it's a single tenant system, is it really an issue if someone wants to XSS themself?  I know I'm not going to convince you with these arguments.  I'm just airing my dissatisfaction with dogmatic security practices that don't consider the individual particulars.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -114,13 +115,16 @@ public ResponseEntity<RecordResponse> getSingleRecord(@PathVariable("instanceId"
 	@PostMapping( "/{instanceId}/tsv/{version}/{recordType}")
 	public ResponseEntity<TsvUploadResponse> tsvUpload(@PathVariable("instanceId") UUID instanceId,
 			   @PathVariable("version") String version, @PathVariable("recordType") RecordType recordType,
-			   @RequestParam(name= "uniqueRowIdentifierColumn", required = false) Optional<String> uniqueRowIdentifierColumn,
+			   @RequestParam(name= "primaryKey", required = false) Optional<String> primaryKey,
                @RequestParam("records") MultipartFile records) throws IOException {
 		validateVersion(version);
 		validateInstance(instanceId);
+		if(recordDao.recordTypeExists(instanceId, recordType)){
+			validatePrimaryKey(instanceId, recordType, primaryKey);
```
aaronkanzer, line 123: Should we check that the `primaryKey` parameter is present here first, since it is an optional parameter in the function definition?

aaronkanzer (approved):

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
aaronkanzer: Discussed offline, default behavior of the `Optional` class works OK here.

PR #151, opened 2022-11-28
AJ-728: Identify pk column in schema apis

sagehen03:
I ran GeneratedClientTests before updating swagger and they passed so I think this change won't break swagger clients.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -618,6 +618,10 @@ public class RecordDao {
 		}
 	}
 
+	public String getPrimaryKeyColumn(RecordType type, UUID instanceId){
+		return cachedQueryDao.getPrimaryKeyColumn(type, instanceId);
+	}
+
 	public boolean recordExists(UUID instanceId, RecordType recordType, String recordId) {
 		return Boolean.TRUE
 				.equals(namedTemplate.queryForObject(
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/model/RecordTypeSchema.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/model/RecordTypeSchema.java
@@ -4,5 +4,5 @@ import org.databiosphere.workspacedataservice.shared.model.RecordType;
 
 import java.util.List;
 
-public record RecordTypeSchema(RecordType name, List<AttributeSchema> attributes, int count, String uniqueRowIdentifier) {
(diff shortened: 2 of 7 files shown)


service/src/test/resources/logback-spring.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
sagehen03, line 1: I'm still battling with log spam in my project directory.  This will turn off file logging completely when running tests. (Unrelated I know).

service/src/main/java/org/databiosphere/workspacedataservice/service/model/RecordTypeSchema.java
```diff
@@ -4,5 +4,5 @@
 
 import java.util.List;
 
-public record RecordTypeSchema(RecordType name, List<AttributeSchema> attributes, int count) {
+public record RecordTypeSchema(RecordType name, List<AttributeSchema> attributes, int count, String uniqueRowIdentifier) {
```
sagehen03, line 7: What do we want to call this value?

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -618,6 +618,10 @@ public Optional<Record> getSingleRecord(UUID instanceId, RecordType recordType,
 		}
 	}
 
+	public String getPrimaryKeyColumn(RecordType type, UUID instanceId){
+		return cachedQueryDao.getPrimaryKeyColumn(type, instanceId);
```
aaronkanzer, line 622: Would this also be applicable to some of the changes in https://github.com/DataBiosphere/terra-workspace-data-service/pull/149 ?

sagehen03: This method exists to make the primary key column available to callers outside of RecordDao.  It should not be quoted.  Let me know if there's something else you're getting at that I'm not picking up on.

service/src/main/java/org/databiosphere/workspacedataservice/service/model/RecordTypeSchema.java
davidangb: FWIW, Entity Service calls it `idName` and TDR calls it `primaryKey` (though TDR is explicitly describing the underlying BigQuery schema). My vote would be for something 1) short-ish, and 2) consistent with the query parameter name, though I realize that might mean changing the query parameter.

How about just `primaryKey`?

davidangb (approved):
:+1:, realizing there may be changes to naming of `uniqueRowIdentifier`

PR #152, opened 2022-11-29
AJ-683: Normalize all boolean values to lowercase for Array of Booleans in DB

aaronkanzer:
Relates to https://broadworkbench.atlassian.net/browse/AJ-683

This ticket introduces logic that allows a user to enter different cases, expressions of boolean values in String format, and have WDS ingest those values as an Array of Booleans in the DB

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
@@ -252,6 +252,9 @@ public class DataTypeInferer {
 			return ARRAY_OF_DATE;
 		}
 		if(ArrayUtils.isNotEmpty(getArrayOfType(val, String[].class))){
+			if(ArrayUtils.isNotEmpty(getArrayOfType(val.toLowerCase(), Boolean[].class))){
+				return ARRAY_OF_BOOLEAN;
+			}
 			return ARRAY_OF_STRING;
 		}
 		if(getArrayOfType(val, String[].class) != null){
@@ -264,7 +267,8 @@ public class DataTypeInferer {
 		try {
 			String escapedValue = replaceLeftRightQuotes(val);
 			if(clazz.getComponentType() == Boolean.class){
-				escapedValue = escapedValue.toLowerCase();
+				// Ensure that potential additional quotes do not surround the boolean values
+				escapedValue = escapedValue.toLowerCase().replaceAll("\"","");
 			}
 			return objectMapper.readValue(escapedValue, clazz);
 		} catch (JsonProcessingException e) {
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
@@ -252,9 +252,6 @@ public class DataTypeInferer {
 			return ARRAY_OF_DATE;
 		}
 		if(ArrayUtils.isNotEmpty(getArrayOfType(val, String[].class))){
-			if(ArrayUtils.isNotEmpty(getArrayOfType(val.toLowerCase(), Boolean[].class))){
-				return ARRAY_OF_BOOLEAN;
-			}
 			return ARRAY_OF_STRING;
 		}
 		if(getArrayOfType(val, String[].class) != null){
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
@@ -48,11 +48,6 @@ public class DataTypeInferer {
 			Map<String, DataTypeMapping> inferred = inferTypes(rcd.getAttributes(), dataSource);
 			for (Map.Entry<String, DataTypeMapping> entry : inferred.entrySet()) {
 				DataTypeMapping inferredType = entry.getValue();
-				if (inferredType == ARRAY_OF_BOOLEAN) {
-					String currentRecordInput = rcd.getAttributeValue(entry.getKey()).toString();
-					String lowerCaseBooleanStr = currentRecordInput.toLowerCase();
-					rcd.getAttributes().putAttribute(entry.getKey(), lowerCaseBooleanStr);
-				}
 				if (result.containsKey(entry.getKey()) && result.get(entry.getKey()) != inferredType) {
 					result.put(entry.getKey(), selectBestType(result.get(entry.getKey()), inferredType));
 				} else {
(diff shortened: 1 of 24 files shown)


davidangb (commented):
The problem with arrays-of-booleans as I see it is that we have a very different implementation for scalars, where we explicitly use `equalsIgnoreCase`:
https://github.com/DataBiosphere/terra-workspace-data-service/blob/c06fd846e6ccc964eb0b0de78feb1f8ba5e48b6c/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java#L179-L181
versus arrays, where we rely on Jackson parsing:
https://github.com/DataBiosphere/terra-workspace-data-service/blob/c06fd846e6ccc964eb0b0de78feb1f8ba5e48b6c/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java#L259-L265

my first instinct was to look for a [Jackson configuration](https://github.com/DataBiosphere/terra-workspace-data-service/blob/c06fd846e6ccc964eb0b0de78feb1f8ba5e48b6c/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java#L39-L42) to allow case-insensitive parsing of Booleans. I don't think such a thing exists, unless we write our own custom deserializer, which … is possible but would be nice to avoid.

Since a config doesn't exist, could you instead isolate the changes in `DataTypeInferer` to just the `getArrayOfType` method? Something like:
```java
	public <T> T[] getArrayOfType(String val, Class<T[]> clazz) {
		try {
			String normalizedVal = replaceLeftRightQuotes(val);
			// enable case-insensitive parsing of Booleans
			if (clazz.getComponentType() == Boolean.class) {
				normalizedVal = normalizedVal.toLowerCase();
			}
			return objectMapper.readValue(normalizedVal, clazz);
		} catch (JsonProcessingException e) {
			return null;
		}
	}
```

See also my comments inline on the test cases. I think this PR is identifying other cases, which are probably out of scope, that may also need fixing in a future PR.

service/src/test/java/org/databiosphere/workspacedataservice/DataTypeInfererTest.java
```diff
@@ -115,6 +115,7 @@ void inferSomeTypes() {
 		assertThat(inferer.inferType("12345A", InBoundDataSource.JSON)).isEqualTo(DataTypeMapping.STRING);
 		assertThat(inferer.inferType(List.of("Hello!"), InBoundDataSource.JSON)).isEqualTo(DataTypeMapping.ARRAY_OF_STRING);
 		assertThat(inferer.inferType(List.of(new BigInteger("12345")), InBoundDataSource.JSON)).isEqualTo(DataTypeMapping.ARRAY_OF_NUMBER);
+		assertThat(inferer.inferType(List.of("true", "False", "TRUE", "fALSE"), InBoundDataSource.JSON)).isEqualTo(DataTypeMapping.ARRAY_OF_BOOLEAN);
```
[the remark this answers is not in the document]
aaronkanzer: I'm slightly confused here -- are we inferring that this test passing would be a bug?

service/src/test/java/org/databiosphere/workspacedataservice/controller/TsvInputFormatsTest.java
```diff
@@ -0,0 +1,159 @@
...
+				Arguments.of(
+						"[\"true\", \"TRUE\", \"tRuE\"]",
+						new Boolean[]{Boolean.TRUE, Boolean.TRUE, Boolean.TRUE},
+						"[\"TRUE\", \"true\", \"True\"]"
+				),
+				Arguments.of(
+						"[\"true\", \"false\", \"true\"]",
+						new Boolean[]{Boolean.TRUE, Boolean.FALSE, Boolean.TRUE},
+						"[\"TRUE\", \"fALSE\", \"True\"]"
+				),
```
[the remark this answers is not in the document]
aaronkanzer: Paired with @sagehen03 for a bit here

We refactored to use your suggestions here, so many thanks :D 

Will follow up with a PR to resolve the buggy behavior here:

service/src/test/java/org/databiosphere/workspacedataservice/controller/TsvInputFormatsTest.java
```diff
@@ -0,0 +1,169 @@
...
+				Arguments.of(" ", " ", " "),
+				Arguments.of("true", Boolean.TRUE, "TRUE"),
+				Arguments.of("TRUE", Boolean.TRUE, "true"),
+				Arguments.of("tRuE", Boolean.TRUE, "true"),
+				Arguments.of("True", Boolean.TRUE, "true"),
+				Arguments.of("false", Boolean.FALSE, "FALSE"),
+				Arguments.of("FALSE", Boolean.FALSE, "false"),
+				Arguments.of("fAlSe", Boolean.FALSE, "false"),
+				Arguments.of("False", Boolean.FALSE, "false"),
+//				Arguments.of("\"false\"", String.valueOf("\"False\""), "\"fAlSe\"") // TODO: Fix in next PR
```
aaronkanzer, line 96: This test fails for now -- will include logic that fixes this in the next PR

davidangb (approved):
thank you!

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -200,7 +202,7 @@ public boolean isNumericValue(String sVal) {
 
 	private JsonNode parseToJsonNode(String val){
 		try {
-			return objectMapper.readTree(val);
+			return objectMapper.readTree(val.toLowerCase());
```
sagehen03, line 205: Can we add a comment here as to why we're lower casing?

sagehen03 (approved):
One request for a comment, but otherwise :+1:

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
aaronkanzer: Added a comment!

davidangb (approved):
still :+1:

service/src/test/java/org/databiosphere/workspacedataservice/controller/TsvInputFormatsTest.java
```diff
@@ -0,0 +1,169 @@
...
+				Arguments.of(
+						"[true, false, true]",
+						new Boolean[]{Boolean.TRUE, Boolean.FALSE, Boolean.TRUE},
+						"[TRUE, fALSE, True]"
+				),
+				Arguments.of(
+						"[\"True\", \"false\", \"true\"]",
+						new String[]{"True", "false", "true"},
+						"[\"TRUE\", \"fALSE\", \"True\"]"
+				),
```
sagehen03, line 111:
```suggestion
				Arguments.of(
						"[\"True\", \"false\", \"true\"]",
						new Boolean[]{Boolean.TRUE, Boolean.FALSE, Boolean.TRUE},
						"[\"TRUE\", \"fALSE\", \"True\"]"
				),
```

PR #153, opened 2022-12-02
AJ-754: Don't fail with a SQL Error when passing list data to a STRING attribute

davidangb:
[AJ-754](https://broadworkbench.atlassian.net/browse/AJ-754)

## Note

This PR eliminates the error described below. If you upload an array of strings into a column that is a scalar string, your array of strings will be serialized into a scalar. For instance, if you upload `["array","of","strings"]` into a scalar column, you will end up with a scalar value `"{array,of,strings}"`.

I am not convinced this is the final behavior we want. However, I consider this PR as a strict improvement over throwing an error, so I suggest we merge it as-is while we debate what the functionality should be.

### Steps to reproduce

1. Create a new WDS instance
2. use the PUT single-record API to create a record with a single attribute whose value is a scalar string
3. use the PUT single-record API to replace that record, this time specifying the same single attribute as an array of strings

### Expected

All requests are 2xx successful.

### Actual

Step 3 - the attempt to upload the array - fails with:
> ```PreparedStatementCallback; bad SQL grammar [insert into \"f482df36-bada-4e19-a534-05f39c8b0f18\".\"thing\"(\"sys_name\", \"foo\") values (?, ?) on conflict (\"sys_name\") do update set \"foo\" = excluded.\"foo\"]; nested exception is org.postgresql.util.PSQLException: Can't infer the SQL type to use for an instance of java.util.ArrayList. Use setObject() with an explicit Types value to specify the type to use.```

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -457,7 +457,7 @@ public class RecordDao {
 		if (Objects.isNull(attVal)) {
 			return null;
 		}
-		if (RelationUtils.isRelationValue(attVal) && typeMapping == DataTypeMapping.STRING) {
+		if (RelationUtils.isRelationValue(attVal) && typeMapping == DataTypeMapping.RELATION) {
 			return RelationUtils.getRelationValue(attVal);
 		}
 		if (attVal instanceof String sVal) {
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -400,7 +400,7 @@ public class RecordDao {
 		if (Objects.isNull(attVal)) {
 			return null;
 		}
-		if (RelationUtils.isRelationValue(attVal)) {
+		if (RelationUtils.isRelationValue(attVal) && typeMapping == DataTypeMapping.STRING) {
 			return RelationUtils.getRelationValue(attVal);
 		}
 		if (attVal instanceof String sVal) {
@@ -427,6 +427,9 @@ public class RecordDao {
 		}
 		if(typeMapping.isArrayType()){
 			return getArrayValues(attVal, typeMapping);
+		}
+		if(attVal instanceof List<?> list && typeMapping == DataTypeMapping.STRING){
+			return list.stream().map(Object::toString).collect(Collectors.joining(","));
 		}
 		return attVal;
 	}
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -439,7 +439,7 @@ public class RecordDao {
 			return getArrayValues(attVal, typeMapping);
 		}
 		if(attVal instanceof List<?> list && typeMapping == DataTypeMapping.STRING){
-			return list.stream().map(Object::toString).collect(Collectors.joining(","));
+			return "{"+ list.stream().map(Object::toString).collect(Collectors.joining(",")) +"}";
 		}
 		return attVal;
 	}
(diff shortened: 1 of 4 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -410,7 +410,7 @@ private Object getValueForSql(Object attVal, DataTypeMapping typeMapping) {
 		if (Objects.isNull(attVal)) {
 			return null;
 		}
-		if (RelationUtils.isRelationValue(attVal)) {
+		if (RelationUtils.isRelationValue(attVal) && typeMapping == DataTypeMapping.STRING) {
```
sagehen03, line 413: I thought we need to be more robust in how we handle some of the casting.  The second part of this check should be replaced with `typeMapping == DataTypeMapping.RELATION` once the array of relations PR merges.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -438,6 +438,9 @@ private Object getValueForSql(Object attVal, DataTypeMapping typeMapping) {
 		if(typeMapping.isArrayType()){
 			return getArrayValues(attVal, typeMapping);
 		}
+		if(attVal instanceof List<?> list && typeMapping == DataTypeMapping.STRING){
```
sagehen03, line 441: This change will cause the list data to be stringified and successfully stored.  Alternatively we could throw an error and inform the user that they're trying to add array data to plain string column.

service/src/test/java/org/databiosphere/workspacedataservice/dao/RecordDaoTest.java
```diff
@@ -85,7 +85,7 @@ void batchDeleteAndTestRelationsFunkyPrimaryKeyValues(){
 		String sample_id = "Sample ID";
 		String recordId = "1199";
 		Record testRecord = new Record(recordId, referencedRt, RecordAttributes.empty());
-		Map<String, DataTypeMapping> schema = Map.of("attr1", DataTypeMapping.STRING);
+		Map<String, DataTypeMapping> schema = Map.of("attr1", DataTypeMapping.RELATION);
```
davidangb, line 88: the changes from `STRING` to `RELATION` in this test are a result of https://github.com/DataBiosphere/terra-workspace-data-service/pull/153#discussion_r1041495970.

aaronkanzer:
You mention in the PR's description:

```
I am not convinced this is the final behavior we want. However, I consider this PR as a strict improvement over throwing an error, so I suggest we merge it as-is while we debate what the functionality should be.
```

What alternative behavior would we want instead?

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -430,6 +430,29 @@ void nullAndNonNullArraysShouldChooseProperType() throws Exception {
 		assertEquals("ARRAY_OF_NUMBER", actual.attributes().get(0).datatype());
 	}
 
+	@Test
+	@Transactional
+	void scalarFollowedByArray() throws Exception {
```
aaronkanzer, line 435: I'm wondering if it would be regression-proof just to assert on the structure of the value added via these `PUT` calls vs the next `GET` of the value?

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -485,6 +485,9 @@ private Object getValueForSql(Object attVal, DataTypeMapping typeMapping) {
...
+			return "{"+ list.stream().map(Object::toString).collect(Collectors.joining(",")) +"}";
```
aaronkanzer, line 489: I'm not familiar with this syntax -- what does the `+` on both sides of the string here mean? Is it just a syntax for unpacking the list?

davidangb: this is just string concatenation, prepending "{" and appending "}" to the comma-delimited list

davidangb:
@aaronkanzer
re: what behavior we want in the end, I'm not sure. I filed AJ-758 to track this, but I think it'll be a wider discussion and I don't want to block this PR on that discussion.

re: your unit test comment, I added more assertions to that test, is that what you meant?

aaronkanzer (approved):

jordansafer (approved):

PR #155, opened 2022-12-08
better error message for TSVs that contain duplicate IDs [AJ-721]

davidangb:
Uploading a TSV that contains duplicate ids now yields an error message of "TSVs cannot contain duplicate primary key values."

Per refinement, scope of this ticket is to make the error message better. If we want to allow duplicate PKs in a TSV, we'll do that in a future PR/ticket.

# New Error Message:
Screenshot 12-8-2022 at 05 10 PM

# Old Error Message:
Screenshot 12-8-2022 at 05 14 PM

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
@@ -123,7 +123,7 @@ public class BatchWriteService {
 		CSVParser rows = csvFormat.parse(is);
 		String leftMostColumn = rows.getHeaderNames().get(0);
 		List<Record> batch = new ArrayList<>();
-		HashSet<String> recordIds = new HashSet<>();
+		HashSet<String> recordIds = new HashSet<>(); // this set may be slow for very large TSVs
 		boolean firstUpsertBatch = true;
 		Map<String, DataTypeMapping> schema = null;
 		String uniqueIdentifierAsString = primaryKey.orElse(leftMostColumn);
@@ -156,11 +156,6 @@ public class BatchWriteService {
 				}
 				recordDao.batchUpsert(instanceId, recordType, batch, schema);
 				batch.clear();
-				// If we want to enforce that record ids are unique within the entire TSV,
-				// instead of within a batch, simply remove the next line.
-				// This would have performance implications for very large TSVs: the `recordIds` HashSet
-				// would need to hold ALL recordIds, reducing its efficiency.
-				recordIds.clear();
 			}
 		}
 		if (firstUpsertBatch) {
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
@@ -131,19 +131,22 @@ public class BatchWriteService {
 		for (CSVRecord row : rows) {
 			Map<String, Object> m = (Map) row.toMap();
 			m.remove(uniqueIdentifierAsString);
-			String recordId = row.get(uniqueIdentifierAsString);
-			// validate that all record ids in this TSV are unique
-			if (!recordIds.add(recordId)) {
-				throw new InvalidTsvException("TSVs cannot contain duplicate primary key values");
-			}
 			changeEmptyStringsToNulls(m);
+			String recordId;
 			try {
+				recordId = row.get(uniqueIdentifierAsString);
 				batch.add(new Record(recordId, recordType, new RecordAttributes(m)));
 			} catch (IllegalArgumentException ex) {
 				LOGGER.error("IllegalArgument exception while reading tsv", ex);
 				throw new InvalidTsvException(
 						"Uploaded TSV is either missing the " + primaryKey
 								+ " column or has a null or empty string value in that column");
+			}
+			// validate that all record ids in this TSV are unique
+			// N.B. this happens after the try/catch block above, because
+			// that block enforces the recordId is not null/empty as part of the "new Record()" constructor
+			if (!recordIds.add(recordId)) {
+				throw new InvalidTsvException("TSVs cannot contain duplicate primary key values");
 			}
 			recordsProcessed++;
 			if (batch.size() >= batchSize) {
@@ -153,8 +156,8 @@ public class BatchWriteService {
 				}
 				recordDao.batchUpsert(instanceId, recordType, batch, schema);
 				batch.clear();
-				// remove the following line to ensure that record ids are unique within the entire TSV,
-				// instead of within a batch; this would have performance implications for very large TSVs.
+				// remove the following line to enforce that record ids are unique within the entire TSV,
+				// instead of within a batch. This would have performance implications for very large TSVs.
 				recordIds.clear();
 			}
 		}
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
@@ -30,6 +30,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -122,6 +123,7 @@ public class BatchWriteService {
 		CSVParser rows = csvFormat.parse(is);
 		String leftMostColumn = rows.getHeaderNames().get(0);
 		List<Record> batch = new ArrayList<>();
+		HashSet<String> recordIds = new HashSet<>();
 		boolean firstUpsertBatch = true;
 		Map<String, DataTypeMapping> schema = null;
 		String uniqueIdentifierAsString = primaryKey.orElse(leftMostColumn);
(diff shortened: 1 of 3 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -122,16 +123,22 @@ public int uploadTsvStream(InputStreamReader is, UUID instanceId, RecordType rec
...
 		String uniqueIdentifierAsString = primaryKey.orElse(leftMostColumn);
 		int recordsProcessed = 0;
 		for (CSVRecord row : rows) {
 			Map<String, Object> m = (Map) row.toMap();
 			m.remove(uniqueIdentifierAsString);
+			String recordId = row.get(uniqueIdentifierAsString);
+			// validate that all record ids in this TSV are unique
+			if (!recordIds.add(recordId)) {
+				throw new InvalidTsvException("TSVs cannot contain duplicate primary key values");
+			}
```
davidangb, line 138: I originally added the dupe-check to `RecordDao.getInsertBatchArgs()` to ensure that anything calling `getInsertBatchArgs()` benefited from the validation - but there's only one caller and it felt more appropriate to do it up here. What do you think, reviewer?

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -146,6 +153,9 @@ public int uploadTsvStream(InputStreamReader is, UUID instanceId, RecordType rec
 				}
 				recordDao.batchUpsert(instanceId, recordType, batch, schema);
 				batch.clear();
+				// remove the following line to ensure that record ids are unique within the entire TSV,
+				// instead of within a batch; this would have performance implications for very large TSVs.
+				recordIds.clear();
```
davidangb, line 158: I think this addresses the actual intent of the ticket, if not the letter … reviewer, do you think we should ensure uniqueness across the entire TSV?

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -130,14 +132,22 @@ public int uploadTsvStream(InputStreamReader is, UUID instanceId, RecordType rec
...
 				throw new InvalidTsvException(
 						"Uploaded TSV is either missing the " + primaryKey
 								+ " column or has a null or empty string value in that column");
 			}
+			// validate that all record ids in this TSV are unique
+			// N.B. this happens after the try/catch block above, because
+			// that block enforces the recordId is not null/empty as part of the "new Record()" constructor
+			if (!recordIds.add(recordId)) {
+				throw new InvalidTsvException("TSVs cannot contain duplicate primary key values");
+			}
```
davidangb, line 150: I originally added the dupe-check to `RecordDao.getInsertBatchArgs()` to ensure that anything calling `getInsertBatchArgs()` benefited from the validation - but there's only one caller and it felt more appropriate to do it up here. What do you think, reviewer?

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
calypsomatic: Why wouldn't we want uniqueness across the entire TSV?

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -146,6 +156,9 @@ public int uploadTsvStream(InputStreamReader is, UUID instanceId, RecordType rec
 				}
 				recordDao.batchUpsert(instanceId, recordType, batch, schema);
 				batch.clear();
+				// remove the following line to enforce that record ids are unique within the entire TSV,
```
aaronkanzer, line 159: This docstring reads a bit funky -- as if someone meant to `remove` the following line of code -- any chance we could re-word it?

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
aaronkanzer: What do you mean by `if not the letter`?

aaronkanzer: Would lack of uniqueness for an identifier of each row be problematic in whatever downstream code/workflow the user invokes? A couple random thought regarding this (curious to get @bethsheets 's thoughts here too)

• Would this be problematic for some Cromwell workflow? Or some interaction in RStudio/Jupyter/Galaxy?
• Do TSVs in this space often have some duplication between rows?

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -130,14 +132,22 @@ public int uploadTsvStream(InputStreamReader is, UUID instanceId, RecordType rec
 			Map<String, Object> m = (Map) row.toMap();
 			m.remove(uniqueIdentifierAsString);
 			changeEmptyStringsToNulls(m);
+			String recordId;
 			try {
-				batch.add(new Record(row.get(uniqueIdentifierAsString), recordType, new RecordAttributes(m)));
+				recordId = row.get(uniqueIdentifierAsString);
```
aaronkanzer, line 137: Could we throw the logic of the following here? Seems we might as well not perform the action of adding to the batch if we already know the `recordId` would be problematic below

```
if (!recordIds.add(recordId)) {
   throw new InvalidTsvException("TSVs cannot contain duplicate primary key values");
}
```

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
aaronkanzer: I agree -- makes sense where this logic is being invoked!

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
davidangb: > Why wouldn't we want uniqueness across the entire TSV?

enforcing uniqueness across the entire TSV would mean that the `HashSet<String> recordIds` has an entry for every single recordId. If the TSV has, say, a million rows, this becomes a large set, starting to push on memory and performance. Everything else streams, or works in batches; this would require a single object to hold information about all records, even if it's just the id for those records.

> What do you mean by `if not the letter`?

The Jira ticket talks about "if more than one row share the same primary key," but I've actually implemented "if more than one row _in the same batch_ share the same primary key". But since TSV upload works fine when there are duplicates in different batches, it seemed ok.

> Would lack of uniqueness for an identifier of each row be problematic …

If the TSV contains duplicate rows (across different batches), every row beyond the first simply updates the pre-existing record. We don't end up with duplicate ids in the database.

calypsomatic: I was thinking about the user perspective - since they don't know about batching, wouldn't they expect the entire tsv to act as one batch and not have later rows overwrite earlier ones?

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
davidangb: I messed around with a bunch of different order-of-operations here, and this seemed the best approach. Here are the considerations:
* `row.get(uniqueIdentifierAsString)` will throw `IllegalArgumentException` if the user specified a pk column or the pre-existing record type specifies the pk column, but the TSV doesn't contain that column at all
* the `new Record` in `batch.add(new Record(recordId …` will throw `IllegalArgumentException` if the TSV contains the pk column, but the row in question has an empty string or just whitespace for the id

So, we need a try/catch around both of those operations, and both of those should result in the same "Uploaded TSV is either missing …" message.

The only way I could make your suggestion work is to have a try/catch around `row.get(uniqueIdentifierAsString)`, then check for duplicate record ids, then have a second try/catch around `batch.add(new Record(recordId …`.

In a failure case, when there exists a duplicate id, the PR as-is results in one extraneous call to `batch.add(new Record(recordId …`. But in the success case, when the TSV is valid, we avoid processing an extra try/catch for each row. It felt better to optimize for the success case than for the failure case. This also feels optimal for a case where the first 99,999 rows are valid but the 100,000th row is a duplicate; this will result overall in one useless call to `batch.add` but avoid 99,999 extra try/catches.

Like I said, though, I messed around with it a bunch - if you see a way to make this more efficient, I welcome the suggestion!

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
davidangb: updated!

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
aaronkanzer: Makes sense! I think as it stands now is good -- thanks for looking at the alternatives.

davidangb:
@calypsomatic @aaronkanzer now updated to require unique record ids across the entire TSV!

calypsomatic (approved):

aaronkanzer (approved):

PR #156, opened 2022-12-20
AJ-784: TSV code-path refactor - TSV processing in sync with JSON processing

davidangb:
Putting this out there for initial review - I'd love to get eyes on this before it gets too too big.

Proposal: https://docs.google.com/document/d/1px91fzlmdWT7KE3e_a_mNX0elQ6sZuLTLoZWqyUmHdk/edit#heading=h.pk3m7dw191dk

## Known Differences 

_n/a_

## Follow-on work

_for future tickets:_

1. Change TSV download to use jackson-dataformats-csv and remove org.apache.commons:commons-csv altogether
2. Break TSV-related tests out of RecordControllerMockMvcTest and into a new standalone test class; consolidate tests as appropriate
3. Add an explicit secondary-detection step early after deserialization to look for dates, datetimes, relations, etc; consolidate  or delete code that is no longer necessary such as in DataTypeInferer
4. change the "is*" methods in DataTypeInferer to return their value so we don't parse twice? This may be premature optimization; we'd need some perf profiling to know if this is worth it or not


service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -141,97 +139,54 @@ private void validateRelationsAndAddColumns(UUID instanceId, RecordType recordTy
...
-			}
-			schema = createOrUpdateSchema(instanceId, recordType, batch, uniqueIdentifierAsString);
+
+		// extract column names from the schema
+		List<String> colNames;
+		FormatSchema formatSchema = tsvIterator.getParser().getSchema();
+		if (formatSchema instanceof CsvSchema actualSchema) {
+			colNames = actualSchema.getColumnNames();
+		} else {
+			throw new InvalidTsvException("Could not determine primary key column; unexpected schema type:" + formatSchema.getSchemaType());
```
davidangb, line 156: I don't expect this `else` clause to ever fire - I'm mostly using the `if … instanceof` syntax as a means of safely casting the `FormatSchema` to `CsvSchema`.

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -244,6 +199,35 @@ public int consumeWriteStream(InputStream is, UUID instanceId, RecordType record
...
+	}
+
+	@WriteTransaction
+	public int batchWriteTsvStream(Stream<Record> upsertRecords, UUID instanceId, RecordType recordType, Optional<String> primaryKey) {
+		try (StreamingWriteHandler streamingWriteHandler = new TsvStreamWriteHandler(upsertRecords)) {
+			return consumeWriteStream(streamingWriteHandler, instanceId, recordType, primaryKey);
+		} catch (IOException e) {
+			throw new BadStreamingWriteRequestException(e);
+		}
+	}
```
davidangb, line 229: note that `batchWriteJsonStream` and `batchWriteTsvStream` are very similar, but they use `JsonStreamWriteHandler` and `TsvStreamWriteHandler` respectively.

service/src/main/java/org/databiosphere/workspacedataservice/service/StreamingWriteHandler.java
```diff
@@ -14,22 +14,8 @@
 import java.util.ArrayList;
 import java.util.List;
 
-public class StreamingWriteHandler implements Closeable {
+public interface StreamingWriteHandler extends Closeable {
```
davidangb, line 17: changed to an interface and moved the previous JSON-only implementation into JsonStreamWriteHandler.

service/src/test/java/org/databiosphere/workspacedataservice/controller/TsvInputFormatsTest.java
```diff
@@ -109,11 +109,6 @@ private static Stream<Arguments> provideInputFormats() {
 						new Boolean[]{Boolean.TRUE, Boolean.FALSE, Boolean.TRUE},
 						"[TRUE, fALSE, True]"
 				),
-				Arguments.of(
-						"[\"True\", \"false\", \"true   \"]",
-						new Boolean[]{Boolean.TRUE, Boolean.FALSE, Boolean.TRUE},
-						"[\"TRUE\", \"fALSE\", \"True    \"]"
-				),
```
davidangb, line 116: removed; I don't think we want to support this.

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
aaronkanzer: A mini nitpick -- Is it worthwhile consolidating them and passing the `new <StreamingWriteHandler> newStreamingHandler` as an arg? Rather than having two very similar functions?

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -81,39 +80,15 @@ public DataTypeMapping selectBestType(DataTypeMapping existing, DataTypeMapping
...
 	//libreoffice at least uses left and right quotes which cause problems when we try to parse as JSON
-	private String replaceLeftRightQuotes(String val){
+	public String replaceLeftRightQuotes(String val){
 		return val.replaceAll("[“”]", "\"");
 	}
 
+	/* This is secondary detection. The JSON and TSV deserializers have created String objects, but those
+		Strings may represent dates, datetimes, etc. So, we inspect those Strings here.
+	 */
+	// TODO: create an explicit deserialization step that creates dates, datetimes, etc. and simplify here.
```
aaronkanzer, line 91: Just flagging this `TODO` -- wasn't sure if you wanted to include updates for it here

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -182,13 +154,8 @@ public boolean isValidBoolean(String sVal) {
...
 	}
 
-	public DataTypeMapping inferType(Object val, InBoundDataSource dataSource) {
-		if (dataSource == InBoundDataSource.TSV) {
-			return inferTypeForTsvSource(val);
-		} else if (dataSource == InBoundDataSource.JSON) {
-			return inferTypeForJsonSource(val);
-		}
-		throw new IllegalArgumentException("Unhandled inbound data source " + dataSource);
+	public DataTypeMapping inferType(Object val) {
```
aaronkanzer, line 157: Do you need to wrap `inferTypeForJsonSource` in a function here?

It doesn't seem that `inferType` is a part of an interface (?), thus I think we could simplify here to just have `inferTypeForJsonSource` as a call?

service/src/main/java/org/databiosphere/workspacedataservice/service/JsonStreamWriteHandler.java
```diff
@@ -0,0 +1,79 @@
...
+	 * 
+	 * @param numRecords
+	 * @return
+	 * @throws IOException
+	 */
+	public WriteStreamInfo readRecords(int numRecords) throws IOException {
+		int recordsProcessed = 0;
+		List<Record> result = new ArrayList<>(numRecords);
+		OperationType lastOp = null;
+		if (savedRecord != null) {
```
aaronkanzer, line 50: Could you explain this logic a bit? What would be the case of where we would enter here? And why would we need to add a Record? Would the Record not have been added prior for some reason?

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -53,15 +48,18 @@ public class BatchWriteService {
...
 	private final ObjectMapper objectMapper;
 
+	private final ObjectReader tsvReader;
+
 	private final RecordService recordService;
 
 	private static final Logger LOGGER = LoggerFactory.getLogger(BatchWriteService.class);
 
-	public BatchWriteService(RecordDao recordDao, @Value("${twds.write.batch.size:5000}") int batchSize, DataTypeInferer inf, ObjectMapper objectMapper, RecordService recordService) {
+	public BatchWriteService(RecordDao recordDao, @Value("${twds.write.batch.size:5000}") int batchSize, DataTypeInferer inf, ObjectMapper objectMapper, ObjectReader tsvReader, RecordService recordService) {
```
aaronkanzer, line 57: Just out of curiosity, why 5,000 for our batch size?

service/src/main/java/org/databiosphere/workspacedataservice/service/JsonStreamWriteHandler.java
```diff
@@ -0,0 +1,79 @@
...
+			result.add(savedRecord.getRecord());
+			lastOp = savedRecord.getOperation();
+			savedRecord = null;
+		}
+		// order matters in this condition, we don't want to advance the parser (call
+		// nextToken()) unless we're
+		// ready to consume the next BatchOperation
+		while (recordsProcessed < numRecords && parser.nextToken() != JsonToken.END_ARRAY && parser.hasCurrentToken()) {
```
aaronkanzer, line 58: What happens if the amount of tokens in the `parser` exceeds the allowable batch size? Is that possible in this case? (e.g. is the logic in this `while` duplicative?)

service/src/main/java/org/databiosphere/workspacedataservice/service/StreamingWriteHandler.java
```diff
@@ -14,22 +14,8 @@
 import java.util.ArrayList;
 import java.util.List;
 
-public class StreamingWriteHandler implements Closeable {
```
aaronkanzer, line 17: Would `AutoaCloseable` be a more robust base class to use? https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html

service/src/main/java/org/databiosphere/workspacedataservice/service/StreamingWriteHandler.java
```diff
@@ -14,22 +14,8 @@
 import java.util.ArrayList;
```
aaronkanzer, line 14: (Github won't let me comment on other lines for some reason ha) but I think some of these imports can be removed :D

service/src/main/java/org/databiosphere/workspacedataservice/service/TsvStreamWriteHandler.java
```diff
@@ -0,0 +1,88 @@
...
+
+		// if a primary key is specified, check if it is present in the TSV
+		if (primaryKey.isPresent() && !colNames.contains(primaryKey.get())) {
+			throw new InvalidTsvException(
+					"Uploaded TSV is either missing the " + primaryKey
+							+ " column or has a null or empty string value in that column");
+		}
+
+		// if primary key is not specified, use the leftmost column
+		String resolvedPK = primaryKey.orElseGet( () -> colNames.get(0) );
```
aaronkanzer, line 57: I was under the assumption that the 1st column is the PK? Is that not the case anymore? Or rather is this some type of safeguard?

service/src/main/java/org/databiosphere/workspacedataservice/service/TsvStreamWriteHandler.java
```diff
@@ -0,0 +1,88 @@
...
+	 */
+	public WriteStreamInfo readRecords(int numRecords) throws IOException {
+		List<Record> result = new ArrayList<>(numRecords);
+		for (int i = 0; i < numRecords && recordSpliterator.tryAdvance(result::add); i++) {
+			// noop; the action happens in result:add
+		}
+		return new WriteStreamInfo(result, OperationType.UPSERT);
+	}
+
+	@Override
```
aaronkanzer, line 82: Do you need the `Override` here if nothing is closed? Seems `Closable` upstream in the MRO should cover this?

service/src/main/java/org/databiosphere/workspacedataservice/tsv/TsvConverter.java
```diff
@@ -0,0 +1,43 @@
...
+            throw new InvalidTsvException(
+                    "Uploaded TSV is either missing the " + primaryKey
+                            + " column or has a null or empty string value in that column");
+        }
+        row.remove(primaryKey);
+        return new Record(recordId.toString(), recordType, row);
+    }
+
+    public Stream<Record> rowsToRecords(Stream<RecordAttributes> rows, RecordType recordType, String primaryKey) {
+        HashSet<String> recordIds = new HashSet<>(); // this set may be slow for very large TSVs
```
aaronkanzer, line 33: Just curious -- why a `HashSet` as the data structure here?

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/RecordAttributes.java
```diff
@@ -97,6 +97,11 @@ public RecordAttributes putAttributeIfAbsent(String key, Object value) {
 		return this;
 	}
 
+	public RecordAttributes remove(String key) {
```
aaronkanzer, line 100: Perhaps a nitpick, but could we name this something less generic, as it isn't an override/is our custom code? Would just help with readability, I think.

service/src/main/java/org/databiosphere/workspacedataservice/tsv/TsvDeserializer.java
```diff
@@ -0,0 +1,190 @@
...
+ */
+public class TsvDeserializer extends StdDeserializer<RecordAttributes> {
+
+    public TsvDeserializer(DataTypeInferer inferer, ObjectMapper objectMapper) {
+        this(null);
+        this.inferer = inferer;
+        this.objectMapper = objectMapper;
+    }
+
+    public TsvDeserializer(Class<?> vc) {
```
aaronkanzer, line 37: What does `vc` stand for here?

service/src/main/java/org/databiosphere/workspacedataservice/service/TsvStreamWriteHandler.java
calypsomatic: From the swagger page you can specify which column is the PK.  From the ui, there's no place to do that, so in that case or if none is specified in swagger, we take the first column.

service/src/main/java/org/databiosphere/workspacedataservice/service/JsonStreamWriteHandler.java
```diff
@@ -0,0 +1,79 @@
...
+
+	private final JsonParser parser;
+
+	private BatchOperation savedRecord;
+
+	private final InputStream inputStream;
+
+	public JsonStreamWriteHandler(InputStream inputStream, ObjectMapper objectMapper) throws IOException {
+		this.inputStream = inputStream;
+		JsonFactory factory = new JsonFactory(objectMapper.copy());
```
calypsomatic, line 30: Just curious, why a copy of the `objectMapper`?

service/src/main/java/org/databiosphere/workspacedataservice/tsv/TsvConverter.java
```diff
@@ -0,0 +1,43 @@
...
+import org.databiosphere.workspacedataservice.shared.model.Record;
+import org.databiosphere.workspacedataservice.shared.model.RecordAttributes;
+import org.databiosphere.workspacedataservice.shared.model.RecordType;
+import org.springframework.stereotype.Service;
+
+import java.util.HashSet;
+import java.util.stream.Stream;
+
+/**
+ * Resposible for converting RecordAttributes generated by the TsvDeserializer
```
calypsomatic, line 14: Nano-nit: typo of "responsible"

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
davidangb: I would prefer to handle this in a follow-on PR (this is number 3 in the "follow-on work" section of this PR's description) … since it _can_ be independent of this PR, I'd like to prevent scope creep.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
davidangb: good point, when I deleted `inferTypeForTsvSource` it made this extraneous. I'll update.

service/src/test/java/org/databiosphere/workspacedataservice/controller/TsvInputFormatsTest.java
aaronkanzer: Why don't we want to support this?

service/src/main/java/org/databiosphere/workspacedataservice/service/JsonStreamWriteHandler.java
davidangb: This code hasn't changed in this PR - it's copied from the previous `StreamingWriteHandler` into here, so I am inclined to not change it as part of this PR.

What's happening here is that the loop needs to store a bit of state. The JSON batch API allows each item in the batch to be either `upsert` or `delete`. The (pre-existing) code for this loops through the inbound JSON until _either_ we reach the batch-size limit _or_ the current item is a different (upsert, delete) operation than the previous item. In other words, each batch needs to be all the same operation. The `savedRecord` and `lastOp` variables allow the loop to know about both the previous item and the current item so it can do that comparison.

service/src/main/java/org/databiosphere/workspacedataservice/service/TsvStreamWriteHandler.java
aaronkanzer: got it -- thanks!

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
davidangb: That merged back in #77 … at the time, it was an educated guess that performed well. We could certainly quantify if a different batch size performs better, but it's doing ok as it is right now.

service/src/main/java/org/databiosphere/workspacedataservice/service/StreamingWriteHandler.java
davidangb: `Closeable` extends `AutoCloseable` :) But `Closeable` seems better - its `close()` method is defined as idempotent, and it throws `IOException` which is narrower than `AutoCloseable`'s `Exception`.

service/src/main/java/org/databiosphere/workspacedataservice/service/StreamingWriteHandler.java
davidangb: good catch, will clean up.

service/src/main/java/org/databiosphere/workspacedataservice/service/TsvStreamWriteHandler.java
davidangb: The `implements StreamingWriteHandler` requires an override to exist. BUT … this comment made me realize that we actually do need to ensure a couple things are closed. I'm updating.

service/src/main/java/org/databiosphere/workspacedataservice/tsv/TsvConverter.java
davidangb: The objective is that as we're looping through records in the TSV, we want to detect if any of the records has a duplicate ID. So we need some data structure that can hold all the IDs and quickly check for duplicates. Storing the IDs in a `HashSet` and using its `add` method to detect a duplicate seemed efficient. This code originally merged in #155 and has just been moved around in this PR.

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/RecordAttributes.java
davidangb: renamed to `removeAttribute`

service/src/main/java/org/databiosphere/workspacedataservice/tsv/TsvDeserializer.java
davidangb: it stands for `_valueClass`, in the parent `StdDeserializer`. The `super` constructor we call here is defined as `protected StdDeserializer(Class<?> vc) {…` so I'm just mirroring it here. Actually just allowed IntelliJ to generate this constructor and it smartly mirrored the super's signature.

service/src/main/java/org/databiosphere/workspacedataservice/service/JsonStreamWriteHandler.java
aaronkanzer: Would that necessitate this logic to be `||` instead of `&&`?

service/src/main/java/org/databiosphere/workspacedataservice/service/JsonStreamWriteHandler.java
davidangb: I … am not sure. This code is pre-existing and moved from the previous `StreamingWriteHandler`. I suspect that since we're creating a `JsonFactory` / `JsonParser` out of it, it's safer to create those from a copy instead of reusing the exact same instance.

service/src/main/java/org/databiosphere/workspacedataservice/tsv/TsvConverter.java
davidangb: fixed :)

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
davidangb: I took a look at this and my opinion is to leave it as-is. The `batchWriteJsonStream` / `batchWriteTsvStream` are a concise place to implement the try-with-resources and a concise place to add the `@WriteTransaction` annotation. I could go either way on this, though - lmk if you feel strongly about refactoring this.

service/src/main/java/org/databiosphere/workspacedataservice/service/JsonStreamWriteHandler.java
aaronkanzer: got it -- thanks for the explanation!

service/src/main/java/org/databiosphere/workspacedataservice/service/StreamingWriteHandler.java
aaronkanzer: makes sense!

service/src/main/java/org/databiosphere/workspacedataservice/tsv/TsvConverter.java
aaronkanzer: got it -- thanks!

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
aaronkanzer: No strong opinions! 📈

service/src/main/java/org/databiosphere/workspacedataservice/tsv/TsvDeserializer.java
```diff
@@ -0,0 +1,190 @@
...
+                return new BigDecimal(val);
+            }
+        }
+        // JSON objects
+        // TODO: change the "is*" methods in inferer to return their value so we don't parse twice?
+        if (inferer.isValidJson(val)) {
+            try {
+                return objectMapper.readValue(val, new TypeReference<Map<String, Object>>(){});
+            } catch (JsonProcessingException jpe) {
+                // this shouldn't happen; if inferer.isValidJson(val) passes, so should the .readValue
```
aaronkanzer, line 122: Why the `try/catch` here if "this shouldn't happen"?

service/src/main/java/org/databiosphere/workspacedataservice/tsv/TsvDeserializer.java
```diff
@@ -0,0 +1,190 @@
...
+                return val;
+            }
+        }
+
+        // arrays
+        String smartQuotesRemoved = inferer.replaceLeftRightQuotes(val);
+        if (inferer.isArray(smartQuotesRemoved.toLowerCase())) {
```
aaronkanzer, line 129: Is the reason to do `toLowerCase()` if the string is bool values?

service/src/test/java/org/databiosphere/workspacedataservice/tsv/TsvJsonArgumentsProvider.java
```diff
@@ -0,0 +1,127 @@
...
+                Arguments.of("😍😎😺",     "😍😎😺",        true),
+                Arguments.of("12345A",     "12345A",        true),
+                    Arguments.of("\uD83D\uDCA9\u0207", "\uD83D\uDCA9\u0207", true),
+
+                // strings that look like other data types
+                Arguments.of("2021-10-03",          "2021-10-03",           true),
+                Arguments.of("2021-10-03T19:01:23", "2021-10-03T19:01:23",  true),
+                Arguments.of("terra-wds:/type/id",  "terra-wds:/type/id",   true),
+
+                // TODO: these quoted-string tests are failing; fix this somehow.
```
aaronkanzer, line 65: Assuming this could be tackled in a follow-up PR?

service/src/test/java/org/databiosphere/workspacedataservice/tsv/TsvJsonEquivalenceTest.java
```diff
@@ -0,0 +1,93 @@
...
+                {"attributes":{"col1":%s}}
+                """.formatted(jsonValue);
+
+        RecordAttributes tsvAttributes = readTsv(tsv);
+        RecordAttributes jsonAttributes = mapper.readValue(json, RecordRequest.class).recordAttributes();
+
+        Object tsvActual = tsvAttributes.getAttributeValue("col1");
+        Object jsonActual = jsonAttributes.getAttributeValue("col1");
+
+        //noinspection rawtypes
```
aaronkanzer, line 74: I'm not quite sure what this docstring means?

service/src/main/java/org/databiosphere/workspacedataservice/service/JsonStreamWriteHandler.java
davidangb: no, I think the logic is correct as-is. I've added a unit test at `RecordControllerMockMvcTest.batchWriteInsertShouldSucceedIfMultipleBatches`, if that helps with confidence.

service/src/main/java/org/databiosphere/workspacedataservice/service/TsvStreamWriteHandler.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+/**
+ * Stream-reads inbound TSV data using a Jackson CsvMapper and WDS's TsvConverter;
+ *  returns a WriteStreamInfo containing a batch of Records.
+ */
+public class TsvStreamWriteHandler implements StreamingWriteHandler {
+
+	private Spliterator<Record> recordSpliterator;
```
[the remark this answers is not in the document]
davidangb: yes; will update

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -1175,6 +1175,25 @@ void batchWriteInsertShouldSucceed() throws Exception {
...
+			String recordId = "record_" + i;
+			Record record = new Record(recordId, newBatchRecordType,
+					new RecordAttributes(Map.of("key", "value_" + i)));
+			ops.add(new BatchOperation(record, OperationType.UPSERT));
+		}
+
+		mockMvc.perform(post("/{instanceid}/batch/{v}/{type}", instanceId, versionId, newBatchRecordType)
+						.content(mapper.writeValueAsString(ops))
+						.contentType(MediaType.APPLICATION_JSON)).andExpect(jsonPath("$.recordsModified", is(totalRecords)))
+				.andExpect(jsonPath("$.message", is("Huzzah"))).andExpect(status().isOk());
```
aaronkanzer, line 1194: Huzzah! 🎉

service/src/main/java/org/databiosphere/workspacedataservice/service/JsonStreamWriteHandler.java
aaronkanzer: thanks for this -- yes, definitely helps!

service/src/main/java/org/databiosphere/workspacedataservice/tsv/TsvDeserializer.java
```diff
@@ -0,0 +1,190 @@
...
+        if (node.isObject()) {
+            Map<String, Object> attrs = new HashMap<>(node.size() * 2); // oversize the map to avoid resizing
+            Iterator<Map.Entry<String, JsonNode>> fields = node.fields();
+            while (fields.hasNext()) {
+                Map.Entry<String, JsonNode> field = fields.next();
+                Object attrValue = this.nodeToObject(field.getValue());
+                attrs.put(field.getKey(), attrValue);
+            }
+            return new RecordAttributes(attrs);
+        } else {
```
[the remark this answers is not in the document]
davidangb: This is another case of "we should never hit the else condition."

Because `StdDeserializer` is common/generic across a bunch of Jackon libraries, its `deserialize` method accepts a `JsonParser`, and the `JsonParser.getCodec().readTree(parser);` returns a generic `JsonNode`.

However, specific to `jackson-dataformat-csv`, the call will always return an object, not any other kind of node.

The else clause is here just as a failsafe - if something went wrong inside the library and we got an error, I want to have a log entry explaining it instead of just a mysterious error that is difficult to debug.

service/src/main/java/org/databiosphere/workspacedataservice/tsv/TsvDeserializer.java
davidangb: I could remove it if you think it's objectionable … without the try/catch, I'd need to add `throws JsonProcessingException` to the method signature, and to the signature of calling methods, until we did have a try/catch. I also like having the try/catch here because if the implementation of `inferer.isValidJson()` changes, we're protected.

service/src/main/java/org/databiosphere/workspacedataservice/tsv/TsvDeserializer.java
davidangb: yes - exactly. I'm adding a code comment with pseudocode to help make this understandable.

service/src/test/java/org/databiosphere/workspacedataservice/tsv/TsvJsonArgumentsProvider.java
davidangb: I want to do some manual testing and see the behavior difference in practice, we might need to tackle it in this PR if it's too much of a big deal!

service/src/test/java/org/databiosphere/workspacedataservice/tsv/TsvJsonEquivalenceTest.java
davidangb: Good call Phil, I got lazy on this one - fixing!

service/src/test/java/org/databiosphere/workspacedataservice/controller/TsvInputFormatsTest.java
davidangb: The previous behavior doesn't match the spec:

> To explicitly create a string in WDS with a value that would otherwise resolve to a relation, number, boolean, date, or datetime, wrap your string in double quotes. For instance: "123" or "true".

from [DataBiosphere/terra-workspace-data-service/wiki/TSV-functionality-and-syntax#strings](https://github.com/DataBiosphere/terra-workspace-data-service/wiki/TSV-functionality-and-syntax#strings)

calypsomatic (approved):
Great work!  Given that there are a few follow-on tickets for the rough edges, I think this is good to go - modulo maybe some cleaning up of unused imports and other small things highlighted by sonarcloud.

service/src/test/java/org/databiosphere/workspacedataservice/tsv/TsvJsonArgumentsProvider.java
aaronkanzer: sounds good!

aaronkanzer:
I think this docstring might need to be updated: https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java#L107

davidangb:
@aaronkanzer good catch on that code comment; updated.

service/src/main/java/org/databiosphere/workspacedataservice/service/TsvStreamWriteHandler.java
aaronkanzer: Just bumping this -- did you ever want to update anything else here?

davidangb: I did, in this commit: [`9df5fbe` (#156)](https://github.com/DataBiosphere/terra-workspace-data-service/pull/156/commits/9df5fbee5aacd76d7d68e93b4087b8f249ca9718)

jordansafer (approved):

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -249,6 +170,28 @@ public int consumeWriteStream(InputStream is, UUID instanceId, RecordType record
 		return recordsAffected;
 	}
 
+
+
+	// try-with-resources wrapper for JsonStreamWriteHandler; calls consumeWriteStream.
+	@WriteTransaction
```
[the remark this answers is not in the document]
davidangb: I am removing `@WriteTransaction` from `batchWriteJsonStream` and `batchWriteTsvStream`, but leaving it on `consumeWriteStream`. If we moved it all the way down to `createOrModifyRecordType`, then we'd have separate transactions for each batch in the request, and we could end up writing only part of the request.

service/src/main/java/org/databiosphere/workspacedataservice/tsv/TsvConverter.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.util.HashSet;
+import java.util.stream.Stream;
+
+/**
+ * Responsible for converting RecordAttributes generated by the TsvDeserializer
+ * into fully-formed Records.
+ */
+@Service
+public class TsvConverter {
```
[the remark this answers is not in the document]
davidangb: what if I did away with TsvConverter entirely, and moved those two methods inside TsvStreamWriteHandler?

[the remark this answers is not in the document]
davidangb: I consolidated them without waiting for your answer :)

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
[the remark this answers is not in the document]
davidangb: huh, removing those transactions caused test failures. I reversed and left `@WriteTransaction` on `batchWriteJsonStream` and `batchWriteTsvStream`, but removed it from `consumeWriteStream`.

aaronkanzer (approved):

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -511,6 +511,25 @@ void tsvUploadUsesFirstColumnAsPrimaryKey() throws Exception {
 		assertEquals("greeting", schema.primaryKey());
 	}
 
+	@Test
+	@Transactional
+	void tsvUploadUsesSpecifiedColumnAsPrimaryKey() throws Exception {
+		String pk = "embedded characters";
+		MockMultipartFile file = new MockMultipartFile("records", "test.tsv", MediaType.TEXT_PLAIN_VALUE,
+				RecordControllerMockMvcTest.class.getResourceAsStream("/small-test-no-sys.tsv"));
+		String recordType = "noPrimaryKeySpecified";
```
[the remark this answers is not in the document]
davidangb: updated!

PR #157, opened 2023-01-03
Aj 735 delete array of relations

calypsomatic:
[AJ-735: Delete arrays of relations](https://broadworkbench.atlassian.net/browse/AJ-735)
Deleting a record should remove the values from the relation-array table; deleting a record type should drop the join table altogether.

[AJ-735]: https://broadworkbench.atlassian.net/browse/AJ-735?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AJ-735]: https://broadworkbench.atlassian.net/browse/AJ-735?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -741,10 +741,8 @@ public class RecordDao {
 	@CacheEvict(value = PRIMARY_KEY_COLUMN_CACHE, key = "{ #recordType.name, #instanceId.toString()}")
 	public void deleteRecordType(UUID instanceId, RecordType recordType) {
 		List<Relation> relationArrayCols = getRelationArrayCols(instanceId, recordType);
-		if (!relationArrayCols.isEmpty()){
-			for (Relation rel : relationArrayCols){
-				namedTemplate.getJdbcTemplate().update("drop table " + getQualifiedJoinTableName(instanceId, rel.relationColName(), recordType));
-			}
+		for (Relation rel : relationArrayCols){
+			namedTemplate.getJdbcTemplate().update("drop table " + getQualifiedJoinTableName(instanceId, rel.relationColName(), recordType));
 		}
 		try {
 			namedTemplate.getJdbcTemplate().update("drop table " + getQualifiedTableName(recordType, instanceId));
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -139,21 +139,21 @@ public class RecordDao {
 		}
 	}
 
-	private String getJoinTableName(String relationColumnName, RecordType fromTable){
+	public String getJoinTableName(String relationColumnName, RecordType fromTable){
 		//Use RESERVED_NAME_PREFIX to ensure no collision with user-named tables.
 		//RecordType name has already been sql-validated
 		return quote(RESERVED_NAME_PREFIX + fromTable.getName() + "_" + SqlUtils.validateSqlString(relationColumnName, ATTRIBUTE));
 	}
 
-	private String getQualifiedJoinTableName(UUID instanceId, String relationColumnName, RecordType fromTable){
+	public String getQualifiedJoinTableName(UUID instanceId, String relationColumnName, RecordType fromTable){
 		return quote(instanceId.toString()) + "." + getJoinTableName(relationColumnName, fromTable);
 	}
 
-	private String getFromColumnName(RecordType referringRecordType){
+	public String getFromColumnName(RecordType referringRecordType){
 		return "from_" + referringRecordType.getName() + "_key";
 	}
 
-	private String getToColumnName(RecordType referencedRecordType){
+	public String getToColumnName(RecordType referencedRecordType){
 		return "to_" + referencedRecordType.getName() + "_key";
 
 	}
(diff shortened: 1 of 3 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/dao/RecordDaoTest.java
```diff
@@ -230,6 +262,74 @@ void testDeleteRelatedRecord() {
 		assertThrows(ResponseStatusException.class, () -> {
 			recordDao.deleteSingleRecord(instanceId, recordType, "referencedRecord");
 		}, "Exception should be thrown when attempting to delete related record");
+
+		//Record should not have been deleted
+		assert(recordDao.getSingleRecord(instanceId, recordType, "referencedRecord").isPresent());
+
```
calypsomatic, line 268: Accidentally started writing the wrong test, realized it was useful anyway, and also updated the other similar one, even though they're not directly relevant to this ticket.

service/src/test/java/org/databiosphere/workspacedataservice/dao/RecordDaoTest.java
```diff
@@ -230,6 +262,74 @@ void testDeleteRelatedRecord() {
...
+		recordDao.batchUpsert(instanceId, relationArrayType, Collections.singletonList(recordWithRelationArray), schema);
+
+		//Normally insert into join is called from the service level, so when working directly with the dao must call it manually
+		recordDao.insertIntoJoin(instanceId, arrayRelation, relationArrayType, List.of(new RelationValue(recordWithRelationArray, referencedRecord), new RelationValue(recordWithRelationArray, referencedRecord2)));
+
+		recordDao.deleteSingleRecord(instanceId, relationArrayType, "recordWithRelationArr");
+
+		//Record should have been deleted
+		assert(recordDao.getSingleRecord(instanceId, recordType, "recordWithRelationArr").isEmpty());
+		//TODO check that values are removed from join table? is it worth it to write a dao method just for this test?
```
calypsomatic, line 332: Would really like to check the join tables in all these tests, but there are no existing dao methods to look at them outside of activity inside other methods.  Is it worth it to make extra dao methods just for tests?

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -565,6 +573,26 @@ private String getInsertColList(List<String> existingTableSchema) {
 	public void batchDelete(UUID instanceId, RecordType recordType, List<Record> records) {
 		List<String> recordIds = records.stream().map(Record::getId).toList();
 		try {
+			List<Relation> relationArrayCols = getRelationArrayCols(instanceId, recordType);
+			if (!relationArrayCols.isEmpty()){
+				//Remove values from join table first
+				for (Relation rel : relationArrayCols) {
+					namedTemplate.getJdbcTemplate().batchUpdate(
+							"delete from" + getQualifiedJoinTableName(instanceId, rel.relationColName(), recordType) + " where " + quote(getFromColumnName(recordType)) + " = ?",
```
aaronkanzer, line 581: Just for clarity, what does the ` = ?` syntax mean in this query?

calypsomatic: The ? is a placeholder that gets replaced by the values defined in the `setValues()` function before the sql is actually run.

service/src/test/java/org/databiosphere/workspacedataservice/dao/RecordDaoTest.java
davidangb: For DB queries that will only be called from tests, I see two different approaches:
1. Create a new `TestDao` (feel free to rename) that holds test-only methods; this can live in the source tree under `src/test` so it doesn't affect any runtime code. I'd go this approach if you find the need to reuse these test-only methods across test classes.
2. Add an `@Autowired` `NamedParameterJdbcTemplate namedTemplate` directly to this test class, and execute db queries using that `namedTemplate`. You could create reusable helper method right in this class if the method is used in multiple test cases, but only within this class.

And - I agree, these tests (and the ones in `RecordControllerMockMvcTest`) should validate that they do actually remove the rows from the join table when deleting a record, or that they delete the join tables entirely when deleting a type.

davidangb (commented):
one comment inline about db queries within tests. Syntax of the code looks good.

However, one design-level consideration: instead of manually deleting the join table rows when the "from" row is deleted, how about setting up the join table with `ON DELETE CASCADE` and letting Postgres handle it transparently? That should remove some complexity from the Java code. Of course you'd still have to handle dropping the table when necessary, and deleting the join table rows when the attribute is set to null/empty but the "from" row still exists.

davidangb (approved):
ty!

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -736,6 +740,12 @@ Map<String, RecordType> getRelationColumnsByName(List<Relation> referenceCols) {
 	 */
 	@CacheEvict(value = PRIMARY_KEY_COLUMN_CACHE, key = "{ #recordType.name, #instanceId.toString()}")
 	public void deleteRecordType(UUID instanceId, RecordType recordType) {
+		List<Relation> relationArrayCols = getRelationArrayCols(instanceId, recordType);
+		if (!relationArrayCols.isEmpty()){
```
sagehen03, line 744: Do we need this check?  I think getRelationArrayCols(instanceId, recordType) will return an empty list which will make the for loop a no-op.

sagehen03 (approved):
Looks good, I end to end tested deleting a record and record type.  I'm pretty sure you don't need the list is not empty check I commented on earlier, but I won't hold this up on that account.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
calypsomatic: Good point!

PR #158, opened 2023-01-04
AJ-719: Create PRs in Cromwhelm with automerge label

sagehen03:
- ready to test
- AJ-55: this is merely a test
- create new branch fix up warnings
- AJ-99: just a test
- AJ-99: just another test

[AJ-99]: https://broadworkbench.atlassian.net/browse/AJ-99?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AJ-99]: https://broadworkbench.atlassian.net/browse/AJ-99?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Set up JDK
         uses: actions/setup-java@v2
         with:
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -39,7 +39,7 @@ jobs:
         ports: [ "5432:5432" ]
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         # Needed by sonar to get the git history for the branch the PR will be merged into.
         with:
           fetch-depth: 0
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -120,4 +120,4 @@ jobs:
           git config --global user.email "broadbot@broadinstitute.org"
           git commit -am "${{ steps.ensure-jira-id.outputs.JIRA_ID }} Auto update WDS version to $HELM_NEW_TAG"
           git push -u origin ${{ steps.ensure-jira-id.outputs.JIRA_ID }}-auto-update-${{ steps.setHash.outputs.git_short_sha }}
-          gh pr create --title "Auto update WDS version to $HELM_NEW_TAG" --body "Testing, don't merge" --label "automerge"
+          gh pr create --title "${{ steps.ensure-jira-id.outputs.JIRA_ID }}: auto update WDS version to $HELM_NEW_TAG" --body "Testing, don't merge" --label "automerge"
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -104,14 +104,14 @@ jobs:
           
       - name: Clone Cromwhelm and create PR for latest WDS version
         uses: actions/checkout@v3
-        if: ${{ steps.ensure-jira-id.outputs.JIRA_ID != "missing" }}
+        if: startsWith(${{ steps.ensure-jira-id.outputs.JIRA_ID }},  "AJ-")
         with:
           repository: broadinstitute/cromwhelm
           token: ${{ secrets.BROADBOT_TOKEN }} # Has to be set at checkout AND later when pushing to work
           path: cromwhelm
           
       - name: Update WDS image tag for CBAS Helm Chart in Cromwhelm
-        if: ${{ steps.ensure-jira-id.outputs.JIRA_ID != "missing" }}
+        if: startsWith(${{ steps.ensure-jira-id.outputs.JIRA_ID }},  "AJ-")
         env:
           BROADBOT_TOKEN: ${{ secrets.BROADBOT_TOKEN }}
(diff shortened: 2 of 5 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
```
sagehen03, line 14: This change and the handful of others like it is to remove deprecation warnings around node version  https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

.github/workflows/publish-docker.yml
```diff
@@ -20,7 +20,7 @@ env:
   GOOGLE_PROJECT: broad-dsp-gcr-public
 
 jobs:
-  publish-job:
+  publish-docker-job:
```
sagehen03, line 23: The job that publishes the wds java client is also called publish-job so I changed the name here to disambiguate.

.github/workflows/publish-docker.yml
```diff
@@ -32,6 +32,17 @@ jobs:
...
+        id: find-jira-id
+        run: |
+          set +e
+          JIRA_ID=$(echo '${{ github.event.head_commit.message }}' | grep -iEo 'AJ-[0-9]+' || '')
+          if [[ -z "$JIRA_ID" ]]; then
+            echo ::set-output name=JIRA_ID::"missing"
+          else 
+            echo ::set-output name=JIRA_ID::${JIRA_ID}
+          fi
+          set -e
```
sagehen03, line 45: This is a bit yuck.  I added the `set +e` so that if grep does not find a jira id in the commit message then we still continue with the rest of the code (setting the step output to missing).  This also means that the some other runtime error won't terminate execution early.  I haven't been able to think of a plausible case where that could cause a problem, but I'm generally not a fan of the carpet bombing approach.

calypsomatic (approved):

.github/workflows/publish-docker.yml
pshapiro4broad: There is probably a way to do this using `grep`, but I know it can be done with `awk`:
```
JIRA_ID=$(echo '${{ github.event.head_commit.message }}' | awk 'match($0, /[Aa][Jj]-[0-9]+/) {print substr($0, RSTART, RLENGTH)}')
```

pshapiro4broad: That is, using `awk` for this will return an exit code of `0` in both cases so you don't need `set +-e` anymore.

.github/workflows/publish-docker.yml
```diff
@@ -92,24 +103,30 @@ jobs:
...
-      - name: Update WDS image tag for CBAS Helm Chart in cromwhelm
+      - name: Create Cromwhelm PR for latest WDS version
         env:
           BROADBOT_TOKEN: ${{ secrets.BROADBOT_TOKEN }}
+          GH_TOKEN: ${{ secrets.BROADBOT_TOKEN }}
         run: |
           set -e
+          JIRA_ID=${{ steps.find-jira-id.outputs.JIRA_ID }}
+          if [[ $JIRA_ID == "missing" ]]; then
+            exit 0;
```
aaronkanzer, line 120: Why do we want to `exit 0` here?

Are we saying that if a PR does not have a JIRA ID, then it shouldn't update the Helm chart at that point in time?

calypsomatic: Yes - without a JIRA ID, then it wouldn't be compliant.

aaronkanzer: Should it be an `exit 1` then?

calypsomatic: Oh wait.  Hmm.

aaronkanzer:
While we are bumping `v2` to `v3` in some of the Github Actions steps in this PR, should we also just bump:

https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/.github/workflows/build-and-test.yml#L16
https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/.github/workflows/build-and-test.yml#L53
https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/.github/workflows/release-python-client.yml#L26

.github/workflows/publish-docker.yml
calypsomatic: I guess it would stop the execution so it wouldn't continue with creating a new PR or anything that goes on in the following lines, but without blowing up the whole publish job.

aaronkanzer: Would we want to `echo` some output signifying that there should be a JIRA ID? 

Or rather I guess I'm still somewhat confused if we would want a Github Actions pipeline to fail because of this?

calypsomatic: Yeah maybe it should fail so we know what happened.

davidangb (approved):
I'm fine with this PR as-is, assuming you've tested it and it is working in practice.

If we were to rewrite this at any point, I'd suggest separating the create-PR-in-cromwhelm functionality out of `publish-docker.yml` and into its own standalone action; this would make it easier to debug and there would be less worry about what happens if any of the steps failed. Cromwhelm has a decent example of this: https://github.com/broadinstitute/cromwhelm/blob/main/.github/workflows/update-and-publish-helm-chart.yml (which also looks for any two-character Jira prefix, like IA- or WF-).

PR #161, opened 2023-01-10
[AJ-668] default instance creation

jordansafer:
WDS changes for default instance creation. With these changes, WDS will attempt to read the workspaceId environment variable provided by Leo, and if it's available it'll create an instance where instanceId = workspaceId


README.md
```diff
@@ -89,6 +89,14 @@ From the command line, run
 ```bash
 ./gradlew test
 ```
+To run one test suite, run
+```bash
+./gradlew test --tests '*TestSuite'
```
calypsomatic, line 94: Just to help dumb people like me, can there be a real-life example here?   I'm thinking `TestSuite` refers to the java class, right, such as `RecordDaoTest`, but I'm not sure so if it was spelled out for me here that would be nice.

service/build.gradle
```diff
@@ -65,6 +65,7 @@ tasks.named('test') {
 		events = ["passed", "failed", "skipped", "started"]
 		exceptionFormat = "full"
 	}
+	environment "WORKSPACE_ID", "123e4567-e89b-12d3-a456-426614174000"
```
calypsomatic, line 68: Is this just for tests?  Does it need to be marked such in some way?

service/src/test/java/org/databiosphere/workspacedataservice/dao/RecordDaoTest.java
```diff
@@ -53,6 +80,17 @@ void cleanUp(){
...
+	@Test
+	void defaultSchemaIsCreated() {
+		LOGGER.info("Default workspace id loaded as {}", workspaceId);
+		UUID defaultInstanceId = UUID.fromString(workspaceId);
+		assertTrue(recordDao.instanceSchemaExists(defaultInstanceId));
+	}
+
+	@Test
+	void workspaceIDNotProvidedNoExceptionThrown() {
+		RecordDao noWorkspaceRecordDao = new RecordDao(namedTemplate, templateForStreaming, dataTypeInferer, objectMapper, cachedQueryDao, "UNDEFINED");
```
calypsomatic, line 92: This just makes sure an exception isn't thrown?  Does it make sense to do something like `assertDoesNotThrow(() -> new RecordDao(namedTemplate, templateForStreaming, dataTypeInferer, objectMapper, cachedQueryDao, "UNDEFINED"));`

service/build.gradle
jordansafer: It _should_ only be for tests since it's under the test task, and for our usecase, this shouldn't be an issue since environment variables don't get build into jib, and the test task won't be run during bootRun/non-dockerized testing

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -81,12 +82,31 @@ public class RecordDao {
...
+		LOGGER.info("Default workspace id loaded as {}", workspaceId);
+
+		try {
+			UUID instanceId = UUID.fromString(workspaceId);
+			if (!instanceSchemaExists(instanceId)) {
+				createSchema(instanceId);
+				LOGGER.info("Creating default schema id succeeded for workspaceId {}", workspaceId);
+			}
+		} catch (IllegalArgumentException e) {
+			LOGGER.warn("Workspace id could not be parsed provided, a default schema won't be created. Provided id: {}", workspaceId);
```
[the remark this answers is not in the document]
jordansafer: Oops just parsed, 'provided' shouldn't be here, I was struggling with word choice :)

service/build.gradle
calypsomatic: Ah sorry the `tasks.named(test)` part wasn't shown in the excerpt.

davidangb (approved):
lgtm, thank you!

calypsomatic (approved):

PR #162, opened 2023-01-12
abandon terra-common-lib [AJ-767]

davidangb:
followup to #160 .

This PR removes https://github.com/DataBiosphere/terra-common-lib as a dependency, and copies (as source code) just the pieces of TCL we really need. Including TCL as a whole had undesirable side effects such as resetting logging config; we also had to intentionally disable some TCL features such as Liquibase. By copying just the pieces we need, we keep WDS slim and without unpredictable side effects.

The drawback is that the code copied in here will not benefit from TCL fixes/enhancements, if those ever happen. An ideal solution would be to fix/repackage TCL, but that's larger in scope than I'd like to tackle right now.

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -35,15 +35,14 @@ dependencies {
 	implementation 'org.springframework.boot:spring-boot-starter-jdbc'
 	implementation 'org.springframework.boot:spring-boot-starter-web'
 	implementation 'org.springframework.retry:spring-retry:1.3.4'
+	implementation 'org.aspectj:aspectjweaver:1.8.9' // required by spring-retry, not used directly by WDS
 	implementation 'org.apache.commons:commons-lang3'
 	implementation 'org.apache.commons:commons-csv:1.9.0'
 	implementation 'com.google.guava:guava:31.1-jre'
 	implementation 'org.postgresql:postgresql'
 	implementation 'org.springframework.boot:spring-boot-starter-actuator'
 	implementation 'org.webjars:webjars-locator-core:0.52'
-	implementation ('bio.terra:terra-common-lib:0.0.77-SNAPSHOT') {
-		exclude group: 'org.broadinstitute.dsde.workbench', module: 'sam-client_2.12'
-	}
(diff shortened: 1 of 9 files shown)


service/build.gradle
```diff
@@ -35,15 +35,14 @@ dependencies {
...
 	implementation 'org.apache.commons:commons-lang3'
 	implementation 'org.apache.commons:commons-csv:1.9.0'
 	implementation 'com.google.guava:guava:31.1-jre'
 	implementation 'org.postgresql:postgresql'
 	implementation 'org.springframework.boot:spring-boot-starter-actuator'
 	implementation 'org.webjars:webjars-locator-core:0.52'
-	implementation ('bio.terra:terra-common-lib:0.0.77-SNAPSHOT') {
-		exclude group: 'org.broadinstitute.dsde.workbench', module: 'sam-client_2.12'
-	}
+	annotationProcessor 'org.springframework.boot:spring-boot-configuration-processor'
```
davidangb, line 45: this annotation processor include is not strictly necessary, but some of the code was complaining in IntelliJ about it being missing, so I added in here as a rider to this PR.

service/src/main/java/bio/terra/common/README.md
```diff
@@ -0,0 +1,5 @@
+The classes under `bio.terra.common` are copied, as source code, from
+the `terra-common-lib` [github repository](https://github.com/DataBiosphere/terra-common-lib).
+
+We do this because importing `terra-common-lib` as a whole has side effects such as changing
+logging config, but we want the `@ReadTransaction` and `@WriteTransaction` annotation support.
```
davidangb, line 5: reviewer: I didn't make any changes to the 6 files I'm adding in bio.terra.common, and this code was already reviewed when merging into TCL. So, review these files at the level that makes you comfortable!

calypsomatic (approved):
I didn't look too carefully at the copied code, but I tested the transactions and those still work.  Can we have a ticket in the backlog to revisit this someday in hopes of less code copying?

jordansafer (approved):

davidangb:
@calypsomatic good call, I made AJ-773.

PR #163, opened 2023-01-12
Sentry via Logback [AJ-763]

davidangb:
This PR configures [sentry-logback](https://docs.sentry.io/platforms/java/guides/logback/) in WDS. We create a new Sentry-specific logback appender, set some variables on the `Sentry` object, and then magically Sentry finds exceptions in our logs and sends them to the WDS-specific Sentry dashboard.

Using the logback approach for Sentry should make this fully compatible with [Application Insights instrumentation](https://learn.microsoft.com/en-us/azure/azure-monitor/app/java-spring-boot) when we add that in the future.

See the dashboard at https://sentry.io/organizations/broad-institute/projects/wds/?issuesType=all&project=4504299946835968.

The `samurl` and `lz_mrg` environment variables depend on https://github.com/broadinstitute/cromwhelm/pull/172 which in turn (partially) depends on https://github.com/DataBiosphere/leonardo/pull/3114 .  The `releasename` variable depends on https://github.com/broadinstitute/cromwhelm/pull/173.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
@@ -6,10 +6,10 @@ import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.PropertySource;
-import org.springframework.context.annotation.PropertySources;
 
 @Configuration
-@PropertySources({@PropertySource("classpath:git.properties"), @PropertySource("classpath:application.properties")})
+@PropertySource("classpath:git.properties")
+@PropertySource("classpath:application.properties")
 public class SentryInitializer  {
 
 	@Value("${sentry.dsn}")
@@ -29,7 +29,7 @@ public class SentryInitializer  {
 
 	@Bean
 	public SmartInitializingSingleton initialize() {
-		return () -> {
+		return () ->
         Sentry.init(options -> {
 				options.setDsn(dsn);
 				options.setEnvironment(env);
@@ -37,6 +37,5 @@ public class SentryInitializer  {
 				options.setRelease(release);
 				options.setTag("workspaceId", workspaceId);
 			});
-		};
 	}
 }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
@@ -7,6 +7,9 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.PropertySource;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 @Configuration
 @PropertySource("classpath:git.properties")
 @PropertySource("classpath:application.properties")
@@ -39,15 +42,30 @@ public class SentryInitializer  {
 			});
 	}
 
-	String urlToEnv(String samUrl){
-		String env = "dev";
-		if (samUrl != null){
-			int dsde_loc = samUrl.indexOf("dsde-");
-			int broad_loc = samUrl.indexOf(".broad");
-			if (dsde_loc > -1 && broad_loc > -1 && dsde_loc != broad_loc){
-				env = samUrl.substring(dsde_loc+5,broad_loc);
-			}
+	private final Pattern SAM_ENV_PATTERN = Pattern.compile("\\.dsde-(\\p{Alnum}+)\\.");
+	private final String DEFAULT_ENV = "unknown";
+
+	/**
+	 * Extracts an environment (e.g. "dev" or "prod") from a Sam url.
+	 * Looks for ".dsde-${env} and returns ${env} if found.
+	 * Also looks for BEEs and returns the bee name if found.
+	 * Else, returns "unknown".
+	 *
+	 * @param samUrl the url to Sam
+	 * @return the environment as parsed from the Sam url
+	 */
+	protected String urlToEnv(String samUrl) {
+		if (samUrl == null) {
+			return DEFAULT_ENV;
 		}
-		return env;
+		Matcher matcher = SAM_ENV_PATTERN.matcher(samUrl);
+		boolean found = matcher.find();
+		if (found) {
+			return matcher.group(1);
+		} else if (samUrl.endsWith("bee.envs-terra.bio")) {
+			return samUrl.split("\\.")[1];
+		} else {
+			return DEFAULT_ENV;
+		}
 	}
 }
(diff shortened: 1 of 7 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
```diff
@@ -0,0 +1,53 @@
...
+        Sentry.init(options -> {
+				options.setDsn(dsn);
+				options.setEnvironment(urlToEnv(samurl));
+				options.setServerName(serverName);
+				options.setRelease(release);
+				options.setTag("workspaceId", workspaceId);
+			});
+	}
+
+	String urlToEnv(String samUrl){
```
calypsomatic, line 42: Is there a smarter way to parse this url?  Is dev a good default?

service/src/main/resources/application.properties
```diff
@@ -28,6 +28,10 @@ twds.streaming.fetch.size=5000
 # Workspace Id for launching instance
 twds.instance.workspace-id=${WORKSPACE_ID:}
 
+sentry.dsn=https://e59ecdd940784bd2922f25a0f3197ffd@o54426.ingest.sentry.io/4504299946835968
+sentry.samurl=${SAM_URL:} 
+sentry.servername=${LZ_MRG:}
```
calypsomatic, line 33: Sentry gets a bit mad if there is no server name 
image
Do we care?  What's a good default?

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
jordansafer: You could use a regex with a capture group. Feel free to bother me for details on this!

service/src/main/resources/application.properties
jordansafer: I'm thinking "wds" or the wds proxy url. I think we'd need to pass the proxy Url in from Leo if we wanted to use it.

jordansafer (approved):

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
davidangb: A PR to your PR! #165

service/src/main/resources/application.properties
davidangb: I suggest using the Leo "release name" as the Sentry server name - that should be a unique-per-workspace identifier for the app. Then, add an additional tag (just like you're doing for workspaceId) to hold the MRG name.

calypsomatic: Where do I find the Leo release name?  (Looking for it but if you just know that would help)

davidangb: I thiiiiink it's this one: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala#L415 but I'm not certain. And if it is `fullnameOverride`, you'd still need to pipe it through cromwhelm to WDS.

service/src/main/resources/logback-spring.xml
```diff
@@ -37,6 +43,7 @@
         <root level="info">
             <appender-ref ref="RollingFile" />
             <appender-ref ref="Console" />
+            <appender-ref ref="Sentry" />
```
davidangb, line 46: do you need to add the appender-ref to the `org.databiosphere.workspacedataservice` and `org.springframework.web` loggers too?

davidangb (commented):
one code comment inline.

The other question I have is a design-level one: should we send exceptions to Sentry for all environments? Or just prod? Or maybe just the live envs (dev/alpha/staging/prod) but not bees or local? I know Sentry can help us filter what we send to it, but is there even any point in sending local or bee exceptions?

The follow-on question would be *how* to disable sending for certain environments …

service/src/main/resources/application.properties
calypsomatic: I *think* that gets populated into "app.fullname" which is something like `coa-sxb43e-coa-rls`.  Is that what you had in mind or am I looking at the wrong thing?

davidangb: the `sxb43e` is what I'm thinking of, though the entire `coa-sxb43e-coa-rls` would be just fine. Apparently that's a unique-per-workspace ID for the specific app deployment, so it could both help us find the app AND it could help us find k8s resources for the app.

you're right it is populated into `app.fullname`! But you might wanna make a dedicated env var for it in WDS; the only place we're feeding it in (I think) is as part of the postgres hostname: https://github.com/broadinstitute/cromwhelm/blob/8ff2d1b0b3e194ce5f5db8061ac818807762c30d/terra-batch-libchart/templates/_wds.tpl#L31-L32

calypsomatic:
> one code comment inline.
> 
> The other question I have is a design-level one: should we send exceptions to Sentry for all environments? Or just prod? Or maybe just the live envs (dev/alpha/staging/prod) but not bees or local? I know Sentry can help us filter what we send to it, but is there even any point in sending local or bee exceptions?
> 
> The follow-on question would be _how_ to disable sending for certain environments …

Looks like there's a[ `beforeSend` filter ](https://docs.sentry.io/platforms/javascript/configuration/filtering/#filtering-error-events) which should make it possible to only send for certain environments.  As to whether we should - I wouldn't want to send for local, but I think other live environments might help in tracking down when a problem started.  Not sure about bees.  Another question would be, assuming we're filtering on the env parameter, does that include if it ends up as `unknown`?

davidangb:
@calypsomatic my $.02 would be to send exceptions for dev|alpha|staging|prod, but no other environments - so, not local, not bees. If we can't detect the environment based on the Sam url  - so it is "unknown" - hm, I'd suggest we DO send exceptions for "uknown" because it could indicate a problem in environment-parsing. But we might want to modify that decision in the future.

As for how to implement, a couple other things to check:
* what is the effect of setting the DSN to "" in the environments for which we don't want to send exceptions?
* what is the effect of setting `options.setSampleRate(null);` in the environments for which we don't want to send exceptions?

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
```diff
@@ -0,0 +1,80 @@
...
+	protected String urlToEnv(String samUrl) {
+		if (samUrl == null) {
+			return DEFAULT_ENV;
+		}
+		Matcher matcher = SAM_ENV_PATTERN.matcher(samUrl);
+		boolean found = matcher.find();
+		if (found) {
+			return matcher.group(1);
+		} else if (samUrl.endsWith("bee.envs-terra.bio")) {
+			return samUrl.split("\\.")[1];
```
davidangb, line 75: nit: if we aren't sending anything to Sentry for bees (which I agree with), then there's no need to extract the bee name here - this can also just be "unknown"

calypsomatic (approved):

PR #166, opened 2023-01-23
swagger cleanup

davidangb:
A collection of swagger fixes:
* add `operationId`s for `/version` and `/status`; this eliminates two messages during compile.
* resolve all errors/warnings as reported by IntelliJ

Compilation warnings on this branch vs. main:

Screen Shot 2023-01-23 at 10 37 44 AM

### DIFF
--- a/service/src/main/resources/static/swagger/openapi-docs.yaml
+++ b/service/src/main/resources/static/swagger/openapi-docs.yaml
@@ -32,6 +32,7 @@ paths:
   /status:
     get:
       summary: Gets health status for WDS -- generated via Spring Boot Actuator (see https://docs.spring.io/spring-boot/docs/current/actuator-api/htmlsingle/#health for details)
+      operationId: statusGet
       tags:
         - General WDS Information
       responses:
@@ -40,6 +41,7 @@ paths:
   /version:
     get:
       summary: Gets related git and build version info for WDS -- generated via Spring Boot Actuator (see https://docs.spring.io/spring-boot/docs/current/actuator-api/htmlsingle/#info for details)
+      operationId: versionGet
       tags:
         - General WDS Information
       responses:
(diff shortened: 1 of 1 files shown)


service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -15,7 +15,7 @@ info:
   termsOfService: https://github.com/DataBiosphere/terra-workspace-data-service
   license:
     name: BSD
-    url: http://opensource.org/licenses/BSD-3-Clause
+    url: https://opensource.org/licenses/BSD-3-Clause
```
davidangb, line 18: IntelliJ gives a warning on this about HTTP being insecure, so I changed this

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -32,6 +32,7 @@ paths:
   /status:
     get:
       summary: Gets health status for WDS -- generated via Spring Boot Actuator (see https://docs.spring.io/spring-boot/docs/current/actuator-api/htmlsingle/#health for details)
+      operationId: statusGet
```
davidangb, line 35: this `operationId` - and the one for `/version` - were just plain missing. In this PR, I specify the same `operationId` values as what was previously being autogenerated. That way, the Java and Python clients maintain the same function names, but we eliminate the compile warning.

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -87,10 +89,8 @@ paths:
         $ref: '#/components/requestBodies/RecordRequestBody'
       responses:
         200:
-          description: Record updated
```
davidangb, line 90: `description`s all moved into the components section where they belong.

calypsomatic:
What about that MalformedUrl exception?  Any insight into what that is?  Can we fix that too while we're at it?

davidangb:
@calypsomatic the `MalformedURLException` is tricker and I don't have a good solution for it just yet - I'd like to clean that up but don't want to block this PR on it.

calypsomatic (approved):

jordansafer (approved):

PR #170, opened 2023-01-26
AJ-675: list-instances API

davidangb:
new API to list all instances inside a given WDS service.

The UI troubleshooter can use this to determine if the default workspace instance exists.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -81,6 +81,9 @@ public class RecordDao {
 	private final ObjectMapper objectMapper;
 	private final CachedQueryDao cachedQueryDao;
 
+	@Value("${spring.datasource.username}")
+	private String wdsDbUser;
+
 	public RecordDao(NamedParameterJdbcTemplate namedTemplate,
 					 @Qualifier("streamingDs") NamedParameterJdbcTemplate templateForStreaming, DataTypeInferer inf, ObjectMapper objectMapper, CachedQueryDao cachedQueryDao,
 					 @Value("${twds.instance.workspace-id}") String workspaceId) {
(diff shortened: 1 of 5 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -115,6 +118,14 @@ public boolean instanceSchemaExists(UUID instanceId) {
 				new MapSqlParameterSource("workspaceSchema", instanceId.toString()), Boolean.class));
 	}
 
+	public List<UUID> listInstanceSchemas() {
+		List<String> schemas = namedTemplate.getJdbcTemplate()
+				.queryForList("select schema_name from information_schema.schemata " +
+						"where schema_owner = ? order by schema_name",
```
calypsomatic, line 124: Is including the username forward-looking to when it might make a difference what db user is making a call?  Right now it's pretty much always the same one so searching for all schemas is the same is searching for all schemas owned by the wdsdbuser.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -115,6 +118,14 @@ public boolean instanceSchemaExists(UUID instanceId) {
...
+						String.class, wdsDbUser);
+		return schemas.stream().map(UUID::fromString).toList();
+	}
```
calypsomatic, line 127: Just some wild/idle speculation - If there's ever a case that *somehow* a schema ended up in the db that didn't conform to a UUID format, then presumably this `map` function would throw up.  Probably not worth the extra code to check!

davidangb:
@calypsomatic to both of your comments - if I exclude the `where schema_owner = 'wds'` clause, I get this list of schemas:
```
[6d88b93d-e6a8-4a62-8703-33ac9406a717, information_schema, pg_catalog, public]
```

the `information_schema, pg_catalog`, and `public` schemas are owned by the user "postgres".

So - I think we do need the where clause … and, including the where clause prevents any problems with casting the instance name to a UUID.

I agree that at some point in the future if/when we allow non-UUID instance names, this will break; the create- and delete-instance APIs will also break! So I'm not to worried about it; we'll revamp at that time.

aaronkanzer (approved):

calypsomatic (approved):

PR #172, opened 2023-01-31
AJ-734 Update arrays of relations

calypsomatic:
See [AJ-734](https://broadworkbench.atlassian.net/browse/AJ-734)
Adds support to update array-of-relations attributes.  Current behavior fails on updating attribute datatype of relations, so this is reinforced with arrays of relations and given clearer messaging.

[AJ-734]: https://broadworkbench.atlassian.net/browse/AJ-734?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -350,6 +350,11 @@ public class RecordDao {
 		}
 	}
 
+	public void removeFromJoin(UUID instanceId, Relation column, RecordType fromType, List<String> recordIds){
+			namedTemplate.update("delete from " + getQualifiedJoinTableName(instanceId, column.relationColName(), fromType) + " where "
+					+ quote(getFromColumnName(fromType)) + "in (:recordIds)", new MapSqlParameterSource("recordIds", recordIds));
+	}
+
 	public void batchUpsert(UUID instanceId, RecordType recordType, List<Record> records,
 							Map<String, DataTypeMapping> schema){
 		batchUpsert(instanceId, recordType, records, schema, cachedQueryDao.getPrimaryKeyColumn(recordType, instanceId));
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
@@ -78,6 +78,11 @@ public class BatchWriteService {
 		for (Map.Entry<String, MapDifference.ValueDifference<DataTypeMapping>> entry : differenceMap.entrySet()) {
 			String column = entry.getKey();
 			MapDifference.ValueDifference<DataTypeMapping> valueDifference = entry.getValue();
+			//Don't allow updating relation columns
+			if (valueDifference.leftValue() == DataTypeMapping.ARRAY_OF_RELATION ||
+					valueDifference.leftValue() == DataTypeMapping.RELATION ){
+				throw new InvalidRelationException("Unable to update a relation or array of relation attribute type");
+			}
 			DataTypeMapping updatedColType = inferer.selectBestType(valueDifference.leftValue(),
 					valueDifference.rightValue());
 			recordDao.changeColumn(instanceId, recordType, column, updatedColType);
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
@@ -38,6 +38,8 @@ public class RecordService {
         Map<Relation, List<RelationValue>> relationArrayValues = getAllRelationArrayValues(records, relationArrays);
         recordDao.batchUpsert(instanceId, recordType, records, requestSchema, primaryKey);
         for (Map.Entry<Relation, List<RelationValue>> rel : relationArrayValues.entrySet()) {
+            //remove existing values from join table and replace with new ones
+            recordDao.removeFromJoin(instanceId, rel.getKey(), recordType, rel.getValue().stream().map(relVal -> relVal.fromRecord().getId()).toList());
             recordDao.insertIntoJoin(instanceId, rel.getKey(), recordType, rel.getValue());
         }
     }
--- a/service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
@@ -2,10 +2,9 @@ package org.databiosphere.workspacedataservice.controller;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.databiosphere.workspacedataservice.TestUtils;
+import org.databiosphere.workspacedataservice.dao.RecordDao;
 import org.databiosphere.workspacedataservice.service.RelationUtils;
-import org.databiosphere.workspacedataservice.service.model.AttributeSchema;
-import org.databiosphere.workspacedataservice.service.model.DataTypeMapping;
-import org.databiosphere.workspacedataservice.service.model.RecordTypeSchema;
+import org.databiosphere.workspacedataservice.service.model.*;
 import org.databiosphere.workspacedataservice.service.model.exception.InvalidNameException;
 import org.databiosphere.workspacedataservice.service.model.exception.InvalidRelationException;
 import org.databiosphere.workspacedataservice.service.model.exception.MissingObjectException;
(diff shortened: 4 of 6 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -78,6 +78,11 @@ public Map<String, DataTypeMapping> addOrUpdateColumnIfNeeded(UUID instanceId, R
 		for (Map.Entry<String, MapDifference.ValueDifference<DataTypeMapping>> entry : differenceMap.entrySet()) {
 			String column = entry.getKey();
 			MapDifference.ValueDifference<DataTypeMapping> valueDifference = entry.getValue();
+			//Don't allow updating relation columns
+			if (valueDifference.leftValue() == DataTypeMapping.ARRAY_OF_RELATION ||
+					valueDifference.leftValue() == DataTypeMapping.RELATION ){
+				throw new InvalidRelationException("Unable to update a relation or array of relation attribute type");
```
calypsomatic, line 84: If the column is already a relation or array of relations, refuse to change it to something else.  Would welcome better wording.
If the column is being changed TO a relation or array of relation, it already fails (before this PR) with "It looks like you're attempting to assign a relation to an existing attribute that was not configured for relations".

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -1330,4 +1332,166 @@ void datetimeAttributeShouldBeHumanReadable() throws Exception {
...
+		List<String> actualAttrValue = assertInstanceOf(List.class, recordResponse.recordAttributes().getAttributeValue("relArrAttr"));
+		assertIterableEquals(relArr, actualAttrValue);
+
+		//Join table should not have been updated
+		List<String> joinVals = getRelationArrayValues(instanceId, "relArrAttr", recordWithRelationArray, recordType);
+		assertIterableEquals(List.of("record_0", "record_1"), joinVals);
+	}
+
+	//TODO: A better way to do this?
+	private List<String> getRelationArrayValues(UUID instanceId, String columnName, Record record, RecordType toRecordType) {
```
calypsomatic, line 1491: I'd really like to check the values in the join table, which requires querying the database directly.  I dealt with this earlier, in the [RecordDaoTest](https://github.com/DataBiosphere/terra-workspace-data-service/blob/ae0bb4122303a07039eddcae9f08396bde3f5bdb/service/src/test/java/org/databiosphere/workspacedataservice/dao/RecordDaoTest.java#L678) by adding a separate template to query the database.  I just repeated that here to get tests going.  Would probably now be worth it to extract these methods out into a test utility, but would also welcome other suggestions for a way (or place) to check join tables in tests.

service/src/test/java/org/databiosphere/workspacedataservice/dao/TestDao.java
```diff
@@ -0,0 +1,39 @@
...
+public class TestDao {
+
+    @Autowired
+    NamedParameterJdbcTemplate namedTemplate;
+
+    @Autowired
+    RecordDao recordDao;
+
+    public boolean joinTableExists(UUID instanceId, String tableName, RecordType referringRecordType) {
+        //This method gives us the name in quotes, need to be able to strip them off
```
calypsomatic, line 24: These methods were in the `RecordDaoTest` class with their own `NamedParameterJdbcTemplate`; since I needed to use them in a separate test class I pulled them out into a new file.  There were a few tests in `RecordDaoTest` that were using the `NamedParameterJdbcTemplate` directly; not sure if they want or need to get refactored as well.

jordansafer (approved):

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -65,6 +58,13 @@ class RecordControllerMockMvcTest {
 
 	private static String versionId = "v0.2";
 
+	@Autowired
+	TestDao testDao;
+
+	@Autowired
+	RecordDao recordDao;
```
aaronkanzer, line 65: I don't think this is being used anywhere in the test file?

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -2,10 +2,10 @@
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.databiosphere.workspacedataservice.TestUtils;
+import org.databiosphere.workspacedataservice.dao.RecordDao;
```
aaronkanzer, line 5: Relates to https://github.com/DataBiosphere/terra-workspace-data-service/pull/172/files#r1097938479

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -350,6 +350,11 @@ public void insertIntoJoin(UUID instanceId, Relation column, RecordType recordTy
 		}
 	}
 
+	public void removeFromJoin(UUID instanceId, Relation column, RecordType fromType, List<String> recordIds){
+			namedTemplate.update("delete from " + getQualifiedJoinTableName(instanceId, column.relationColName(), fromType) + " where "
```
aaronkanzer, line 354: Is the approach here to delete values from a DB table and then re-add? Rather than using some form of `UPDATE` SQL call?

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
calypsomatic: Huh, yeah, not sure what happened there.  Will remove.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
calypsomatic: Right, this is just to avoid the extra work of calculating the difference between existing values and what they should be.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
```diff
@@ -38,6 +38,8 @@ public void prepareAndUpsert(UUID instanceId, RecordType recordType, List<Record
         Map<Relation, List<RelationValue>> relationArrayValues = getAllRelationArrayValues(records, relationArrays);
         recordDao.batchUpsert(instanceId, recordType, records, requestSchema, primaryKey);
         for (Map.Entry<Relation, List<RelationValue>> rel : relationArrayValues.entrySet()) {
+            //remove existing values from join table and replace with new ones
+            recordDao.removeFromJoin(instanceId, rel.getKey(), recordType, rel.getValue().stream().map(relVal -> relVal.fromRecord().getId()).toList());
```
aaronkanzer, line 42: Is this call necessary if the table is empty though? Rather this is the first time data is inserted?

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
aaronkanzer: I think something like the following might reduce the amount of writes you'd need to do to and from the DB via a `DELETE` call:

```
SELECT <some values>
FROM <array-of-relations-table>
WHERE value NOT IN ( SELECT value FROM <array-of-relations-table> WHERE <value> = <something-else>)
```

Let me know though if you think this might be much more arduous -- I think the efficiency might be worth a refactor.

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
aaronkanzer: Not sure about wording -- but wondering if you could reference some args from the code into the exception?

Therefore if two of these exceptions happened, we could differentiate the two?

calypsomatic: I think we've been avoiding putting names of attributes or other identifying information into the error messages for security reasons.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
calypsomatic: I don't think I follow where this is supposed to go or what call it's supposed to replace.

aaronkanzer: Sorry, I meant that something like this could be used to determine the difference, then alter the rows, instead of a full `DELETE` then a re-insertion of some of those rows.

My thought was to have this happen in the code that this comment is referencing: https://github.com/DataBiosphere/terra-workspace-data-service/pull/172#discussion_r1097950465

Don't want to nitpick too much, so if you think that it is fine as, then can ✅ !

aaronkanzer: that makes sense -- will approve then!

PR #176, opened 2023-02-09
AJ-678 File datatype in WDS

calypsomatic:
See [AJ-678](https://broadworkbench.atlassian.net/browse/AJ-678)
Adds a FILE type to record attribute data types.  Only recognizes azure storage container (https://[].blob.core.windows.net/[]) or drs (drs://[]) files.

[AJ-678]: https://broadworkbench.atlassian.net/browse/AJ-678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -304,21 +304,11 @@ public class RecordController {
 		relationCols.addAll(recordDao.getRelationCols(instanceId, recordType));
 		Map<String, RecordType> relations = relationCols.stream()
 				.collect(Collectors.toMap(Relation::relationColName, Relation::relationRecordType));
-//		List<AttributeSchema> attrSchema = schema.entrySet().stream().sorted(Map.Entry.comparingByKey())
-//				.map(entry -> createAttributeSchema(entry.getKey(), entry.getValue(), relations.get(entry.getKey())))
-//				.toList();
 		List<AttributeSchema> attrSchema = schema.entrySet().stream().sorted(Map.Entry.comparingByKey())
 				.map(entry -> new AttributeSchema(entry.getKey(), entry.getValue().toString(), relations.get(entry.getKey())))
 				.toList();
 		int recordCount = recordDao.countRecords(instanceId, recordType);
 		return new RecordTypeSchema(recordType, attrSchema, recordCount, recordDao.getPrimaryKeyColumn(recordType, instanceId));
-	}
-
-	private AttributeSchema createAttributeSchema(String name, DataTypeMapping datatype, RecordType relation) {
-		if (relation == null) {
-			return new AttributeSchema(name, datatype.toString(), null);
-		}
-		return new AttributeSchema(name, "STRING".equals(datatype.toString()) ? "RELATION" : "ARRAY_OF_RELATION", relation);
 	}
 
 	private static void validateVersion(String version) {
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -304,8 +304,11 @@ public class RecordController {
 		relationCols.addAll(recordDao.getRelationCols(instanceId, recordType));
 		Map<String, RecordType> relations = relationCols.stream()
 				.collect(Collectors.toMap(Relation::relationColName, Relation::relationRecordType));
+//		List<AttributeSchema> attrSchema = schema.entrySet().stream().sorted(Map.Entry.comparingByKey())
+//				.map(entry -> createAttributeSchema(entry.getKey(), entry.getValue(), relations.get(entry.getKey())))
+//				.toList();
 		List<AttributeSchema> attrSchema = schema.entrySet().stream().sorted(Map.Entry.comparingByKey())
-				.map(entry -> createAttributeSchema(entry.getKey(), entry.getValue(), relations.get(entry.getKey())))
+				.map(entry -> new AttributeSchema(entry.getKey(), entry.getValue().toString(), relations.get(entry.getKey())))
 				.toList();
 		int recordCount = recordDao.countRecords(instanceId, recordType);
 		return new RecordTypeSchema(recordType, attrSchema, recordCount, recordDao.getPrimaryKeyColumn(recordType, instanceId));
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -549,7 +549,7 @@ public class RecordDao {
 
 	private Object[] getListAsArray(List<?> attVal, DataTypeMapping typeMapping) {
 		switch (typeMapping){
-			case ARRAY_OF_STRING, ARRAY_OF_RELATION, ARRAY_OF_DATE, ARRAY_OF_DATE_TIME, ARRAY_OF_NUMBER, EMPTY_ARRAY:
+			case ARRAY_OF_STRING, ARRAY_OF_FILE, ARRAY_OF_RELATION, ARRAY_OF_DATE, ARRAY_OF_DATE_TIME, ARRAY_OF_NUMBER, EMPTY_ARRAY:
 				return attVal.stream().map(Object::toString).toList().toArray(new String[0]);
 			case ARRAY_OF_BOOLEAN:
 				//accept all casings of True and False if they're strings
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -248,7 +248,7 @@ public class RecordDao {
 	public Map<String, DataTypeMapping> getExistingTableSchema(UUID instanceId, RecordType recordType) {
 		MapSqlParameterSource params = new MapSqlParameterSource(INSTANCE_ID, instanceId.toString());
 		params.addValue("tableName", recordType.getName());
-		String sql = "select column_name, udt_name::regtype as data_type from INFORMATION_SCHEMA.COLUMNS " +
+		String sql = "select column_name,coalesce(domain_name, udt_name::regtype::varchar) as data_type from INFORMATION_SCHEMA.COLUMNS " +
 				"where table_schema = :instanceId and table_name = :tableName";
 		return getTableSchema(sql, params);
 	}
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
@@ -103,10 +103,6 @@ public class DataTypeInferer {
 		// when we load from TSV, numbers are converted to strings, we need to go back
 		// to numbers
 		String sVal = val.toString();
-		if (isFileType(sVal)){
-			return FILE;
-		}
-
 
 		if (isNumericValue(sVal)) {
 			return NUMBER;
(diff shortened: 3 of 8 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -305,19 +305,12 @@ private RecordTypeSchema getSchemaDescription(UUID instanceId, RecordType record
...
 				.collect(Collectors.toMap(Relation::relationColName, Relation::relationRecordType));
 		List<AttributeSchema> attrSchema = schema.entrySet().stream().sorted(Map.Entry.comparingByKey())
-				.map(entry -> createAttributeSchema(entry.getKey(), entry.getValue(), relations.get(entry.getKey())))
+				.map(entry -> new AttributeSchema(entry.getKey(), entry.getValue().toString(), relations.get(entry.getKey())))
 				.toList();
 		int recordCount = recordDao.countRecords(instanceId, recordType);
 		return new RecordTypeSchema(recordType, attrSchema, recordCount, recordDao.getPrimaryKeyColumn(recordType, instanceId));
 	}
 
-	private AttributeSchema createAttributeSchema(String name, DataTypeMapping datatype, RecordType relation) {
```
calypsomatic, line 314: Turns out that https://broadworkbench.atlassian.net/browse/AJ-835 was actually caused by a failure of the record dao to appropriately find the data type of a column. When that was fixed, we no longer needed this extra method to extract the relation.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -248,7 +248,7 @@ public List<String> getAllAttributeNames(UUID instanceId, RecordType recordType)
 	public Map<String, DataTypeMapping> getExistingTableSchema(UUID instanceId, RecordType recordType) {
 		MapSqlParameterSource params = new MapSqlParameterSource(INSTANCE_ID, instanceId.toString());
 		params.addValue("tableName", recordType.getName());
-		String sql = "select column_name, udt_name::regtype as data_type from INFORMATION_SCHEMA.COLUMNS " +
+		String sql = "select column_name,coalesce(domain_name, udt_name::regtype::varchar) as data_type from INFORMATION_SCHEMA.COLUMNS " +
```
calypsomatic, line 251: See above comment: when `getTableSchema` was split into `getExistingTableSchema` and `getExistingTableSchemaLessPrimaryKey`, the `coalesce` was only carried over to one of them - likely a result of concurrent work.  Adding it in here lets `describeRecordType` return the correct result without extra work.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -335,4 +346,11 @@ public RelationCollection findRelations(List<Record> records, Map<String, DataTy
 		}
 		return new RelationCollection(relations, relationArrays);
 	}
+
+	private boolean isFileType(String possibleFile){
+		//https://[].blob.core.windows.net/sc-[] or drs://[]
+		Pattern AZURE_FILE_PATTERN = Pattern.compile("https://[a-z0-9]*.blob.core.windows.net/sc-[a-z0-9-]*", Pattern.CASE_INSENSITIVE);
```
aaronkanzer, line 352: Is the `sc-*` guaranteed? 

At least in our TSV example in Terra, we have a link such as `https://sa226344b664da26ad6863.blob.core.windows.net/file1.cram`

Perhaps `blob.core.windows.net` could be sufficient?

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -680,11 +680,14 @@ components:
...
           "relationAttr": "terra-wds:/target-type/target-id",
+          "fileAttr": "https://account_name.blob.core.windows.net/container-1/blob1",
           "arrayString": ["green", "red"],
           "arrayBoolean": [true, false],
           "arrayNumber": [12821.112, 0.12121211, 11],
           "arrayDate": ["2022-11-03"],
-          "arrayDateTime": ["2022-11-03T04:36:20"]
+          "arrayDateTime": ["2022-11-03T04:36:20"],
+          "arrayRelation": ["terra-wds:/target-type/target-id-1", "terra-wds:/target-type/target-id-2"],
+          "arrayFile": ["drs://drs.example.org/file_id_1", "https://account_name.blob.core.windows.net/container-2/blob2"]
```
aaronkanzer, line 690: Similar to this comment - https://github.com/DataBiosphere/terra-workspace-data-service/pull/176/files#r1101856663

The Azure file example here does not container the `sc-*` -- if we keep the `sc-*` should it reflect in this test data?

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -102,6 +103,10 @@ public DataTypeMapping inferTypeForTsvSource(Object val) {
 		// when we load from TSV, numbers are converted to strings, we need to go back
 		// to numbers
 		String sVal = val.toString();
+		if (isFileType(sVal)){
```
aaronkanzer, line 106: Is this logic repetitive per logic in `getTypeMappingFromString`?

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -680,11 +680,14 @@ components:
           "numericAttr": 123,
           "booleanAttr": true,
           "relationAttr": "terra-wds:/target-type/target-id",
+          "fileAttr": "https://account_name.blob.core.windows.net/container-1/blob1",
```
aaronkanzer, line 683: Similar to https://github.com/DataBiosphere/terra-workspace-data-service/pull/176/files#r1101858179

service/src/main/resources/static/swagger/openapi-docs.yaml
calypsomatic: Good point - when I was implementing the pattern I was going off what was written in the ticket, but for the swagger I wanted to be as absolutely general as possible so I actually went and checked the azure storage container site to see what to generally expect.  Unless somebody (@bethsheets or @davidangb ) knows any better, I'll change the pattern to be more general.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
calypsomatic: Sure looks that way!  Good catch!

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -335,4 +342,11 @@ public RelationCollection findRelations(List<Record> records, Map<String, DataTy
 		}
 		return new RelationCollection(relations, relationArrays);
 	}
+
+	private boolean isFileType(String possibleFile){
+		//https://[].blob.core.windows.net/[] or drs://[]
```
aaronkanzer, line 347: Do we need to take into account `dos://` prefixes too for Azure workspaces? 

Seems this exists in Terra UI -- wasn't sure if this is GCP specific though

Cc @bethsheets

davidangb: `dos://` is the precursor to `drs://` so it seems safe to support - though if we are actively trying to _retire` DOS URIs then we should leave it out. Agree this is a Beth question!

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -335,4 +342,11 @@ public RelationCollection findRelations(List<Record> records, Map<String, DataTy
...
+		Pattern AZURE_FILE_PATTERN = Pattern.compile("https://[a-z0-9]*.blob.core.windows.net/[a-z0-9.\\/\\-_]*", Pattern.CASE_INSENSITIVE);
+		Pattern DRS_FILE_PATTERN = Pattern.compile("drs://[a-z0-9.\\/\\-_]*", Pattern.CASE_INSENSITIVE);
+		return AZURE_FILE_PATTERN.matcher(possibleFile).matches() || DRS_FILE_PATTERN.matcher(possibleFile).matches();
```
davidangb, line 350: a few different thoughts here:
1. Move the `AZURE_FILE_PATTERN` and `DRS_FILE_PATTERN` variables to the class level. That way, we're only compiling the regex (which is relatively expensive) once at class-instantiation time, instead of doing it for every call to `isFileType`.
2. I don't think it's worth trying to match the tail end of the url. As-is, the pattern would fail to match if the url included e.g. "?foo=bar" or escape characters like "my%20file.pdf". If we're sticking with regex, look at the `^` metacharacter to match the beginning of a string - that should simplify the pattern.
3. Maybe regex isn't the right solution at all? We could use URL parsing instead, something like:

```java
URL fileUrl;
		try {
			fileUrl = new URL(possibleFile);
		} catch (MalformedURLException mue) {
			return false;
		}

		return fileUrl.getProtocol().equalsIgnoreCase("drs") ||
				(fileUrl.getProtocol().equalsIgnoreCase("https") && fileUrl.getHost().toLowerCase().endsWith(".blob.core.windows.net"));
```

on the other hand, if we stuck with regex we could potentially move the regex patterns to application.properties to make it easy to change in the future …

service/src/main/resources/static/swagger/openapi-docs.yaml
davidangb: I agree on leaving it pretty generic here

service/src/test/java/org/databiosphere/workspacedataservice/DataTypeInfererTest.java
```diff
@@ -129,6 +133,7 @@ void inferSomeTypes() {
 		assertThat(inferer.inferType(List.of(RelationUtils.createRelationString(RecordType.valueOf("testType"),"recordId"), RelationUtils.createRelationString(RecordType.valueOf("testType"),"recordId2"), RelationUtils.createRelationString(RecordType.valueOf("testType"),"recordId3")), InBoundDataSource.JSON)).isEqualTo(DataTypeMapping.ARRAY_OF_RELATION);
 		assertThat(inferer.inferType(List.of(RelationUtils.createRelationString(RecordType.valueOf("testType"),"recordId"), "not a relation string", RelationUtils.createRelationString(RecordType.valueOf("testType"),"recordId3")), InBoundDataSource.JSON)).isEqualTo(DataTypeMapping.ARRAY_OF_STRING);
 		assertThat(inferer.inferType(RelationUtils.createRelationString(RecordType.valueOf("testType"),"recordId3"), InBoundDataSource.JSON)).isEqualTo(DataTypeMapping.RELATION);
+		assertThat(inferer.inferType("https://lz1a2b345c67def8a91234bc.blob.core.windows.net/sc-7ad51c5d-eb4c-4685-bffe-62b861f7753f/file.pdf", InBoundDataSource.JSON)).isEqualTo(DataTypeMapping.FILE);
```
davidangb, line 136: following my other comment about the regex patterns, try adding some query params and url-escapes to some of the test cases.

davidangb (commented):
a couple comments inline on the implementation. I admit I'm still having a hard time with the requirements of this ticket - I will raise those outside of this PR thread.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
aaronkanzer: @bethsheets we need your wisdom!

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
calypsomatic: I think I started out using URI, ran into some tiny problem, and so switched to regex.  URL doesn't work because "drs" is not a recognized protocol.  I switched back to using URI.

service/src/test/java/org/databiosphere/workspacedataservice/DataTypeInfererTest.java
calypsomatic: Done!

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -335,4 +343,20 @@ public RelationCollection findRelations(List<Record> records, Map<String, DataTy
 		}
 		return new RelationCollection(relations, relationArrays);
 	}
+
+	private boolean isFileType(String possibleFile){
+		URI fileUri;
+		try {
+			fileUri = new URI(possibleFile);
+			//Many non-URI strings will parse without exception but have no scheme or host
+			if (fileUri.getScheme() == null || fileUri.getHost() == null){
```
jordansafer, line 352: minor nit, could simplify to:
```java
return fileUri.getScheme() != null && fileUri.getHost() != null;
```

jordansafer (approved):

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
calypsomatic: Well wouldn't want to return at this point since we still need to check what the scheme and host are.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
bethsheets: I have never heard of DOS, and if we are actively trying to retire it and its not a GA4GH standard, I doubt it would ever be used when we eventually ingest Azure hosted data into TDR that is handed off to WDS. I say we don't include dos://.

PR #177, opened 2023-02-13
[AJ-497] refactor transactions

jordansafer:
Changes:
- business logic is migrated from the RecordController to the RecordOrchestratorService
- transactions are moved down into the business logic service
- testing for recordController business logic

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -1,25 +1,16 @@
 package org.databiosphere.workspacedataservice.controller;
 
-import bio.terra.common.db.ReadTransaction;
 import bio.terra.common.db.WriteTransaction;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import org.apache.commons.csv.CSVPrinter;
 import org.databiosphere.workspacedataservice.dao.RecordDao;
-import org.databiosphere.workspacedataservice.service.*;
-import org.databiosphere.workspacedataservice.service.model.*;
-import org.databiosphere.workspacedataservice.service.model.exception.MissingObjectException;
-import org.databiosphere.workspacedataservice.service.model.exception.NewPrimaryKeyException;
+import org.databiosphere.workspacedataservice.service.RecordOrchestratorService;
+import org.databiosphere.workspacedataservice.service.model.RecordTypeSchema;
 import org.databiosphere.workspacedataservice.shared.model.BatchResponse;
-import org.databiosphere.workspacedataservice.shared.model.Record;
-import org.databiosphere.workspacedataservice.shared.model.RecordAttributes;
 import org.databiosphere.workspacedataservice.shared.model.RecordQueryResponse;
 import org.databiosphere.workspacedataservice.shared.model.RecordRequest;
 import org.databiosphere.workspacedataservice.shared.model.RecordResponse;
 import org.databiosphere.workspacedataservice.shared.model.RecordType;
 import org.databiosphere.workspacedataservice.shared.model.SearchRequest;
 import org.databiosphere.workspacedataservice.shared.model.TsvUploadResponse;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
@@ -39,11 +30,9 @@ import org.springframework.web.servlet.mvc.method.annotation.StreamingResponseBo
 
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.io.OutputStreamWriter;
-import java.util.*;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
+import java.util.List;
+import java.util.Optional;
+import java.util.UUID;
 
 @RestController
 public class RecordController {
@@ -57,7 +46,6 @@ public class RecordController {
 	}
 
 	@PatchMapping("/{instanceId}/records/{version}/{recordType}/{recordId}")
-	@WriteTransaction
 	public ResponseEntity<RecordResponse> updateSingleRecord(@PathVariable("instanceId") UUID instanceId,
 			@PathVariable("version") String version, @PathVariable("recordType") RecordType recordType,
 			@PathVariable("recordId") String recordId, @RequestBody RecordRequest recordRequest) {
@@ -67,7 +55,6 @@ public class RecordController {
 	}
 
 	@GetMapping("/{instanceId}/records/{version}/{recordType}/{recordId}")
-	@ReadTransaction
 	public ResponseEntity<RecordResponse> getSingleRecord(@PathVariable("instanceId") UUID instanceId,
 			@PathVariable("version") String version, @PathVariable("recordType") RecordType recordType,
 			@PathVariable("recordId") String recordId) {
@@ -87,7 +74,6 @@ public class RecordController {
 	}
 
 	@GetMapping("/{instanceId}/tsv/{version}/{recordType}")
-	// TODO: enable read transaction
 	public ResponseEntity<StreamingResponseBody> streamAllEntities(@PathVariable("instanceId") UUID instanceId,
 			@PathVariable("version") String version, @PathVariable("recordType") RecordType recordType) {
 		StreamingResponseBody responseBody =
@@ -98,7 +84,6 @@ public class RecordController {
 	}
 
 	@PostMapping("/{instanceid}/search/{version}/{recordType}")
-	@ReadTransaction
 	public RecordQueryResponse queryForRecords(@PathVariable("instanceid") UUID instanceId,
 			@PathVariable("recordType") RecordType recordType,
 			@PathVariable("version") String version,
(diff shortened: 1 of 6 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -0,0 +1,286 @@
...
+        int recordCount = recordDao.countRecords(instanceId, recordType);
+        return new RecordTypeSchema(recordType, attrSchema, recordCount, recordDao.getPrimaryKeyColumn(recordType, instanceId));
+    }
+
+    private AttributeSchema createAttributeSchema(String name, DataTypeMapping datatype, RecordType relation) {
+        if (relation == null) {
+            return new AttributeSchema(name, datatype.toString(), null);
+        }
+        return new AttributeSchema(name, "STRING".equals(datatype.toString()) ? "RELATION" : "ARRAY_OF_RELATION", relation);
+    }
```
davidangb, line 285: `createAttributeSchema` is unused, and isn't present in the `main` branch - delete it? Where did it come from? Maybe an artifact of rebasing?

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -120,130 +68,50 @@ public ResponseEntity<TsvUploadResponse> tsvUpload(@PathVariable("instanceId") U
...
 	}
 
 	@GetMapping("/instances/{version}")
-	@ReadTransaction
 	public ResponseEntity<List<UUID>> listInstances(@PathVariable("version") String version) {
-		validateVersion(version);
+		RecordOrchestratorService.validateVersion(version);
 		List<UUID> schemaList = recordDao.listInstanceSchemas();
 		return new ResponseEntity<>(schemaList, HttpStatus.OK);
 	}
```
davidangb, line 108: `listInstances`, `createInstance`, and `deleteInstance` are the last three routes whose implementation doesn't call out to a service. Could you adjust these three as well, to keep things clean? Then there'd be no transactions at all in `RecordController`. If you don't want to put the implementations in `RecordOrchestrationService` you could put them in a new `InstanceService`; that might be cleaner anyway.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -120,130 +68,50 @@ public ResponseEntity<TsvUploadResponse> tsvUpload(@PathVariable("instanceId") U
 			   @PathVariable("version") String version, @PathVariable("recordType") RecordType recordType,
 			   @RequestParam(name= "primaryKey", required = false) Optional<String> primaryKey,
                @RequestParam("records") MultipartFile records) throws IOException {
-		validateVersion(version);
-		validateInstance(instanceId);
-		if(recordDao.recordTypeExists(instanceId, recordType)){
-			validatePrimaryKey(instanceId, recordType, primaryKey);
-		}
-		int recordsModified = batchWriteService.batchWriteTsvStream(records.getInputStream(), instanceId, recordType, primaryKey);
+		int recordsModified = recordOrchestratorService.tsvUpload(instanceId, version, recordType, primaryKey, records);
```
davidangb, line 71: nit: move the `// N.B. transaction ..` comment at line 66 into `RecordOrchestratorService.tsvUpload` instead

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -0,0 +1,286 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@Service
+public class RecordOrchestratorService { // TODO give me a better name
```
davidangb, line 44: > // TODO give me a better name

should all this code merge into `RecordService`? Would that end up being too big?

davidangb (commented):
a couple small comments inline, otherwise looking good and thank you so much for these changes.

You mentioned you couldn't really make transactions smaller - I think that by excluding from the transaction the lines of code that write the API responses to the caller, you've eliminated the biggest worry - those writes can hang/be slow and so that's a big win.

I admit I didn't review RecordOrchestratorService too deeply; I assume most of it is cut-and-paste.

I could imagine work in the future to move some of the unit tests out of `RecordControllerMockMvcTest` and down to the `RecordOrchestratorService` level.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
jordansafer: Oops, yes. An artifact of rebasing now removed!

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
jordansafer: Agreed, and moving them down to an `InstanceService` to keep things cleaner/separated by function

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
jordansafer: It will create a circular dependency because this code makes calls to BatchWriteService and BatchWriteService makes calls to RecordService. We could change the name of RecordService, or shift the existing business logic in RecordService up to BatchWriteService or down into the DAOs. Ultimately, I think we do want this layer to handle the business logic above BatchWriteService, although I find the naming a bit confusing.

service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
```diff
@@ -0,0 +1,50 @@
...
+@Service
+public class InstanceService {
+
+    private final RecordDao recordDao;
+
+    public InstanceService(RecordDao recordDao) {
+        this.recordDao = recordDao;
+    }
+
+    public List<UUID> listInstances(String version) {
```
davidangb, line 24: nit: give this a `@ReadTransaction` (which, admittedly, was missing from the API layer)

service/src/test/java/org/databiosphere/workspacedataservice/service/InstanceServiceTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Test
+    void testCreateAndValidateInstance() {
+        instanceService.createInstance(INSTANCE, VERSION);
+        instanceService.validateInstance(INSTANCE);
+
+        try {
+            instanceService.validateInstance(UUID.fromString("000e4444-e22b-22d1-a333-426614174000"));
+            // Test should not reach this line
+            assert(false);
```
davidangb, line 42: nit: instead of `assert(false)`, you could do: `fail("validateInstance should have thrown an error");` (here and in the `deleteInstance` test)

davidangb (approved):
two super small comments, giving this a proactive :+1:

service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
jordansafer: My understanding is that we don't need a `@ReadTransaction` for this API because there's only a single SQL request. In the other read requests, we have 2+ transactions, for example: 1 read to check that the instance exists, 2nd read to describe a record type from the instance. In this case we only list instances, so I don't think we benefit from a transaction.

service/src/test/java/org/databiosphere/workspacedataservice/service/InstanceServiceTest.java
jordansafer: Will do!!

service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
davidangb: That's a fair point. I am thinking more from the developer standpoint, the presence of `@ReadTransaction` makes it clear to anyone reading the code that this method contains a read-only database query. I'm fine either way - your call on whether or not to change anything.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordUtils.java
```diff
@@ -0,0 +1,14 @@
+package org.databiosphere.workspacedataservice.service;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.server.ResponseStatusException;
+
+public class RecordUtils {
+    public static final String VERSION = "v0.2";
+
+    public static void validateVersion(String version) {
```
calypsomatic, line 9: Just trying to think through how this will work with different versions - will we have a separate `RecordUtils` class per version as well as separate `RecordOrchestratorService` classes that each import from a different `RecordUtils` class?

service/src/test/java/org/databiosphere/workspacedataservice/service/InstanceServiceTest.java
calypsomatic: You could also use an `assertThrows` for all your expected errors, such as is used [here](https://github.com/DataBiosphere/terra-workspace-data-service/blob/e78fe0a6e28afd021b9feae26eae785a178b9892/service/src/test/java/org/databiosphere/workspacedataservice/dao/RecordDaoTest.java#L380)

calypsomatic (approved):

PR #181, opened 2023-02-27
AJ-870 - update python version from 3.7 to 3.8 for swagger client generation

yuliadub:
Ran through tests with generating the wds client using python 3.8 and didnt see any issues with any of the python versions. 

I could not replicate the behavior that I saw with the client not loading in python 3.10, not sure if when I originally saw it earlier it was some sort of fluke or the changes that we made in the swagger definition fixed something.


davidangb (approved):
code change looks good; did not test it myself.

calypsomatic (approved):

PR #182, opened 2023-03-03
AJ 813 - Add python client documentation

yuliadub:
Added examples of how python client can be used, along with instructions on how one can generate a full set of docs if needed. 

can use this link to view: https://github.com/DataBiosphere/terra-workspace-data-service/blob/aj-813-python-docs/docs/WDS%20Python%20Client.md

Created a separate task to further explore auto generated client docs and if it would make sense to include them in the repo. Linked this pull request in that task so the example if not lost.


docs/WDS Python Client.md
```diff
@@ -0,0 +1,182 @@
...
+azure_token = cli_token[0].replace('"', '')
+```
+
+### WDS Endpoint
+
+WDS endpoint URL can be aquired from Leo APIs and should soon be available in a enviromental variable in the notebook VM. 
+
+However, if for some reason it is not, or you are looking to write data into a different workspace data table that you also have access to (remember this will only work if the token you generated has access to that workspace), it can be aquired by calling the following code
+
+Replace env with "dev" or "prod", based on where your workspace is running. Workspace Id can be located either on the Data tab by clinking "data table status" or from the following env variable in the notebook: os.environ['WORKSPACE_ID']
```
aaronkanzer, line 45: `clicking` not `clinking` 🥂

docs/swagger-gen-docs/README.md
```diff
@@ -0,0 +1,150 @@
...
+ - [StatusResponse](docs/StatusResponse.md)
+ - [TsvUploadResponse](docs/TsvUploadResponse.md)
+ - [VersionResponse](docs/VersionResponse.md)
+
+
+## Documentation For Authorization
+
+ All endpoints do not require authorization.
+
+## Author
```
aaronkanzer, line 146: Don't forget to give yourself credit! :D

docs/swagger-gen-docs/README.md
```diff
@@ -0,0 +1,150 @@
...
+```python
+from __future__ import print_function
+
+import time
+import wds_client
+from wds_client.rest import ApiException
+from pprint import pprint
+
+# Defining the host is optional and defaults to http://..
+# See configuration.py for a list of all supported configuration parameters.
```
aaronkanzer, line 64: Where might we find the `configuration.py`?

docs/swagger-gen-docs/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+```python
+from __future__ import print_function
+
+import time
+import wds_client
+from wds_client.rest import ApiException
+from pprint import pprint
+
+# Defining the host is optional and defaults to http://..
```
aaronkanzer, line 63: Why might I need to alter this value?

docs/swagger-gen-docs/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+# Enter a context with an instance of the API client
+with wds_client.ApiClient(configuration) as api_client:
+    # Create an instance of the API class
+    api_instance = wds_client.GeneralWDSInformationApi(api_client)
+    
+    try:
+        # Gets health status for WDS -- generated via Spring Boot Actuator (see https://docs.spring.io/spring-boot/docs/current/actuator-api/htmlsingle/#health for details)
+        api_response = api_instance.status_get()
+        pprint(api_response)
```
aaronkanzer, line 79: `print` not `pprint`

docs/swagger-gen-docs/README.md
```diff
@@ -0,0 +1,150 @@
...
+ - [SearchSortDirection](docs/SearchSortDirection.md)
+ - [StackTraceElement](docs/StackTraceElement.md)
+ - [StatusResponse](docs/StatusResponse.md)
+ - [TsvUploadResponse](docs/TsvUploadResponse.md)
+ - [VersionResponse](docs/VersionResponse.md)
+
+
+## Documentation For Authorization
+
+ All endpoints do not require authorization.
```
aaronkanzer, line 144: Is this true?

docs/swagger-gen-docs/README.md
yuliadub: it would be in the actual files that come with the client - only way to look at it would be pip install and then do a pip show of the path where wds_client was installed.

docs/swagger-gen-docs/README.md
yuliadub: this would need to be set to the WDS proxy url, it needs to be set before the client can be used. Since this file is auto generated I am not sure I could add that distinct details in there

docs/swagger-gen-docs/README.md
yuliadub: not sure why its defaulting to that (again, auto generated by "openapi-generator-client generate")  ...have spent most of the day trying to figure it out and nothing :D

docs/swagger-gen-docs/README.md
yuliadub: it is not, but openapi-generator-client thinks so. trying to figure out how to change or exclude this section and no luck so far

docs/WDS Python Client.md
```diff
@@ -0,0 +1,182 @@
...
+# if running outside of Terra notebook VM, will need to set this to the desired workspace Id 
+current_workspaceId = os.environ['WORKSPACE_ID'];
+```
+
+## WDS Available Client APIs
+
+You can check out the swagger generated documentation for all available APIs [**here**](swagger-gen-docs/README.md). Below find a few specific examples for the most common use cases. 
+
+## General WDS Info
+General WDS Info client allows to check the version of WDS that is running in a given workspace. It can also check the status of WDS, as well as its subcomponents.
```
bethsheets: suggest - "General WDS Info client checks the version of WDS that is running"

docs/WDS Python Client.md
```diff
@@ -0,0 +1,182 @@
...
+# this returns a VersionResponse object, to get the build version: response.build.version
+# you can also just simply print the whole response as an str: print(response)
+response = generalInfo_instance.version_get()
+print(response.build.version)
+
+response = generalInfo_instance.status_get()
+print(response.status)
+```
+
+### Records
```
bethsheets, line 104: Should we define "Record" vs "recordtype" vs schema? I am a little confused by the terminology. 

In WDS, a record is a row of a data table, and a recordtype is the name of the data table. Is recordtype is same as schema?

bethsheets (changes requested):
Looks good! I have a few suggestions. Thanks!

docs/WDS Python Client.md
```diff
@@ -0,0 +1,182 @@
...
+
+You will need a user entity that has access to the WDS client you need to reach, you will have this access if you have access to the workspace where data that analysis is being run on is present. 
+
+To authenticate in terminal, using azure cli with a user account 
+
+```
+az login --use-device-code
+```
+
+This will provide a link and a code that one can visit in a browser to log in. Use the same account to authenticate with that you would lob into azure to view the subscription of where your Terra billing is deployed to. If you are unsure if you have an account that has that permission, autenticate in the context of a notebook running inside of Terra.
```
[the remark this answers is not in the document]
yuliadub: that is a great question - I checked and all of you do have direct subscription access, at least in dev. Not sure if that is by design or part of testing, but in theory it can be possible/granted by the owner of the subscription.

docs/WDS Python Client.md
yuliadub: I added a few clarifications in the Records and Shema sections, let me know if that helps!

yuliadub:
@aaronkanzer resolved your comments for now but will make sure to add them to the task I created for the auto gen doc tracking!

docs/WDS Python Client.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+### Auth set up
+
+To authenticate your user entity to access a WDS data table, you must have permission to access the workspace where the data table is stored.
+
+To authenticate in terminal, use azure cli with a user account:
+
+```
+az login --use-device-code
```
aaronkanzer, line 26: Perhaps we should mention for now that `WDS` is only supported for `Azure` workspaces then?

docs/WDS Python Client.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+To upgrade since it will not happen automatically
+```
+pip install wds-client --upgrade
+```
+
+### Auth set up
+
+To authenticate your user entity to access a WDS data table, you must have permission to access the workspace where the data table is stored.
```
aaronkanzer, line 21: Do I have to have any access to the WDS table, or `Reader`, `Writer`, `Owner`, etc?

docs/WDS Python Client.md
```diff
@@ -0,0 +1,190 @@
...
+
+```
+!az login --identity --allow-no-subscriptions
+cli_token = !az account get-access-token | jq .accessToken
+azure_token = cli_token[0].replace('"', '')
+```
+
+### WDS Endpoint
+
+WDS endpoint URL can be acquired from Leo APIs and should soon be available in a environmental variable in the notebook VM.
```
aaronkanzer, line 41: Is Leo a DSP-internal term? I might suggest just explaining what it is if these docs are somewhat external

docs/WDS Python Client.md
```diff
@@ -0,0 +1,190 @@
...
+
+### WDS Endpoint
+
+WDS endpoint URL can be acquired from Leo APIs and should soon be available in a environmental variable in the notebook VM. 
+
+However, if for some reason it is not, or you are looking to write data into a different workspace data table that you also have access to (remember this will only work if the token you generated has access to that workspace), it can be acquired by calling the following code
+
+Replace env with "dev" or "prod", based on where your workspace is running. Workspace Id can be located either on the Data tab by clicking "Data Table Status" or from the following env variable in the notebook: os.environ['WORKSPACE_ID'] 
+```
+def get_wds_url(workspaceId):
```
aaronkanzer, line 47: Just a thought, but should we make this a utility function somewhere in the `wds-client` package?

docs/swagger-gen-docs/README.md
davidangb: why not `pprint`, though? `pprint` would result in much nicer formatting when outputting objects.
https://amalgjose.com/2021/05/09/what-is-the-difference-between-print-and-pprint-in-python/

docs/swagger-gen-docs/README.md
davidangb: this is true as far as our openapi definition goes, and therefore as far as any generated client thinks. All our auth is implemented in the relay listener, above/outside of WDS.

Note: this will change as of #183 and future PRs, but as of this writing it is true

docs/WDS Python Client.md
davidangb: not as of this writing. AJ-666 and AJ-667 will be the first tickets where this becomes true (and has Leo-level multi-user apps as a prerequisite)

docs/WDS Python Client.md
davidangb, line 41: nit: I believe it is "environment" variable, not "environmental" variable

davidangb, line 41: thinking out loud: I am not convinced we should/will be able to add the WDS endpoint URL as an environment variable in the notebook.

The logic to determine which WDS app is the correct one is simple in the most common cases, but also has a lot of edge cases. What if the first WDS app errored out and the user deployed a second one? And so forth.

In this doc, I might sidestep it entirely, and say something generic like "find the WDS app in the list of apps that Leo returns, then take its `proxyUrls.wds` value.`

docs/WDS Python Client.md
```diff
@@ -0,0 +1,190 @@
...
+!az login --identity --allow-no-subscriptions
+cli_token = !az account get-access-token | jq .accessToken
+azure_token = cli_token[0].replace('"', '')
+```
+
+### WDS Endpoint
+
+WDS endpoint URL can be acquired from Leo APIs and should soon be available in a environmental variable in the notebook VM. 
+
+However, if for some reason it is not, or you are looking to write data into a different workspace data table that you also have access to (remember this will only work if the token you generated has access to that workspace), it can be acquired by calling the following code
```
davidangb, line 43: what does "a different workspace data table that you also have access to" mean here? I don't understand this use case.

docs/WDS Python Client.md
```diff
@@ -0,0 +1,190 @@
...
+        return response.text
+    
+    print(f"Successfully retrieved details.")
+
+    return json.loads(response.text)
+
+# the response will return all proxy urls for the workspace
+response = get_wds_url(current_workspaceId)
+# specifically grab the wds one
+wds_url = response[0]['proxyUrls']['wds']
```
davidangb, line 68: as discussed in the previous comment, using `[0]` to grab the first app in the response is not going to be correct in all cases.

docs/WDS Python Client.md
```diff
@@ -0,0 +1,190 @@
...
+print(wds_url)
+```
+
+Once you have bearer token and the WDS endpoint url, you are all set to create the WDS Api Client. Do note that token will expire and will be renewed every 60 minutes. 
+To call WDS, you will always need to provide a version and workspace Id. 
+
+```
+import wds_client
+api_client = wds_client.ApiClient(header_name='Authorization', header_value="Bearer " + azure_token)
+api_client.configuration.host = wds_url
```
davidangb, line 78: thinking out loud: when we get to AJ-879 check here to see if a `api_client.configuration.httpUserAgent ` setting exists (or similar, I don't know the exact name)

docs/WDS Python Client.md
```diff
@@ -0,0 +1,190 @@
...
+# if running outside of Terra notebook VM, will need to set this to the desired workspace Id 
+current_workspaceId = os.environ['WORKSPACE_ID'];
+```
+
+## WDS Available Client APIs
+
+You can check out the swagger generated documentation for all available APIs by following instruction to generate it [**here**](#Generate-Full-Client-Documentation). Below find a few specific examples for the most common use cases. 
+
+## General WDS Info
+General WDS Info client checks the version of WDS that is running.
```
davidangb, line 90: nit:
``` suggestion
General WDS Info client checks the version and current status of WDS.
```

docs/WDS Python Client.md
```diff
@@ -0,0 +1,190 @@
...
+
+```
+generalInfo_instance = wds_client.GeneralWDSInformationApi(api_client)
+
+# this returns a VersionResponse object, to get the build version: response.build.version
+# you can also just simply print the whole response as an str: print(response)
+response = generalInfo_instance.version_get()
+print(response.build.version)
+
```
davidangb, line 99: nit: add code comment e.g.
```suggestion

# this returns the current status of WDS; if WDS is down it indicates where
# the underlying failure may be
```

docs/WDS Python Client.md
```diff
@@ -0,0 +1,190 @@
...
+print(records)
+testType = pd.read_csv(records, sep='\t')
+testType.head()
+```
+
+Example of upload_tsv:
+
+
+```
+tsv_file_name = "TestType_uploaded".tsv";
```
davidangb, line 143: nit: think you have an extra quote in there:
```suggestion
tsv_file_name = "TestType_uploaded.tsv";
```

docs/WDS Python Client.md
```diff
@@ -0,0 +1,190 @@
...
+### Schema
+
+Schema defines the column name and types inside a Data Table. The code below assumes that the workspace has a Data Table with the name "testType" that contains data in it. The code to create the record type and add data to it was covered in Records section.
+
+Example of describe_record_type:
+
+```
+schema_instance = wds_client.SchemaApi(api_client)
+
+# get data specifics for a specific data set vs all
```
davidangb, line 161:
```suggestion
# get data specifics for a specific table
```

docs/WDS Python Client.md
```diff
@@ -0,0 +1,190 @@
...
+ent_types = schema_instance.describe_record_type(current_workspaceId, version, 'testType')
+print ("name:", ent_types.name ,"count:", ent_types.count)
+```
+
+Example of describe_all_record_types:
+
+```
+schema_instance = wds_client.SchemaApi(api_client)
+
+# get workspace data specifics, such as what records exist
```
davidangb, line 171:
```suggestion
# get workspace data specifics, such as what tables exist

```

docs/WDS Python Client.md
```diff
@@ -0,0 +1,190 @@
...
+workspace_ent_type = schema_instance.describe_all_record_types(current_workspaceId, version)
+for t in workspace_ent_type:
+    print ("name:", t.name ,"count:", t.count)
+```
+
+### Instances
+
+Currently WDS only supports a single WDS instance by workspace, however in the future there will be support for multiple instances per workspace and these functions will be helpful on checking the state of running WDS instances.
```
davidangb, line 179:
```suggestion
Currently the Terra UI only supports a single WDS instance per workspace, however in the future there will be support for multiple instances per workspace and these functions will be helpful on checking the state of running WDS instances. 
```

docs/swagger-gen-docs/README.md
aaronkanzer: Wow ha I didn't know of `pprint`, but that is awesome! I'd prefer to use that for sure

docs/WDS Python Client.md
yuliadub: great point! added this right at the top

docs/swagger-gen-docs/README.md
yuliadub: wow I also didnt know! Thanks for the context David!

docs/WDS Python Client.md
yuliadub: after chatting more, removing this for now as it is more of an "advanced used scenario"

docs/WDS Python Client.md
yuliadub: renamed and added a link to the repo that has a link to swagger - let me know if this looked better!

docs/WDS Python Client.md
yuliadub: after chatting with David more, took away this for now - only leaving instructions around authenticating as notebook identity

docs/WDS Python Client.md
yuliadub: let me know if this is what you were thinking?

docs/WDS Python Client.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+To upgrade since it will not happen automatically
+```
+pip install wds-client --upgrade
+```
+
+### Auth set up
+
+To authenticate inside a notebook in Terra, leveraging the managed identity the notebook is running under the following commands need to be executed. These commands will acquire a token that will be used by the client to connect to WDS.
```
jordansafer, line 23: small nit! Can the first sentence be a little simpler or 2 sentences? I don't totally follow, I think the idea is we need to run these commands from the notebook to get the managed identity?

jordansafer (approved):

PR #183, opened 2023-03-06
AJ-666: Sam integration for create/delete instance

davidangb:
Enable Sam integration for the create-instance and delete-instance APIs.

Highlights of this PR:
* adds an optional `workspaceId` query parameter to the create-instance API. When omitted, assumes that the workspaceId and its child WDS instance have the same id.
* adds [Bearer authentication](https://swagger.io/docs/specification/authentication/bearer-authentication/) to the swagger definitions for create- and delete-instance.
* when creating a WDS instance:
  * check if the current user has write permission on the parent workspace. If so:
  * create a `wds-instance` Sam resource to represent permissions for this instance
* when deleting a WDS instance:
  * check if the current user has delete permission on that instance. If so:
  * delete the `wds-instance` Sam resource
* introduces the Sam client as a dependency
* makes heavy use of Mockito in unit tests

See also these PRs in Sam:
* https://github.com/broadinstitute/sam/pull/886
* https://github.com/broadinstitute/sam/pull/1013

I've reviewed the 11 Sonar code smells and I disagree with all of them 🤷 

TODOs:
- [x] remove Sam call from inside Postgres transaction
- [ ] investigate disallowing `setResourceParent` on `wds-instance` Sam resource type
- [x] managed identity/human user compatibility

Soon, I will make follow-on tickets for:
- [x] how to handle missing/empty/invalid `SAM_URL`
- [x] diagram of WDS/Sam permission model
- [x] default instance create at startup needs to execute as an identity, should move to a dedicated startup bean
- [x] retries for Sam calls

Questions for reviewers:
1. What do you think we should do with missing/empty/invalid `SAM_URL` values?
2. This PR requires a Sam connection (just like we require a Postgres connection) for local development. How can we make that easier for devs?

WDS Deployment Architecture - Deployment Architecture (1)


service/build.gradle
```diff
@@ -44,6 +44,11 @@ dependencies {
 	implementation 'org.webjars:webjars-locator-core:0.52'
 	implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-csv:2.14.2'
 	implementation 'io.sentry:sentry-logback:6.12.1'
+
+	// Terra libraries
+	implementation group: 'org.broadinstitute.dsde.workbench', name: 'sam-client_2.13', version: '0.1-b7e47d2'
```
aaronkanzer, line 49: Why declare a specific version (rather than a range of versions, etc.?)

Will this require us to manually check and edit `sam` over time?

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamClientSupport.java
```diff
@@ -0,0 +1,101 @@
...
+     * Perform a Sam client request with logging and exception handling, and return the result of that request.
+     *
+     * @param samFunction the Sam client request to perform
+     * @param loggerHint short string to include for all log entries for this request
+     * @return the result of the Sam client request
+     * @param <T> the return type of the Sam client request
+     * @throws SamException on most exceptions thrown by the Sam client request
+     * @throws AuthorizationException on a 401 or 403 from the Sam client request
+     */
+    <T> T executeSamRequest(SamFunction<T> samFunction, String loggerHint) throws SamException, AuthorizationException {
```
aaronkanzer, line 40: Small nit: is there any reason to name this and the function below both `executeSamRequest`?

service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
```diff
@@ -27,19 +36,47 @@ public List<UUID> listInstances(String version) {
     }
 
     @WriteTransaction
-    public void createInstance(UUID instanceId, String version) {
+    public void createInstance(UUID instanceId, String version, Optional<UUID> workspaceId) {
         validateVersion(version);
+
+        UUID samResourceId = instanceId; // id of "wds-instance" Sam resource we will create
+        UUID samParentResourceId = workspaceId.orElse(instanceId); // id of "workspace" Sam resource to use as the parent of "wds-instance"
```
aaronkanzer, line 43: How could the parent ID ever be a UUID that references the `wds` instance -- could we not assume that the workspace ID would always need to be present? Rather, should we be concerned if a parent workspace ID is not present?

service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
```diff
@@ -27,19 +36,47 @@ public List<UUID> listInstances(String version) {
     }
 
     @WriteTransaction
```
aaronkanzer, line 38: (Me being super lazy and not looking up the source code...heh)

I assume that this decorated makes the transactions atomic here? Just want to make sure that the writes to Sam DB and WDS DBs are both (or neither) executed

service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
```diff
@@ -27,19 +36,47 @@ public List<UUID> listInstances(String version) {
...
         }
+
+        // create `wds-instance` resource in Sam, specifying workspace as parent
+        samDao.createInstanceResource(samResourceId, samParentResourceId);
+        // create instance schema in Postgres
         recordDao.createSchema(instanceId);
     }
 
     @WriteTransaction
     public void deleteInstance(UUID instanceId, String version) {
```
aaronkanzer, line 64: What's the reasoning behind including the `workspaceId` for the creation, and not the delete? 

Would this make the inclusion of the `workspaceId` in `createInstance` superfluous?

service/src/main/java/org/databiosphere/workspacedataservice/service/model/exception/AuthorizationException.java
```diff
@@ -0,0 +1,10 @@
+package org.databiosphere.workspacedataservice.service.model.exception;
+
+import org.springframework.http.HttpStatus;
+
+public class AuthorizationException extends SamException {
+
+	public AuthorizationException(String message) {
+		super(HttpStatus.FORBIDDEN, message);
```
aaronkanzer, line 8: It seems that we are also checking for `401`s in terms of permissions

Do you want to reference a `HttpStatus.UNAUTHORIZED`? Or do you think this suffices enough?

service/src/test/java/org/databiosphere/workspacedataservice/service/InstanceServiceSamExceptionTest.java
```diff
@@ -0,0 +1,333 @@
...
+                () -> instanceService.deleteInstance(instanceId, VERSION),
+                "deleteInstance should throw if caller does not have permission to create wds-instance resource in Sam"
+        );
+        allInstances = instanceService.listInstances(VERSION);
+        assertTrue(allInstances.contains(instanceId), "instanceService.deleteInstance should not have deleted the instances.");
+    }
+
+    // implementation of tests that expect SamException
+    private void doSamCreateAndDeleteTest(UUID instanceId, int expectedSamExceptionCode) {
+        doSamCreateTest(instanceId, expectedSamExceptionCode);
```
aaronkanzer, line 300: I'm slightly confused with this test (perhaps it could just be the naming...)

It seems that `doSamCreateTest` is only used for tests that throw exceptions? Wouldn't an instance need to be present for it to be deleted at first? Why not use either `doSamCreateTest` or `doSamDeleteTest` only, especially since `doSamDeleteTest` circumvents creation via `ercordDao.createSchema(instanceId)`?

service/src/test/java/org/databiosphere/workspacedataservice/sam/MockSamResourcesApi.java
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Mock for the Sam Client ResourcesApi for use in unit tests.
+ * Always returns true for all permission checks.
+ * Never throws any Exceptions.
+ */
+public class MockSamResourcesApi extends ResourcesApi {
+
+    @Override
+    public Boolean resourcePermissionV2(String resourceTypeName, String resourceId, String action) {
```
aaronkanzer, line 14: Is there any need to test this if it returns `false`? Or is this purely here for mocking purposes?

service/src/main/java/org/databiosphere/workspacedataservice/sam/BearerTokenFilter.java
```diff
@@ -0,0 +1,57 @@
...
+public class BearerTokenFilter implements Filter {
+    private static final Logger LOGGER = LoggerFactory.getLogger(BearerTokenFilter.class);
+
+    public static final String ATTRIBUTE_NAME_TOKEN = "bearer-token-attribute";
+
+    private static final String BEARER_PREFIX = "Bearer ";
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException {
+        if (request instanceof HttpServletRequest httpRequest) {
```
aaronkanzer, line 39: What if it is not an `instanceof HttpServletRequest httpRequest`, do you still want to proceed with `chain.doFilter(request, response)` or do you want an `else` clause in this function?

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamClientFactory.java
```diff
@@ -0,0 +1,64 @@
...
+                .getAttribute(ATTRIBUTE_NAME_TOKEN, SCOPE_REQUEST);
+        // add the user's bearer token to the client
+        if (!Objects.isNull(token)) {
+            LOGGER.debug("setting access token for Sam request");
+            apiClient.setAccessToken(token.toString());
+        } else {
+            LOGGER.warn("No access token found for Sam request.");
+        }
+        // return the client
+        return apiClient;
```
aaronkanzer, line 49: Does the `apiClient` need an `AccessToken` and a `BasePath` in order to be operational? Just curious since there is no exception handling between this function and `getResourcesApi`...

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -105,8 +105,9 @@ public ResponseEntity<List<UUID>> listInstances(@PathVariable("version") String
 
 	@PostMapping("/instances/{version}/{instanceId}")
 	public ResponseEntity<String> createInstance(@PathVariable("instanceId") UUID instanceId,
-			@PathVariable("version") String version) {
-		instanceService.createInstance(instanceId, version);
+			@PathVariable("version") String version,
+			@RequestParam(name= "workspaceid", required = false) Optional<UUID> workspaceId) {
+		instanceService.createInstance(instanceId, version, workspaceId);
```
aaronkanzer, line 110: Just thinking out loud -- if a `Bearer` token is missing from the request, does the request even need to hit the backend? Could it not just be blocked from a proxy such as `nginx`?

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamClientSupport.java
davidangb: These are overloaded methods - they do the exact same thing, just accept different arguments (`SamFunction<T>` vs  `VoidSamFunction`). Think of it like `Logger.info(String msg)` and `Logger.info(String msg, Throwable t)`.

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /**
+     * Check if the current user has permission to delete a "wds-instance" resource from Sam.
+     * Implemented as a check for delete permission on the resource.
+     *
+     * @param instanceId the id of the "wds-instance" resource to be deleted
+     * @return true if the user has permission
+     */
+    @Override
```
yuliadub, line 43: could there ever be a case that the instance that we try to delete would be not in the workspace we are trying to delete it from? asking since instance creation does take a parentWorkspaceId which makes me Queston why Delete doesn't.

service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
davidangb: The logic is:
* if the user supplies both an `instanceId` and a `workspaceId`, use them as-is.
* if the user supplies only an `instanceId`, assume that the parent workspace has the same id as the instance; i.e. use the value of `instanceId` for both the instance and the workspace.

In lower classes such as `SamDao.createInstanceResource()`, I require both ids to be present. Here in `InstanceService` is where I'm implementing the defaulting - if the user didn't specify `workspaceId` then I fill it in here.

Am I understanding your question correctly to mean that you think we should require the end user to always specify both, even if they are the same? That would be a backwards-incompatible API change (which probably isn't that big a deal) - but imho isn't necessary to impose on the user.

service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
davidangb: No, this would not roll back the Sam call; that's something we'd have to implement manually. This transaction annotation is purely about Postgres.

That said, good callout here - we shouldn't include an external http call inside a postgres transaction, as it can keep the transaction open for longer than it needs! I will fix that.

service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
yuliadub: ha! left a similar comment up higher, although it seems like including it in both places would make more sense to me. I think it is better to be safe and include more data during deletion vs not (so we never accidentally delete a wds not in the right workspace somehow)

davidangb: This reflects the calls we need to make to Sam:
* when creating the Sam resource, we need to tell Sam both the id of the `wds-instance` resource itself, as well as the id of its parent `workspace` resource.
* when deleting a Sam resource, we only need the id of the resource itself; we don't need to specify the parent's id to delete it.

service/src/main/java/org/databiosphere/workspacedataservice/sam/BearerTokenFilter.java
```diff
@@ -0,0 +1,57 @@
...
+            String authString = httpRequest.getHeader(HttpHeaders.AUTHORIZATION);
+            if (!Objects.isNull(authString) && authString.startsWith(BEARER_PREFIX)) {
+                String token = authString.replaceFirst(BEARER_PREFIX, "");
```
yuliadub, line 42: how will this work when I am trying to create or delete an instance from a jupyter notebook? would it be assumed that the identity of the notebook would have sufficient  permissions or would I need to use another identity to execute instance creation/deletion?

service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
davidangb: It might be helpful here to look at the Sam API to delete a resource: https://sam.dsde-dev.broadinstitute.org/#/Resources/deleteResourceV2. It accepts only two parameters: `resourceTypeName` and `resourceId`.

service/src/test/java/org/databiosphere/workspacedataservice/service/InstanceServiceSamExceptionTest.java
davidangb: Of all the test functions in this file:
* 3 call only `doSamCreateTest`. These test handling for exceptions thrown by the Sam call to create a resource.
* 3 call only `doSamDeleteTest`. These test handling for exceptions thrown by the Sam call to delete a resource.
* 3 call `doSamCreateAndDeleteTest`, which is just a helper to call both `doSamCreateTest` and `doSamDeleteTest`. These test handling for exceptions thrown by the Sam call to checking permissions, which we do prior to actually trying to create or delete a resource.

Are you suggesting that the 3 tests that call `doSamCreateAndDeleteTest` should actually be 6 tests, and call `doSamCreateTest`\`doSamDeleteTest` separately?

service/src/test/java/org/databiosphere/workspacedataservice/sam/MockSamResourcesApi.java
davidangb: This class is purely a mock that is used within tests; there are no tests against this specifically.

service/src/main/java/org/databiosphere/workspacedataservice/sam/BearerTokenFilter.java
davidangb: The `chain.doFilter(request, response)` is outside of the `if` clause and thus will always happen.

Beyond that - if `instanceof HttpServletRequest` is false, there is no guarantee that HTTP headers exist at all, and in fact `getHeader()` is only present on `HttpServletRequest`, not on `ServletRequest` … so in that corner case there's nothing to do.

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamClientFactory.java
davidangb: It definitely needs a `basePath`, else it will try to contact Sam at `http://localhost` and will fail.

It technically does not need an `accessToken`, though in practice if we try to make calls to Sam without one, Sam will return 401. However, that's exactly what I want to happen - in WDS, we just grab the token from the inbound request and pass it to Sam without validating it, and let Sam do the validation. There's no need for WDS to have extensive logic to validate tokens.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
davidangb: The approach I was taking here is that WDS doesn't validate or enforce anything about the token. Upstream of WDS, the Relay Listener validates the token by making a call to Sam; downstream of WDS, Sam validates any tokens we send to it. I didn't think WDS *also* needed do any validation, though I admit "isBlank" is a pretty simple and lightweight validation.

That said, requiring a token here would also impact local development and unit tests - or we'd need more conditional logic to support those use cases.

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
davidangb: There's no feature in the product that would move an instance from one workspace to another. However, it is technically possible for an enterprising user:
1. create the WDS instance, which creates a Sam resource with a parent workspace
2. manually call Sam's [setResourceParent](https://sam.dsde-dev.broadinstitute.org/#/Resources/setResourceParent) API, which changes the parent.

I would like to investigate disallowing the `setResourceParent` call; I need to consult with the Sam team about that.

Nonetheless, I think we still want to delete the resource, regardless if anyone changed its parent in the meantime.

service/src/main/java/org/databiosphere/workspacedataservice/service/model/exception/AuthorizationException.java
davidangb: I agree, I am looping together Sam 401s and 403s into a single WDS 403. I can change this.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -99,6 +99,9 @@ public RecordDao(NamedParameterJdbcTemplate namedTemplate,
 	private void createDefaultInstanceSchema(String workspaceId) {
 		LOGGER.info("Default workspace id loaded as {}", workspaceId);
 
+		// TODO: execute this as the WDS managed identity so it can call Sam
```
jordansafer, line 102: Is there a JIRA ticket we can tie this TODO to?

service/src/main/java/org/databiosphere/workspacedataservice/service/model/exception/AuthorizationException.java
davidangb: Just pushed support for `AuthenticationException` as well as `AuthorizationException`.

service/src/main/java/org/databiosphere/workspacedataservice/service/model/exception/SamException.java
```diff
@@ -0,0 +1,15 @@
...
+import org.springframework.http.HttpStatus;
+import org.springframework.web.server.ResponseStatusException;
+
+public class SamException extends ResponseStatusException {
+
+	public SamException(HttpStatus status, String message) {
+		super(status, message);
+	}
+
+	public SamException(HttpStatus status, String message, Throwable cause) {
```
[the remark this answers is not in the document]
davidangb: removed, good catch

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
```diff
@@ -0,0 +1,93 @@
...
+        parent.setResourceTypeName(RESOURCE_NAME_WORKSPACE);
+        parent.setResourceId(parentWorkspaceId.toString());
+
+        CreateResourceRequestV2 createResourceRequest = new CreateResourceRequestV2();
+        createResourceRequest.setResourceId(instanceId.toString());
+        createResourceRequest.setParent(parent);
+        createResourceRequest.setAuthDomain(Collections.emptyList());
+
+        VoidSamFunction samFunction = () -> samClientFactory.getResourcesApi().createResourceV2(RESOURCE_NAME_INSTANCE, createResourceRequest);
+        executeSamRequest(samFunction, "createInstanceResource");
```
jordansafer, line 76: The naming for executeSamRequest feels a little vague to me. This reminds me a lot of wrapped function calls where we'll wrap a call with `withErrorHandling`, and I'm wondering if we could call this something like `withSamErrorHandling(() -> apiCallHere()`.

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
```diff
@@ -0,0 +1,93 @@
...
+     */
+    @Override
+    public boolean hasDeleteInstancePermission(UUID instanceId) {
+        return hasPermission(RESOURCE_NAME_INSTANCE, instanceId.toString(), ACTION_DELETE,
+                "hasDeleteInstancePermission");
+    }
+
+    // helper implementation for permission checks
+    private boolean hasPermission(String resourceType, String resourceId, String action, String loggerHint) {
+        SamFunction<Boolean> samFunction = () -> samClientFactory.getResourcesApi()
```
jordansafer, line 51: Since we always use the client factory with `samClientFactory.getResourceApi()`, could we move this to the constructor, and replace it with something along the lines of 'samApi.resourcePermissionV2()`? I may be wrong, we may use samClientFactory in a different way that I'm missing

service/build.gradle
davidangb: It's kinda bugged me that we have some dependencies that DON'T specify a version! https://blog.danlew.net/2015/09/09/dont-use-dynamic-versions-for-your-dependencies/ is a good description of my thoughts.

service/build.gradle
```diff
@@ -44,6 +44,11 @@ dependencies {
...
+	implementation 'com.squareup.okhttp3:okhttp:4.10.0' // required by Sam client
```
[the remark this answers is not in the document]
davidangb: yeah, potentially it's something that could be included in the Sam client, but it's also not that big a deal to add the extra line here, and as a silver lining it makes the `okhttp3` dependency visible and obvious

service/src/main/java/org/databiosphere/workspacedataservice/sam/BearerTokenFilter.java
davidangb: This is a great question - I need to look into that! This will also be relevant for the `TODO` of creating the default instance upon WDS startup, which will happen as a managed identity instead of a human user.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
davidangb: this is on me - I need to file tickets for the 3 issues listed in the PR description, I will do that!

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
davidangb: I agree it's a bit awkward/verbose code. I couldn't think of a better solution (and would love to hear ideas!), because:
* `SamDao`/`HttpSamDao` is a Spring-managed bean, which will get reused across many requests
* `ResourcesApi` must be re-created for each request, since its `ApiClient` contains the user's access token
* `SamClientFactory`/`HttpSamClientFactory` are Spring-managed beans

As I write this, I realize … I could also just set a `@Scope` for the `SamDao` bean to make sure it gets reused.

The other significant issue with `SamClientFactory` is that I've built a lot of tests using `MockSamClientFactory` to return a `MockResourcesApi` - possibly there's a solution for that too.

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
davidangb: renamed to `withSamErrorHandling`!

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
jordansafer: I'm ok to leave this as is, the only solutions I can think of add a lot more complexity than they save

jordansafer (approved):

service/build.gradle
aaronkanzer: Thanks for the share here -- makes sense!

service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
aaronkanzer: Perhaps to simplify -- what would the scenario be that the parent workspace has the same ID as the instance? (Perhaps I am confusing myself with the meaning of `instance` again here haha)

service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
aaronkanzer: Do we want to just bump the docstring here to say something like `instanceId === `resourceId` in Sam-land?

service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
davidangb: The default instance that WDS creates on startup uses the workspace id: https://github.com/DataBiosphere/terra-workspace-data-service/blob/71353cb25941db18e2833a888f39d85a69c83493/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java#L96

The Terra UI uses the workspaceId when specifying which WDS instance to query. Here's one example:
https://github.com/DataBiosphere/terra-ui/blob/2a22a0fcc193fdd7c9895d7e5df81dbfe6b19e82/src/libs/ajax/data-table-providers/WdsDataTableProvider.ts#L262

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
davidangb: made AJ-897, AJ-898, AJ-899

yuliadub:
(wrote this yesterday but never sent, whooops!) 

note to self: likely should create a task to test that this functionality works as expected for python Client (or maybe just include tests for this in the automated test task) 

As far as questions, here are my thoughts: 

**What do you think we should do with missing/empty/invalid SAM_URL values?**
I think failing the call makes sense  - since if SAM is missing, there is no way to check permissions period - as long as it had a very clear error message that SAM was not reached and what the url that was provided was (empty vs localhost/etc)

I guess the question is, if SAM url is missing on workspace creation vs say a user creating a second instance, should behavior between these 2 scenario be different? 
Should workspace creation fail altogether if WDS cant create its instance? (maybe?) Can a workspace be useful if WDS doesnt launch and what would be the way to recover?  (trigger instance creation from notebook or offer a button to create WDS on the data tab page?) But then again, if WDS didnt get created the first time due to a SAM issue, would the underlying secondary call even work or something fundamentally is wrong and wont recover on its own? I talked myself into the fact that not having a valid SAM url should prevent a workspace from being usable. :D 

**This PR requires a Sam connection (just like we require a Postgres connection) for local development. How can we make that easier for devs?**

I assume we want to hit existing Sam vs deploying our own testing instance, so could there be config variables that specifies the Sam url at runtime? (I know this affects BEE creation, since this value likely will need to be configured/changed for that but maybe if it is provided as an env variable as well as config that may be okay?) Alternatively, could we do it so when running on localhost authorization always returns true (like today when running locally no auth for the api is required), not sure if something like that is possible.

service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
davidangb: just pushed this change!

service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
davidangb: added javadoc to the method!

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamClientFactory.java
davidangb: I am hoping to defer that work to AJ-898, especially because it involves a bunch of figuring out what we actually want to do. Are you ok with deferring it?

aaronkanzer: Yep -- ok with that!

yuliadub (approved):

aaronkanzer (approved):

service/src/main/java/org/databiosphere/workspacedataservice/sam/BearerTokenFilter.java
davidangb: Checked with some folks - the managed identity used by apps/notebooks is linked to your human-user account, and thus everything should just magically work!

PR #184, opened 2023-03-09
AJ-689: Include Status check to Sam in WDS

aaronkanzer:
Relates to https://broadworkbench.atlassian.net/browse/AJ-679

Builds out functionality to understand whether or not SAM is healthy -- WDS relies on SAM being available as a mechanism of checking permissions for multiple HTTP request types (POST, DELETE) that WDS does upon its tables


calypsomatic:
As far as I can tell, this is showing up in the statusAPi as "statusService":
```
"statusService": {
      "status": "UP"
    }
```
Should it be identified as 'sam'?  And do we need to add it to the `StatusResponse` schema in swagger?

aaronkanzer:
> As far as I can tell, this is showing up in the statusAPi as "statusService":
> 
> ```
> "statusService": {
>       "status": "UP"
>     }
> ```
> 
> Should it be identified as 'sam'? And do we need to add it to the `StatusResponse` schema in swagger?

Ah good point -- I originally was thinking that I would make this abstract, but now I'm wondering if we should rename/not-abstract and make it all `SamStatusService` class, etc.

Let me see if I can make "status" as "samStatus"

service/src/test/java/org/databiosphere/workspacedataservice/service/StatusServiceTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.mockito.Mockito;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.actuate.health.Health;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+
+import java.math.BigDecimal;
```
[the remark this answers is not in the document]
aaronkanzer: Hmmm not sure how that got in there -- removed!

service/src/main/java/org/databiosphere/workspacedataservice/service/StatusService.java
```diff
@@ -0,0 +1,40 @@
...
+    protected void doHealthCheck(Health.Builder builder) throws Exception {
+
+        SystemStatus samStatus = samDao.getSystemStatus();
+
+        if(samStatus.getOk()) {
+            builder.up();
+        } else {
+            builder.down();
+            LOGGER.warn("The SAM instance that WDS is requesting is currently down. Details: {}", samStatus);
+        }
```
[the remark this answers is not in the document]
aaronkanzer: I'm not sure about the alternative here in terms of using `spring-boot-actuator` and determining the health of an another microservice (with `builder.up()`? As well as if `builder.withDetail()` adds any further utility here unless we use a specific subsystem as the determinant of Sam's health vs WDS health?

There are the values Sam returns that aggregate to the overall health:

```
    systems: {GoogleGroups=class SubsystemStatus {
        ok: true
        messages: null
    }, GooglePubSub=class SubsystemStatus {
        ok: true
        messages: null
    }, GoogleIam=class SubsystemStatus {
        ok: true
        messages: null
    }, Database=class SubsystemStatus {
        ok: true
        messages: null
    }}
```

I'm curious if you have any alternatives in mind -- my research so far on the interwebs doesn't reveal much -- unless we do some more intelligent logic on the k8s probes...

service/src/main/java/org/databiosphere/workspacedataservice/service/StatusService.java
```diff
@@ -0,0 +1,40 @@
...
+import org.springframework.boot.actuate.health.AbstractHealthIndicator;
+import org.springframework.boot.actuate.health.Health;
+import org.springframework.stereotype.Service;
+
+/**
+* Service that performs health checks on microservices that WDS relies on (such as SAM)
+ *
+ * See here for more details how we handle health checks in WDS: https://www.baeldung.com/spring-boot-health-indicators
+ */
+@Service
```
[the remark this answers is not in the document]
aaronkanzer: Updated! I kept the function call name of `doHealthCheck`, but I renamed the class to `SamStatusService`

service/src/main/java/org/databiosphere/workspacedataservice/service/StatusService.java
```diff
@@ -0,0 +1,40 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(StatusService.class);
+
+    public StatusService(SamDao samDao) {
+        this.samDao = samDao;
+    }
+
+    @Override
+    protected void doHealthCheck(Health.Builder builder) throws Exception {
+
+        SystemStatus samStatus = samDao.getSystemStatus();
```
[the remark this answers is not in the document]
aaronkanzer: I _think_ I did this correctly (added the `Cacheable`) decorator -- let me know if you had something else in mind though/or if my usage is incorrect...

service/src/test/java/org/databiosphere/workspacedataservice/sam/MockSamClientFactory.java
```diff
@@ -10,4 +11,9 @@ public class MockSamClientFactory implements SamClientFactory {
     public ResourcesApi getResourcesApi() {
         return new MockSamResourcesApi();
     }
+
+    public StatusApi getStatusApi() {
+        return new StatusApi();
```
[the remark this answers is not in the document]
aaronkanzer: Is the concern if `StatusApi` ever changes its interface, that these tests will break silently?

service/src/main/java/org/databiosphere/workspacedataservice/service/StatusService.java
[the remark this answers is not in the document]
davidangb: My suggestion, in code, is to do something like this:
```java
    public void doHealthCheck(Health.Builder builder) {
        // we don't want a problem with the Sam connection to take WDS down entirely. So,
        // we always call builder.up() here, but we include the actual Sam status in the builder detail.
        builder.up();

        try {
            SystemStatus samStatus = samDao.getSystemStatus();
            builder.withDetail("ok", samStatus.getOk());
        } catch (Exception e) {
            builder.withDetail("connectionError", e.getMessage());
        }
    }
```
this should result in:
* WDS's overall status remains UP even if the call to Sam fails, or if Sam returns not-OK
* WDS's `/status` response includes information about Sam's current status, even though the information doesn't change the overall UP/DOWN state.

Does this help explain?

service/src/test/java/org/databiosphere/workspacedataservice/sam/MockSamClientFactory.java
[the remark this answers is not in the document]
davidangb: the concern is that a developer who is expecting a fully-mocked object in their unit tests would end up with a live runtime object instead. I think we can let this go for now; until/unless anyone is using `StatusApi` outside of the health-check code, this won't be a problem.

service/src/main/java/org/databiosphere/workspacedataservice/service/StatusService.java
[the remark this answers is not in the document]
davidangb: my suggestion for this:

move the `@Cacheable` annotation into `HttpSamDao`, and add an additional method there to enforce a TTL for the cache so it's not eternal:

**HttpSamDao.java**
```java
    /**
     * Gets the System Status of Sam. Using @Cacheable, will reach out to Sam no more than once every 5 minutes.
     * See also emptySamStatusCache()
     */
    @Cacheable(value = "samStatus")
    public SystemStatus getSystemStatus() {
        SamFunction<SystemStatus> samFunction = () -> samClientFactory.getStatusApi().getSystemStatus();
        return withSamErrorHandling(samFunction, "getSystemStatus");
    }

    /**
     * Clears the samStatus cache every 5 minutes, to ensure we get fresh results from Sam
     * every so often. See also getSystemStatus()
     */
    @CacheEvict(value = "samStatus", allEntries = true)
    @Scheduled(fixedRate = 5, timeUnit = TimeUnit.MINUTES)
    public void emptySamStatusCache() {
        LOGGER.debug("emptying samStatus cache");
    }
```

note that for `@Scheduled` to work here, you'll also have to add `@EnableScheduling` to `WorkspaceDataServiceApplication`, just like it already has `@EnableCaching`

service/src/main/java/org/databiosphere/workspacedataservice/service/StatusService.java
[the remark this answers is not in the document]
aaronkanzer: This worked great 🎉 -- thanks for the suggestions, I incorporated it into the code

service/src/test/java/org/databiosphere/workspacedataservice/sam/MockSamClientFactory.java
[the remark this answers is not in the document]
aaronkanzer: Sounds good -- I'd be surprised if this was used outside of the health-check code, so happy to leave as is

service/src/test/java/org/databiosphere/workspacedataservice/sam/SamStatusServiceTest.java
```diff
@@ -0,0 +1,51 @@
...
+    @BeforeEach
+    void beforeEach() throws ApiException {
+        // return the mock StatusApi from the mock SamClientFactory
+        given(mockSamClientFactory.getStatusApi()).willReturn(mockStatusApi);
+        Mockito.clearInvocations(mockStatusApi);
+        Mockito.clearInvocations(mockHealthBuilder);
+    }
+
+    @Test
+    void testSamHealthyCall() throws Exception {
```
aaronkanzer, line 42: @calypsomatic @davidangb 

I couldn't figure out how to mock the unhealthy case here (e.g. when an exception is thrown in this case) -- tried a handful of combinations with Exception throwing by Mockito, with no avail 😞 

Curious if we are OK preceding as is, or if there is some magic to unit test the unhealthy case...

calypsomatic: What doesn't work?  Mockito doesn't throw an exception?

calypsomatic: Maybe something more like [here](https://github.com/DataBiosphere/terra-workspace-data-service/pull/186#discussion_r1136013888)?

davidangb: > in many cases
could this be because of the caching? (I haven't tried it, just putting out a guess)

aaronkanzer: Yeah I thought that might be the case, but including `Mockito.clearAllCaches();` didn't seem to resolve within the `beforeEach()` call 🤔

aaronkanzer: I tried this AM to clear the cache in any manner possible using `CacheEvict`, etc. with no avail 😞 

Wondering if anyone else would be willing to take a hack here -- otherwise, I'm OK with not having the specific unit test here, as it seems we are mocking just about everything -- let me know -- thanks

aaronkanzer:
> Hmm... I tried calling status with a fake SAM_URL and it said UP but with an error message:
> 
> ```
> "Sam": {
>       "status": "UP",
>       "details": {
>         "connectionError": "500 INTERNAL_SERVER_ERROR \"Error from Sam: java.net.UnknownHostException: sam.nonono.broadinstitute.org: nodename nor servname provided, or not known\""
>       }
> ```
> 
> Seems like it shouldn't say UP? Or is this the proper response because we always call `builder.up()`?

Good point -- I've included `builder.down()` in the `Exception` clause -- I hope we don't have a wrong SAM_URL in general though... 👻

calypsomatic:
> Good point -- I've included `builder.down()` in the `Exception` clause -- I hope we don't have a wrong SAM_URL in general though... 👻
One last small question - when Sam is DOWN the status response returns 503.  I'm assuming it's automatic but if everything else is up should the status response be 503?  Not sure 200 makes sense either, just wondering.

service/src/test/java/org/databiosphere/workspacedataservice/sam/SamStatusServiceTest.java
davidangb: for your experimentation, try adding an `@Autowired CacheManager cacheManager;` to this class, then play with:

```java
    @BeforeEach
    public void beforeEach() throws ApiException {
        Cache samCache = cacheManager.getCache("samStatus");
        Cache.ValueWrapper vw = samCache.get("getSystemStatus");
        LOGGER.info("cache entry before clear is: " + (Objects.isNull(vw) ? "null" : vw.get()));
        // samCache.clear(); // try commenting/uncommenting this line to see different test behavior
        vw = samCache.get("getSystemStatus");
        LOGGER.info("cache entry after clear is: " + (Objects.isNull(vw) ? "null" : vw.get()));
        …
```

davidangb: oh whoops - the previous comment also requires setting a constant cache key in `HttpSamDao`, like so:
```java
@Cacheable(value = "samStatus", key="'getSystemStatus'")
```

calypsomatic:
> Good catch
> 
> So I updated the code to return the actual Sam status in the `details` section of the response -- it seems that Spring Boot Health Actuator aggregates the `status` field specifically, so we can't set that down, and keep the overall status as 'up'
> 
ARGH:
```
  "Sam": {
      "status": "UP",
      "details": {
        "status": "DOWN",
        "connectionError": "500 INTERNAL_SERVER_ERROR \"Error from Sam: java.net.UnknownHostException: sam.nonono.broadinstitute.org: nodename nor servname provided, or not known\""
      }
```
Let's just keep it the way it was I guess

service/src/test/java/org/databiosphere/workspacedataservice/sam/SamStatusServiceTest.java
aaronkanzer: @davidangb Turns out that this made unit tests fail in other scenarios with Sam auth, e.g. `testCreateAndValidateInstance` in which the errors are non-descript.

Speficially:

```
                    Caused by:
                    org.springframework.jdbc.CannotGetJdbcConnectionException: Failed to obtain JDBC Connection; nested exception is org.postgresql.util.PSQLException: FATAL: remaining connection slots are reserved for non-replication superuser connections
                        at app//org.springframework.jdbc.datasource.DataSourceUtils.getConnection(DataSourceUtils.java:83)
                        at app//org.springframework.jdbc.datasource.init.DatabasePopulatorUtils.execute(DatabasePopulatorUtils.java:52)
                        ... 124 more

                        Caused by:
                        org.postgresql.util.PSQLException: FATAL: remaining connection slots are reserved for non-replication superuser connections
                            at app//org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2675)
                            at app//org.postgresql.core.v3.QueryExecutorImpl.readStartupMessages(QueryExecutorImpl.java:2787)
                            at app//org.postgresql.core.v3.QueryExecutorImpl.<init>(QueryExecutorImpl.java:173)
                            at app//org.postgresql.core.v3.ConnectionFactoryImpl.openConnectionImpl(ConnectionFactoryImpl.java:290)
                            at app//org.postgresql.core.ConnectionFactory.openConnection(ConnectionFactory.java:49)
                            at app//org.postgresql.jdbc.PgConnection.<init>(PgConnection.java:223)
                            at app//org.postgresql.Driver.makeConnection(Driver.java:402)
                            at app//org.postgresql.Driver.connect(Driver.java:261)
                            at app//com.zaxxer.hikari.util.DriverDataSource.getConnection(DriverDataSource.java:138)
                            at app//com.zaxxer.hikari.pool.PoolBase.newConnection(PoolBase.java:364)
                            at app//com.zaxxer.hikari.pool.PoolBase.newPoolEntry(PoolBase.java:206)
                            at app//com.zaxxer.hikari.pool.HikariPool.createPoolEntry(HikariPool.java:476)
                            at app//com.zaxxer.hikari.pool.HikariPool.checkFailFast(HikariPool.java:561)
                            at app//com.zaxxer.hikari.pool.HikariPool.<init>(HikariPool.java:115)
                            at app//com.zaxxer.hikari.HikariDataSource.getConnection(HikariDataSource.java:112)
                            at app//org.springframework.jdbc.datasource.DataSourceUtils.fetchConnection(DataSourceUtils.java:159)
                            at app//org.springframework.jdbc.datasource.DataSourceUtils.doGetConnection(DataSourceUtils.java:117)
                            at app//org.springframework.jdbc.datasource.DataSourceUtils.getConnection(DataSourceUtils.java:80)

```

I'm inclined to timebox and move on from this specific test and remove it (I'm not sure if we gain much upside), but if strong opinions are here, happy to continue hacking

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamStatusService.java
```diff
@@ -0,0 +1,42 @@
...
+        // we don't want a problem with the Sam connection to take WDS down entirely. So,
+        // we always call builder.up() here, but we include the actual Sam status in the builder detail.
+        builder.up();
+
+        try {
+            SystemStatus samStatus = samDao.getSystemStatus();
+            builder.withDetail("ok", samStatus.getOk());
+        } catch (Exception e) {
+            // SAM "status" will still return "UP"; however, details.status should be evaluated instead to determine the health of Sam
+            builder.withDetail("status", "DOWN");
```
calypsomatic, line 37: This status detail isn't the real status:
`"Sam": {
      "status": "UP",
      "details": {
        "status": "DOWN",
        "connectionError": "500 INTERNAL_SERVER_ERROR \"Error from Sam: java.net.UnknownHostException: sam.nonono.broadinstitute.org: nodename nor servname provided, or not known\""
      }`
I'm ok with putting it back the way it was before and letting it be a 503.

aaronkanzer: I think a `503` though would fail the k8s status check -- e.g. if Sam is down, we still want to grant some form of Read access?

calypsomatic: Hmm but `{ "status": "UP", "details": { "status": "DOWN"}}` is really confusing.  Is there some non-contradictory way we can indicate that it's down?

davidangb: Given [this Slack discussion](https://broadinstitute.slack.com/archives/C046RMBQ2R3/p1678898917958389), I don't think there is anything that will take WDS out of operation if the status API reports down. Therefore I suggest reverting the previous change and just reporting Sam as down and calling `builder.down()` in that case. This is probably worth testing empirically , though - make sure WDS remains reachable if /status reports down.

aaronkanzer: updated here!

calypsomatic (approved):

service/src/test/java/org/databiosphere/workspacedataservice/sam/SamStatusServiceTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.springframework.boot.actuate.health.Health;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.cache.Cache;
+import org.springframework.cache.CacheManager;
+
+import static org.mockito.BDDMockito.given;
+import static org.mockito.Mockito.*;
+
+@SpringBootTest
```
davidangb, line 20: after playing around with this a bit, I think a much cleaner solution to the caching issues in this test is to specify:
```java
@SpringBootTest(properties = "spring.cache.type=NONE")
```
and remove the `CacheManager` and all interaction with it

davidangb: also, using #188 as a guide, I'd love to reduce the `classes` this `@SpringBootTest` loads, specifically to eliminate it starting up a Hikari connection pool. This was pretty tricky in the other PR and had a lot of trial and error; if that effort becomes large and ugly, let's save it for future work

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
```diff
@@ -87,6 +96,25 @@ public void deleteInstanceResource(UUID instanceId) {
...
+    public SystemStatus getSystemStatus() {
+        SamFunction<SystemStatus> samFunction = () -> samClientFactory.getStatusApi().getSystemStatus();
+        return withSamErrorHandling(samFunction, "getSystemStatus");
+    }
+
+    /**
+     * Clears the samStatus cache every 5 minutes, to ensure we get fresh results from Sam
+     * every so often. See also getSystemStatus()
+     */
+    @CacheEvict(value = "samStatus", allEntries = true)
```
davidangb, line 113: since the `@Cacheable` above specifies a single key, you could/should also specify that key here instead of `allEntries = true`

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
```diff
@@ -87,6 +96,25 @@ public void deleteInstanceResource(UUID instanceId) {
...
+    @Scheduled(fixedRate = 5, timeUnit = TimeUnit.MINUTES)
```
davidangb, line 114: I wonder if the "5" here is worth pulling into application.properties config to make it easier to tweak down the line. What do you think?

davidangb: does the example in https://www.baeldung.com/spring-setting-ttl-value-cache#setting-tll-for-cacheable work? See the section "4. Setting TLL for @Cacheable"

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamStatusService.java
```diff
@@ -0,0 +1,45 @@
...
+
+    public SamStatusService(SamDao samDao) {
+        this.samDao = samDao;
+    }
+
+    @Override
+    public void doHealthCheck(Health.Builder builder) {
+        // we don't want a problem with the Sam connection to take WDS down entirely. So,
+        // we always call builder.up() here, but we include the actual Sam status in the builder detail.
+        builder.up();
```
davidangb, line 30: this call to `up()` and its code comment are now extraneous/incorrect

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamStatusService.java
```diff
@@ -0,0 +1,41 @@
...
+
+        try {
+            SystemStatus samStatus = samDao.getSystemStatus();
+            builder.withDetail("ok", samStatus.getOk());
```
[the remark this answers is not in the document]
aaronkanzer: Removed!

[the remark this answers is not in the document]
aaronkanzer: Evaluated further -- just left 

```
if (samStatus.getOk()) {
                builder.up();
            } else {
                builder.down();
            }
```

and removed Exception calling / updated unit tests

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamStatusService.java
aaronkanzer: Updated!

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
aaronkanzer: Updated!

service/src/test/java/org/databiosphere/workspacedataservice/sam/SamStatusServiceTest.java
aaronkanzer: I might want to split that into follow-on, just to get this initial code merged if that's alright!

aaronkanzer:
https://github.com/DataBiosphere/terra-workspace-data-service/pull/184#discussion_r1140451440

@davidangb Will try this on a follow-up PR!

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamStatusService.java
```diff
@@ -0,0 +1,36 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(SamStatusService.class);
+
+    public SamStatusService(SamDao samDao) {
+        this.samDao = samDao;
+    }
+
+    @Override
+    public void doHealthCheck(Health.Builder builder) {
+        SystemStatus samStatus = samDao.getSystemStatus();
```
davidangb, line 28: I think this still needs a try/catch here - without that, what is the behavior if `samDao.getSystemStatus();` throws an error? I saw you removed the unit test too, assuming you put the try/catch back I think you should also put the unit test back

aaronkanzer: Updated

davidangb (commented):
I am good with the code changes in this PR. Just want to make sure you have tested it empirically inside a real cluster and that returning non-200 from /status does not cause WDS to become unavailable inside that cluster.

davidangb (approved):
thank you!

PR #186, opened 2023-03-14
AJ-667 Add sam permission check to write requests

calypsomatic:
[See AJ-667](https://broadworkbench.atlassian.net/browse/AJ-667)
All requests that involve a write transaction need to check with Sam to see if user has write permission for the wds-instance.
Note this PR also includes some refactoring to maintain transactions.


service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
```diff
@@ -1,22 +1,25 @@
...
+import org.springframework.web.server.ResponseStatusException;
 
 import java.sql.SQLException;
 import java.util.*;
 import java.util.stream.Collectors;
 
+import static org.databiosphere.workspacedataservice.service.model.ReservedNames.RESERVED_NAME_PREFIX;
+
 @Service
 public class RecordService {
```
calypsomatic, line 25: Note: all the changes in this class are just code moved from other classes, no real changes.

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -57,86 +57,6 @@ public BatchWriteService(RecordDao recordDao, @Value("${twds.write.batch.size:50
 		this.recordService = recordService;
 	}
 
-	public Map<String, DataTypeMapping> addOrUpdateColumnIfNeeded(UUID instanceId, RecordType recordType,
```
calypsomatic, line 60: This code was moved to `RecordService.java`

davidangb (commented):
:+1: for the code changes in here - nice job!

What do you think about writing unit tests directly against `validateAndPermissions`? If we had nice thorough coverage against that method, it would ease the need for coverage against higher-level methods or APIs.

calypsomatic:
> 👍 for the code changes in here - nice job!
> 
> What do you think about writing unit tests directly against `validateAndPermissions`? If we had nice thorough coverage against that method, it would ease the need for coverage against higher-level methods or APIs.

Makes a lot of sense!  Let me try.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
```diff
@@ -116,4 +119,159 @@ private String convertSchemaDiffToErrorMessage(
...
+
+    public void validatePrimaryKey(UUID instanceId, RecordType recordType, Optional<String> primaryKey) {
+        if (primaryKey.isPresent() && !primaryKey.get().equals(recordDao.getPrimaryKeyColumn(recordType, instanceId))) {
+            throw new NewPrimaryKeyException(primaryKey.get(), recordType);
+        }
+    }
+
+    @WriteTransaction
+    public boolean deleteSingleRecord(UUID instanceId, RecordType recordType, String recordId) {
+        return recordDao.deleteSingleRecord(instanceId, recordType, recordId);
```
aaronkanzer, line 269: Do we need to check if the user `hasDeleteInstancePermission` here?

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
```diff
@@ -116,4 +119,159 @@ private String convertSchemaDiffToErrorMessage(
...
+    }
+
+    @WriteTransaction
+    public void deleteRecordType(UUID instanceId, RecordType recordType) {
```
aaronkanzer, line 273: Similar to: https://github.com/DataBiosphere/terra-workspace-data-service/pull/186/files#r1135887894

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
calypsomatic: `deleteInstancePermission` is permission to delete the entire instance.  Deleting a single record (or table) is just making a write/change *within* the instance.

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorSamTest.java
```diff
@@ -0,0 +1,323 @@
...
+
+    @AfterEach
+    void cleanUp() {
+        recordDao.dropSchema(INSTANCE);
+    }
+
+    @Test
+    void testUpdateRecordNoPermission() throws ApiException {
+
+        //create record type
```
aaronkanzer, line 78: There are a handful of repeated code used to `create record type` -- would you want to include some type of fixture function for these?

calypsomatic: See updated code!

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorSamTest.java
```diff
@@ -0,0 +1,95 @@
...
+    @Test
+    void testValidateAndPermissionWithPermission() throws ApiException {
+
+        given(mockResourcesApi.resourcePermissionV2(anyString(), anyString(), anyString()))
+                .willReturn(true);
+
+        assertDoesNotThrow(() -> recordOrchestratorService.validateAndPermissions(INSTANCE, VERSION),
+                "validateAndPermissions should not throw if caller has write permission in Sam"
+        );
+    }
```
davidangb, line 94: Could I poke for just one more unit test that ensures `validateAndPermissions()` will throw if `resourcePermissionV2()` throws? Something like the suggestion below.

To get this test to work you'll also have to change `Mockito.clearInvocations` to `Mockito.reset` in the BeforeEach, otherwise the `mockResourcesApi` will keep throwing for your other unit tests, too.

```suggestion
    }
    
    @Test
    void testValidateAndPermissionWhenException() throws ApiException {

        given(mockResourcesApi.resourcePermissionV2(anyString(), anyString(), anyString()))
                .willThrow(new ApiException(0, "intentional failure for unit test")); // 0 indicates a failed connection

        assertThrows(SamException.class,
                () -> recordOrchestratorService.validateAndPermissions(INSTANCE, VERSION),
                "validateAndPermissions should throw if caller does not have write permission in Sam");
    }
```

calypsomatic: Done!

davidangb (approved):
thank you!!

aaronkanzer (approved):

PR #188, opened 2023-03-16
AJ-679: move @bean producers into separate classes

davidangb:
Using `@SpringBootTest` without any qualifiers in our unit tests meant that each unit test instantiated the full set of available `@Bean`s, including the two Hikari datasource connection pools.

Given the number of unit tests we have, this eventually consumed the max connections (default 100) allowed by Postgres, and tests would intermittently fail with `Failed to obtain JDBC Connection`. In #184 this error was very easy to make happen, though it happened inconsistently on different unit tests. Additionally, raising the `max_connections` flag on Postgres makes the error stop, but that's a band-aid solution.

In this PR:
* separate the `@Bean`-producing methods into a few different `*Config` classes : `DataSourceConfig`, `DataTypeInfererConfig`, etc. Previously these all lived in `WorkspaceDataServiceApplication`. No changes to this code; just moving it.
* Update the `@SpringBootTest` annotation on some unit tests to specify only the config classes they need. Most importantly, remove `DataSourceConfig` from some of these unit tests so they don't instantiate a Hikari thread pool.
* Remove `@SpringBootTest` entirely from a few tests.

Discovered as part of AJ-679 and blocks that ticket; does not resolve that ticket.

### DIFF
--- a/service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
@@ -1,5 +1,6 @@
 package org.databiosphere.workspacedataservice.service;
 
+import org.databiosphere.workspacedataservice.dao.DataSourceConfig;
 import org.databiosphere.workspacedataservice.dao.RecordDao;
 import org.databiosphere.workspacedataservice.service.model.RecordTypeSchema;
 import org.databiosphere.workspacedataservice.service.model.exception.MissingObjectException;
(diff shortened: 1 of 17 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
```diff
@@ -1,5 +1,6 @@
 package org.databiosphere.workspacedataservice.service;
 
+import org.databiosphere.workspacedataservice.dao.DataSourceConfig;
 import org.databiosphere.workspacedataservice.dao.RecordDao;
```
calypsomatic, line 4: This is imported but not used.  Did you mean to use it here?

davidangb: good catch, fixed

aaronkanzer (approved):

PR #190, opened 2023-03-17
Aj 897 create sam resource

calypsomatic:
See [AJ-897](https://broadworkbench.atlassian.net/browse/AJ-897)
When WDS starts up, it will read the azure managed identity client id (see https://github.com/broadinstitute/cromwhelm/pull/213) and workspace id from the environment, call azure for an access token under the managed identity, and create a wds-instance resource in Sam as well as a default instance schema in WDS with the given workspaceId.
This PR modifies the SamDao and SamFactoryClient to accept a token for calls, passing in null when none is given.  Only this initializing call has a token; other calls to sam will be finding a token from the http header request.

[AJ-897]: https://broadworkbench.atlassian.net/browse/AJ-897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
```diff
@@ -86,6 +87,15 @@ public void deleteInstanceResource(UUID instanceId) {
         VoidSamFunction samFunction = () -> samClientFactory.getResourcesApi().deleteResourceV2(RESOURCE_NAME_INSTANCE, instanceId.toString());
         withSamErrorHandling(samFunction, "deleteInstanceResource");
     }
+    @Override
+    public boolean instanceResourceExists(UUID instanceId){
```
yuliadub, line 91: why do we need to use Sam to check if an instance exists? dont we only need to check when creating/deleting?

yuliadub:
I was thinking more about instance creation for local testing - and I was wondering, could we spin up the postgress image locally such that it already has an instance to begin with? that sort of saves us from needing to check sam permission to do local testing, since Sam permission is more of an integration test anyways. 

for mocking integration tests, all my thoughts are around some sort of hard coding somewhere which I am not sure would be ideal (and as I type them I am talking myself out of them :D) : 
- hard code a specific client id behavior to only work when running in dev(or test) environment. requires to have a concept of environment while code is running, not great since client id is visible and maybe could be leveraged in a bad way
- check if service is running against local host and if it is, skip auth checks (again not great since one may be tricked that service is running on localhost when it is not, so maybe this plus an env check?) or only allow this when no instance is present
- maintain a test managed identity that has a very low scope access to just create WDS instances when its called from gradle/github action that can have a client id/generate tokens that can be used with the code as is (not sure if it is possible to scope this down so it cant be abused...)

anyways, going to do more thinking on this...

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
calypsomatic: We only call `SamDao.instanceResourceExists` when creating or deleting a wds instance.   This checks for a wds-instance resource in Sam, which will manage permissions for the instance in WDS.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -0,0 +1,57 @@
...
+    public void initializeInstance() {
+        LOGGER.info("Default workspace id loaded as {}", workspaceId);
+
+        try {
+            UUID instanceId = UUID.fromString(workspaceId);
+            String token = managedIdentityDao.getAzureCredential();
+            // create `wds-instance` resource in Sam if it doesn't exist
+            if (!samDao.instanceResourceExists(instanceId, token)){
+                LOGGER.info("Creating wds-resource for workspaceId {}", workspaceId);
+                //TODO what should the parent id be?
```
davidangb, line 42: this is the default instance, so we use the same UUID for the `wds-resource` and the parent. Your call to `createInstanceResource()` is correct.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -0,0 +1,57 @@
...
+                samDao.createInstanceResource(instanceId, instanceId, token);
+            }
+            if (!instanceDao.instanceSchemaExists(instanceId)) {
+                instanceDao.createSchema(instanceId);
+                LOGGER.info("Creating default schema id succeeded for workspaceId {}", workspaceId);
+            }
```
davidangb, line 48: nit: I suggest adding `else` clauses to both of these `if`s, and log something like:
```
LOGGER.debug("Default schema for workspaceId {} already exists; skipping creation.", workspaceId);
```

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -0,0 +1,57 @@
...
+        } catch (IllegalArgumentException e) {
+            LOGGER.warn("Workspace id could not be parsed, a default schema won't be created. Provided id: {}", workspaceId);
+        } catch (
+        DataAccessException e) {
+            LOGGER.error("Failed to create default schema id for workspaceId {}", workspaceId);
+        }
+        //TODO what errors do i need to catch from sam
```
davidangb, line 55: See `HttpSamClientSupport` - you'll want to catch SamException, AuthenticationException, AuthorizationException.

service/src/main/java/org/databiosphere/workspacedataservice/dao/ManagedIdentityDao.java
```diff
@@ -0,0 +1,40 @@
...
+    //TODO what to do for local development
+    @Value("${CLIENT_ID:}")
+    private String clientId;
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ManagedIdentityDao.class);
+
+    /**
+     * The default credential will use the user assigned managed identity with the specified client ID.
+     */
+    public String getAzureCredential() {
```
davidangb, line 25: possibly add `StringUtils.isNotBlank(clientId)` validation before attempting to get the identity?

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
```diff
@@ -83,10 +98,27 @@ public void createInstanceResource(UUID instanceId, UUID parentWorkspaceId) {
...
+        return instanceResourceExists(instanceId, null);
+    }
+
+    @Override
+    public boolean instanceResourceExists(UUID instanceId, String token){
+        SamFunction<List<UserResourcesResponse>> samFunction = () -> samClientFactory.getResourcesApi()
+                .listResourcesAndPoliciesV2(RESOURCE_NAME_INSTANCE);
+        List<UserResourcesResponse> resources = withSamErrorHandling(samFunction, "instanceResourceExists");
+        return resources.stream().filter(resource -> resource.getResourceId().equals(instanceId.toString())).count() > 0;
+    }
```
davidangb, line 121: brainstorming a bit - it feels awkward to list all `wds-instance` resources from Sam, then filter the result looking for one particular id. Instead, what if you checked if the user has the "read" action on the specific instance? If that returns 404, then you can consider the resource as non-existent.

Internally to Sam, `listResourcesAndPoliciesV2` checks for permission on all resources of the specified type, so you'd be optimizing to instead only check for one specified resource.

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamConfig.java
```diff
@@ -5,6 +5,8 @@
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Primary;
+import org.springframework.context.annotation.Profile;
```
davidangb, line 9: nit: unused imports

service/src/test/java/org/databiosphere/workspacedataservice/InstanceInitializerBeanTest.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Nested
+    @TestPropertySource(properties = {"twds.instance.workspace-id="})
+    class NoWorkspaceIdTest {
+        @Test
+        void workspaceIDNotProvidedNoExceptionThrown() {
+            assertDoesNotThrow(() -> instanceInitializerBean.initializeInstance());
+        }
+    }
```
davidangb, line 85: for this and the previous test, you could `verify` that the mock's methods to create things were NOT called

service/src/test/java/org/databiosphere/workspacedataservice/InstanceInitializerBeanTest.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.List;
+import java.util.UUID;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.BDDMockito.given;
+
+@ActiveProfiles({"mock-sam","mock-instance-dao", "local"})
+@TestPropertySource(properties = {"twds.instance.workspace-id=90e1b179-9f83-4a6f-a8c2-db083df4cd03"})
+@SpringBootTest
```
davidangb, line 30: using #188 as an example, could you specify just the `classes` you need here? If you're using the mock instance dao, I don't think you need the database pools, for instance.

service/src/test/java/org/databiosphere/workspacedataservice/controller/CorsLiveMockMvcTest.java
```diff
@@ -5,6 +5,7 @@
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
 import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.ActiveProfiles;
```
davidangb, line 8: nit: unused import

service/src/test/java/org/databiosphere/workspacedataservice/controller/FullStackMDCRequestResponseTest.java
```diff
@@ -12,6 +12,7 @@
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
 import org.springframework.http.ResponseEntity;
+import org.springframework.test.context.ActiveProfiles;
```
davidangb, line 15: nit: unused import

service/src/test/java/org/databiosphere/workspacedataservice/dao/PostgresInstanceDaoTest.java
```diff
@@ -13,9 +13,9 @@ public class PostgresInstanceDaoTest {
...
     NamedParameterJdbcTemplate namedTemplate;
 
-    @Test
-    void workspaceIDNotProvidedNoExceptionThrown() {
-        assertDoesNotThrow(() -> new PostgresInstanceDao(namedTemplate, "UNDEFINED"));
-    }
+//    @Test
+//    void workspaceIDNotProvidedNoExceptionThrown() {
+//        assertDoesNotThrow(() -> new PostgresInstanceDao(namedTemplate, "UNDEFINED"));
+//    }
```
davidangb, line 19: nit: delete this entirely

service/src/test/java/org/databiosphere/workspacedataservice/dao/RecordDaoTest.java
```diff
@@ -22,6 +22,7 @@
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.test.context.ActiveProfiles;
```
davidangb, line 25: unused import

service/src/test/java/org/databiosphere/workspacedataservice/dao/RecordDaoTest.java
```diff
@@ -91,12 +92,12 @@ void cleanUp(){
...
-		LOGGER.info("Default workspace id loaded as {}", workspaceId);
-		UUID defaultInstanceId = UUID.fromString(workspaceId);
-		assertTrue(instanceDao.instanceSchemaExists(defaultInstanceId));
-	}
+//	@Test
+//	void defaultSchemaIsCreated() {
+//		LOGGER.info("Default workspace id loaded as {}", workspaceId);
+//		UUID defaultInstanceId = UUID.fromString(workspaceId);
+//		assertTrue(instanceDao.instanceSchemaExists(defaultInstanceId));
+//	}
```
davidangb, line 100: delete

service/src/test/java/org/databiosphere/workspacedataservice/dao/RecordDaoTest.java
```diff
@@ -111,9 +112,9 @@ void listInstances() {
 		List<UUID> actualInitialSchemas = instanceDao.listInstanceSchemas();
 
 		// check that the default schema exists - see also defaultSchemaIsCreated() above
-		UUID defaultInstanceId = UUID.fromString(workspaceId);
-		assertTrue(actualInitialSchemas.contains(defaultInstanceId),
-				"initial schema list should contain default instance");
+//		UUID defaultInstanceId = UUID.fromString(workspaceId);
+//		assertTrue(actualInitialSchemas.contains(defaultInstanceId),
+//				"initial schema list should contain default instance");
```
davidangb, line 117: delete

service/src/test/java/org/databiosphere/workspacedataservice/sam/BearerTokenFilterTest.java
```diff
@@ -10,6 +10,7 @@
 import org.springframework.mock.web.MockFilterChain;
 import org.springframework.mock.web.MockHttpServletRequest;
 import org.springframework.mock.web.MockHttpServletResponse;
+import org.springframework.test.context.ActiveProfiles;
```
davidangb, line 13: Ill just say you should check the whole PR for unused imports :)

davidangb (commented):
comments inline!

service/src/test/java/org/databiosphere/workspacedataservice/dao/PostgresInstanceDaoTest.java
calypsomatic: Yeah sorry, hadn't done cleanup yet.  Will do!

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamDao.java
```diff
@@ -25,6 +25,10 @@ public interface SamDao {
      * Sam action name for delete permission
      */
     String ACTION_DELETE = "delete";
+    /**
+     * Sam action name for delete permission
```
yuliadub, line 29: "read" instead of "delete" :)

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamClientFactory.java
```diff
@@ -54,11 +59,15 @@ private ApiClient getApiClient() {
...
-    public ResourcesApi getResourcesApi() {
-        ApiClient apiClient = getApiClient();
+    public ResourcesApi getResourcesApi(String token) {
+        ApiClient apiClient = getApiClient(token);
         ResourcesApi resourcesApi = new ResourcesApi();
         resourcesApi.setApiClient(apiClient);
         return resourcesApi;
     }
 
+    public ResourcesApi getResourcesApi() {
```
yuliadub, line 69: is there a specific reason to keep this vs just use getResourcesApi(token)?

calypsomatic: Probably not!  I was thinking I would leave it so I didn't have to refactor more than I had to, but I think this doesn't need to be used.  I'll take another look, thanks!

jordansafer (approved):

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -0,0 +1,64 @@
...
+                LOGGER.warn("No token acquired from azure managed identity; wds-instance resource and default schema not created");
+            }
+
+        } catch (IllegalArgumentException e) {
+            LOGGER.warn("Workspace id could not be parsed, a default schema won't be created. Provided id: {}", workspaceId);
+        } catch (
+        DataAccessException e) {
+            LOGGER.error("Failed to create default schema id for workspaceId {}", workspaceId);
+        } catch (SamException e) {
+            LOGGER.error("Exception thrown from sam, wds-instance resource and default schema not created : {}", e.getMessage());
```
davidangb, line 61: There is one corner case I can think of, in which a Sam `wds-instance` resource exists, but the managed identity does not have permission to it.

In this case, `samDao.instanceResourceExists()` will return false, so this code will attempt to call `samDao. createInstanceResource`, which will return … a 409.

Your code will log this with "Exception thrown from sam, wds-instance resource and default schema not created : A resource of this type and name already exists".

I'm just thinking out loud - do you think this corner case is worth a special logging statement or any special handling? I don't have a good sense if this will ever happen with any frequency.

davidangb (approved):
one TOL comment inline, otherwise looks good!!

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
calypsomatic: Hmm, in that case would we want to create the schema instance in wds?  It seems like if the managed identity doesn't have permission to the sam resource then it will continue to cause trouble, so I would guess no, in which case I think the current behavior is probably ok.

PR #192, opened 2023-03-20
Aj 816 - Adding WDS python client tests

yuliadub:
Corresponding github action: https://github.com/DataBiosphere/terra-workspace-data-service/actions/workflows/test-python-client.yml. All the tests run and pass (yay)

A few outstanding things:

- need to understand if any adjustments will need to be made once auth is required for creating instances (today I rely on local boot to create an instance for me if the env workspace_id is present)
- the test github action is stand alone but it is reusing some of the configurations other actions use, going to take a look and see if it makes sense to pull certain steps to be contained into common yamls. 
- There are likely more test cases to be added - tried to focus on the core functionality, looking for feedback if there is anything I have forgotten
- Kept all the tests/helpers/etc in a single python file to keep it simple. Let me know your thoughts on this vs separating into multiple files 
- any other suggestions/thoughts very welcome! 

jira link [here](https://broadworkbench.atlassian.net/browse/AJ-816).


.github/workflows/test-python-client.yml
```diff
@@ -0,0 +1,83 @@
+name: Test Python client
+
+on:
+  push:
+    branches:
```
yuliadub, line 5: will need to make sure I remove this before merging, should run on all branches when the yaml file is changed but that is it

.github/workflows/test-python-client.yml
```diff
@@ -0,0 +1,85 @@
...
+      - name: Set up JDK
+        uses: actions/setup-java@v2
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+          cache: 'gradle'
+
+      - name: Run WDS service and tests
+        run: |
+            export WORKSPACE_ID=c9ed46a9-fc9b-4130-a608-cf3760ed70a0
```
calypsomatic, line 80: This is just a random UUID?

yuliadub: correct - would it would make sense to generate a new one every time?

calypsomatic: Nah just checking

.github/workflows/test-python-client.yml
```diff
@@ -0,0 +1,85 @@
...
+        run: openapi-generator-cli version-manager set 4.3.1
+
+      - name: Generate Python client
+        run: |
+          openapi-generator-cli generate \
+          -i service/src/main/resources/static/swagger/openapi-docs.yaml \
+          -g python \
+          -o wds-client \
+          --additional-properties=projectName=wds-client,packageName=wds_client,packageVersion=1.0.0 \
+          --skip-validate-spec
```
davidangb, line 59: I agree with things you said in the PR description about consolidating workflows. I bet this could get consolidated with `release-python-client.yml` so that:
* we only generate the client once, and we're testing the same generated code that we would push to pypi
* we only publish a new client if the swagger defs changed
* we only publish a new client if the client tests pass

.github/workflows/test-python-client.yml
```diff
@@ -0,0 +1,85 @@
...
+          java-version: '17'
+          distribution: 'temurin'
+          cache: 'gradle'
+
+      - name: Run WDS service and tests
+        run: |
+            export WORKSPACE_ID=c9ed46a9-fc9b-4130-a608-cf3760ed70a0
+            ./gradlew --build-cache build -x test
+            ./gradlew bootRun &
+            sleep 5
```
davidangb, line 83: in my testing of the mock-sam stuff, `sleep 5` wasn't always long enough. Could just increase this, or do something fancy like polling the port until you get a 200 back - what do you think?

.github/workflows/test-python-client.yml
yuliadub: Edited so that tests run on pr creation/updates, and then publish only happens when changes are pushed/merged into main (tested this best I can i.e without merging to main and it should work) via workflow trigger

currently, we trigger the publish workflow when we have a new tag - a new tag is generated each time a new version of WDS is built. In some ways, it makes sense to keep this behavior since all changes (even ones not affecting the client code directly, but potentially affecting the kind of data may be returned from WDS) is always aligned on the same version number. 

alternatively, we can have the python client to generate its own version number, that is not related to WDS. After thinking about it more, I think my suggestion would be to keep the workflow as is, publishing a new version of client each time any code changes are made to WDS, not just specific 

I looked into what it would take to only trigger the python action if the yaml file changed, but havent yet figured out a simple way to do that.

calypsomatic (approved):
Looks great!

davidangb (approved):
love it, thank you!

PR #194, opened 2023-03-21
AJ-911: 'cached plan must not change result type' error

davidangb:
In this PR:
* add logging for retried transactions via `@ReadTransaction`/`@WriteTransaction`
* adds a new `@RetryableApi` annotation, which will retry the entire implementation of an API call if that call encountered a `SQLException` and uses the same logging implementation as transactions
* marks all non-streaming APIs with `@RetryableApi`
* adds a unit test for retrying transactions
* adds a unit test for concurrent schema changes, which is the root issue in AJ-911

The errors thrown by concurrent schema changes are not retried inside the transaction annotations, and I could not achieve success on the unit test by trying to add those exceptions to the transaction annotations, even if I upped the number/duration of retries there.

Retrying at the API level feels blunt and like a giant hammer. In the case of `SQLException`s that are valid and non-transient, this would also increase the latency for APIs because we retry multiple times and will get an error each time.

I would like to narrow down and precisely target the specifics of concurrent schema changes, possibly consider app-level solutions like synchronizing all Java methods that execute DDL, but that is proving to be a rabbit hole. Putting this PR up for review as-is with the hope of improving the implementation in a future PR.

### DIFF
--- a/service/src/main/java/bio/terra/common/retry/transaction/TransactionRetryInterceptorConfiguration.java
+++ b/service/src/main/java/bio/terra/common/retry/transaction/TransactionRetryInterceptorConfiguration.java
@@ -2,6 +2,7 @@ package bio.terra.common.retry.transaction;
 
 import bio.terra.common.retry.CompositeBackOffPolicy;
 import org.aopalliance.intercept.MethodInterceptor;
+import org.databiosphere.workspacedataservice.retry.RetryLoggingListener;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.classify.BinaryExceptionClassifier;
 import org.springframework.context.annotation.Bean;
@@ -11,8 +12,10 @@ import org.springframework.retry.backoff.BackOffPolicy;
 import org.springframework.retry.backoff.ExponentialBackOffPolicy;
 import org.springframework.retry.backoff.UniformRandomBackOffPolicy;
 import org.springframework.retry.interceptor.RetryInterceptorBuilder;
+import org.springframework.retry.interceptor.RetryOperationsInterceptor;
 import org.springframework.retry.policy.CompositeRetryPolicy;
 import org.springframework.retry.policy.SimpleRetryPolicy;
+import org.springframework.retry.support.RetryTemplate;
 
 import java.util.LinkedHashMap;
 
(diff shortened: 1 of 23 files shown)


service/src/main/java/bio/terra/common/retry/transaction/TransactionRetryInterceptorConfiguration.java
```diff
@@ -27,10 +30,18 @@ public class TransactionRetryInterceptorConfiguration {
...
+    BackOffPolicy backOffPolicy = createTransactionBackOffPolicy(config);
+
+    RetryTemplate retryTemplate = RetryTemplate.builder()
+            .customBackoff(backOffPolicy)
+            .customPolicy(retryPolicy)
+            .withListener(new RetryLoggingListener())
+            .build();
+
+    RetryOperationsInterceptor interceptor = RetryInterceptorBuilder.stateless().build();
+    interceptor.setRetryOperations(retryTemplate);
```
aaronkanzer, line 43: (Apologies if I'm misundertstanding how this exactly works) -- do all of these need to be created each time? Or could they be retrieved via a global variable somewhere and referenced in this function?

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -1,5 +1,6 @@
 package org.databiosphere.workspacedataservice.controller;
```
aaronkanzer, line 1: (Wasn't sure where to leave this comment) -- should `createInstance` and `deleteInstance` also have the `RetryableAPI` decorator?

I don't _think_ they are streaming endpoints (?)

davidangb: `createInstance` and `deleteInstance` are not idempotent in the case of an error, since there's no rollback for the Sam API calls - so I left the retry off those APIs.

service/src/main/java/bio/terra/common/retry/transaction/TransactionRetryInterceptorConfiguration.java
davidangb: Since `transactionRetryInterceptor` is a `@Bean`, it should be a Spring-managed singleton. I honestly don't know the exact behavior since both `@ReadTransaction` and `@WriteTransaction` specify `@Retryable(`interceptor = "transactionRetryInterceptor"` … it might still use just one bean for both, or it could create two separate singleton beans. Either way it shouldn't create all this stuff for each invocation of a transaction.

aaronkanzer: Sounds good -- yeah my only concern was it potentially not being a singleton -- 📈

jordansafer (approved):

PR #202, opened 2023-04-07
AJ-939 Recognize clone mode

calypsomatic:
See [AJ-939](https://broadworkbench.atlassian.net/browse/AJ-939)
Adds the expectation of an environment variable containing the source workspace id.  If this variable is present, WDS will start up in "clone mode" - i.e. initialize cloning of the source workspace id (to be developed later).
See also supporting PRs:
https://github.com/broadinstitute/rawls/pull/2316
https://github.com/DataBiosphere/leonardo/pull/3281
https://github.com/broadinstitute/cromwhelm/pull/232
https://github.com/broadinstitute/terra-helmfile/pull/3986

[AJ-939]: https://broadworkbench.atlassian.net/browse/AJ-939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -28,8 +31,39 @@ public InstanceInitializerBean(SamDao samDao, InstanceDao instanceDao, ManagedId
...
+                //by default this will continue and return false, thereby creating a default schema - do we want to do that?
+                //If we return true instead, we'll encounter errors down the line...
+            }
+        }
+        return false;
+
+    }
+
+    public void initCloneMode(){
+        LOGGER.info("Beginning clone...");
```
yuliadub, line 53: small nit - since cloning wont happen yet, maybe it makes sense to just print "Starting in clone mode" or something similar?

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -28,8 +31,39 @@ public InstanceInitializerBean(SamDao samDao, InstanceDao instanceDao, ManagedId
...
+    }
+
     public void initializeInstance() {
         LOGGER.info("Default workspace id loaded as {}", workspaceId);
+        LOGGER.info("Source workspace id loaded as {}", sourceWorkspaceId);
```
yuliadub, line 58: I had a somewhat crazy idea - would it be an awful idea to just overwrite default id to be the source id in clone mode? that way we dont have to do any renaming when we do the restore...? or could that cause issues down the line that I am not thinking of?

calypsomatic: Then it wouldn't match the workspace Id, which is the whole reason why we care which id it is.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
calypsomatic: Sure

service/src/test/java/org/databiosphere/workspacedataservice/InstanceInitializerWithSourceWorkspaceTest.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Test
+    void sourceWorkspaceIDCorrect() {
+        boolean cloneMode = instanceInitializerBean.isInClodeMode();
+        assert(cloneMode);
+    }
+
+//        @Test
+//        void sourceWorkspaceIDInvalid() {
```
calypsomatic, line 44: I have not had success having different property values in a single test class; in theory you should be able to do it with [inner classes](https://github.com/spring-projects/spring-framework/issues/18951#issuecomment-1153547515) but that has not worked for me.  I'm a bit reluctant to make a handful of extra classes just to test one property; if anyone has a solution I'd love to hear it.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
yuliadub: that is assuming that we do the comparison to know if current WDS is empty or not the first time WDS starts up? 

what I was thinking is, when WDS goes to start up for the first time, if there is a source workspace Id, we know that it is starting in clone mode - so why not make that the default id (I guess we can check before hand just in case if the WDS db is empty or not before doing this). Then, if the WDS restarts in the future, the flow would be something like: if there is a source workspace Id, start up in clone mode, check the state of back up/restore, if it is good - start up with no errors, if it is not, check state of db (is it running or corrupted) and either start it as empty or not start and show error.

Since in the end, once the backup/restore completes successfully and we rename the schema for the default instance, we would affectively be doing the same thing?

let me know if I am making any sense at all :D

calypsomatic: Either you're not making sense or I'm just dense.  The idea is, the difference between clone mode and non-clone mode should be that we set defaultId = sourceId?  But then we'd lose default id and won't be able to rename the schema to match the workspace.  Or what are you suggesting I do here?

davidangb: External callers like CBAS and Terra UI look for an instance where instanceid == workspaceid. If we used instanceid == sourceid here, those callers wouldn't find the data. I think it makes sense to rename the instance after cloning … or (future-looking) change our API signatures so it's optional for the caller to specify an instanceid, then we can handle defaulting on our own.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -28,8 +31,37 @@ public InstanceInitializerBean(SamDao samDao, InstanceDao instanceDao, ManagedId
         this.managedIdentityDao = managedIdentityDao;
     }
 
+    public boolean isInClodeMode() {
+        if (sourceWorkspaceId != null && !"".equals(sourceWorkspaceId.trim())){
```
davidangb, line 35: check out `StringUtils.isNotBlank` (in `org.apache.commons.lang3`) as a convenience here

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -28,8 +31,37 @@ public InstanceInitializerBean(SamDao samDao, InstanceDao instanceDao, ManagedId
...
+        LOGGER.info("Starting in clone mode...");
+    }
+
     public void initializeInstance() {
         LOGGER.info("Default workspace id loaded as {}", workspaceId);
+        LOGGER.info("Source workspace id loaded as {}", sourceWorkspaceId);
+        if (isInClodeMode())
+            initCloneMode();
+        initializeDefaultInstance(); //TODO Wrap this in an else once cloning is implemented
+    }
```
davidangb, line 62: re: your question about unit testing below … 

if you accepted an argument to `isInCloneMode` like:
```java
    public boolean isInClodeMode(String input) {
        if (StringUtils.isNotBlank(input)) {
```
and then passed in the value from here:
```java
if (isInClodeMode(sourceWorkspaceId))
```
then `isInClodeMode()` would be unit-testable without needing to muck with `@TestPropertySource(properties …`

davidangb (commented):
lgtm, some comments on style and unit-testability

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -28,8 +32,38 @@ public InstanceInitializerBean(SamDao samDao, InstanceDao instanceDao, ManagedId
...
+        if (StringUtils.isNotBlank(sourceWorkspaceId)){
+            LOGGER.info("Source workspace id found, checking database");
+            try {
+                UUID.fromString(sourceWorkspaceId);
+                //TODO: this is a placeholder for checking that the db has already been cloned;
+                //In the future it could check the clone status,
+                //But for now assuming if we've created a workspace schema, work is done
+                return !instanceDao.instanceSchemaExists(UUID.fromString(workspaceId));
+            } catch (IllegalArgumentException e) {
+                LOGGER.warn("Workspace id could not be parsed, unable to clone DB. Provided source workspace id: {}, default workspace id: {}", sourceWorkspaceId, workspaceId);
```
calypsomatic, line 45: I realized I don't distinguish here between whether the `sourceWorkspaceId` or the default `workspaceId` fails - presumably we want different behavior so I'd need two `try`-`catch`es, but I'm waffling over exactly what behavior.  If `sourceWorkspaceId` cannot be parsed, do we throw an error and abort cloning as well as not creating default schema, or do we simply create an empty default schema?  What about if the default `workspaceId` can't be parsed?

davidangb: my take on this:
* if `workspaceId` is invalid, neither create a default instance nor attempt any clone operations. In #206 we may end up not even starting WDS in this case.
* if `workspaceId` is good, but `sourceWorkspaceId` is missing or invalid, create an empty default instance (but don't do any clone stuff)
* if `workspaceId` and `sourceWorkspaceId` are both good, attempt all the cloning stuff, rename the `sourceWorkspaceId` instance to `workspaceId` as the last step of the clone

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -28,8 +32,42 @@ public InstanceInitializerBean(SamDao samDao, InstanceDao instanceDao, ManagedId
         this.managedIdentityDao = managedIdentityDao;
     }
 
+    public boolean isInClodeMode(String sourceWorkspaceId) {
```
aaronkanzer, line 35: This function never returns `true` -- is the goal to have it return true in where `return !instanceDao.instanceSchemaExists(UUID.fromString(workspaceId));` is called as a `TODO` in this function?

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -28,8 +32,42 @@ public InstanceInitializerBean(SamDao samDao, InstanceDao instanceDao, ManagedId
...
+                LOGGER.warn("Workspace id could not be parsed, unable to clone DB. Provided default workspace id: {}", workspaceId);
+                return false;
+            }
+        }
+        return false;
+
+    }
+
+    public void initCloneMode(){
+        LOGGER.info("Starting in clone mode...");
```
aaronkanzer, line 59: Is this a placeholder for something? Otherwise, can we just move the logger statement to a pre-existing function?

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
calypsomatic: If there's not already a default schema then it will return true but in the future we might use the cloning status instead.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
calypsomatic: Yes, placeholder!  In later PRs it will actually do the clone!

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
aaronkanzer, line 35: Minor typo as well `isInCloneMode` instead of `isInClodeMode`

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -28,8 +32,42 @@ public InstanceInitializerBean(SamDao samDao, InstanceDao instanceDao, ManagedId
...
+                //TODO: this is a placeholder for checking that the db has already been cloned;
+                //In the future it could check the clone status,
+                //But for now assuming if we've created a workspace schema, work is done
+                return !instanceDao.instanceSchemaExists(UUID.fromString(workspaceId));
+            } catch (IllegalArgumentException e) {
+                LOGGER.warn("Workspace id could not be parsed, unable to clone DB. Provided default workspace id: {}", workspaceId);
+                return false;
+            }
+        }
+        return false;
```
aaronkanzer, line 54: Should we throw a log statement here if we get to this point? Kinda signals something spooky if we never got a `true` or `false` or an `Exception` prior?

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
calypsomatic: Ha!  And that's propagated everywhere!  How did I not see it??

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
calypsomatic: Good idea

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
yuliadub: +1 on what David said!

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
yuliadub: ignore everything I was trying to say :D

davidangb (approved):
lgtm, thank you!

yuliadub (approved):

yuliadub:
lgtm too! thanks for getting all this figured out!

PR #203, opened 2023-04-10
AJ-963: temporarily disable Sam operations

davidangb:
Temporarily disable all Sam resource operations until AJ-964 and WM-1862 land.

This PR replaces the `ResourcesApi` from the Sam client with a noop class that returns true for all permission checks and does not create or delete any Sam resources.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/sam/SamConfig.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/sam/SamConfig.java
@@ -17,6 +17,9 @@ public class SamConfig {
     @Value("${SAM_URL:}")
     private String samUrl;
 
+    @Value("${sam.enabled:true}")
+    private boolean isSamEnabled;
+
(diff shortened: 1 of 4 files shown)


yuliadub (approved):

cahrens (approved):

calypsomatic (approved):

davidangb:
@cahrens I pushed a new commit after you approved, if you care - thank you!

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamConfig.java
```diff
@@ -17,6 +17,9 @@ public class SamConfig {
     @Value("${SAM_URL:}")
     private String samUrl;
 
+    @Value("${sam.enabled:true}")
```
cahrens, line 20: It seems like it might be safer to make the default value false.

yuliadub (approved):

PR #206, opened 2023-04-12
AJ-964: use workspace Sam resource, not wds-instance

davidangb:
In this PR:
* don't create or delete `wds-instance` Sam resources when creating/deleting WDS instances, including at startup time
* write APIs check permissions against the `workspace` Sam resource, not against `wds-instance` child resources
* we always check permission against the workspace specified in the `WORKSPACE_ID` env var, regardless of the WDS instance id
* _**remove the workspaceId query parameter from the create-instance API, since it is now obsolete and we rely on the `WORKSPACE_ID` env var instead**_
* if the `WORKSPACE_ID` env var can't be parsed to a UUID at startup, inject a `FailingSamDao` class instead of `HttpSamDao` - this will return false for all permission checks, log the failure reasons, and show the failure reason in WDS' `/status` API
* simplify a bunch of method signatures since we don't need the instanceId anymore when calling Sam

Impacts:
* You must be a workspace writer to upload a TSV, write or delete records, or create a WDS instance 
* You must be a workspace owner to delete a WDS instance
* _**As a developer running WDS locally, you'll want to set the `WORKSPACE_ID` env var or the `twds.instance.workspace-id` Spring property to the UUID of one of your workspaces**_
* If the `WORKSPACE_ID` env var is not set correctly, all write APIs will fail with auth errors

Open questions:
* ~I've left `hasCreateInstancePermission`, `hasDeleteInstancePermission`, and `hasWriteInstancePermission` as separate methods, though we could collapse them all into one method. Should we collapse these?~ Decision: leave these as-is
* ~since we're always checking the same workspaceid, does it even make sense to pass the instanceid into the Sam methods? Should we clean that up and simplify signatures?~ Decision: these are now cleaned up.
* ~should we remove `ManagedIdentityDao` and the `azure-identity` library entirely? Will we need those before too long when we get to multi-user WDS and have an app identity?~ Decision: leave this in place; we'll need it soon
* ~I'm not satisfied with the explosion in complexity of Sam integration - all the different Sam beans and configuration settings. I'd like to simplify that and make it easier to understand … is that worth doing in this PR, or as a follow-on?~ Decision: do this in a future PR - maybe after we re-enable Sam

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
@@ -111,9 +111,8 @@ public class RecordController {
 
 	@PostMapping("/instances/{version}/{instanceId}")
 	public ResponseEntity<String> createInstance(@PathVariable("instanceId") UUID instanceId,
-			@PathVariable("version") String version,
-			@RequestParam(name= "workspaceid", required = false) Optional<UUID> workspaceId) {
-		instanceService.createInstance(instanceId, version, workspaceId);
+			@PathVariable("version") String version) {
+		instanceService.createInstance(instanceId, version);
 		return new ResponseEntity<>(HttpStatus.CREATED);
 	}
 
--- a/service/src/main/java/org/databiosphere/workspacedataservice/sam/FailingSamDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/sam/FailingSamDao.java
@@ -1,0 +1,73 @@
+package org.databiosphere.workspacedataservice.sam;
+
+import org.broadinstitute.dsde.workbench.client.sam.model.SystemStatus;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.UUID;
+
+/**
+ * This class will be used instead of HttpSamDao if the WORKSPACE_ID env var
+ * contains a non-UUID value, including being blank.
+ * <p>
+ * This class will fail all permission checks and log a warning each time. Its
+ * objective is to disallow any write operations that require permissions, while still
+ * allowing WDS to start up and for users to read any data already in WDS.
+ */
+public class FailingSamDao implements SamDao {
+
+    private final String errorMessage;
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    public FailingSamDao(String errorMessage) {
+        this.errorMessage = errorMessage;
+    }
+
+    @Override
+    public boolean hasCreateInstancePermission(UUID parentWorkspaceId) {
+        logWarning();
+        return false;
+    }
+
+    @Override
+    public boolean hasCreateInstancePermission(UUID parentWorkspaceId, String token) {
+        logWarning();
+        return false;
+    }
+
+    @Override
+    public boolean hasDeleteInstancePermission(UUID instanceId) {
+        logWarning();
+        return false;
+    }
+
+    @Override
+    public boolean hasDeleteInstancePermission(UUID instanceId, String token) {
+        logWarning();
+        return false;
+    }
+
+    @Override
+    public boolean hasWriteInstancePermission(UUID instanceId) {
+        logWarning();
+        return false;
+    }
+
+    @Override
+    public boolean hasWriteInstancePermission(UUID instanceId, String token) {
+        logWarning();
+        return false;
+    }
+
+    @Override
+    public SystemStatus getSystemStatus() {
+        throw new RuntimeException("Sam integration failure: " + errorMessage);
+    }
+
+    private void logWarning() {
+        logger.warn("Sam permission check failed. {}", errorMessage);
+    }
+
+
+}
--- a/service/src/main/java/org/databiosphere/workspacedataservice/sam/FailingSamDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/sam/FailingSamDao.java
@@ -4,8 +4,6 @@ import org.broadinstitute.dsde.workbench.client.sam.model.SystemStatus;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.UUID;
-
 /**
  * This class will be used instead of HttpSamDao if the WORKSPACE_ID env var
  * contains a non-UUID value, including being blank.
@@ -25,37 +23,37 @@ public class FailingSamDao implements SamDao {
     }
 
     @Override
-    public boolean hasCreateInstancePermission(UUID parentWorkspaceId) {
+    public boolean hasCreateInstancePermission() {
         logWarning();
         return false;
     }
 
     @Override
-    public boolean hasCreateInstancePermission(UUID parentWorkspaceId, String token) {
+    public boolean hasCreateInstancePermission(String token) {
         logWarning();
         return false;
     }
 
     @Override
-    public boolean hasDeleteInstancePermission(UUID instanceId) {
+    public boolean hasDeleteInstancePermission() {
         logWarning();
         return false;
     }
 
     @Override
-    public boolean hasDeleteInstancePermission(UUID instanceId, String token) {
+    public boolean hasDeleteInstancePermission(String token) {
         logWarning();
         return false;
     }
 
     @Override
-    public boolean hasWriteInstancePermission(UUID instanceId) {
+    public boolean hasWriteInstancePermission() {
         logWarning();
         return false;
     }
 
     @Override
-    public boolean hasWriteInstancePermission(UUID instanceId, String token) {
+    public boolean hasWriteInstancePermission(String token) {
         logWarning();
         return false;
     }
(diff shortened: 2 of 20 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/sam/FailingSamDao.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Override
+    public boolean hasCreateInstancePermission(UUID parentWorkspaceId, String token) {
+        logWarning();
+        return false;
+    }
+
+    @Override
+    public boolean hasDeleteInstancePermission(UUID instanceId) {
```
aaronkanzer, line 40: Per your question in the description -- I like the idea of separating `DELETE` from all other write privileges -- just feels a bit more safe in terms of granting something that could have major implications like being able to delete an entire instance.

service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
```diff
@@ -50,11 +49,12 @@ public List<UUID> listInstances(String version) {
     public void createInstance(UUID instanceId, String version, Optional<UUID> workspaceId) {
         validateVersion(version);
 
-        UUID samResourceId = instanceId; // id of "wds-instance" Sam resource we will create
-        UUID samParentResourceId = workspaceId.orElse(instanceId); // id of "workspace" Sam resource to use as the parent of "wds-instance"
+        // id of workspace in which this instance exists
+        // if not specified, assume the workspace id is equal to the instance id
+        UUID containingWorkspaceId = workspaceId.orElse(instanceId);
```
aaronkanzer, line 54: Mini-nit -- what is `containing` mean exactly in this case? Is it the same as `current`, etc?

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
```diff
@@ -62,55 +60,18 @@ public boolean hasDeleteInstancePermission(UUID instanceId) {
     }
     @Override
     public boolean hasDeleteInstancePermission(UUID instanceId, String token) {
-        return hasPermission(RESOURCE_NAME_INSTANCE, instanceId.toString(), ACTION_DELETE,
+        return hasPermission(RESOURCE_NAME_WORKSPACE, instanceId.toString(), ACTION_WRITE,
```
aaronkanzer, line 63: Just flagging in terms of if we want to separate `ACTION_DELETE` and `ACTION_WRITE`

calypsomatic:
> Open questions:
> 
> * since we're always checking the same workspaceid, does it even make sense to pass the instanceid into the Sam methods? Should we clean that up and simplify signatures?
I think it makes sense to simplify.
> * should we remove `ManagedIdentityDao` and the `azure-identity` library entirely? Will we need those before too long when we get to multi-user WDS and have an app identity?
If we need them we can maybe get them back from git history - but that might be a bit of a hassle.  Leaving them there presumably doesn't do any harm.
> * I'm not satisfied with the explosion in complexity of Sam integration - all the different Sam beans and configuration settings. I'd like to simplify that and make it easier to understand … is that worth doing in this PR, or as a follow-on?
Follow on PR!

service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
davidangb: correct - "the workspace which contains this WDS". I can try to find a better name.

service/src/main/java/org/databiosphere/workspacedataservice/sam/FailingSamDao.java
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * This class will be used instead of HttpSamDao if the WORKSPACE_ID env var
+ * contains a non-UUID value, including being blank.
+ * <p>
+ * This class will fail all permission checks and log a warning each time. Its
+ * objective is to disallow any write operations that require permissions, while still
+ * allowing WDS to start up and for users to read any data already in WDS.
+ */
+public class FailingSamDao implements SamDao {
```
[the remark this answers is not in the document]
davidangb: No, Sam is not down entirely for this class. I'm injecting this bean if we can't parse the WORKSPACE_ID env var, which in turn means we can't determine which Sam resource to query for permissions. Instead of sending queries to Sam for some dummy resourceID like "n/a", I've substituted this bean which bypasses the Sam queries and just returns false. I tried to explain this in the javadoc for the class - I can work to rewrite that and make it clearer.

service/src/main/java/org/databiosphere/workspacedataservice/sam/FailingSamDao.java
```diff
@@ -0,0 +1,73 @@
...
+
+    private final String errorMessage;
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    public FailingSamDao(String errorMessage) {
+        this.errorMessage = errorMessage;
```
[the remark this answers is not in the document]
davidangb: There are two callers to this constructor in `SamConfig`, which pass in either the workspaceId or an exception message - do you think those need changing?

service/src/main/java/org/databiosphere/workspacedataservice/sam/FailingSamDao.java
davidangb: To be concrete, are you suggesting we check the "delete" action on the workspace resource instead of the "write" action to see if users are allowed to delete a WDS instance? Looking at Sam's resource.conf, this means that only workspace owners will be able to delete WDS instances. Workspace writers will be denied. Is that what you're thinking?

https://github.com/broadinstitute/sam/blob/8e252402449bb929e248698aaa8e6cd620bf2088/src/main/resources/reference.conf#L51

yuliadub: I also like the idea of separating delete and I think only owners should be able to delete wds instances.

yuliadub:
> Open questions:
> 
> * I've left `hasCreateInstancePermission`, `hasDeleteInstancePermission`, and `hasWriteInstancePermission` as separate methods, though we could collapse them all into one method. Should we collapse these?
> * since we're always checking the same workspaceid, does it even make sense to pass the instanceid into the Sam methods? Should we clean that up and simplify signatures?
> * should we remove `ManagedIdentityDao` and the `azure-identity` library entirely? Will we need those before too long when we get to multi-user WDS and have an app identity?
> * I'm not satisfied with the explosion in complexity of Sam integration - all the different Sam beans and configuration settings. I'd like to simplify that and make it easier to understand … is that worth doing in this PR, or as a follow-on?

- I think we should keep as is if permissions between create/delete/write are different. however, if permissions are the same then collapsing them would make sense 
- I saw simplify
- I think keep those since we likely will need them soon enough
- I say follow-on, it is not so overly complex that I think it needs to happen now

service/src/main/java/org/databiosphere/workspacedataservice/sam/FailingSamDao.java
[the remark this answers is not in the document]
aaronkanzer: Hmmm I don't know if those need to change -- was just thinking in the context of providing more specific information to us than just an error message

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -111,9 +111,8 @@ public ResponseEntity<List<UUID>> listInstances(@PathVariable("version") String
 
 	@PostMapping("/instances/{version}/{instanceId}")
 	public ResponseEntity<String> createInstance(@PathVariable("instanceId") UUID instanceId,
-			@PathVariable("version") String version,
-			@RequestParam(name= "workspaceid", required = false) Optional<UUID> workspaceId) {
-		instanceService.createInstance(instanceId, version, workspaceId);
+			@PathVariable("version") String version) {
```
calypsomatic, line 114: Do we anticipate any issues with changing the params for this API?  Was anybody using that?

yuliadub: I think its unlikely this will have impact since we haven't really advertised that you can have more than one instance of WDS (and UI only shows one) - I did include a call in the python docs to list instances but didnt include anything about creating/deleting them. 

But this makes think of what versioning looks like since in my mind this is a bit of a "breaking" change even if it doesnt have impact yet

yuliadub (approved):

yuliadub:
lgtm, thank you for the thorough comments on everything!

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
calypsomatic: I agree that it's not really going to be a problem... but we sure could stand to deal with versioning better.

calypsomatic (approved):

PR #210, opened 2023-04-14
Spring Boot 3, openapi-generator, okhttp-gson

davidangb:
DRAFT, work in progress; posting to get opinions and initial thoughts.

1. Upgrade to Spring Boot 3. One of the biggest changes in Spring Boot 3 / Spring 6 is the move from Java EE to Jakarta EE: https://www.baeldung.com/spring-boot-3-spring-6-new#jakarta-ee-9.
2. The move to Jakarta EE causes problems with generating the Java client via swagger-codegen, so this PR also switches to openapi-generator which is more modern anyway, and is what we use to generate the Python client.
3. Since we're making notable changes, also switch the Java client from `jersey2` to `okhttp-gson`, which is the default for openapi-generator and is the same library used by the Sam client.

TODO:
- [x] resolve issues with logging and Spring Boot 3 - necessary to keep supporting Sentry. The pre-existing color coding in our logback-spring.xml files breaks in Spring Boot 3.
- [x] resolve new ProblemDetail response structure. Should we align on ProblemDetail by using custom exception handler, and update the swagger definition? Should we revert to not using ProblemDetail at all? Suggestion: disable ProblemDetail for now, evaluate turning it on in a follow-on PR
- [x] is it possible to revert the argument-order changes in the client? Otherwise, users of the client - CBAS - will need to update their code to use the new client. You can see the argument-order changes in GeneratedClientTests.java.
- [ ] resolve issues with `datarepo-client`, which still needs Java EE, not Jakarta EE
- [ ] code cleanup and self-review
- [ ] manual testing


service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -625,14 +625,6 @@ components:
...
           "arrayFile": ["drs://drs.example.org/file_id_1", "https://account_name.blob.core.windows.net/container-2/blob2"]
         }
-    RecordAttributeDefinition:
-      required:
-        - name
-      type: object
-      properties:
-        name:
-          type: string
-          description: Attribute name
```
davidangb, line 635: this was unused

davidangb:
> Why did the argument order change?

openapi-generator has a different default ordering than swagger-codegen, specifically around placing form/body parameters last instead of first. I found the setting - `prependFormOrBodyParameters: "true"` - that changes argument order to be the same as it was before, and added it to this PR!

client/swagger.gradle
```diff
@@ -1,50 +1,45 @@
...
+	implementation 'jakarta.ws.rs:jakarta.ws.rs-api:3.1.0'
 }
 
+openApiGenerate {
+	inputSpec = "$rootDir//service/src/main/resources/static/swagger/openapi-docs.yaml".toString()
+	outputDir = "$buildDir/generated".toString()
+	generatorName = 'java'
+	library = 'okhttp-gson' // the default
+	configOptions = [
+			// useRuntimeException: "true", // TODO: a change, but a good one?
```
aaronkanzer, line 22: Just for clarity, what would this change give us in the context of WDS?

client/swagger.gradle
```diff
@@ -1,50 +1,45 @@
...
-	]
+			prependFormOrBodyParameters: "true", // orders arguments in the Java client same as previous swagger-codegen
 
-	rawOptions = ["--ignore-file-override", "${projectDir}/.swagger-codegen-ignore"]
+			dateLibrary: "java8" // TODO: this is the default; delete?
+	]
+	invokerPackage = "${artifactGroup}.client".toString()
+	modelPackage = "${artifactGroup}.model".toString()
+	apiPackage = "${artifactGroup}.api".toString()
+//	ignoreFileOverride = '${projectDir}/.swagger-codegen-ignore'
```
aaronkanzer, line 34: Just flagging that this is commented out -- wasn't sure if you want to delete or not

service/src/test/java/org/databiosphere/workspacedataservice/GeneratedClientTests.java
```diff
@@ -5,7 +5,7 @@
 import org.databiosphere.workspacedata.api.SchemaApi;
 import org.databiosphere.workspacedata.client.ApiClient;
 import org.databiosphere.workspacedata.client.ApiException;
-import org.databiosphere.workspacedata.model.RecordAttributes;
+//import org.databiosphere.workspacedata.model.RecordAttributes;
```
aaronkanzer, line 8: Can we fully delete?

service/src/test/java/org/databiosphere/workspacedataservice/controller/FullStackRecordControllerTest.java
```diff
@@ -103,7 +103,7 @@ void testBadRecordTypeNames() throws JsonProcessingException {
 					"/{instanceId}/records/{version}/{recordType}/{recordId}", HttpMethod.PUT, requestEntity,
 					ErrorResponse.class, instanceId, versionId, badName, "sample_1");
 			ErrorResponse err = response.getBody();
-			assertThat(err.getMessage()).containsPattern("Record Type .* or contain characters besides letters");
+//			assertThat(err.getMessage()).containsPattern("Record Type .* or contain characters besides letters");
```
aaronkanzer, line 106: Can we delete as well?

service/src/main/java/org/databiosphere/workspacedataservice/controller/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,38 @@
...
+        return ResponseEntity.badRequest().body(problemDetail);
+    }
+
+    @Override
+    protected ResponseEntity<Object> handleNoHandlerFoundException(NoHandlerFoundException ex, HttpHeaders headers, HttpStatusCode status, WebRequest request) {
+        System.err.println("OVERRIDDEN handleNoHandlerFoundException RUNNING ... ");
+        ProblemDetail problemDetail = ProblemDetail.forStatusAndDetail(HttpStatus.NOT_FOUND, ex.getMessage());
+        return new ResponseEntity<>(problemDetail, HttpStatus.NOT_FOUND);
+
+    }
```
davidangb, line 36: new in Spring Boot 3 / Spring Framework 6 is support for problem details; see `https://github.com/spring-projects/spring-framework/issues/27052` and https://datatracker.ietf.org/doc/html/rfc7807.

From my testing, it seems that problem details are used in most Spring errors, but certain errors create the old-style errors. This exception handler catches two of those casess - 404s when a route doesn't match, and 400s when a request parameter can't be cast to the expected type - and returns problem details for them.

This is necessary for our swagger definition - and therefore our clients - to properly interpret the response payload.

As an alternative, we could disable problem details and keep the old payloads (though I wonder if we'll have the reverse case and see a few places where problem details slip through)

davidangb: Update: I've disabled Problem Details as a whole; now, this exception handler is only responsible for generating the same error message for `MethodArgumentTypeMismatchException` as Spring Boot 2 did.

davidangb:
superseded by #440

PR #211, opened 2023-04-18
AJ-996 Adding ability to launch child process

yuliadub:
Adding ability for WDS to launch a child process that will be used by back/restore for cloning. 

Jira ticket [here](https://broadworkbench.atlassian.net/browse/AJ-996).

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
@@ -1,5 +1,7 @@
 package org.databiosphere.workspacedataservice.process;
 
+import org.databiosphere.workspacedataservice.service.model.exception.LaunchProcessException;
+
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
--- a/service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
@@ -12,7 +12,7 @@ import java.util.List;
 import java.util.Map;
 
 /** This class provides utility methods for launching local child processes. */
-class LocalProcessLauncher {
+public class LocalProcessLauncher {
     private Process process;
 
     public enum Output {
--- a/service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
@@ -13,9 +13,8 @@ import java.util.List;
 import java.util.Map;
 
 /** This class provides utility methods for launching local child processes. */
-public class LocalProcessLauncher {
+class LocalProcessLauncher {
     private Process process;
-    private static final Logger LOGGER = LoggerFactory.getLogger(LocalProcessLauncher.class);
 
     public enum Output {
         OUT,
@@ -58,7 +57,6 @@ public class LocalProcessLauncher {
         try {
             process = procBuilder.start();
         } catch (IOException ioEx) {
-            LOGGER.error("Error launching local process", ioEx);
             throw new LaunchProcessException("Error launching local process", ioEx);
         }
     }
@@ -80,7 +78,6 @@ public class LocalProcessLauncher {
 
             return processOutput.toString().trim();
         } catch (IOException ioEx) {
-            LOGGER.error("Error streaming output of child process", ioEx);
             throw new LaunchProcessException("Error streaming output of child process", ioEx);
         }
     }
@@ -102,7 +99,6 @@ public class LocalProcessLauncher {
         try {
             return process.waitFor();
         } catch (InterruptedException intEx) {
-            LOGGER.error("Error waiting for child process to terminate", intEx);
             Thread.currentThread().interrupt();
             throw new LaunchProcessException("Error waiting for child process to terminate", intEx);
         }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
@@ -21,11 +21,6 @@ class LocalProcessLauncher {
     }
 
     /**
-     * LocalProcessLauncher will be used to run processes.
-     */
-    public LocalProcessLauncher() {}
-
-    /**
      * Executes a command in a separate process from the current working directory (i.e. the same
      * place as this Java process is running).
      *
--- a/service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
@@ -1,5 +1,8 @@
 package org.databiosphere.workspacedataservice.process;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
@@ -12,6 +15,7 @@ import java.util.Map;
 /** This class provides utility methods for launching local child processes. */
 public class LocalProcessLauncher {
     private Process process;
+    private static final Logger LOGGER = LoggerFactory.getLogger(LocalProcessLauncher.class);
 
     public enum Output {
         OUT,
@@ -54,7 +58,8 @@ public class LocalProcessLauncher {
         try {
             process = procBuilder.start();
         } catch (IOException ioEx) {
-            throw new RuntimeException("Error launching local process", ioEx);
+            LOGGER.error("Error launching local process", ioEx);
+            throw new LaunchProcessException("Error launching local process", ioEx);
         }
     }
 
@@ -70,12 +75,13 @@ public class LocalProcessLauncher {
             String line;
             StringBuilder processOutput = new StringBuilder();
             while ((line = bufferedReader.readLine()) != null) {
-                processOutput.append(line + System.lineSeparator());
+                processOutput.append(line).append(System.lineSeparator());
             }
 
             return processOutput.toString().trim();
         } catch (IOException ioEx) {
-            throw new RuntimeException("Error streaming output of child process", ioEx);
+            LOGGER.error("Error streaming output of child process", ioEx);
+            throw new LaunchProcessException("Error streaming output of child process", ioEx);
         }
     }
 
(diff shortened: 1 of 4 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
```diff
@@ -0,0 +1,106 @@
...
+     * @param command the command and arguments to execute
+     * @param envVars the environment variables to set or overwrite if already defined
+     * @param workingDirectory the working directory to launch the process from
+     */
+    public void launchProcess(
+            List<String> command, Map<String, String> envVars, Path workingDirectory) {
+        // build and run process from the specified working directory
+        ProcessBuilder procBuilder = new ProcessBuilder(command);
+        if (workingDirectory != null) {
+            procBuilder.directory(workingDirectory.toFile());
```
calypsomatic, line 50: Is there a default working directory?  Or if `workingDirectory` is null will it just end up throwing an error when it tries to run without one?

service/src/test/java/org/databiosphere/workspacedataservice/process/LocalProcessLaunchTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class LocalProcessLaunchTest {
```
calypsomatic, line 13: I'd like to see a test where something goes wrong to see that it errors correctly

service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
yuliadub: the default working directory, if not specified would be where the service executable is being run from, so it will still run in the case where process doesnt use any specific files that it wont be able to find based on directory context.

service/src/test/java/org/databiosphere/workspacedataservice/process/LocalProcessLaunchTest.java
```diff
@@ -0,0 +1,89 @@
...
+        // block until the child process exits
+        int exitCode = localProcessLauncher.waitForTerminate();
+        assertEquals(0, exitCode);
+        assertThat(output).isEqualTo("Hello World");
+        assertThat(error).isEmpty();
+    }
+
+    @Test
+    void runSimpleCommandCauseError() {
+        List<String> processCommand = new ArrayList<>();
```
calypsomatic, line 57: This test failed when I ran it locally - because the LocalProcessLauncher threw an error, as it should.  I think maybe the GHA runner actually does have pg_dump installed but I don't locally?  May need to try something else that won't run correctly, and then you can verify the method throws an error with something like `assertThrows(LocalProcessLaunchTest.class,
                () ->localProcessLauncher.launchProcess(processCommand, null))
        );`.  See .e.g [here](https://github.com/DataBiosphere/terra-workspace-data-service/blob/f53b1dfc2b69fdc81479012596ee923dd029586f/service/src/test/java/org/databiosphere/workspacedataservice/service/InstanceServiceNoPermissionSamTest.java#L64) for an example

service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
```diff
@@ -0,0 +1,108 @@
...
+
+    public enum Output {
+        OUT,
+        ERROR
+    }
+
+    /**
+     * LocalProcessLauncher will be used to run processes.
+     */
+    public LocalProcessLauncher() {}
```
calypsomatic, line 26: [Sonarcloud is mad](https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&sinceLeakPeriod=true&branch=aj-996-launch-process&id=DataBiosphere_workspace-data-service) about this line - I think you can just remove it; if the constructor doesn't take any parameters or do anything you don't need to include it in the class.

service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
```diff
@@ -0,0 +1,108 @@
...
+            StringBuilder processOutput = new StringBuilder();
+            while ((line = bufferedReader.readLine()) != null) {
+                processOutput.append(line).append(System.lineSeparator());
+            }
+
+            return processOutput.toString().trim();
+        } catch (IOException ioEx) {
+            throw new LaunchProcessException("Error streaming output of child process", ioEx);
+        }
+    }
```
davidangb, line 85: could `getOutputForProcessFromStream()` be private? It looks like it is only called from `getOutputForProcess()` - is there a reason for it to be public?

service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
```diff
@@ -0,0 +1,108 @@
...
+     *
+     *  @param type specifies which process stream to get data from
+     * */
+    public String getOutputForProcess(Output type) {
+        if (type == Output.ERROR) {
+            return getOutputForProcessFromStream(process.getErrorStream());
+        }
+
+        return getOutputForProcessFromStream(process.getInputStream());
+    }
```
davidangb, line 97: this is just style, so feel free to ignore: I would just make two separate methods `getErrorText()` and `getOutputText()` (or whatever you want to name them), which don't accept any arguments. Then you could do away with the `Output` enum entirely!

service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
```diff
@@ -0,0 +1,108 @@
...
+
+    /** Block until the child process terminates, then return its exit code. */
+    public int waitForTerminate() {
+        try {
+            return process.waitFor();
+        } catch (InterruptedException intEx) {
+            Thread.currentThread().interrupt();
```
davidangb, line 104: is this `interrupt()` necessary? If the code has entered this catch clause, hasn't it already been interrupted?

davidangb (commented):
a few comments inline!

service/src/test/java/org/databiosphere/workspacedataservice/process/LocalProcessLaunchTest.java
yuliadub: ah using pg_dump is probably not the best test case - but I see whats going on: 

-  if pg_dump is installed and you run the command just like that you should get an error: "pg_dump: error: connection to server on socket "/var/run/postgresql/.s.PGSQL.5433" failed: No such file or directory". 
- if it is not installed, then process launcher throws as exception saying "Error launching local process" since it cant find the binary to launch it

I was trying to come up with a command that would always fail the same when it is run incorrectly (but it is actually a real command)

changed both of the failing tests around, and used your suggestion for asserting the exception instead of the backwards way I was doing it before

service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
yuliadub: ah I mean for it to be private but was messing with something else and forgot to change it back! thanks for spotting!

service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
yuliadub: sonarcloud was mad about it if it wasnt there but I dont think it is needed

service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
yuliadub: I go back and forth on this but personally like having the enum vs two functions :D

service/src/test/java/org/databiosphere/workspacedataservice/process/LocalProcessLaunchTest.java
davidangb: if you need a command that always fails, does [false](https://linux.die.net/man/1/false) work for the tests?

service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * Executes a command in a separate process from the current working directory (i.e. the same
+     * place as this Java process is running).
+     *
+     * @param command the command and arguments to execute
+     * @param envVars the environment variables to set or overwrite if already defined
+     */
+    public void launchProcess(List<String> command, Map<String, String> envVars) {
+        launchProcess(command, envVars, null);
+    }
```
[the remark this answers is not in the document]
yuliadub: my idea was the as long as you have a local process instance, it would actually be okay to keep changing the commands and running them one after the other vs having to create multiple instances of local process (since in theory you only launch the command once and then you are likely done with it, and all you want is the output from that one command which you will get back if you stream output before writing the next command). 

with that said if in our context it does make sense to switch to mapping one command to one process then the change makes sense but I am not sure I understand the advantage

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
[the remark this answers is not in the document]
davidangb: I won't hold the PR on this. I am thinking pretty defensively here, from the perspective of making a "framework" for other code to use, and thinking of ways the calling code could mis-use `LocalProcessLauncher`. We can handle that in code review as we actually start consuming this class.

service/src/main/java/org/databiosphere/workspacedataservice/process/LocalProcessLauncher.java
```diff
@@ -0,0 +1,103 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Map;
+
+/** This class provides utility methods for launching local child processes. */
+class LocalProcessLauncher {
```
davidangb, line 15: this class (and its default constructor) are `package` visibility since nothing is specified here. Make this `public` so it can be used from other packages?

davidangb (approved):
one last comment about class visibility, otherwise :+1:!

PR #212, opened 2023-04-18
AJ-935: Create functionality for pg_dump in WDS for Azure Postgres instances

aaronkanzer:
Relates to https://broadworkbench.atlassian.net/browse/AJ-935

[yuliadub notes] Marking this PR ready for review, a few things: 
- WDS now has an API to launch a pg_dump process and output the results of the dump into a file directly into Azure blob storage
- currently, all configuration (db user, password, host, etc) is expected to be passed via variables available to the WDS service (similar to how we pass in workspace Id today) - some of these values will need to be passed to WDS pod when it is starting (todo: need to create task for this once I confirm that is correct, unless those values are already available somewhere or should just go into WDS config)
- Azure blob storage currently takes in a connection string to be able to write to blob storage (later to be replaced with sas token), as part of the ticket that covers adding sas token to WDS
- Image is working when tested locally in docker, have not yet pulled this into a BEE but likely would be a good idea before this is merged. Since we are changing from distroless to a Dockerfile, do we need to make any changes anywhere else to tell leo or would we simply change our base image in WDS gradle config to be the one we create and we would need to update it manually in the future (also, do we need app sec approval?) **update** image was approved by appsec
- struggling a bit with tests for new backup process - unit tests seem hard (or more useless?), since pretty much everything needs to be mocked and is not really testing anything major (can add a test to generate a back up file but it would have to be locally vs on blob storage, and similarly can check that process launches but wont be pg_dump specific since the build machine is unlikely to have it). Feels like this would be a great candidate for an integration test, where we would spin up a db and install pg_dump on the machine that is doing the test (and for now we could mock the blob storage part to be local storage but still check that the contents of the file are what is expected)! **update** integration tests have been added for testing pg_dump


service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
```diff
@@ -0,0 +1,84 @@
...
+                    dbName
+            );
+            Process process = processBuilder.start();
+
+            int exitCode = process.waitFor();
+            if (exitCode == 0) {
+
+                // Upload the backup file to Azure Blob Storage
+                BlobClient blobClient = getAzureBlobClient(connectionString, containerName, workspaceId);
+                try (FileInputStream backupInputStream = new FileInputStream(backupFile)) {
```
aaronkanzer, line 49: Flag for Yulia in Slack conversation

service/src/main/java/org/databiosphere/workspacedataservice/controller/BackupController.java
```diff
@@ -0,0 +1,29 @@
...
+@RestController
+public class BackupController {
+
+    private final BackupService backupService;
+
+    public BackupController(BackupService backupService) {
+        this.backupService = backupService;
+    }
+
+    // TODO: We may not need the workspaceId
```
aaronkanzer, line 21: Will remove comment

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
```diff
@@ -0,0 +1,87 @@
...
+                "pg_dump", "--help"
+        );
+
+        InputStream pgDumpOutput = localProcessLauncher.launchProcess(command, null, null);
+        BufferedReader reader = new BufferedReader(new InputStreamReader(pgDumpOutput));
+
+        String line;
+        while ((line = reader.readLine()) != null) {
+            System.out.println(line);
+        }
```
aaronkanzer, line 64: Will remove this block -- this is just used to verify that pg_dump command is observed in logs

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
```diff
@@ -0,0 +1,87 @@
...
+        LocalDateTime now = LocalDateTime.now();
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd_HH-mm-ss");
+        String timestamp = now.format(formatter);
+        String blobName = workspaceId.toString() + "-" + timestamp + ".sql";
+
+//        String dbHost = System.getenv("WDS_DB_HOST");
+//        String dbPort = System.getenv("WDS_DB_PORT");
+//        String dbUser = System.getenv("WDS_DB_USER");
+//        String dbName = System.getenv("WDS_DB_NAME");
+//        String dbPassword = System.getenv("WDS_DB_PASSWORD")
```
aaronkanzer, line 47: Eventually we will use these instead of the values below

service/src/main/java/org/databiosphere/workspacedataservice/controller/BackupController.java
```diff
@@ -0,0 +1,29 @@
...
+    @PostMapping("/backup/azure")
```
aaronkanzer, line 22: Note for Slack update - Hitting this endpoint will now invoke the ProcessBuilder -> call `pg_dump --help` -> std.out values that convey the client is available

service/build.gradle
```diff
@@ -96,3 +99,10 @@ jacocoTestReport {
 		xml.required = true
 	}
 }
+
+bootRun {
+	doFirst {
+		def localPostgresPath = "/opt/homebrew/bin/pg_dump"
+		environment "PATH", "${localPostgresPath}:${System.getenv('PATH')}"
+	}
+}
```
davidangb, line 108: this looks like an artifact of Aaron's testing in his local environment; it's not needed for the real implementation.

service/src/main/java/org/databiosphere/workspacedataservice/controller/BackupController.java
davidangb: I agree we don't need a workspaceId parameter here

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
```diff
@@ -0,0 +1,101 @@
...
+
+        LocalDateTime now = LocalDateTime.now();
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd_HH-mm-ss");
+        String timestamp = now.format(formatter);
+        String blobName = workspaceId.toString() + "-" + timestamp + ".sql";
+
+        String dbHost = System.getenv("WDS_DB_HOST");
+        String dbPort = System.getenv("WDS_DB_PORT");
+        String dbUser = System.getenv("WDS_DB_USER");
+        String dbName = System.getenv("WDS_DB_NAME");
```
davidangb, line 41: this env is not currently set for WDS; this is new

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
```diff
@@ -0,0 +1,101 @@
...
+        String dbPassword = System.getenv("WDS_DB_PASSWORD");
+
+        Map<String, String> command = new LinkedHashMap<>();
+        command.put("/usr/bin/pg_dump", null);
```
davidangb, line 45: even though the `apt-get install -y postgresql-client-15.2` is failing, the `us.gcr.io/broad-dsp-gcr-public/base/jre:17-debian` image already has `pg_dump` pre-installed!! It's at `/usr/bin/pg_dump`.

Sadly, the pre-installed version of pg_dump is 12.14, which doesn't work against higher postgres versions. It results in an error of:

```
pg_dump: error: server version: 13.1 (Debian 13.1-1.pgdg100+1); pg_dump version: 12.14 (Ubuntu 12.14-0ubuntu0.20.04.1)
pg_dump: error: aborting because of server version mismatch
```

Dockerfile
```diff
@@ -6,8 +6,16 @@
...
+RUN apt-get update && apt-get dist-upgrade -y
+RUN apt-get install lsb-core -y
+RUN apt-get install gnupg2 ca-certificates wget -y
+# use the latest postgres repository
+RUN sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list'
+# install the postgres public key
+RUN wget -qO- https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -
+# refresh the repository
+RUN apt-get update
+RUN apt-get install postgresql-client -y
```
davidangb, line 18: this works but is less than ideal; GCR scanning shows lots of vulnerabilities for this image. We'll need to figure out how to make a better base image.

service/build.gradle
```diff
@@ -21,7 +21,7 @@ repositories {
 jib {
 	from {
 		// see https://github.com/broadinstitute/dsp-appsec-blessed-images/tree/main/jre
-		image = "us.gcr.io/broad-dsp-gcr-public/base/jre:17-distroless"
+		image = "us.gcr.io/broad-dsp-gcr-public/sandbox/wds-debian-davidan:latest"
```
davidangb, line 24: I have built the `Dockerfile` above and pushed it to `us.gcr.io/broad-dsp-gcr-public/sandbox/wds-debian-davidan:latest` so the jib build can use it as a base.

Dockerfile
davidangb: Resolved by skipping installation of `lsb-core`. We now have the same number of vulns as the appsec blessed base image.

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
```diff
@@ -0,0 +1,101 @@
...
+        LocalDateTime now = LocalDateTime.now();
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd_HH-mm-ss");
+        String timestamp = now.format(formatter);
+        String blobName = workspaceId.toString() + "-" + timestamp + ".sql";
+
+        String dbHost = System.getenv("WDS_DB_HOST");
+        String dbPort = System.getenv("WDS_DB_PORT");
+        String dbUser = System.getenv("WDS_DB_USER");
+        String dbName = System.getenv("WDS_DB_NAME");
+        String dbPassword = System.getenv("WDS_DB_PASSWORD");
```
davidangb, line 42: I think all the env vars in here should move to Spring properties

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
davidangb, line 45: the path `/usr/bin/pg_dump` should be a Spring property, which would make it easy to override during local development

Dockerfile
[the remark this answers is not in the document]
yuliadub: I found simple command on debian to get pg_dump installed, let me know if those changes work or if it still makes sense to move to alpine

service/build.gradle
yuliadub: I may need help understand how all these dockerfiles chain up together

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
yuliadub: let me know if I did that right!

Dockerfile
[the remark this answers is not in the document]
yuliadub: scratch that - ran into the same versioning issue (it was indeed version 12), reverted back to what you had!

yuliadub:
sadly all Azure blob storage classes are final, making them hard to mock for unit tess. https://github.com/mockito/mockito/wiki/What's-new-in-Mockito-2#unmockable

tests for backup seem a lot more like integration tests (have the ability to connect to azure blob storage and will need pg_dump on machine where the test is running), would it make sense to make a separate task for those and move on or continue to work on them as part of this PR?

service/build.gradle
```diff
@@ -22,7 +22,7 @@ repositories {
 jib {
 	from {
 		// see https://github.com/broadinstitute/dsp-appsec-blessed-images/tree/main/jre
-		image = "us.gcr.io/broad-dsp-gcr-public/base/jre:17-distroless"
+		image = "us.gcr.io/broad-dsp-gcr-public/sandbox/wds-debian-yuliadub:latest"
```
yuliadub, line 25: need to adjust this before merging

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -14,6 +14,8 @@ servers:
   - url: ../
     description: Relative to the current swagger page
 tags:
+  - name: Backup
+    description: Backup APIs
```
yuliadub, line 18: need to add auth - but I thinking of doing that when we start using the api at cloning since its nice to keep it open for testing at this time. Any concerns with that approach?

service/src/test/java/org/databiosphere/workspacedataservice/storage/LocalFileStorage.java
```diff
@@ -0,0 +1,24 @@
...
+
+    public void streamOutputToBlobStorage(InputStream fromStream, String blobName) {
+        File targetFile = new File("backup.txt");
+        try(OutputStream outStream = new FileOutputStream(targetFile)) {
+            try (BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(fromStream, StandardCharsets.UTF_8))) {
+                int line;
+                while ((line = bufferedReader.read()) != -1) {
+                    outStream.write(line);
+                }
+            }
```
davidangb, line 19: this block is copied between `AzureBlobStorage` and `LocalFileStorage`; you could make a reusable method a la:
```java
    public void writeStream(InputStream fromStream, OutputStream toStream) throws IOException {
        try (BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(fromStream, StandardCharsets.UTF_8))) {
            int line;
            while ((line = bufferedReader.read()) != -1) {
                toStream.write(line);
            }
        }
    }
```
is that worth it?

davidangb (commented):
initial comments inline!

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        return true;
+    }
+
+    public static String GenerateBackupFilename(String workspaceId) {
+        LocalDateTime now = LocalDateTime.now();
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd_HH-mm-ss");
+        String timestamp = now.format(formatter);
+        return  workspaceId + "-" + timestamp + ".sql";
```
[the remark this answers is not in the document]
yuliadub: they are nested under backup "container" [directory ](https://github.com/DataBiosphere/terra-workspace-data-service/blob/0fa93b233b5d6f300824c4cce479c40f9de6804f/service/src/main/java/org/databiosphere/workspacedataservice/storage/AzureBlobStorage.java#L48)at this time. Will double check if that does or does not show up in the action terra UI.

service/src/test/java/org/databiosphere/workspacedataservice/storage/LocalFileStorage.java
yuliadub: if the output streams they were using would be the same, it would be great - but blob storage uses a special "BlobOutputStream" output stream so that makes this less worth it I think.... :(

service/src/main/java/org/databiosphere/workspacedataservice/controller/BackupController.java
```diff
@@ -0,0 +1,31 @@
...
+    private final BackupService backupService;
+    private final AzureBlobStorage storage;
+
+    public BackupController(BackupService backupService) {
+        this.storage = new AzureBlobStorage();
+        this.backupService = backupService;
+    }
+
+    @PostMapping("/backup")
+    @RetryableApi
```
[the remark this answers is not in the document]
yuliadub: 1. Good callout! added
2. yeah, I was going back and forth as well, I think for now - lets keep it as is, once we hook this up and do it on cloning, we can experiment if there would be an easy way to check if a back up is already in progress/etc and there may be better retry guardrails that can be put in place. For now I am not 100% sure what that looks like yet

service/src/main/java/org/databiosphere/workspacedataservice/storage/AzureBlobStorage.java
```diff
@@ -0,0 +1,50 @@
...
+import org.databiosphere.workspacedataservice.service.model.exception.LaunchProcessException;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+
+public class AzureBlobStorage implements BackUpFileStorage {
+    private static String backUpContainerName = "backup";
```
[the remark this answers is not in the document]
yuliadub: need to confirm this but my suspicion is the answer is no, it should not be visible to the user - only specific containers inside blob storage are exposed to user.

service/src/main/resources/static/swagger/openapi-docs.yaml
yuliadub: not at this time, since blob storage is relying to have a connection string to be present as an env variable, but we switch to using the access token, you are correct.

service/build.gradle
yuliadub: need to update this to be a more official image name and edit the main release GAH to build and push this image each time we release

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
```diff
@@ -0,0 +1,101 @@
...
+                commandList.add(entry.getValue());
+            }
+        }
+        commandList.add("-v");
+        commandList.add("-w");
+
+        return commandList;
+    }
+
+    public String GenerateBackupFilename() {
```
yuliadub, line 95: part of me wonders if this should live along side blob storage inside its class?

yuliadub:
some unit tests are failing with "nested exception is org.postgresql.util.PSQLException: ERROR: permission denied for schema", cant seem to replicate this locally - at first I thought this may be caused by unit/integration tests running at the same time with the same db configurations (even thought they shouldn't be related) but that is not it....hmmm, any help/guidance appreciated. It's the RecordOrchestratorServiceTest that seems to be unhappy

.github/workflows/build-and-integration-test.yml
```diff
@@ -0,0 +1,90 @@
...
+          java-version: '17'
+          distribution: 'temurin'
+          cache: 'gradle'
+
+      - name: Cache SonarCloud packages
+        uses: actions/cache@v3
+        with:
+          path: ~/.sonar/cache
+          key: ${{ runner.os }}-sonar
+          restore-keys: ${{ runner.os }}-sonar
```
davidangb, line 65: since we have a sonarqube scan in `build-and-test.yml`, we don't also need it here, can just delete this step.

.github/workflows/build-and-integration-test.yml
```diff
@@ -0,0 +1,90 @@
...
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports: [ "5432:5432" ]
+
+    steps:
+      - uses: actions/checkout@v3
+        # Needed by sonar to get the git history for the branch the PR will be merged into.
+        with:
+          fetch-depth: 0
```
davidangb, line 43: I think you can remove the sonarqube step (see comment below), so you could also remove `fetch-depth: 0`

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/BackupServiceIntegrationTest.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.io.File;
+import java.util.Scanner;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@SpringBootTest(properties = "spring.cache.type=NONE")
+public class BackupServiceIntegrationTest {
+    @Autowired
+    private BackupService mockBackupService;
```
davidangb, line 17: nit: if I'm understanding correctly, this isn't a mock, is it? It's the real `BackupService` implementation. Remove "mock" from the name?

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/BackupServiceIntegrationTest.java
```diff
@@ -0,0 +1,26 @@
...
+
+    private LocalFileStorage storage = new LocalFileStorage();
+
+    @Test
+    void testBackupAzureWDS() throws Exception {
+        var response = mockBackupService.backupAzureWDS(storage, "v0.2");
+        assertTrue(response.backupStatus());
```
davidangb, line 24: nit: include the backup error message in the cases this test fails, that will make test-debugging a lot easier:
```suggestion
        assertTrue(response.backupStatus(), response.message());
```

service/src/integrationTest/resources/integrationTest.properties
davidangb, line 1: I don't believe `integrationTest.properties` or `service/src/test/resources/test.properties` are actually used by anything - are they necessary?

service/src/test/java/org/databiosphere/workspacedataservice/storage/LocalFileStorage.java
davidangb: `BlobOutputStream` is a subclass of `OutputStream` so I think it would still work - I wouldn't hold the PR for this though

service/src/test/java/org/databiosphere/workspacedataservice/service/BackupServiceTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.util.List;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+@SpringBootTest(properties = "spring.cache.type=NONE")
+@TestPropertySource("classpath:test.properties")
+class BackupServiceTest {
+    @Autowired
+    private BackupService mockBackupService;
```
davidangb, line 16: also here, this isn't a mock, remove "mock" from the name

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
```diff
@@ -0,0 +1,101 @@
...
+    @Value("${twds.pg_dump.port:}")
+    private String dbPort;
+
+    @Value("${twds.pg_dump.host:}")
+    private String dbHost;
+
+    @Value("${twds.pg_dump.path:}")
+    private String pgDumpPath;
+
+    @WriteTransaction
```
davidangb, line 46: remove `@WriteTransaction` here - this isn't a db transaction - `pg_dump` has its own transaction semantics

davidangb (commented):
see comments inline, very close! I also see a unit test failed (it doesn't fail for me locally), do you know what's up with that?

client/build.gradle
```diff
@@ -5,6 +5,7 @@ plugins {
 
 	id 'com.jfrog.artifactory' version '4.18.2'
 	id 'org.hidetake.swagger.generator'
+	id 'jvm-test-suite'
```
davidangb, line 8: I don't believe this is necessary.

service/build.gradle
```diff
@@ -112,3 +116,25 @@ tasks.register("pactTests", Test) {
 		showStandardStreams = true
 	}
 }
+
+testing {
+    suites {
+        integrationTest(JvmTestSuite) {
+            dependencies {
+        		implementation project()
+        		implementation project(':client')
```
davidangb, line 125: I think you can replace these two lines with:
```suggestion
        		implementation project(':service')
```

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
yuliadub: actually this is specific to backup, so going to keep it here

client/build.gradle
yuliadub: good catch

yuliadub:
> see comments inline, very close! I also see a unit test failed (it doesn't fail for me locally), do you know what's up with that?

finally got all those fixed! was caused by the issue of password being used for unit specific record unit tests being "wds" but me overwriting it with "postgres" lower down. may make a ticket to optimize all of it so it is easier to track in the future!

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.*;
+
+import static org.databiosphere.workspacedataservice.service.RecordUtils.validateVersion;
+
+@Service
+public class BackupService {
+    private static final Logger LOGGER = LoggerFactory.getLogger(BackupService.class);
+
+    //TODO: in the future this will shift to "twds.instance.source-workspace-id"
+    @Value("${twds.instance.workspace-id:}")
```
calypsomatic, line 25: Maybe a dumb question, but why isn't is source-workspace-id now?

calypsomatic: Makes sense!  Thank you

.github/workflows/build-and-integration-test.yml
```diff
@@ -0,0 +1,80 @@
...
+      - name: Build, assemble, and integration test
+        id: build-test
+        run: ./gradlew integrationTest
+
+      # The SonarQube scan is done here, so it can upload the coverage report generated by the tests.
+      - name: SonarQube scan
+        env:
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: ./gradlew --build-cache sonarqube
```
davidangb, line 73: remove this sonarqube scan in favor of the one in `build-and-test.yml`

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
davidangb: > would the destination wds directly try to connect to the wds source db to get the backup?

pretty sure this won't be feasible, especially when source and destination are in different MRGs

> would the destination wds call into source wds to trigger a backup to happen on the source wds side?

I think this is going to be the only realistic approach.

I'm definitely game for experimentation and seeing what's possible, just voicing what I think will be the result!

service/src/test/resources/test.properties
davidangb, line 1: I don't believe this is actually used; delete it?

davidangb (approved):
yay thank you for pushing this through!

PR #213, opened 2023-04-20
AJ-762 address set-output and save-state deprecation warnings

calypsomatic:
See [AJ-762](https://broadworkbench.atlassian.net/browse/AJ-762)
This PR updates the version of actions/cache to address deprecation of set-output and save-state.

[AJ-762]: https://broadworkbench.atlassian.net/browse/AJ-762?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -57,14 +57,14 @@ jobs:
           cache: 'gradle'
 
       - name: Cache SonarCloud packages
-        uses: actions/cache@v1
+        uses: actions/cache@v3
         with:
           path: ~/.sonar/cache
           key: ${{ runner.os }}-sonar
           restore-keys: ${{ runner.os }}-sonar
 
       - name: Cache Gradle packages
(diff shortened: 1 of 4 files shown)


aaronkanzer:
Do will still need to handle the explicit usage of `set-output` in `publish-docker.yaml`?

calypsomatic:
> Do will still need to handle the explicit usage of `set-output` in `publish-docker.yaml`?

Yep!  I must not have been very careful going through all the actions looking for the warnings.  Good catch, thank you!

davidangb (commented):
I think the tag.yaml action will need an update too, based on the output of this recent run: https://github.com/DataBiosphere/terra-workspace-data-service/actions/runs/4746784424

davidangb (approved):

PR #215, opened 2023-04-26
AJ-1005: Create Dockerfile to include postgres client for prod WDS

aaronkanzer:
We need to invoke `pg_dump` in WDS during cloning of workspaces -- this Dockerfile allows us to include the Postgres client -- it extends from app-sec blessed code here: https://github.com/broadinstitute/dsp-appsec-blessed-images/blob/main/jre/Dockerfile.17-alpine

Reminder:
All PRs merged into main will generate a PR in https://github.com/broadinstitute/cromwhelm to update the WDS image deployed to kubernetes. 
After your merge, you must go to this repo and approve this generated PR and make sure it merges.
That merge will then generate a PR in https://github.com/DataBiosphere/leonardo.  This may or may not automerge; be sure to watch it to ensure it merges.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@
 FROM us.gcr.io/broad-dsp-gcr-public/base/jre:17-alpine
 
 # Add postgres client for pg_dump command
-RUN apk add --no-cache postgresql-client
+RUN apk add --no-cache postgresql-client~=15.2
 
 # Temp storage location for pg_dump outputs on Azure backups
 VOLUME /backup
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,28 +1,9 @@
-# https://github.com/docker-library/docs/blob/master/eclipse-temurin/README.md#creating-a-jre-using-jlink
+### This Dockerfile is used to extend the app-sec blessed Docker image for JRE-17
+### We extend the Dockerfile within WDS to include a PostgreSQL client -- the client is primarily used
+### to faciliate the cloning of a WDS instance/workspace in Terra
 
-########## start of code directly from https://github.com/broadinstitute/dsp-appsec-blessed-images/blob/main/jre/Dockerfile.17-alpine
-FROM eclipse-temurin:17-jdk-alpine as jre-build
-
-# Create a custom Java runtime
-RUN apk add --no-cache binutils && \
-    $JAVA_HOME/bin/jlink \
-         --add-modules java.base \
-         --strip-debug \
-         --no-man-pages \
-         --no-header-files \
-         --compress=2 \
-         --output /javaruntime
-
-# Define your base image
-FROM alpine
-
-RUN apk --no-cache -U upgrade
-
-ENV JAVA_HOME=/opt/java/openjdk
-ENV PATH="${JAVA_HOME}/bin:${PATH}"
-
-COPY --from=jre-build /javaruntime $JAVA_HOME
-########## end of code directly from https://github.com/broadinstitute/dsp-appsec-blessed-images/blob/main/jre/Dockerfile.17-alpine
+### Sourced from https://github.com/broadinstitute/dsp-appsec-blessed-images/blob/main/jre/Dockerfile.17-alpine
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# Define your base image
+FROM alpine
+
+RUN apk --no-cache -U upgrade
+
+ENV JAVA_HOME=/opt/java/openjdk
+ENV PATH="${JAVA_HOME}/bin:${PATH}"
+
+COPY --from=jre-build /javaruntime $JAVA_HOME
+########## end of code directly from https://github.com/broadinstitute/dsp-appsec-blessed-images/blob/main/jre/Dockerfile.17-alpine
```
aaronkanzer, line 25: We could just do `FROM us.gcr.io/broad-dsp-gcr-public/base/jre:17-alpine` but I figured that this would ensure that the contents of our Dockerfile are reliably unchanged -- curious to get others thoughts

calypsomatic (approved):
👍 assuming appsec approval

Dockerfile
```diff
@@ -0,0 +1,12 @@
+### This Dockerfile is used to extend the app-sec blessed Docker image for JRE-17
+### We extend the Dockerfile within WDS to include a PostgreSQL client -- the client is primarily used
+### to faciliate the cloning of a WDS instance/workspace in Terra
+
+### Sourced from https://github.com/broadinstitute/dsp-appsec-blessed-images/blob/main/jre/Dockerfile.17-alpine
+FROM us.gcr.io/broad-dsp-gcr-public/base/jre:17-alpine
+
+# Add postgres client for pg_dump command
+RUN apk add --no-cache postgresql-client
```
davidangb, line 9: is it worth pinning this to a specific version of the client, so we're not surprised by an upgrade that changes behavior? I'm fine either way, just want to make sure you've considered it.

davidangb (approved):
one question for consideration, otherwise :+1:

Dockerfile
aaronkanzer: Do we think that the Postgres instance version we are using alongside WDS will be a moving target? It seems that the client and server components that Postgres uses for its protocol can run into compatibility issues if there versioning are far apart.

If its not a moving target, I'll definitely pin the version here alongside what we are using (or rather it might just need to be something we periodically update)

I agree that versioning is better here 💪

dinvlad:
LGTM!

Dockerfile
davidangb: I doubt we'll change the server-side Postgres version too often - especially after we move to the LZ Postgres server instead of in-k8s server.

Dockerfile
```diff
@@ -0,0 +1,12 @@
+### This Dockerfile is used to extend the app-sec blessed Docker image for JRE-17
+### We extend the Dockerfile within WDS to include a PostgreSQL client -- the client is primarily used
+### to faciliate the cloning of a WDS instance/workspace in Terra
+
+### Sourced from https://github.com/broadinstitute/dsp-appsec-blessed-images/blob/main/jre/Dockerfile.17-alpine
+FROM us.gcr.io/broad-dsp-gcr-public/base/jre:17-alpine
+
+# Add postgres client for pg_dump command
+RUN apk add --no-cache postgresql-client~=15.2
```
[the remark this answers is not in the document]
aaronkanzer: from the k8s pod

davidangb (approved):
looks good, wonder if you want to pin the `postgres-client` version as you had in a previous commit, but I wouldn't hold the PR for that

calypsomatic (approved):

aaronkanzer:
> looks good, wonder if you want to pin the `postgres-client` version as you had in a previous commit, but I wouldn't hold the PR for that

I realized that our `postgres` images in k8s-deployed-via-Leo apps are pinned to `latest`, so I figured here I'd continue the client to be lock-step with it, thus not pinning.

I agree we _should_ pin both though -- I'd message `dsp-azure` for now.

PR #216, opened 2023-04-26
AJ-958: liquibase, internal WDS state for instances

davidangb:
IN THIS PR:
* enable liquibase for schema migrations
* move the creation of file&relation domains (datatypes) from data.sql to liquibase
* disable data.sql processing via `spring.sql.init.mode=never`
* delete data.sql since we're not using it anymore
* add a liquibase changeset to create the `sys_wds` schema
* add a liquibase changeset to create the `sys_wds.instance` table to power create/delete/list/exists instances
* change create/delete/list/exists instance methods to use `sys_wds.instance` instead of `information_schema.schemata`

TODO/QUESTIONS:
- [ ] _Can we move Liquibase's `databasechangelog` and `databasechangeloglock` tables into the `sys_wds` schema?_ Liquibase does not support this natively because of a chicken-or-egg problem: Liquibase needs these tables to exist in order to run changesets, but we need to run a changeset to create the schema. We'd have to use something like https://github.com/lbruun/Pre-Liquibase, which requires Spring Boot 3. Perhaps there's another solution somewhere. The `data.sql` approach runs _after_ liquibase, so that's not an option.
- [ ] _Can/should we move the file/relation domains into sys_wds?_ This is possible; it does require code changes to WDS Java classes. Is this worth it, since we're already relying on the default `public` schema for our Liquibase tables?
- [ ] _Should we use XML, YAML, JSON, or SQL format for our changelogs?_ I chose YAML from personal preference but open to change. See https://docs.liquibase.com/concepts/changelogs/home.html for examples.


calypsomatic:
Are we caring about backwards compatibility at all or are we still at the stage where we can just through out any WDS apps that existed before this?

davidangb:
> Are we caring about backwards compatibility at all or are we still at the stage where we can just through out any WDS apps that existed before this?

Good point! Upgrading to this liquibase-enabled version of WDS should work fine for everything *unless* we moved the file/relation domains into `sys_wds`. That's enough of a reason for me to NOT move them.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresInstanceDao.java
```diff
@@ -30,32 +31,33 @@ public PostgresInstanceDao(NamedParameterJdbcTemplate namedTemplate) {
...
-        return schemas.stream().map(s -> safeParseUUID(s))
-                .filter(Objects::nonNull).toList();
+        return namedTemplate.getJdbcTemplate().queryForList(
+                "select id from sys_wds.instance order by id",
+                UUID.class
+        );
     }
 
     @Override
+    @WriteTransaction
```
aaronkanzer, line 47: Just to fully understand, why does this decorator introduced now, and not when the prior code for `create schema...` existed?

Is the decorator specific to `insert` statements?

davidangb: We technically don't need the `@WriteTransaction` here, because we already annotate transactions inside `InstanceService`. I'm adding them here for clarity and visibility, and in case anything other than `InstanceService` decides to call these methods.

As for why now … before this PR, there was only a single database statement inside `createSchema` and `dropSchema`. As of this PR, those methods both execute two statements. Because I want the two statements to be grouped together into a single atomic unit, I annotate the transaction. When it was just one statement we could rely on the inherent auto-transaction.

aaronkanzer: makes sense -- thanks!

davidangb:
> Could you explain a bit more for why we'd want to do the following:

>> Can we move Liquibase's databasechangelog and databasechangeloglock tables into the sys_wds schema?

As it stands right now in this PR, we have "WDS stuff" split across two schemas: the `instance` table is inside the `sys_wds` schema, but the `databasechangelog` table, `databasechangeloglock` table, and the file/relation domains are inside the `public` schema. I think it would be cleaner if all the "WDS stuff" was all inside a single schema. But:

* moving the file/relation domains into sys_wds is problematic for upgrading older WDSes
* moving databasechangelog/databasechangeloglock into sys_wds is problematic due to the chicken/egg problem
* moving the instance table into the public schema is quite possible, but then we're centralizing on the public schema instead of sys_wds, which might be less clear for anyone exploring the database

calypsomatic:
> As it stands right now in this PR, we have "WDS stuff" split across two schemas: the `instance` table is inside the `sys_wds` schema, but the `databasechangelog` table, `databasechangeloglock` table, and the file/relation domains are inside the `public` schema. I think it would be cleaner if all the "WDS stuff" was all inside a single schema. But:

I agree it would be cleaner, but overall not a big deal, so given the problems I'm happy leaving them as is.

aaronkanzer (approved):

PR #217, opened 2023-04-30
AJ-968: prefer artifactory caching proxy

davidangb:
Adds the Broad Artifactory caching proxy as the first repository, with maven central as the second.

This will help with any maven central outages.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,7 +13,10 @@ plugins {
     id 'jacoco'
 }
 
-repositories { mavenCentral() }
+repositories {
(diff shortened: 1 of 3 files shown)


aaronkanzer:
Seems we declare `mavenCentral()` in the general build.gradle and the client/build.gradle -- do you want to extend the value you added here there as well?

calypsomatic (approved):

davidangb:
> Seems we declare mavenCentral() in the general build.gradle and the client/build.gradle -- do you want to extend the value you added here there as well?

@aaronkanzer I just updated those builds too! thanks for finding that.

PR #218, opened 2023-04-30
AJ-952: transactions must be called from another class

davidangb:
See the first item at https://medium.com/javarevisited/spring-transactional-mistakes-everyone-did-31418e5a6d6b:

> @Transactional is powered by [Aspect-Oriented Programming](https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#aop). Therefore, processing occurs when a bean is called from another bean.

We had two cases where annotated transactions were called from within the same class; this PR removes those.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
@@ -1,6 +1,5 @@
 package org.databiosphere.workspacedataservice.service;
 
-import bio.terra.common.db.WriteTransaction;
 import org.databiosphere.workspacedataservice.dao.InstanceDao;
 import org.databiosphere.workspacedataservice.sam.SamDao;
 import org.databiosphere.workspacedataservice.service.model.exception.AuthorizationException;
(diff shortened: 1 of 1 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
```diff
@@ -81,11 +75,6 @@ public void deleteInstance(UUID instanceId, String version) {
         }
 
         // delete instance schema in Postgres
-        deleteInstanceFromDatabase(instanceId);
-    }
-
-    @WriteTransaction
-    void deleteInstanceFromDatabase(UUID instanceId) {
         instanceDao.dropSchema(instanceId);
     }
```
davidangb, line 79: for these changes, note that `PostgresInstanceDao.createSchema()` and `PostgresInstanceDao. dropSchema()` already both specify `@WriteTransaction`. So, the `@WriteTransaction`s removed in this PR were redundant. I am removing them for clarity and to ensure nobody would try to depend on them in the future.

calypsomatic (approved):

aaronkanzer (approved):

PR #221, opened 2023-05-02
Aj 1008 snapshot permission

calypsomatic:
Reminder:
All PRs merged into main will generate a PR in https://github.com/broadinstitute/cromwhelm to update the WDS image deployed to kubernetes. 
After your merge, you must go to this repo and approve this generated PR and make sure it merges.
That merge will then generate a PR in https://github.com/DataBiosphere/leonardo.  This may or may not automerge; be sure to watch it to ensure it merges.


service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -407,7 +409,23 @@ paths:
             'application/json':
               schema:
                 $ref: '#/components/schemas/ErrorResponse'
-
+  /{instanceid}/snapshots/{v}/{snapshotid}:
```
yuliadub, line 412: should we make a separate task to add tests for this new call for the python client tests/etc?

service/src/main/java/org/databiosphere/workspacedataservice/controller/DataRepoController.java
```diff
@@ -0,0 +1,32 @@
...
+public class DataRepoController {
+
+    private final DataRepoDao dataRepoDao;
+
+    public DataRepoController(DataRepoDao dataRepoDao) {
+        this.dataRepoDao = dataRepoDao;
+    }
+
+    @PostMapping("/{instanceId}/snapshots/{version}/{snapshotId}")
+    @RetryableApi
```
davidangb, line 21: I'm not sure if this will be safe to retry in its eventual implementation - totally fine for now though

service/src/main/java/org/databiosphere/workspacedataservice/datarepo/DataRepoDao.java
```diff
@@ -0,0 +1,31 @@
...
+    public boolean hasSnapshotPermission(UUID snapshotId) {
+        LOGGER.debug("Checking for permission for snapshot {} ...", snapshotId);
+        try {
+            dataRepoClientFactory.getRepositoryApi().retrieveSnapshot(snapshotId, Collections.emptyList());
+            //If we get no error, then the user has permission
+            return true;
+        } catch (Exception e){
+            LOGGER.error(e.getMessage());
+            return false;
+        }
```
davidangb, line 28: How about catching `ApiException`, inspecting its `.getCode()`, and returning false for a 401 or 403 - otherwise rethrowing the error?

I realize this is an early spike for the API, but I think it would be good to distinguish permission errors from other errors like a timeout or having an invalid TDR URL

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -458,6 +476,13 @@ components:
       schema:
         type: string
         default: v0.2
+    snapshotIdPathParam:
+      name: snapshotid
+      in: path
+      description: snapshot id
+      required: true
+      schema:
+        type: string
```
davidangb, line 485: I suggest
```
                  type: string
                  format: uuid
```
to indicate that snapshot id is a UUID

service/src/test/java/org/databiosphere/workspacedataservice/controller/DataRepoControllerMockMvcTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @MockBean
+    DataRepoClientFactory mockDataRepoClientFactory;
+
+    RepositoryApi mockRepositoryApi = Mockito.mock(RepositoryApi.class);
+
+    @BeforeEach
+    void beforeEach() {
+        given(mockDataRepoClientFactory.getRepositoryApi()).willReturn(mockRepositoryApi);
+        Mockito.clearInvocations(mockRepositoryApi);
```
davidangb, line 44: `clearInvocations` isn't necessary here, nothing is verifying the invocations

service/src/test/java/org/databiosphere/workspacedataservice/datarepo/DataRepoDaoTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @MockBean
+    DataRepoClientFactory mockDataRepoClientFactory;
+
+    RepositoryApi mockRepositoryApi = Mockito.mock(RepositoryApi.class);
+
+    @BeforeEach
+    void beforeEach() {
+        given(mockDataRepoClientFactory.getRepositoryApi()).willReturn(mockRepositoryApi);
+        Mockito.clearInvocations(mockRepositoryApi);
```
davidangb, line 34: clearInvocations also not necessary here

davidangb (commented):
initial comments inline!

service/src/main/java/org/databiosphere/workspacedataservice/datarepo/HttpDataRepoClientFactory.java
```diff
@@ -0,0 +1,53 @@
...
+    public HttpDataRepoClientFactory(String dataRepoUrl) {
+        this.dataRepoUrl = dataRepoUrl;
+    }
+
+    private ApiClient getApiClient() {
+        // create a new data repo client
+        ApiClient apiClient = new ApiClient();
+
+        // initialize the client with the url to data repo
+        if (StringUtils.isNotBlank(dataRepoUrl)) {
```
aaronkanzer, line 30: Do you need the `isNotBlank` condition here? Aren't we always creating a new ApiClient in this code?

davidangb (approved):
Could you update README.md to include info about setting `DATA_REPO_URL` when running locally? Otherwise :+1:, thank you!

service/src/main/java/org/databiosphere/workspacedataservice/datarepo/HttpDataRepoClientFactory.java
calypsomatic: This is to make sure it doesn't blow up if the data repo URL isn't present for some reason.

.github/workflows/release-python-client.yml
```diff
@@ -84,6 +84,7 @@ jobs:
       - name: Run WDS service and tests
         run: |
             export SAM_URL=http://localhost:9889
+            export DATA_REPO_URL=http://localhost:9889
```
aaronkanzer, line 87: Is this intentional to have the same URL as the SAM_URL value above?

calypsomatic: Yes - rather than set up multiple nginxs for mocking, I'm using the same one and just adding an additional path in `nginx.conf` to handle TDR

aaronkanzer (approved):

PR #222, opened 2023-05-02
AJ-1006: re-enable Sam

davidangb:
This is PR 1 of 2 to re-enable Sam. This one is the quick fix to turn Sam back on, but it doesn't yet delete any of the custom code for disabled Sam.

Once we are confident in Sam working via the CBAS integration, I'll make another PR to delete the crufty code.


service/build.gradle
```diff
@@ -74,6 +74,8 @@ tasks.named('test') {
 	testLogging {
 		events = ["passed", "failed", "skipped", "started"]
 		exceptionFormat = "full"
+		// should command-line test output show stdout/stderr?
+		showStandardStreams = false
```
davidangb, line 78: this is the default, so this PR doesn't change anything - but I find myself constantly toggling this on/off during development and so having the placeholder here is helpful to me!

service/src/test/java/org/databiosphere/workspacedataservice/service/PermissionsStatusServiceTest.java
```diff
@@ -18,12 +18,14 @@
...
 import org.springframework.boot.test.mock.mockito.MockBean;
 import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.TestPropertySource;
 
 import static org.mockito.BDDMockito.given;
 import static org.mockito.Mockito.*;
 
 @DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_CLASS)
 @SpringBootTest(properties = "spring.cache.type=NONE")
+@TestPropertySource(properties = {"twds.instance.workspace-id=123e4567-e89b-12d3-a456-426614174000"}) // example uuid from https://en.wikipedia.org/wiki/Universally_unique_identifier
```
davidangb, line 28: see comment in `application-mock-sam.properties` for why we need this

davidangb: I would have to add `@ActiveProfiles(profiles = "mock-sam")` to this test for that to work - which would change the behavior of this test.

PR #224, opened 2023-05-04
AJ-1006: remove the disabled-sam code paths

davidangb:
#222 turned Sam integration back on.

This PR removes the code paths that check whether or not Sam integration is enabled. In other words, Sam is always on - regardless of config - in this PR.


yuliadub (approved):

davidangb:
> Could you remove this too?

@aaronkanzer arg can't believe I missed that one! Thanks for finding; now removed.

aaronkanzer (approved):

PR #227, opened 2023-05-05
AJ-915: disable automerge PRs to Cromwhelm

davidangb:
Since Terra is now autodeploying the standalone WDS, and we will shortly be removing WDS from Cromwhelm, let's stop autocreating PRs to update Cromwhelm on every PR merge.


aaronkanzer (approved):

aaronkanzer:
I'm wondering if we could also remove the push Image to ACR?

Seems we are referencing the GCR image in the terra-helmfile repo: https://github.com/broadinstitute/terra-helmfile/blob/master/charts/wds/values.yaml#L47

yuliadub (approved):

PR #233, opened 2023-05-12
AJ-769: move to ehcache

davidangb:
In this PR:
1. Replace the simple in-memory ConcurrentHashMap cache provider - the "Simple" provider - with ehcache. This gives us way more features and configurability.
2. Modify the SamDao interface and implementation classes to play nice with ehcache. Previously we were caching the `org.broadinstitute.dsde.workbench.client.sam.model.SystemStatus` object, which is not serializable (ehcache requires serializable values); now we just cache a Boolean
3. Remove the eviction of Sam cache via a `@Scheduled` invocation and replace with a TTL on the cache itself
4. Add debug-level logging for cache operations

This PR is pre-work for [AJ-769](https://broadworkbench.atlassian.net/browse/AJ-769). In that ticket, I expect to make pretty heavy use of a new cache and would like control over that cache's size limits and TTL.

more info:
https://springframework.guru/using-ehcache-3-in-spring-boot/
https://www.baeldung.com/spring-boot-ehcache

[AJ-769]: https://broadworkbench.atlassian.net/browse/AJ-769?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/resources/ehcache.xml
+++ b/service/src/main/resources/ehcache.xml
@@ -41,7 +41,7 @@
     </cache>
 
     <cache alias="primaryKeys" uses-template="default">
-        <key-type>java.util.List</key-type>
+        <key-type>java.util.ArrayList</key-type>
         <value-type>java.lang.String</value-type>
         <expiry>
             <ttl unit="hours">24</ttl>
--- a/service/src/main/resources/ehcache.xml
+++ b/service/src/main/resources/ehcache.xml
@@ -1,0 +1,56 @@
+<config
+        xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance'
+        xmlns='http://www.ehcache.org/v3'
+        xsi:schemaLocation="
+            http://www.ehcache.org/v3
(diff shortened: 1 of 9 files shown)


service/src/main/resources/ehcache.xml
```diff
@@ -0,0 +1,56 @@
...
+            <offheap unit="MB">1</offheap>
+        </resources>
+    </cache>
+
+    <cache alias="primaryKeys" uses-template="default">
+        <key-type>java.util.List</key-type>
+        <value-type>java.lang.String</value-type>
+        <expiry>
+            <ttl unit="hours">24</ttl>
+        </expiry>
```
davidangb, line 48: this cache was previously infinite - I've now imposed a 24-hour TTL on it.

calypsomatic (approved):

yuliadub (approved):

PR #238, opened 2023-05-17
AJ-769: activity logging

davidangb:
Enables activity logging for all write operations, as part of control M1 for production readiness. I end up with log entries like this:
```
2023-05-17 15:25:31,925 INFO  [a50716bb-eb7b-4003-b571-26ee004d52cf] org.databiosphere.workspacedataservice.activitylog.ActivityLogger: user 213400982345703434599 created 1 instance(s) with id(s) [b0a945bd-a737-4017-a519-786392023a0d]
2023-05-17 15:25:58,307 INFO  [eee6fd6f-b108-4db9-8844-b8ae0b92ba83] org.databiosphere.workspacedataservice.activitylog.ActivityLogger: user 213400982345703434599 created 1 record(s) of type mytype with id(s) [123]
2023-05-17 15:26:05,042 INFO  [db12c397-3b56-412c-8846-6670f3358970] org.databiosphere.workspacedataservice.activitylog.ActivityLogger: user 213400982345703434599 deleted 1 record(s) of type mytype with id(s) [123]
```

The implementation in this PR is a bit extra and over-designed. I am looking ahead to a time when we are also saving activity log entries to a db and eventually expose them to end users, so end users can see the history/provenance of their data. It's overkill for now, when all we need is log entries.

Additional changes in this PR:
* Changed logging such that responses from Sam aren't logged unless we turn on trace-level logging. Sam responses can and do contain user email addresses, and we don't want to log those.
* Updated SamDaos to include the call to resolve a token to a userid
* Added a cache for resolving tokens to Sam userids, so we're not continually pinging Sam for this info.
* Moved the snapshot-import implementation out of DataRepoController and into a new DataRepoService, because this matches the design for our other controllers.

ATTENTION REVIEWER: I suggest reviewing this "top down" - that is, read the unit tests in `LogStatementTest` as a specification of what this code is supposed to do; then read the changes in `InstanceService`, `DataRepoService`, and `RecordOrchestratorService` to see how methods add entries to the activity log; then, read the files in the `org.databiosphere.workspacedataservice.activitylog` package and the Sam changes to see how it's all implemented under the covers.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/activitylog/ActivityEventBuilder.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/activitylog/ActivityEventBuilder.java
@@ -2,6 +2,8 @@ package org.databiosphere.workspacedataservice.activitylog;
 
 import org.databiosphere.workspacedataservice.sam.SamDao;
 import org.databiosphere.workspacedataservice.shared.model.RecordType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.web.context.request.RequestContextHolder;
 
 import java.util.UUID;
@@ -22,6 +24,8 @@ public class ActivityEventBuilder {
     private RecordType recordType;
     private int quantity;
     private String[] ids;
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ActivityEventBuilder.class);
 
     /**
      * Constructor.
@@ -50,7 +54,8 @@ public class ActivityEventBuilder {
             }
 
         } catch (Exception e) {
-            this.subject = "???";
+            LOGGER.warn("Error resolving user token to id via Sam: " + e.getMessage(), e);
+            this.subject = "(unknown due to exception)";
         }
         return this;
     }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/activitylog/ActivityEventBuilder.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/activitylog/ActivityEventBuilder.java
@@ -76,6 +76,11 @@ public class ActivityEventBuilder {
         this.action = ActivityModels.Action.MODIFY;
         return this;
     }
+    public ActivityEventBuilder linked() {
+        this.action = ActivityModels.Action.LINK;
+        return this;
+
+    }
     // OBJECT BUILDERS
 
     public ActivityEventBuilder instance() {
@@ -93,6 +98,10 @@ public class ActivityEventBuilder {
         return this;
     }
 
+    public ActivityEventBuilder snapshotReference() {
+        this.thing = ActivityModels.Thing.SNAPSHOT_REFERENCE;
+        return this;
+    }
     // RECORD TYPE BUILDERS
     public ActivityEventBuilder withRecordType(RecordType recordType) {
         this.recordType = recordType;
--- a/service/src/main/java/org/databiosphere/workspacedataservice/activitylog/ActivityEventBuilder.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/activitylog/ActivityEventBuilder.java
@@ -3,7 +3,6 @@ package org.databiosphere.workspacedataservice.activitylog;
 import org.databiosphere.workspacedataservice.sam.SamDao;
 import org.springframework.web.context.request.RequestContextHolder;
 
-import java.util.Objects;
 import java.util.UUID;
 
 import static org.databiosphere.workspacedataservice.sam.BearerTokenFilter.ATTRIBUTE_NAME_TOKEN;
@@ -18,6 +17,17 @@ public class ActivityEventBuilder {
     private ActivityModels.Action action;
     private ActivityModels.Thing thing;
     private String[] ids;
+
+    /**
+     * Takes a reference to the ActivityLogger that created this event, and which
+     * will eventually persist the event (to a log or to a db table)
+     * @param activityLogger
+     * @param samDao
+     */
+    public ActivityEventBuilder(ActivityLogger activityLogger, SamDao samDao) {
+        this.activityLogger = activityLogger;
+        this.samDao = samDao;
+    }
 
 
     // ===== SUBJECT BUILDERS
@@ -55,11 +65,6 @@ public class ActivityEventBuilder {
         return this;
     }
 
-//    public ActivityEventBuilder setAction(ActivityModels.Action action) {
-//        this.action = action;
-//        return this;
-//    }
-
     // OBJECT BUILDERS
 
     public ActivityEventBuilder instance() {
@@ -75,16 +80,6 @@ public class ActivityEventBuilder {
     public ActivityEventBuilder record() {
         this.thing = ActivityModels.Thing.RECORD;
         return this;
-    }
-//
-//    public ActivityEventBuilder setThing(ActivityModels.Thing thing) {
-//        this.thing = thing;
-//        return this;
-//    }
-
-    public ActivityEventBuilder(ActivityLogger activityLogger, SamDao samDao) {
-        this.activityLogger = activityLogger;
-        this.samDao = samDao;
     }
 
     // ID BUILDERS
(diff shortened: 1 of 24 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
```diff
@@ -83,11 +86,25 @@ public boolean hasWriteInstancePermission(String token) {
...
+                .getUserStatusInfo();
+        return httpSamClientSupport.withRetryAndErrorHandling(samFunction, "getUserInfo");
+    }
+
     /**
      * Gets the up/down status of Sam. Using @Cacheable, will reach out to Sam no more than
      * once every 5 minutes (configured in ehcache.xml).
      */
-    @Cacheable(value = "samStatus", key="'getSystemStatus'", cacheNames = "samStatus")
+    @Cacheable(cacheNames = "samStatus", key="'getSystemStatus'")
```
calypsomatic, line 107: What's the difference between `value` and `cacheNames`?

davidangb: ah sorry I meant to comment on this! `value` is an alias for `cacheNames` so the old code was specifying the same thing twice, I removed the redundancy.
https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/cache/annotation/Cacheable.html#value()

marctalbott (approved):
LGTM, appreciate the reviewing advice!

service/src/main/java/org/databiosphere/workspacedataservice/activitylog/ActivityEventBuilder.java
```diff
@@ -0,0 +1,138 @@
...
+                    .getAttribute(ATTRIBUTE_NAME_TOKEN, SCOPE_REQUEST);
+            if (token != null) {
+                // resolve the token to a user id via Sam
+                this.subject = samDao.getUserId(token.toString());
+            } else {
+                this.subject = "anonymous";
+            }
+
+        } catch (Exception e) {
+            this.subject = "???";
```
yuliadub, line 53: would it make sense to make this be something like "exception"?

service/src/main/java/org/databiosphere/workspacedataservice/activitylog/ActivityLogger.java
```diff
@@ -0,0 +1,75 @@
...
+            if (event.recordType() != null) {
+                sb.append(" of type %s".formatted(event.recordType().getName()));
+            }
+            if (event.ids() != null) {
+                sb.append(" with id(s) %s".formatted(Arrays.toString(event.ids())));
+            }
+            LOGGER.info(sb.toString());
+        }
+    }
+
```
yuliadub, line 73: nit: extra new lines :)

service/src/main/java/org/databiosphere/workspacedataservice/activitylog/ActivityModels.java
```diff
@@ -0,0 +1,42 @@
...
+
+        public String getName() {
+            return name;
+        }
+
+        Thing(String name) {
+            this.name = name;
+        }
+    }
+
```
yuliadub, line 35: some more empty lines in here

service/src/main/java/org/databiosphere/workspacedataservice/activitylog/ActivityEventBuilder.java
davidangb: good call; I changed this and logged a warning

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -158,22 +168,35 @@ public ResponseEntity<RecordResponse> upsertSingleRecord(UUID instanceId,String
...
                                                              RecordRequest recordRequest) {
         validateAndPermissions(instanceId, version);
-        return recordService.upsertSingleRecord(instanceId, recordType, recordId, primaryKey, recordRequest);
+        ResponseEntity<RecordResponse> response = recordService.upsertSingleRecord(instanceId, recordType, recordId, primaryKey, recordRequest);
+
+        if (response.getStatusCode() == HttpStatus.CREATED) {
+            activityLogger.saveEventForCurrentUser(user ->
+                    user.created().record().withRecordType(recordType).withId(recordId));
+        } else {
+            activityLogger.saveEventForCurrentUser(user ->
```
yuliadub, line 177: would it make sense to record the error (should be present in body of response) for those records that do not get created here as well)?

yuliadub: ah wait, never mind, this is created vs updated....not if there is an error :D

yuliadub: I guess my question still remains, what happens if the record creation or update fails, would we want to record that differently for this activity logging?

yuliadub:
I am curious why we need to have the user id in the log statements vs just tracking the workspace id for activities that are happening in those workspaces and generalizing the users? since the user id is not human readable at this time (i.e. I cant easily associate it with me doing the action) - or is it expected that users can translate user id to actual username if they do need to? Basically seems like having that layer for logging is overly complicated and I am unsure of the benefit of it. 

otherwise LGTM!

yuliadub (approved):

service/src/main/java/org/databiosphere/workspacedataservice/controller/DataRepoController.java
```diff
@@ -1,39 +1,34 @@
...
-
-        // createDataRepoSnapshotReference is required to setup policy and will throw exception if policy conflicts
-        workspaceManagerDao.createDataRepoSnapshotReference(snapshot);
-
-        // TODO do the import
-
+        validateVersion(version);
+        // TODO: validate the instance, when it's time to actually write anything to that instance
+        dataRepoService.importSnapshot(instanceId, snapshotId);
         return new ResponseEntity<>(HttpStatus.ACCEPTED);
```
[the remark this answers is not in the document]
davidangb: if `validateVersion` or `dataRepoService.importSnapshot` throw an exception, processing will halt and Spring's exception handlers kick in - the user will see an error. Though I agree - as this API is fleshed out, we'll want more granular error handling.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
davidangb: Hmm - I've taken the approach throughout this PR that I'm only logging if a write succeeded. I think it might get pretty hairy if we tried to add activity logging for all exception cases. Exceptions should already be written to logs - just not via this activity logger. Do you think that's necessary?

davidangb:
@yuliadub 

> I am curious why we need to have the user id in the log statements vs just tracking the workspace id for activities that are happening in those workspaces and generalizing the users? since the user id is not human readable at this time (i.e. I cant easily associate it with me doing the action) - or is it expected that users can translate user id to actual username if they do need to? Basically seems like having that layer for logging is overly complicated and I am unsure of the benefit of it.

For AJ-769, the purpose of this logging is to assist US - or other admins/auditors. In the event of a hack or security breach, we want audit logs that track all the writes, so we can see who did what. If we find anything suspicious, WE can take that Sam user id and correlate it back to a human identity. The extra layer of indirection - logging the userid instead of an email address - means we aren't logging PII (because email addresses count as PII), and that relieves oversight burden.

If/when we decide to expose these activity logs to end users, we'll definitely want to reconsider showing userids vs. email addresses.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
yuliadub: ah that makes sense - I guess I was thinking if we expose this to the user it could at least be useful to have something like "you tried to do a thing but it failed" with short exception message if that is available. But I guess it comes back to user expectations, if users only want to know activity that happened in WDS and errors are not actually making any changes then that makes sense to me.

yuliadub: based on your comment below, I think keeping as is makes perfect sense!

PR #239, opened 2023-05-17
AJ-1010 Add a placeholder table on TDR import

calypsomatic:
See [AJ-1010](https://broadworkbench.atlassian.net/browse/AJ-1010)

If an attempt to link a TDR snapshot passes the permission check, then a tdr-import table will be created (if it doesn't already exist) and rows for each table in the snapshot will be added.  This is a placeholder for future work where the data will actually be imported.

[AJ-1010]: https://broadworkbench.atlassian.net/browse/AJ-1010?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service/src/main/java/org/databiosphere/workspacedataservice/datarepo/DataRepoDao.java
```diff
@@ -3,29 +3,55 @@
...
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
 
 public class DataRepoDao {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(DataRepoDao.class);
+    public static final String TDRIMPORT_TABLE = "tdr-imports";
+    public static final String TDRIMPORT_PRIMARY_KEY = "tablename";
```
calypsomatic, line 24: Here I have exactly one tdr-imports table with only the table name as the primary key.  Although we will be fleshing out the data in the future, do we want to have multiple tables - one for each snapshot?  And if so, what should the naming convention be?

service/src/main/java/org/databiosphere/workspacedataservice/datarepo/DataRepoConfig.java
```diff
@@ -21,8 +22,8 @@ public DataRepoClientFactory getDataRepoClientFactory() {
         }
 
         @Bean
-        public DataRepoDao dataRepoDao(DataRepoClientFactory dataRepoClientFactory) {
-                return new DataRepoDao(dataRepoClientFactory);
+        public DataRepoDao dataRepoDao(DataRepoClientFactory dataRepoClientFactory, RecordDao recordDao) {
+                return new DataRepoDao(dataRepoClientFactory, recordDao);
```
davidangb, line 26: I don't love the idea of one dao relying on another dao. Could you move logic that is currently in DataRepoDao up to DataRepoService, and have DataRepoService depend on both DataRepoDao and RecordDao?

service/src/main/java/org/databiosphere/workspacedataservice/datarepo/DataRepoDao.java
```diff
@@ -3,29 +3,55 @@
...
+        LOGGER.debug("Importing snapshot {}", snapshot.getId());
+        RecordType importType = RecordType.valueOf(TDRIMPORT_TABLE);
+        if (!recordDao.recordTypeExists(instanceId, importType)){
+            recordDao.createRecordType(instanceId, Collections.emptyMap(), importType,
+                    new RelationCollection(Collections.emptySet(), Collections.emptySet()),
+                    TDRIMPORT_PRIMARY_KEY);
+        }
+        List<Record> records = new ArrayList<>();
+        for (TableModel table : snapshot.getTables()){
+            records.add(new Record(table.getName(), importType, RecordAttributes.empty()));
```
davidangb, line 53: ooh - could you add the snapshotId and the current datetime as attributes here? That would help to demo what happened.

also see my previous comment about moving this code up to DataRepoService.

service/src/main/java/org/databiosphere/workspacedataservice/datarepo/DataRepoDao.java
davidangb: I think this is fine for now, this is just an interim/demo-ready implementation.

service/src/main/java/org/databiosphere/workspacedataservice/datarepo/DataRepoDao.java
```diff
@@ -3,29 +3,55 @@
...
+        }
+        recordDao.batchUpsert(instanceId, importType, records, Collections.emptyMap());
```
davidangb, line 55: following up from Slack conversation, yeah, both the creation of the table and the insertion of the records should be activity-logged. That said since this is just demo code I don't care about activity logging … add it if you feel like playing with that code, but it's not a requirement

service/src/main/java/org/databiosphere/workspacedataservice/service/DataRepoService.java
```diff
@@ -27,9 +44,27 @@ public void importSnapshot(UUID instanceId, UUID snapshotId) {
...
         workspaceManagerDao.createDataRepoSnapshotReference(snapshot);
-
         activityLogger.saveEventForCurrentUser(user ->
                 user.linked().snapshotReference().withUuid(snapshotId));
-        // TODO do the import
+
+        //do the import
+        addSnapshot(snapshot, instanceId);
+        activityLogger.saveEventForCurrentUser(user ->
+                user.created().table().withId(snapshot.getName()));
```
davidangb, line 53: I'd move this `activityLogger` statement into `addSnapshot` and put it inside the `if (!recordDao.recordTypeExists … ` clause, otherwise we'd log it for every snapshot import, even if the table already exists

service/src/main/java/org/databiosphere/workspacedataservice/service/DataRepoService.java
```diff
@@ -27,9 +44,27 @@ public void importSnapshot(UUID instanceId, UUID snapshotId) {
...
+        if (!recordDao.recordTypeExists(instanceId, importType)){
+            recordDao.createRecordType(instanceId, Map.of("Snapshot Id", DataTypeMapping.STRING, "Import Time", DataTypeMapping.DATE_TIME), importType,
+                    new RelationCollection(Collections.emptySet(), Collections.emptySet()),
+                    TDRIMPORT_PRIMARY_KEY);
+        }
+        List<Record> records = new ArrayList<>();
+        for (TableModel table : snapshot.getTables()){
+            records.add(new Record(table.getName(), importType, RecordAttributes.empty().putAttribute("Snapshot Id", snapshot.getId()).putAttribute("Import Time", LocalDate.now())));
+        }
+        recordDao.batchUpsert(instanceId, importType, records, Collections.emptyMap());
```
davidangb, line 68: if this were the final implementation, you'd need to log the inserted records after the call to `batchUpsert`:
```java
        if (!records.isEmpty()) {
            activityLogger.saveEventForCurrentUser(user ->
                    user.created().record().withRecordType(importType).ofQuantity(records.size()));
        }
```
since this is interim code, your call on whether or not to add it

davidangb (commented):
minor comments about activity logging, otherwise looking good!

service/src/main/java/org/databiosphere/workspacedataservice/service/DataRepoService.java
```diff
@@ -27,9 +44,27 @@ public void importSnapshot(UUID instanceId, UUID snapshotId) {
...
+        addSnapshot(snapshot, instanceId);
+        activityLogger.saveEventForCurrentUser(user ->
+                user.created().table().withId(snapshot.getName()));
+    }
+
+    public void addSnapshot(SnapshotModel snapshot, UUID instanceId) {
+        LOGGER.debug("Importing snapshot {}", snapshot.getId());
+        RecordType importType = RecordType.valueOf(TDRIMPORT_TABLE);
+        if (!recordDao.recordTypeExists(instanceId, importType)){
+            recordDao.createRecordType(instanceId, Map.of("Snapshot Id", DataTypeMapping.STRING, "Import Time", DataTypeMapping.DATE_TIME), importType,
```
yuliadub, line 60: should we define these as final strings at the top of class? since we end up referencing them in more then one place

service/src/main/java/org/databiosphere/workspacedataservice/service/DataRepoService.java
calypsomatic: Good call!

service/src/main/java/org/databiosphere/workspacedataservice/service/DataRepoService.java
calypsomatic: sure why not

davidangb (approved):
nice! thank you

yuliadub (approved):

PR #240, opened 2023-05-17
AJ-1063: upgrade Sam client and org.json:json

davidangb:
* upgrades `sam-client` to latest version
* puts a constraint on `org.json:json`, which is a transitive dependency of `sam-client`, to use the latest version `20230227`. This addresses CVE-2022-45688.

Note: the fix from #228 does NOT appear necessary in this case, since the `dependencies.constraints` stanza is in service/build.gradle, not build.gradle. This means the POM file generation in the :client project is unaffected.

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -52,7 +52,7 @@ dependencies {
 	implementation 'org.ehcache:ehcache:3.10.8'
 
 	// Terra libraries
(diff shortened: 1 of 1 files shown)


ashanhol (approved):
LGTM! Straightforward. :)

aherbst-broad (approved):

PR #241, opened 2023-05-18
Da clean up warnings

davidangb:
DRAFT, do not review yet.

Clean up as many IntelliJ inspection warnings as I can stomach.

Intention is to have no change to end users.


calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/service/model/DataTypeMapping.java
```diff
@@ -68,24 +65,16 @@ public static DataTypeMapping getArrayTypeForBase(DataTypeMapping baseType){
...
-				return ARRAY_OF_NUMBER;
-			case DATE:
-				return ARRAY_OF_DATE;
-			case DATE_TIME:
-				return ARRAY_OF_DATE_TIME;
-			default:
-				throw new IllegalArgumentException("No supported array type for " + baseType);
-		}
+		return switch (baseType) {
+			case STRING -> ARRAY_OF_STRING;
```
ashanhol, line 69: I didn't know this switch formatting so I learned something from this PR :)

yuliadub (approved):

ashanhol (approved):

yuliadub:
the one thing I noticed is that <a> links references all have "..." as text, I assumed that is expected but I wasnt sure if it should have text about where the link leads instead :D

service/src/main/java/org/databiosphere/workspacedataservice/service/model/DataTypeMapping.java
davidangb: IntelliJ suggested the syntax for me in response to the warning, so it was super easy for me!

davidangb:
@yuliadub 

> the one thing I noticed is that links references all have "..." as text, I assumed that is expected but I wasnt sure if it should have text about where the link leads instead :D

good catch! I have updated. This was an IntelliJ auto-fix and I didn't even check it afterwards, my bad for missing this.

PR #242, opened 2023-05-18
AJ-1063: speed up unit tests

davidangb:
In my environment, unit tests take ~60 seconds. This isn't bad, but we can do better.

In very unscientific testing on my local system with other apps open, I saw the following results. These times are the average of running each test 5 times.

| | Before | After |
|--------|--------|--------|
| `TransactionRetryTest` | 37.62s | 2.38s |
| `ConcurrentDataTypeChangesTest` | 8.9s | 5.7s |
| `RecordOrchestratorSamTest` | 4.2s | 0.2s |

This PR speeds up tests by:
* reducing concurrency for `ConcurrentDataTypeChangesTest`
* reducing retry counts and wait durations for `TransactionRetryTest` and `RecordOrchestratorSamTest`

### DIFF
--- a/service/src/test/java/org/databiosphere/workspacedataservice/retry/TransactionRetryTest.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/retry/TransactionRetryTest.java
@@ -18,7 +18,14 @@ import static org.junit.jupiter.api.Assertions.*;
 
 
 @DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_CLASS)
-@SpringBootTest
+// aggressive retry settings to make this test fast; these are
+// inappropriate for runtime behavior.
+@SpringBootTest(properties = {
+        "terra.common.retry.transaction.slowRetryMaxAttempts=2",
+        "terra.common.retry.transaction.slowRetryInitialInterval=250ms",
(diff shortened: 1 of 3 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/retry/TransactionRetryTest.java
```diff
@@ -18,7 +18,14 @@
 
 
 @DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_CLASS)
-@SpringBootTest
+// aggressive retry settings to make this test fast; these are
+// inappropriate for runtime behavior.
+@SpringBootTest(properties = {
+        "terra.common.retry.transaction.slowRetryMaxAttempts=2",
```
ashanhol, line 24: Do we imagine we might see more test failures from tests that need more retries?

davidangb: I don't think so. These tests explicitly throw exceptions (they don't rely on some deadlock or anything in a real db), and they assert that the transaction was executed > 1 time, so I don't believe there's any variability here.

ashanhol (approved):
Those are some great unofficial benchmark numbers, I'm curious to see the comparison after several pipeline runs :)

PR #243, opened 2023-05-18
Fix Cache[primaryKeys] is closed unit test failures in #239.

davidangb:
Here's the recap of what I changed while trying to eliminate the `java.lang.IllegalStateException: Cache[primaryKeys] is closed` unit test failures in #239.

First, some web searches and Stack Overflow suggested that the `Cache … is closed` error happens when multiple callers attempt to use the same shared cache. Specific to Spring Boot unit tests, it could be multiple unit tests using the same instance of a CacheManager.

1. I started by setting `@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)` on `RecordOrchestratorServiceTest` in 27e54e65c5f80c8fd442eb577b53add13be1a589. This made tests pass, but because it makes a brand new Spring context for every test method, it increased runtime of the test from ~2s to ~16s.
2. I set `spring.cache.type=NONE` on `RecordOrchestratorServiceTest` in 63c8c19c43c21d6ca1361628a6d1ceed4032f35c. This also made tests pass, but felt wrong since runtime code uses cache - it makes the tests use a code path different than runtime.
3. I noticed that `DataRepoControllerMockMvcTest` did not have `@DirtiesContext`. I added it in 3698af7bbd0bbd6fac259e4f6b0e1c1682541bf9. This did not make tests pass.
4. I set `@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_CLASS)` on `RecordOrchestratorServiceTest` in cce40b0fc8c67494da3bc278ca8a0f2c425b265f. This made tests pass, but is also unacceptable because it leaves the Spring context around after this test class finishes;  whichever test class runs next would inherit the context. But, this was a good finding - it indicates that the problem really is that `RecordOrchestratorServiceTest` is reusing cache beans from some previous test class.
5. Finally, I added `@DirtiesContext` to every single test class that also uses `@SpringBootTest` in 4e0ad618137817fb5eb7a7be5b448b938e2e9858. This made tests pass and feels correct, in that each test cleans up after itself. Note that I also changed all instances of `@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_CLASS)` to be just `@DirtiesContext` since `AFTER_CLASS` is the default.

### DIFF
--- a/service/src/test/java/org/databiosphere/workspacedataservice/DataTypeInfererTest.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/DataTypeInfererTest.java
@@ -18,8 +18,10 @@ import org.databiosphere.workspacedataservice.shared.model.RecordType;
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.annotation.DirtiesContext;
 
- @SpringBootTest(classes = {JsonConfig.class, DataTypeInfererConfig.class})
+@DirtiesContext
+@SpringBootTest(classes = {JsonConfig.class, DataTypeInfererConfig.class})
 class DataTypeInfererTest {
 
 	@Autowired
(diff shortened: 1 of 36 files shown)


calypsomatic (approved):
Thanks so much for this investigation!

davidangb:
closing in deference to #244 which merged to main.

PR #245, opened 2023-05-19
AJ-999: incorporate Release Drafter action

davidangb:
part of AJ-999 but does not complete that ticket.

Incorporate the [Release Drafter](https://github.com/marketplace/actions/release-drafter) github action, using recommended config, to help us automatically generate drafts for release notes.

I'm suggesting we let this run for a bit and see what it does for us, then we can make a decision on whether or not it'll benefit us. We'll see, after we gather some data!

### DIFF
--- a/.github/release-drafter-config.yml
+++ b/.github/release-drafter-config.yml
@@ -1,0 +1,33 @@
+exclude-labels:
+  - 'skip-changelog'
+name-template: 'v$RESOLVED_VERSION 🌈'
+tag-template: 'v$RESOLVED_VERSION'
+categories:
+  - title: '🚀 Features'
+    labels:
+      - 'feature'
+      - 'enhancement'
+  - title: '🐛 Bug Fixes'
+    labels:
+      - 'fix'
+      - 'bugfix'
+      - 'bug'
+  - title: '🧰 Maintenance'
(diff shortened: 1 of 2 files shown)


.github/release-drafter-config.yml
```diff
@@ -0,0 +1,33 @@
...
+      - 'feature'
+      - 'enhancement'
+  - title: '🐛 Bug Fixes'
+    labels:
+      - 'fix'
+      - 'bugfix'
+      - 'bug'
+  - title: '🧰 Maintenance'
+    label: 'chore'
+change-template: '- $TITLE @$AUTHOR (#$NUMBER)'
```
yuliadub, line 17: where do author and number come from?

yuliadub (approved):

.github/release-drafter-config.yml
yuliadub: I guess these get pre-populated somehow https://github.com/marketplace/actions/release-drafter#change-template-variables

davidangb: yup, exactly - they're read from the PR.

ashanhol (approved):
I don't know github actions that well but the yml looks fine to me

PR #247, opened 2023-05-19
AJ-1028 Fix PostgreSQL error caused by blank TSV headers

ashanhol:
Fixes [AJ-1028](https://broadworkbench.atlassian.net/browse/AJ-1028)

This PR fixes an issue caused by extra tabs in a TSV file, specifically in the headers. This created empty columns and serializing issues for Postgres (specifically, `ERROR: zero-length delimited identifier`).
This fix is inserted in TsvStreamWriteHandler.java when we're processing records. It iterates through the attributes (which converts the Treemap to a Set since `while` looping and removing directly would give us false negatives if the value `remove()` returns is null) removing any entries with empty keys (aka empty TSV headers). 

We're now seeing happy 200s
image

[AJ-1028]: https://broadworkbench.atlassian.net/browse/AJ-1028?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/shared/model/RecordAttributes.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/shared/model/RecordAttributes.java
@@ -8,6 +8,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.Iterator;
 
 /**
  * Represents the attributes of a Record.
@@ -103,14 +104,18 @@ public class RecordAttributes {
 	}
 
 	public RecordAttributes removeNullHeaders() {
-		// Can't iterate over a TreeMap, so iterate over its set, check for empty keys, 
-		// then remove them from the TreeMap.
+		// Can't iterate over a TreeMap (or remove entries while iterating), so iterate over its 
+		// set, count empty keys, then iterate and remove them from the TreeMap.
 		Set<Map.Entry<String, Object>> attributeSet = this.attributes.entrySet();
+		int i = 0;
 		for(Map.Entry<String, Object> attribute : attributeSet) {
 			if (attribute.getKey().isEmpty()) {
-				this.attributes.remove(attribute.getKey());
+				i++;
 			}
 		}	
+		for(; i > 0; i--) {
+			this.attributes.remove("");
+		}
 		return this;
 	}
 
--- a/service/src/main/java/org/databiosphere/workspacedataservice/shared/model/RecordAttributes.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/shared/model/RecordAttributes.java
@@ -8,7 +8,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.TreeMap;
-import java.util.Iterator;
 
 /**
  * Represents the attributes of a Record.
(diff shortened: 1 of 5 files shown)


davidangb (commented):
Implementation looks good on this. Could you add some unit tests? `RecordAttributesTest` is a good place for testing `removeNullHeaders()` directly; `RecordControllerMockMvcTest` is probably the best place for testing behavior of a TSV with trailing tabs; see for instance `tsvWithDuplicateRowIdsInSameBatch()`.

yuliadub:
could be useful to add a unit test for this scenario in here (@davidangb keep me honest if this is the right place) https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/service/src/test/java/org/databiosphere/workspacedataservice/service/StreamingTest.java

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -331,6 +331,20 @@ void tsvWithDuplicateRowIdsInSameBatch() throws Exception {
...
+		MockMultipartFile file = new MockMultipartFile("records", "null_header.tsv", MediaType.TEXT_PLAIN_VALUE,
+				"""
+					value		
+					foo
+					bar		
+					baz		""".getBytes());
+
+		mockMvc.perform(multipart("/{instanceId}/tsv/{version}/{recordType}", instanceId, versionId, "null-headers")
+				.file(file)).andExpect(status().isOk());
+	}
```
calypsomatic, line 346: maybe we could add a line to this test that gets the resultant record type and make sure it has the correct columns?

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/RecordAttributes.java
```diff
@@ -102,6 +103,22 @@ public RecordAttributes removeAttribute(String key) {
...
+		// Can't iterate over a TreeMap (or remove entries while iterating), so iterate over its 
+		// set, count empty keys, then iterate and remove them from the TreeMap.
+		Set<Map.Entry<String, Object>> attributeSet = this.attributes.entrySet();
+		int i = 0;
+		for(Map.Entry<String, Object> attribute : attributeSet) {
+			if (attribute.getKey().isEmpty()) {
+				i++;
+			}
+		}	
+		for(; i > 0; i--) {
```
calypsomatic, line 116: Not anything to address in this PR, but this makes me wonder - are we allowing multiple columns with the same name?  Or is it just acting weirdly because it doesn't understand blank column names?

ashanhol: It's definitely acting weirdly because of the blank space since the error is
`bad SQL grammar ... \"\" ... nested exception is org.postgresql.util.PSQLException: ERROR: zero-length delimited identifier`
I see a test for duplicate RowIds but not columns so at the very least that's a scenario that seems to be getting through.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
ashanhol: added 😎

calypsomatic (approved):
Looks great!  Thank you!

davidangb (approved):
nice, thank you!!

PR #250, opened 2023-05-22
AJ-1068 Duplicate TSV Column Names Throws Error

ashanhol:
Fixes [Jira Bug](https://broadworkbench.atlassian.net/browse/AJ-1068)

Current behavior: User uploads a TSV with duplicate column names, which is currently allowed, but database cannot support that so it overwrites the first column with the duplicate column. 

Fix: Throw an error if we detect duplicate column names.


davidangb (approved):
woot thanks!

PR #252, opened 2023-05-24
AJ-885 Sam contract tests

calypsomatic:
[AJ-885](https://broadworkbench.atlassian.net/browse/AJ-885)
This PR adds consumer contract tests with Sam as the provider.  Since this is the first time contract tests are being added to WDS, it also updates build.gradle to add pactjvm.

[AJ-885]: https://broadworkbench.atlassian.net/browse/AJ-885?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service/src/test/java/org/databiosphere/workspacedataservice/pact/SamPactTest.java
```diff
@@ -0,0 +1,227 @@
...
+        SystemStatus samStatus = samDao.getSystemStatus();
+        assertTrue(samStatus.getOk());
+    }
+
+    @Test
+    @PactTestFor(pactMethod = "downStatusApiPact")
+    void testSamServiceDown(MockServer mockServer) {
+        SamClientFactory clientFactory = new HttpSamClientFactory(mockServer.getUrl());
+        SamDao samDao = new HttpSamDao(clientFactory, new HttpSamClientSupport(), UUID.randomUUID().toString());
+        assertThrows(SamServerException.class,
```
calypsomatic, line 167: Not sure if this is the right way to do this - because of the way we have the Sam infrastructure set up (SamDao with SamClientFactory and SamClientSupport), I can't directly check the Sam response in these failure cases, so I'm also checking that we're responding correctly by throwing an error.

davidangb (approved):
lgtm!

service/src/test/java/org/databiosphere/workspacedataservice/pact/SamPactTest.java
calypsomatic: Yes, this warning always happens with WDS!  Some day we'll fix it but we've been ignoring it as it's just a warning.  Thanks for taking a look!

ichengchang: Sounds good!

ichengchang (approved):
Looks good!

service/src/test/java/org/databiosphere/workspacedataservice/pact/SamPactTest.java
ichengchang: @calypsomatic Some of the endpoints in your pact tests are not yet covered by the `Sam` provider tests (e.g. those that involve `write` | `delete` permissions). So we will have to add coverage for these endpoints in order for `Sam` to verify the `wds` pact successfully.

service/build.gradle
```diff
@@ -96,3 +98,14 @@ jacocoTestReport {
...
+
+test {
+	systemProperties['pact.rootDir'] = "$buildDir/pacts"
+	systemProperties['pact.provider.version'] = "$project.version"
+}
+
+tasks.register("pactTests", Test) {
+	useJUnitPlatform {
+		includeTags "pact-test"
+	}
```
ichengchang, line 110:
```suggestion
	}
	testLogging {
		showStandardStreams = true
	}
```

ichengchang: Direct pact to send log to console instead of `xml` so we can easily verify the test is doing things that make sense.

ichengchang: You may also want to add following to `logback-spring.xml` under `test\resources`.

```
    <logger name="au.com.dius.pact.consumer" level="DEBUG" additivity="false">
        <appender-ref ref="Console" />
    </logger>
```

service/src/test/java/org/databiosphere/workspacedataservice/pact/SamPactTest.java
calypsomatic: Yes I was planning on addressing provider-side verification next!

ichengchang: 👍

PR #264, opened 2023-06-06
AJ-1099: clean up logging patterns

davidangb:
Changes WDS logging:
* removes the file logger, since we don't need that in prod
* removes color coding from console logs, so logs are easier to read in Azure Portal
* aligns the logging pattern very closely to Spring's default from org/springframework/boot/logging/logback/defaults.xml
* aligns the console appender very closely to Spring's default from org/springframework/boot/logging/logback/console-appender.xml

Here's what logging looks like now:
Screenshot 2023-06-06 at 1 41 26 PM

### DIFF
--- a/service/src/main/resources/logback-spring.xml
+++ b/service/src/main/resources/logback-spring.xml
@@ -1,13 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration>
 
-    <appender name="Console"
-              class="ch.qos.logback.core.ConsoleAppender">
-        <layout class="ch.qos.logback.classic.PatternLayout">
-            <Pattern>
-                %green(%d{ISO8601}) %highlight(%-5level) [%blue(%X{requestId})] %yellow(%C{1.}): %msg%n%throwable
-            </Pattern>
-        </layout>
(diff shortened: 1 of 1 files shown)


calypsomatic:
That huge gap where the request id would go is slightly obnoxious when it's not there - I don't suppose there's anything to do about that?

yuliadub (approved):

davidangb:
> That huge gap where the request id would go is slightly obnoxious when it's not there - I don't suppose there's anything to do about that?

Heh, I waffled on this. Leaving the gap there means the columns are justified - the next column always starts in the same place. But it would be trivial to collapse that space when the requestId doesn't exist and forget about the justification. Do you think we should collapse?

calypsomatic:
> > That huge gap where the request id would go is slightly obnoxious when it's not there - I don't suppose there's anything to do about that?
> 
> Heh, I waffled on this. Leaving the gap there means the columns are justified - the next column always starts in the same place. But it would be trivial to collapse that space when the requestId doesn't exist and forget about the justification. Do you think we should collapse?

Hmm... I see the appeal of keeping it justified.  I'm sure I'd be annoyed either way, so I suppose we can leave it.

calypsomatic (approved):

PR #265, opened 2023-06-06
AJ-1059 compare primary key on tsvs

calypsomatic:
[AJ-1059](https://broadworkbench.atlassian.net/browse/AJ-1059)
If a user tried to upload a tsv to a pre-existing table, but the primary key (leftmost column) in the tsv differed from that in the existing table, they received a hideous error.  If the primary key was passed as a parameter in the API call, then it was checked for conflict before reading the TSV.  But if it was not passed as a parameter, we could only find it out by reading the tsv, at which point we're past the point of checking against the existing table.
I solved this the easy way, by fetching the pre-existing primary key at the conflict check and passing it in to the `TSVStreamWriteHandler`, where previously would only be the parameter-passed key.  This results in a message "Uploaded TSV is either missing the `primaryKey` column or has a null or empty string value in that column" - which may be somewhat less clear than the "The primary key for `recordType` is already set to `oldPk` if you wish to change it you'll need to create a new record type." message we would get at the initial conflict check, but requires a lot less code change - am putting out this PR for discussion.

[AJ-1059]: https://broadworkbench.atlassian.net/browse/AJ-1059?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
@@ -106,7 +106,7 @@ public class RecordOrchestratorService { // TODO give me a better name
                           MultipartFile records) throws IOException {
         validateAndPermissions(instanceId, version);
         if(recordDao.recordTypeExists(instanceId, recordType)){
-            recordService.validatePrimaryKey(instanceId, recordType, primaryKey);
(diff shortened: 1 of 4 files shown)


davidangb (approved):
nice! thank you

yuliadub (approved):

PR #266, opened 2023-06-06
AJ-856: downgrade jackson-dataformat-csv

davidangb:
The version of `jackson-dataformat-csv` we had was not in sync with Spring Boot's version of other Jackson libraries, and this caused some annoying errors, like `java.lang.NoSuchFieldError: USE_FAST_DOUBLE_WRITER`.

This PR aligns the jackson versions and fixes the one syntax compatibility issue that came from the downgrade.


ashanhol (approved):
thank you

calypsomatic (approved):

PR #268, opened 2023-06-06
AJ-856: Migrate TSV Write to use Jackson over Apache

ashanhol:
Fixes https://broadworkbench.atlassian.net/browse/AJ-856

Current behavior: Still using org.apache.commons:commons-csv for TSV writing despite previously switching to Jackson for TSV reading. 

Fix: This PR switches to using Jackson for consistency sake and removes org.apache.commons:commons-csv from build. No noticeable change for users. 
Note: previous TSV Reading PR created custom deserializer class as well as a reader Bean. I found this to be unnecessary as we're already doing the work in recordDao to convert the PostgreSQL info into `Records`.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/TsvSupport.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/TsvSupport.java
@@ -1,19 +1,14 @@
 package org.databiosphere.workspacedataservice.service;
 
+import com.fasterxml.jackson.databind.SequenceWriter;
 import com.fasterxml.jackson.dataformat.csv.CsvMapper;
 import com.fasterxml.jackson.dataformat.csv.CsvSchema;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.SequenceWriter;
-import org.apache.commons.csv.CSVFormat;
-import org.apache.commons.csv.CSVPrinter;
-import org.apache.commons.csv.QuoteMode;
 import org.databiosphere.workspacedataservice.shared.model.Record;
 
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.List;
 import java.util.ArrayList;
-import java.util.function.Consumer;
 import java.util.stream.Stream;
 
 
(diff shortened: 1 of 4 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/controller/TsvDownloadTest.java
```diff
@@ -95,18 +92,15 @@ void batchWriteFollowedByTsvDownload() throws IOException {
...
-		assertThat(iterator.hasNext()).isFalse();
+        MappingIterator<RecordAttributes> tsvIterator = tsvReader.readValues(reader);
+		RecordAttributes recordAttributes = tsvIterator.next();
+		assertThat(recordAttributes.getAttributeValue("description")).isEqualTo("Embedded\tTab");
+		recordAttributes = tsvIterator.next();
+		assertThat(recordAttributes.getAttributeValue("description")).isEqualTo("\n,Weird\n String");
+		assertThat(recordAttributes.getAttributeValue("location")).isEqualTo("Cambridge, \"MA\"");
+		assertThat(recordAttributes.getAttributeValue("unicodeData")).isEqualTo("\uD83D\uDCA9ȇ");
+		assertThat(tsvIterator.hasNext()).isFalse();
 		reader.close();
```
davidangb, line 104: re: failures on this test … I don't know the exact root cause, but here's some info. This test writes the contents of  `batch_write_tsv_data.json`, then downloads those records as TSV and inspects the TSV.

If I edit `batch_write_tsv_data.json` to remove the `jsonObj`, `createdAt`, and `jsonArray` attributes, this test passes. If I leave any of those three attributes in the json, the test fails.

I also see in manual testing that array attributes and date attributes seem to write unexpected newlines to the resultant TSV. I'm guessing these are related.

davidangb (commented):
initial thoughts inline!

service/src/main/java/org/databiosphere/workspacedataservice/service/TsvSupport.java
```diff
@@ -1,50 +1,47 @@
...
-				csvPrinter.println();
-			} catch (IOException e) {
-				throw new RuntimeException(e);
-			}
-		}
+	private static List<Object> RecordToRow(Record record, List<String> headers) {
+		List<Object> row = new ArrayList<Object>();
+		row.add(record.getId());
+		headers.forEach(h -> {
+			row.add(record.getAttributeValue(h));
```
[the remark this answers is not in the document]
ashanhol: Added .withNullValue(""); to the CsvSchema 👍

service/src/main/java/org/databiosphere/workspacedataservice/service/TsvSupport.java
```diff
@@ -1,50 +1,48 @@
...
-					csvPrinter.print(attributeValue != null && attributeValue.getClass().isArray() ? objectMapper.writeValueAsString(attributeValue) : attributeValue);
-				}
-				csvPrinter.println();
-			} catch (IOException e) {
-				throw new RuntimeException(e);
-			}
-		}
+	private static List<String> recordToRow(Record record, List<String> headers) {
+		List<String> row = new ArrayList<String>();
+		//row.add(record.getId());
```
ashanhol, line 42: one test passes, the other fails depending on whether the line is commented or not. (specifically, with it commented out, batchWriteFollowedByTsvDownload passes, and with it uncommented, tsvUploadWithChoosenPrimaryKeyFollowedByDownload passes)

davidangb (commented):
Two Java-specific comments inline, otherwise looking good and thanks for sticking with it!

Also +1 to Bria's comment about figuring out/fixing the Python test failure.

service/src/main/java/org/databiosphere/workspacedataservice/service/TsvSupport.java
```diff
@@ -1,50 +1,59 @@
...
 public class TsvSupport {
 
 	private TsvSupport() {
 	}
 
-	public static CSVFormat getOutputFormat(List<String> headers) {
-		return CSVFormat.DEFAULT.builder().setDelimiter('\t').setQuoteMode(QuoteMode.MINIMAL)
-				.setHeader(headers.toArray(new String[0])).build();
-	}
+	public static void writeCsvToStream (Stream<Record> records, OutputStream stream, List<String> headers) throws IOException {
```
yuliadub, line 22: should we call this write Tsv to stream? :)

ashanhol: aw man good catch

ashanhol: fixed

service/src/main/java/org/databiosphere/workspacedataservice/service/TsvSupport.java
```diff
@@ -1,50 +1,59 @@
...
-		private final List<String> attributeNames;
+		final CsvMapper tsvMapper = CsvMapper.builder()
+		.enable(CsvGenerator.Feature.STRICT_CHECK_FOR_QUOTING)
+		.build();
 
-		private final ObjectMapper objectMapper;
+		SequenceWriter seqW = tsvMapper.writer(tsvHeaderSchema)
+			.writeValues(stream);
+		seqW.write(headers);
+		// First header is Primary Key, and value is stored in record.id. Remove header here and add record.id manually.
```
yuliadub, line 35: extra "and"

yuliadub:
check out the code smells link above, those look like worth fixing ^ otherwise looks good! thanks for figuring this out!

yuliadub (approved):

PR #269, opened 2023-06-12
AJ-1109: more helpful TSV errors

davidangb:
We already have error trapping for specific TSV problems, like duplicate column names. But, we didn't have catch-all error trapping for the exceptions we didn't expect; this PR adds that. Plus, some new known-bad TSVs for use in unit tests.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/TsvStreamWriteHandler.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/TsvStreamWriteHandler.java
@@ -38,7 +38,12 @@ public class TsvStreamWriteHandler implements StreamingWriteHandler {
 
 	public TsvStreamWriteHandler(InputStream inputStream, ObjectReader tsvReader, RecordType recordType, Optional<String> primaryKey) throws IOException {
 		this.inputStream = inputStream;
-		this.tsvIterator = tsvReader.readValues(inputStream);
+		try {
+			this.tsvIterator = tsvReader.readValues(inputStream);
+		} catch (Exception e) {
+			throw new InvalidTsvException("Error reading TSV. Please check the format of your upload. " +
+					"Underlying error is " + e.getClass().getSimpleName() + ": " + e.getMessage());
+		}
 
 		// check for no rows in TSV
 		if (!tsvIterator.hasNext()) {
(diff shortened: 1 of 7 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/service/TsvStreamWriteHandler.java
```diff
@@ -38,7 +38,12 @@ public class TsvStreamWriteHandler implements StreamingWriteHandler {
 
 	public TsvStreamWriteHandler(InputStream inputStream, ObjectReader tsvReader, RecordType recordType, Optional<String> primaryKey) throws IOException {
 		this.inputStream = inputStream;
-		this.tsvIterator = tsvReader.readValues(inputStream);
+		try {
+			this.tsvIterator = tsvReader.readValues(inputStream);
+		} catch (Exception e) {
+			throw new InvalidTsvException("Error reading TSV. Please check the format of your upload. " +
+					"Underlying error is " + e.getClass().getSimpleName() + ": " + e.getMessage());
+		}
```
davidangb, line 46: `tsvReader.readValues()` will throw a runtime exception in certain TSV cases, which wasn't caught anywhere and thus delivered an unhelpful error message.

service/src/main/java/org/databiosphere/workspacedataservice/service/model/exception/BadStreamingWriteRequestException.java
```diff
@@ -12,7 +12,7 @@ public class BadStreamingWriteRequestException extends IllegalArgumentException
 
 	private static final Logger LOGGER = LoggerFactory.getLogger(BadStreamingWriteRequestException.class);
 	public BadStreamingWriteRequestException(IOException ex) {
-		super("The server doesn't understand the request. Please check swagger to make sure you use"
+		super("The server doesn't understand the request. Please verify you are using "
```
davidangb, line 15: bonus fix: let's not point users at "swagger"; they probably don't know what it is and can't view it anyway

service/src/main/java/org/databiosphere/workspacedataservice/service/TsvStreamWriteHandler.java
ashanhol: How did you come by that error? Jc what I missed in my testing

ashanhol (approved):

service/src/main/java/org/databiosphere/workspacedataservice/service/TsvStreamWriteHandler.java
davidangb: > How did you come by that error? Jc what I missed in my testing

I reviewed all the actual end-user errors from production, as caught by Sentry. Then I made reproduction cases for each of those errors and found where they were throwing in the code - this was one of them.

service/src/test/java/org/databiosphere/workspacedataservice/tsv/TsvErrorMessageTest.java
```diff
@@ -0,0 +1,77 @@
...
+    void afterEach() {
+        List<UUID> allInstances = instanceService.listInstances(VERSION);
+        for (UUID id : allInstances) {
+            instanceService.deleteInstance(id, VERSION);
+        }
+    }
+
+    @ParameterizedTest(name = "The malformed TSV file '{0}' should throw a helpful error")
+    @ValueSource(strings = {"tsv-errors/column-separator.tsv", "tsv-errors/empty-header-line.tsv",
+            "tsv-errors/invalid-middle-byte.tsv", "tsv-errors/too-man-entries.tsv",})
```
calypsomatic, line 59: typo: "too-man-entries"

davidangb: thank you! Fixed.

calypsomatic (approved):

PR #270, opened 2023-06-12
AJ-1065: reuse http clients for Sam, WSM, TDR libraries

davidangb:
See https://broadworkbench.atlassian.net/browse/AJ-1065

We want to reuse the low-level http clients for the Sam, WSM, and TDR libraries. This PR does that, following examples previously set such as:

https://github.com/DataBiosphere/terra-workspace-manager/blob/3970810228cb235d507dcdb4a3c54f33fff5c463/service/src/main/java/bio/terra/workspace/service/iam/SamService.java#L106

https://github.com/DataBiosphere/terra-data-catalog/blob/main/common/src/main/java/bio/terra/catalog/rawls/RawlsClient.java

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamClientFactory.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamClientFactory.java
@@ -1,5 +1,6 @@
 package org.databiosphere.workspacedataservice.sam;
 
+import okhttp3.OkHttpClient;
 import org.apache.commons.lang3.StringUtils;
 import org.broadinstitute.dsde.workbench.client.sam.ApiClient;
 import org.broadinstitute.dsde.workbench.client.sam.api.ResourcesApi;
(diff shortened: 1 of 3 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamClientFactory.java
```diff
@@ -22,16 +23,24 @@
...
     private static final Logger LOGGER = LoggerFactory.getLogger(HttpSamClientFactory.class);
 
     public HttpSamClientFactory(String samUrl) {
         this.samUrl = samUrl;
+        this.commonHttpClient = new ApiClient()
+                .getHttpClient()
+                .newBuilder()
+                .build();
+        // TODO: add tracing interceptor for distributed tracing to Sam.
+        // this requires we import terra-common-lib
```
davidangb, line 37: pulling in terra-common-lib has annoying side effects, such as breaking our Sentry integration. I'd like to save that for follow-on work and hopefully figure out a way to do it gracefully.

PR #273, opened 2023-06-14
DRAFT, DO NOT MERGE: proof-of-concept JobRunr for async APIs

davidangb:
this shows how we'd implement async APIs using [JobRunr](https://www.jobrunr.io/en/?tab=developer).

My initial assessment of JobRunr … 

## Pros
* actively-maintained library (https://github.com/jobrunr/jobrunr)
* easy to integrate with Spring Boot
* nice property-file based config options for compatibility with our schemas
* natively understands MDC, so we can correlate async jobs to the request that spawned them
* support for retries
    * we need to make sure our jobs are idempotent if we want to use retries!
* built-in UI dashboard for local debugging
* uses our existing db for persistence
* nice features but I didn't test them:
    * distributed optimistic locking to support multiple replicas/horizontal scaling
    * jobs automatically restart if WDS restarts 
    * multiple options for persistence if we don't want to use the db

## Cons
* requires JobRunr pro ($$$) for programmatic access to job status and job results
    * workaround: query the JobRunr database tables directly. Fragile?
* using our existing db for job persistence complicates cloning a little bit. We don't want to clone job history (or do we?)
    * workaround: ensure we drop/recreate the JobRunr tables after cloning

more reading: https://www.baeldung.com/java-jobrunr-spring

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -55,7 +55,8 @@ dependencies {
 	implementation 'javax.cache:cache-api'
 	implementation 'org.ehcache:ehcache:3.10.8'
 	implementation 'org.hashids:hashids:1.0.3'
-	implementation 'org.jobrunr:jobrunr-spring-boot-2-starter:6.2.1'
+	implementation 'org.jobrunr:jobrunr-spring-boot-2-starter:6.2.1'	// JobRunr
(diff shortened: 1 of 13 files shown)


davidangb:
closing; we will not be using JobRunr

PR #274, opened 2023-06-15
DRAFT, DO NOT MERGE: proof-of-concept Quartz for async APIs

davidangb:
this shows how we'd implement async APIs using [Quartz](https://www.quartz-scheduler.org/).

My initial assessment of Quartz …

## Pros
* well-maintained library (https://github.com/quartz-scheduler/quartz)
* pretty large user community
* native support in Spring Boot: https://docs.spring.io/spring-boot/docs/2.7.12/reference/html/io.html#io.quartz
* property-file based config options
* uses our existing db for persistence
* nice features but I didn't test them:
    * cluster support for multiple WDS replicas/horizontal scaling
    * multiple options for persistence if we don't want to use the db

## Cons
* no built-in support for tracking or retrieving job status
    * we can easily build our own, with full control over statuses. Is this actually a pro?
* using our existing db for job persistence complicates cloning a little bit. We don't want to clone job history (or do we?)
    * workaround: ensure we drop/recreate the Quartz tables after cloning
    * may be able to rely on auto-drop/auto-recreate of Quartz tables to wipe
* complicated to put the Quartz tables in the sys_wds schema
* no native support for MDC, we'd have to implement it
* retry support is primitive
    * Quartz natively supports immediately retrying infinitely
    * we'd have to implement our own backoff and retry-limit policies


davidangb:
closing; this is now implemented for real

PR #275, opened 2023-06-15
AJ-1115 workload identity options

dvoet:
https://broadworkbench.atlassian.net/browse/AJ-1115

Reminder:


service/src/main/resources/application.properties
```diff
@@ -4,7 +4,9 @@ management.endpoints.web.path-mapping.info=version
 management.endpoints.web.path-mapping.health=status
 management.endpoint.health.show-details=ALWAYS
 
-spring.datasource.jdbc-url=jdbc:postgresql://${WDS_DB_HOST:localhost}:${WDS_DB_PORT:5432}/wds?reWriteBatchedInserts=true
+# When running on Azure in k8s with workload identity set ADDITIONAL_JDBC_URL_PARAMS to
+# sslmode=require&authenticationPluginClassName=com.azure.identity.extensions.jdbc.postgresql.AzurePostgresqlAuthenticationPlugin
+spring.datasource.jdbc-url=jdbc:postgresql://${WDS_DB_HOST:localhost}:${WDS_DB_PORT:5432}/${DB_NAME:wds}?reWriteBatchedInserts=true${ADDITIONAL_JDBC_URL_PARAMS}
```
davidangb, line 9: we'll also need `{ADDITIONAL_JDBC_URL_PARAMS}` on the `streaming.query.jdbc-url` definition

davidangb (commented):
one comment inline, this looks much simpler than I had expected/feared it might be!

I think we can get rid of the `ManagedIdentityDao` now too, if we don't need that code for this implementation.

yuliadub:
what David said, my only other question is, what would provide the database endpoint url? would that be something Leo would give to us or is this an api WDS would have to make? since I am not seeing changes in any other db transactions besides cloning maybe that info is already in WDS somewhere and I just dont know where to look.

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
```diff
@@ -43,13 +46,16 @@ public class BackupService {
...
     public BackupResponse backupAzureWDS(BackUpFileStorage storage, String version) {
         try {
             validateVersion(version);
             String blobName = GenerateBackupFilename();
 
             List<String> commandList = GenerateCommandList();
-            Map<String, String> envVars = Map.of("PGPASSWORD", dbPassword);
+            Map<String, String> envVars = Map.of("PGPASSWORD", determinePassword());
 
             LocalProcessLauncher localProcessLauncher = new LocalProcessLauncher();
```
yuliadub, line 60: today this list of commands also takes in the url to the db (which is localhost, set by env param) would the expectation be that we get the flex postgres endpoint from Leo?

dvoet: yes, Leo would populate relevant env vars

davidangb (approved):
thank you!

yuliadub (approved):

yuliadub:
thank you!  properties file looks so much better :)

PR #276, opened 2023-06-15
AJ-999: publish release-drafter release in tag workflow

davidangb:
part of AJ-999 but does not complete that ticket.

We have been using the the [Release Drafter](https://github.com/marketplace/actions/release-drafter) action to collect release notes in a draft release as PRs merge.

This PR adds automation to automatically publish the draft release at the same time we do the other release actions in `tag.yml`.


.github/release-drafter-config.yml
```diff
@@ -1,7 +1,7 @@
 exclude-labels:
   - 'skip-changelog'
 name-template: 'v$RESOLVED_VERSION 🌈'
-tag-template: 'v$RESOLVED_VERSION'
+tag-template: '$RESOLVED_VERSION'
```
davidangb, line 4: bumper creates tags without a "v"; our release should point at an existing tag instead of creating a new one.

.github/workflows/tag.yml
```diff
@@ -57,3 +57,21 @@ jobs:
...
+    steps:
+      - uses: release-drafter/release-drafter@v5
+        with:
+          # override release-drafter's version calculation to use the version generated by bumper
+          version: '${{ needs.tag-job.outputs.new_tag }}'
+          # publish the release; it is no longer draft
+          publish: true
+          config-name: release-drafter-config.yml
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
davidangb, line 77: see similar code in `release-drafter.yml`

calypsomatic (approved):

ashanhol (approved):

PR #277, opened 2023-06-15
auth docker before pushing to GCR

davidangb:
#212 introduced changes to `publish-docker.yml` that broke; see https://github.com/DataBiosphere/terra-workspace-data-service/actions/runs/5282685550/jobs/9557927242 for an example of the error.

This PR changes things around to ensure we auth before attempting to push to GCR.

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -15,7 +15,7 @@ on:
         required: true
       BROADBOT_TOKEN:
         required: true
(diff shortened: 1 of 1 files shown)


ashanhol (approved):

ashanhol:
it is very frustrating that computers do exactly what we tell them to do in the order we specify it.

yuliadub (approved):

PR #279, opened 2023-06-16
AJ-936: Implement Restore for Cloning API

ashanhol:
This PR implements the API to restore a pg_dump to our database. It downloads a (hardcoded) file from Azure to local, and executes a `psql -f` command to restore. 
I had to do a bit of finagling for the integration test. I had to call backup to mock the pg_dump file locally, then scrub the file and database of "workspace-id" to convert that guid to "source-workspace-id" since in lieu of spinning up a new version of wds.


service/src/main/java/org/databiosphere/workspacedataservice/controller/CloningController.java
```diff
@@ -11,10 +13,12 @@
 public class CloningController {
 
     private final BackupService backupService;
+    private final RestoreService restoreService;
     private final AzureBlobStorage storage;
-    public CloningController(BackupService backupService) {
+    public CloningController(BackupService backupService, RestoreService restoreService) {
```
ashanhol, line 18: So most of RestoreService is the same as BackupService (since we're doing the same thing in reverse). I wasn't sure if we really needed a whole separate service or if we should combine BackupService and RestoreService into some single CloningOrchestratorService?

service/src/main/java/org/databiosphere/workspacedataservice/storage/AzureBlobStorage.java
```diff
@@ -11,9 +11,11 @@
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.io.UncheckedIOException;
 import java.nio.charset.StandardCharsets;
 
 public class AzureBlobStorage implements BackUpFileStorage {
+    // TODO: we won't have access to a unique container name, move to workspaceid container with specific path
```
ashanhol, line 18: @yuliadub you mentioned this was work we might have to do so I added a TODO

service/src/main/java/org/databiosphere/workspacedataservice/storage/AzureBlobStorage.java
```diff
@@ -34,6 +36,19 @@ public void streamOutputToBlobStorage(InputStream fromStream, String blobName) {
...
     }
+    
+    @Override
+    public void streamInputFromBlobStorage(String blobName) {
+        // TODO: remove this once connection is switched to be done via SAS token
+        String storageConnectionString = System.getenv("STORAGE_CONNECTION_STRING");
+        BlobContainerClient blobContainerClient = constructBlockBlobClient(backUpContainerName, storageConnectionString);
+        
+        try{
+            blobContainerClient.getBlobClient(blobName).downloadToFile("backup.sql");
```
ashanhol, line 47: I decided to download the file so I could call psql -f <FileName>

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -820,6 +841,15 @@ components:
         message:
           type: string
           description: message regarding backup status
+    RestoreResponse:
```
ashanhol, line 844: This is the same as BackupResponse. Since we're looking to make Backup async I wasn't sure if it'd be right to have the same object represent both backup and restore?

PR #280, opened 2023-06-16
[not ready for review yet] AJ 955 trigger backup when wds starts in clone mode

yuliadub:
<will populate once ready for review>

Reminder:


Dockerfile
```diff
@@ -3,7 +3,7 @@
 ### to faciliate the cloning of a WDS instance/workspace in Terra
 
 ### Sourced from https://github.com/broadinstitute/dsp-appsec-blessed-images/blob/main/jre/Dockerfile.17-debian
-FROM us.gcr.io/broad-dsp-gcr-public/base/jre:17-debian
+FROM us.gcr.io/broad-dsp-gcr-public/sandbox/debian-test-wds
```
yuliadub, line 6: need to fix this

service/src/main/java/org/databiosphere/workspacedataservice/leonardo/HttpLeonardoClientFactory.java
```diff
@@ -0,0 +1,57 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(HttpLeonardoClientFactory.class);
+
+
+    public HttpLeonardoClientFactory(String leoUrl) {
+        this.leoEndpointUrl = leoUrl;
+    }
+
+    private ApiClient getApiClient(String token) {
+        // create a new data repo client
+        ApiClient apiClient = new ApiClient();
```
yuliadub, line 27: could not switch Leo to HTTP client due to: 
Error: class org.glassfish.jersey.client.JerseyClient cannot be cast to class okhttp3.OkHttpClient (org.glassfish.jersey.client.JerseyClient and okhttp3.OkHttpClient are in unnamed module of loader 'app'), unless I used a different client then the HTTP default one. Havent dug into it more yet.

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/BackupServiceIntegrationTest.java
```diff
@@ -1,22 +1,42 @@
...
     void testBackupAzureWDS() throws Exception {
-        var response = backupService.backupAzureWDS(storage, "v0.2");
-        assertTrue(response.backupStatus(), response.message());
+        var trackingId = UUID.randomUUID();
+        var sourceWorkspaceId = UUID.fromString("123e4567-e89b-12d3-a456-426614174001");
+        backupService.backupAzureWDS("v0.2", trackingId, sourceWorkspaceId);
+
+        var response = backupService.checkBackupStatus(trackingId);
+        assertEquals(true, response.backupStatus());
+        assertEquals("COMPLETED", response.state());
```
calypsomatic, line 34: nit - would we want to use the `BackupSchema.BackupState` enum here instead of a string?

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/BackupServiceIntegrationTest.java
```diff
@@ -1,22 +1,42 @@
...
+
+        var backupRecord = backupDao.getBackupStatus(trackingId);
+        assertNotNull(backupRecord);
```
calypsomatic, line 37: Is this part just checking that the response from the service matches the response from the dao?

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -12,62 +20,104 @@
...
-    public InstanceInitializerBean(InstanceDao instanceDao){
+    public InstanceInitializerBean(InstanceDao instanceDao, BackupDao backupDao){
         this.instanceDao = instanceDao;
+        this.backupDao = backupDao;
     }
 
     public boolean isInCloneMode(String sourceWorkspaceId) {
         if (StringUtils.isNotBlank(sourceWorkspaceId)){
-            LOGGER.info("Source workspace id found, checking database");
+            LOGGER.info("DEBUGMODE: Source workspace id found, checking database");
```
calypsomatic, line 46: Did you mean to leave this in here?  Otherwise it could be LOGGER.debug right?

service/src/main/java/org/databiosphere/workspacedataservice/leonardo/HttpLeonardoClientFactory.java
```diff
@@ -0,0 +1,57 @@
...
+
+        // initialize the client with the url to data repo
+        if (StringUtils.isNotBlank(leoEndpointUrl)) {
+            apiClient.setBasePath(leoEndpointUrl);
+        }
+
+        // grab the current user's bearer token (see BearerTokenFilter)
+        if(token.isEmpty()) {
+            Object userToken = RequestContextHolder.currentRequestAttributes()
```
calypsomatic, line 36: Not for this PR but just thinking aloud - we have a lot of copy-pasta between clients that do this endpoint setting and token checking.  Maybe for fun someday we could abstract it out?

service/src/main/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDao.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.List;
+import java.util.Map;
+
+public class LeonardoDao {
+  private final LeonardoClientFactory leonardoClientFactory;
+  private final String workspaceId;
+
+  public LeonardoDao(LeonardoClientFactory leonardoClientFactory, String workspaceId) {
+    this.leonardoClientFactory = leonardoClientFactory;
+    this.workspaceId = workspaceId;
```
calypsomatic, line 16: So the LeoDao for a given WDS is essentially monogamous with the source WDS, since it is created with that workspaceId and stays that way forever?  No complaint, unless there comes some day in the future when we're using Leo for something other than cloning.

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
```diff
@@ -45,13 +47,44 @@ public class BackupService {
...
+                return new BackupResponse(true, BackupSchema.BackupState.COMPLETED.toString(), backup.getFilename(), "Backup successfully completed.");
+            } else if (backup.getState() == BackupSchema.BackupState.ERROR) {
+                return new BackupResponse(true, BackupSchema.BackupState.ERROR.toString(), "", "Backup completed with an error.");
+            } else {
+                return new BackupResponse(false, backup.getState().toString(), "", "Backup still in progress.");
+            }
+        }
+        else {
+            backupDao.updateFilename(UUID.randomUUID(), "hi");
+            return new BackupResponse(false, "", "", "Backup not found.");
```
calypsomatic, line 74: Is the "hi" essentially thrown away then or what?

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
```diff
@@ -110,6 +151,7 @@ public String GenerateBackupFilename() {
         LocalDateTime now = LocalDateTime.now();
         DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd_HH-mm-ss");
         String timestamp = now.format(formatter);
-        return workspaceId + "-" + timestamp + ".sql";
+        return "wdsservice/cloning/backup/" + workspaceId + "-" + timestamp + ".sql";
```
calypsomatic, line 154: nit - do we want this location to a class variable or something - or is that to be dealt with in the ticket where we settle on the location?

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
```diff
@@ -30,18 +30,48 @@ public void createDataRepoSnapshotReference(SnapshotModel snapshotModel) {
...
+      UUID workspaceUUID = UUID.fromString(workspaceId);
+      ResourceList resourceList = resourceApi.enumerateResources(workspaceUUID, 0, 5, ResourceType.AZURE_STORAGE_CONTAINER, null);
+      // note: it is possible a workspace may have more than one storage container associated with it
+      // but currently there is no way to tell which one is the primary except for checking the actual container name
+      var storageUUID = extractResourceId(resourceList);
+      if(storageUUID != null) {
+        CreatedAzureStorageContainerSasToken sasBundle = azureResourceApi.createAzureStorageContainerSasToken(workspaceUUID, storageUUID, null, null, null, null);
+        return sasBundle.getUrl();
+      }
+      else throw new ApiException("Cant locate a storage resource matching workspace Id. ");
```
calypsomatic, line 64: extreme nit: Cant -> Can't

service/src/test/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDaoTest.java
```diff
@@ -0,0 +1,82 @@
...
+        assertEquals(null, url);
+    }
+
+    @Test
+    void testWdsUrlNotFoundWrongKey() throws ApiException {
+        var url = helperMethod("not-wds", AppStatus.RUNNING);
+        assertEquals(null, url);
+    }
+
+    String helperMethod(String wdsKey, AppStatus wdsStatus){
```
calypsomatic, line 70: ... can this have a slightly more specific name?

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/BackupServiceIntegrationTest.java
yuliadub: pretty much, just making sure that whatever got recorded and what got returned matches. (and that it is actually there).

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
yuliadub: nope, did not, good catch - thank you!

service/src/main/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDao.java
yuliadub: yeah I wasnt sure if there was a better way to do this - making this a Bean seemed unnecessary since really the only time we use it is during start up for cloning

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
yuliadub: lol whooops!

service/src/test/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDaoTest.java
yuliadub: I am so bad at names :) let me know if that is better!

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -40,34 +47,77 @@ public boolean isInCloneMode(String sourceWorkspaceId) {
...
-        LOGGER.info("No source workspace id found, initializing default schema");
+        LOGGER.info("No source workspace id found, initializing default schema.");
         return false;
 
     }
 
     public void initCloneMode(){
         LOGGER.info("Starting in clone mode...");
+
+        try {
```
ashanhol, line 72: Can we add that comment clarifying how cloning works? (what is source, what is dest, etc)

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
```diff
@@ -18,13 +21,12 @@
 
 import static org.databiosphere.workspacedataservice.service.RecordUtils.validateVersion;
 
-import com.azure.identity.extensions.jdbc.postgresql.AzurePostgresqlAuthenticationPlugin;
-
 @Service
 public class BackupService {
+    private final BackupDao backupDao;
+    private BackUpFileStorage storage;
```
ashanhol, line 27: We're not passing storage as a param anymore through backupAzureWDS so how do we know whether to use the azure version or the local file storage version?

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
```diff
@@ -45,13 +47,43 @@ public class BackupService {
 
     @Value("${twds.pg_dump.path:}")
     private String pgDumpPath;
-
+
```
ashanhol, line 50: nit: remove whitespace

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
```diff
@@ -45,13 +47,43 @@ public class BackupService {
...
+    public BackupService(BackupDao backupDao, BackUpFileStorage backUpFileStorage) {
+        this.backupDao = backupDao;
+        this.storage = backUpFileStorage;
+    }
+
+    public BackupResponse checkBackupStatus(UUID trackingId) {
+        var backup = backupDao.getBackupStatus(trackingId);
+
+        if(backup !=null) {
+            if (backup.getState() == BackupSchema.BackupState.COMPLETED) {
```
ashanhol, line 64: switch statement?

service/src/main/java/org/databiosphere/workspacedataservice/storage/AzureBlobStorage.java
```diff
@@ -6,22 +6,33 @@
...
+
     @Override
-    public void streamOutputToBlobStorage(InputStream fromStream, String blobName) {
+    public void streamOutputToBlobStorage(InputStream fromStream, String blobName, String workspaceId) {
         // TODO: remove this once connection is switched to be done via SAS token
-        String storageConnectionString = System.getenv("STORAGE_CONNECTION_STRING");
-        BlobContainerClient blobContainerClient = constructBlockBlobClient(backUpContainerName, storageConnectionString);
-
+        LOGGER.info("Creating blob storage client. ");
+        BlobContainerClient blobContainerClient = constructBlockBlobClient(workspaceId);
```
ashanhol, line 34: Can we add a comment example of what we expect the naming format to be? The expectations of this are scattered in a bunch of different files/functions

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
```diff
@@ -30,18 +30,48 @@ public void createDataRepoSnapshotReference(SnapshotModel snapshotModel) {
...
-              new DataRepoSnapshotAttributes()
-                  .instanceName(INSTANCE_NAME)
-                  .snapshot(snapshotModel.getId().toString()))
-              .metadata(
-                  new ReferenceResourceCommonFields()
-                      .cloningInstructions(CloningInstructionsEnum.REFERENCE)
-                      .name("%s_%s".formatted(snapshotModel.getName(), timeStamp))
-              ),
-          UUID.fromString(workspaceId));
+                      .snapshot(
```
ashanhol, line 33: nit: double check this whitespace? it's pushed over pretty far

service/src/main/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDao.java
```diff
@@ -0,0 +1,49 @@
...
+  public String extractWdsUrl(List<ListAppResponse> response) {
+    for(int i=0; i< response.size(); i++) {
+      Map<String, String> proxyUrls = ((Map<String, String>) response.get(i).getProxyUrls());
+      var url = proxyUrls.get("wds");
+      if (url != null && response.get(i).getStatus() == AppStatus.RUNNING) {
+        return url;
+      }
+    }
+    return null;
+  }
```
davidangb, line 48: I took a shot at replicating the logic we have in the UI. This is totally untested and deserves a unit test or two!
```suggestion
  class AppCreationDateComparator implements Comparator<ListAppResponse> {
    @Override
    public int compare(ListAppResponse o1, ListAppResponse o2) {
      return o1.getAuditInfo().getCreatedDate().compareTo(o2.getAuditInfo().getCreatedDate());
    }
    @Override
    public boolean equals(Object obj) {
      return false;
    }
  }

  public String extractWdsUrl(List<ListAppResponse> response) {
    // look for apps of type "WDS" which are in "RUNNING" status.
    // if more than one, choose the earliest-created. This matches Terra UI logic.
    Optional<ListAppResponse> maybeApp = response.stream()
            .filter(app -> AppType.WDS.equals(app.getAppType()) && AppStatus.RUNNING.equals(app.getStatus()))
            .min(new AppCreationDateComparator());

    if (maybeApp.isPresent()) {
      Map<String, String> proxyUrls = ((Map<String, String>) maybeApp.get().getProxyUrls());
      if (proxyUrls != null) {
        String url = proxyUrls.getOrDefault("wds", "");
        if (StringUtils.isNotBlank(url)) {
          return url;
        }
      }
    }
    return null;
  }
```

davidangb (commented):
I haven't finished reviewing, but figured I'd submit comments so far to get you some feedback. I'll continue tomorrow!

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
yuliadub: good catch, I made this change since I noticed a pattern where Controllers usually take Services as params and Services take Daos. Plus I wanted to be able to auto wire and not have to default to one or another implementation for storage. I made a few changes to make sure the code can know which storage interface to leverage:
- in the main service, I actually just added an AzureBlobStorage constructor and since that is the only BackUpFileStorage implementation spring boot magically finds it within me needing to do anything
- in the integration test, I specifically added a new implementation of BackUpFileStorage, so then the integration test now can recognize that there are 2 possible options: mock and azure blob. To tell spring boot which one to use, I created a Bean for BackUpFileStorage which returns a LocalFileStorage object, which I then marked Primary which tells spring boot to use that instead of the Azure blob. 

Let me know if this makes sense.

service/src/main/java/org/databiosphere/workspacedataservice/storage/AzureBlobStorage.java
yuliadub: the GenerateBackupFilename should generally be the only place that determines naming of the backup file, but I agree the storage naming can be confusing. Let me know if the comment I added actually helped.

Dockerfile
yuliadub: this will actually need to become  us.gcr.io/broad-dsp-gcr-public/base/wds-debian-pg-dump:latest since we are building our own image now that has pg_dump that we will use for jib.

service/build.gradle
```diff
@@ -56,19 +60,29 @@ dependencies {
...
 	implementation "bio.terra:datarepo-client:1.476.0-SNAPSHOT"
-	implementation "bio.terra:workspace-manager-client:0.254.717-SNAPSHOT"
+	implementation "bio.terra:workspace-manager-client:0.254.770-SNAPSHOT"
+	implementation project(path: ':client')
+
+	// hk2 is required to use WSM client, but not correctly exposed by the client
+	implementation "org.glassfish.jersey.inject:jersey-hk2:$jersey_version"
+	implementation "org.glassfish.jersey.core:jersey-client:$jersey_version"
+	implementation "org.glassfish.jersey.core:jersey-common:$jersey_version"
+    testImplementation 'junit:junit:4.13.1'
```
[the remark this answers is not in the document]
yuliadub: thanks for catching these, when I was experimenting with how dependencies work I wasnt sure if having it in implementation was enough. or not.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresBackupDao.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Override
+    public String getBackupRequestStatus(UUID sourceWorkspaceId, UUID destinationWorkspaceId) {
+        try {
+            return namedTemplate.getJdbcTemplate().queryForObject(
+                    "select status from sys_wds.backup_requests WHERE sourceworkspaceid = ? and destinationworkspaceid = ?", String.class, sourceWorkspaceId, destinationWorkspaceId);
+        }
+        catch(Exception e) {
+            return null;
+        }
```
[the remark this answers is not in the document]
yuliadub: these where I had the most issues when I was debugging and kept going back and forth on if log statements would be useful here. Landed on adding them for now!

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresBackupDao.java
```diff
@@ -0,0 +1,110 @@
...
+    @Override
+    public boolean backupExists(UUID trackingId) {
+        return Boolean.TRUE.equals(namedTemplate.queryForObject(
+                "select exists(select from sys_wds.backup WHERE id = :trackingId)",
+                new MapSqlParameterSource("trackingId", trackingId), Boolean.class));
+    }
+
+    @Override
+    @WriteTransaction
+    @SuppressWarnings("squid:S2077") // since trackingId must be a UUID, it is safe to use inline
```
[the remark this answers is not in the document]
yuliadub: I am honestly not sure how these appeared, for a second I thought it was copy/paste but dont remember updating the comment. From looking more into it doesnt seem like I need that suppress.

service/src/main/java/org/databiosphere/workspacedataservice/leonardo/HttpLeonardoClientFactory.java
yuliadub: aha, I figured there was something like this! thank you!

env.txt
```diff
@@ -0,0 +1,7 @@
+WORKSPACE_ID=1117a5d3-9f2c-42d0-9961-550cdbe3585a
+SAM_URL=https://sam.dsde-dev.broadinstitute.org/
```
yuliadub, line 2: whoops this is an artifact of my testing, will remove.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -40,34 +47,90 @@ public boolean isInCloneMode(String sourceWorkspaceId) {
             try {
                 UUID.fromString(sourceWorkspaceId);
             } catch (IllegalArgumentException e){
-                    LOGGER.warn("Source workspace id could not be parsed, unable to clone DB. Provided source workspace id: {}", sourceWorkspaceId);
+                    LOGGER.warn("Source workspace id could not be parsed, unable to clone DB. Provided source workspace id: {}.", sourceWorkspaceId);
                     return false;
             }
+
+            if (sourceWorkspaceId == workspaceId) {
+                LOGGER.warn("Source workspace id and current workspace Id can't be the same.");
```
yuliadub, line 55: nit to myself: workspace Id

service/src/main/java/org/databiosphere/workspacedataservice/sourcewds/HttpWorkspaceDataServiceClientFactory.java
```diff
@@ -0,0 +1,57 @@
...
+    public HttpWorkspaceDataServiceClientFactory() {
+        this.commonHttpClient = new bio.terra.workspace.client.ApiClient().getHttpClient();
+    }
+
+    private ApiClient getApiClient(String token, String workspaceDataServiceUrl) {
+        // create a new client
+        ApiClient apiClient = new ApiClient();
+        apiClient.setHttpClient(commonHttpClient);
+
+        // initialize the client with the url to data repo
```
davidangb, line 29: nit: not "data repo"

service/src/main/java/org/databiosphere/workspacedataservice/storage/AzureBlobStorage.java
```diff
@@ -35,20 +56,23 @@ public void streamOutputToBlobStorage(InputStream fromStream, String blobName) {
...
+        // the url we get from WSM already contains the token in it, so no need to specify sasToken separately
+        BlobServiceClient blobServiceClient = new BlobServiceClientBuilder().endpoint(blobstorageDetails).buildClient();
 
-        // if the backup container in storage doesnt already exists, it will need to be created
         try {
-            return blobServiceClient.getBlobContainerClient(containerName);
+            // the way storage containers are set up in a workspace are as follows:
+            // billing project gets a single azure storage account
+            // each workspace gets a container inside of that storage account to keep its data
+            return blobServiceClient.getBlobContainerClient("sc-"+ workspaceId);
```
davidangb, line 70: `AzureBlobStorage` relies on `WorkspaceManagerDao`, and `WorkspaceManagerDao` is tied to WDS's workspaceId; therefore, `AzureBlobStorage` is also tied to the current workspaceId. You could create the container name `"sc-"+ workspaceId` in the constructor and save it as a class variable, instead of coding it up here.

Either way works, this is me thinking out loud, no changes necessary unless you feel like doing it

service/src/main/resources/application.properties
```diff
@@ -52,7 +52,7 @@ twds.pg_dump.dbName=${env.wds.db.name}
 twds.pg_dump.password=${env.wds.db.password}
 twds.pg_dump.host=${env.wds.db.host}
 # When running on Azure in k8s with workload identity set PGDUMP_USE_AZURE_IDENTITY to true
-twds.pg_dump.useAzureIdentity=${PGDUMP_USE_AZURE_IDENTITY:false}
+twds.pg_dump.useAzureIdentity=${PGDUMP_USE_AZURE_IDENTITY:true}
```
davidangb, line 55: does this default need to change?

service/src/main/resources/application.properties
```diff
@@ -71,7 +71,7 @@ api.retry.backoff.multiplier=1.5
 
 datarepourl=${DATA_REPO_URL:}
 workspacemanagerurl=${WORKSPACE_MANAGER_URL:}
-
+leoUrl=${LEONARDO_URL:https://leonardo.yuliadub.bee.envs-terra.bio}
```
davidangb, line 74: oooh your bee!

service/src/main/java/org/databiosphere/workspacedataservice/dao/BackupDao.java
```diff
@@ -0,0 +1,23 @@
...
+public interface BackupDao {
+    BackupSchema getBackupStatus(UUID trackingId);
+
+    String getBackupRequestStatus(UUID sourceWorkspaceId, UUID destinationWorkspaceId);
+
+    boolean backupExists(UUID trackingId);
+
+    void createBackupEntry(UUID trackingId);
+
+    void updateBackupStatus(UUID trackingId, String status);
```
davidangb, line 16: have this accept a `BackupSchema.BackupState` instead of a String?

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
```diff
@@ -60,20 +91,28 @@ public BackupResponse backupAzureWDS(BackUpFileStorage storage, String version)
...
+            LOGGER.info("Starting streaming backup to storage.");
+            storage.streamOutputToBlobStorage(localProcessLauncher.getInputStream(), blobName, workspaceId);
             String error = localProcessLauncher.getOutputForProcess(LocalProcessLauncher.Output.ERROR);
             int exitCode = localProcessLauncher.waitForTerminate();
 
             if (exitCode != 0 && StringUtils.isNotBlank(error)) {
                 LOGGER.error("process error: {}", error);
-                return new BackupResponse(false, error);
+                backupDao.updateBackupStatus(trackingId, BackupSchema.BackupState.ERROR.toString());
+                backupDao.updateBackupRequestStatus(UUID.fromString(workspaceId), BackupSchema.BackupState.ERROR);
```
davidangb, line 103: this should save something to the `error` column in the db, to make it easy to report to users later. May need to increase the `error` column size above 255, too

service/src/main/resources/liquibase/changesets/20230612_backup_table.yaml
```diff
@@ -0,0 +1,29 @@
...
+                  name: id
+                  type: uuid
+                  constraints:
+                    primaryKey: true
+              - column:
+                  name: status
+                  type: varchar(10)
+              - column:
+                  name: error
+                  type: varchar(255)
```
davidangb, line 20: this might need to be longer, if we're saving a stderr stream

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -27,23 +27,49 @@ tags:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/BackupTrackingResponse'
+  /backup/{v}/status/{trackingId}:
+    get:
+      tags:
+        - Cloning
+      parameters:
+        - $ref: '#/components/parameters/trackingIdPathParam'
```
davidangb, line 59: add
```
- $ref: '#/components/parameters/versionPathParam'
```

service/src/test/java/org/databiosphere/workspacedataservice/InstanceInitializerBeanTest.java
```diff
@@ -17,16 +22,21 @@
...
-@SpringBootTest(classes = {InstanceInitializerConfig.class, MockInstanceDaoConfig.class})
+@SpringBootTest(classes = {InstanceInitializerConfig.class, MockInstanceDaoConfig.class, MockBackupDao.class, LeonardoConfig.class, WorkspaceDataServiceConfig.class})
 class InstanceInitializerBeanTest {
 
     @Autowired
     InstanceInitializerBean instanceInitializerBean;
     @SpyBean
     InstanceDao instanceDao;
+    @SpyBean
+    BackupDao backupDao;
```
davidangb, line 36: `backupDao` is unused

service/src/test/java/org/databiosphere/workspacedataservice/InstanceInitializerNoWorkspaceIdTest.java
```diff
@@ -15,17 +21,21 @@
...
 class InstanceInitializerNoWorkspaceIdTest {
 
     @Autowired
     InstanceInitializerBean instanceInitializerBean;
+
     @SpyBean
     InstanceDao instanceDao;
 
+    @SpyBean
+    BackupDao backupDao;
```
davidangb, line 37: backupDao is unused?

davidangb (commented):
phew! A buch more comments, none critical.

service/src/main/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDao.java
yuliadub: thank you!

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
```diff
@@ -49,9 +51,38 @@ public class BackupService {
...
+
+    public void backupAzureWDS(String version, UUID trackingId, UUID requestorWorkspaceId) {
         try {
             validateVersion(version);
+
+            // create an entry to track progress of this backup
+            backupDao.createBackupEntry(trackingId);
+
+            // create an entry to track who requested this backup
+            backupDao.createBackupRequestsEntry(requestorWorkspaceId, UUID.fromString(workspaceId));
```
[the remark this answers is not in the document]
yuliadub: my thought was that this way you can track different things: 
- on the source, it can easy to tell how many time it has been cloned
- on the destination, you can keep track of the original if you were the one that came from a clone

service/src/main/java/org/databiosphere/workspacedataservice/storage/AzureBlobStorage.java
yuliadub: the reason why I havent made that change was actually because I figured we may want to use this function to write from source wds to destination workspace blob storage vs passing the sas token via the api. I kept it as is but I also totally okay with changing as per you suggestion.

service/src/main/resources/application.properties
yuliadub: I only changed to get the deployment into my BEE - this should be changed whenever we enable flex postgres for WDS. if we enable flex postgres and dont change it, things will break.

service/src/main/resources/application.properties
yuliadub: yup, good catch! I need to go ahead and default this - and the extra piece of work that I need to do (and keep not getting to) is to pass the leonardo endpoint url to WDS in terra-helm.

service/src/main/java/org/databiosphere/workspacedataservice/controller/CloningController.java
```diff
@@ -2,28 +2,35 @@
...
-    @PostMapping("/backup/{version}")
-    public ResponseEntity<BackupResponse> createBackup(@PathVariable("version") String version) {
-        BackupResponse response = backupService.backupAzureWDS(storage, version);
-        if(response.backupStatus()) {
-            return new ResponseEntity<>(response, HttpStatus.OK);
-        }
-        
-        return new ResponseEntity<>(response, HttpStatus.INTERNAL_SERVER_ERROR);
+    @PostMapping("/backup/{version}/{requestorWorkspaceId}")
+    public ResponseEntity<BackupTrackingResponse> createBackup(@PathVariable("version") String version, @PathVariable("requestorWorkspaceId") UUID requestorWorkspaceId) {
```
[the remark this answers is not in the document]
yuliadub: I think we could make that change when we start supporting backup/restore? but I dont feel super strongly about it.

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
```diff
@@ -30,18 +30,48 @@ public void createDataRepoSnapshotReference(SnapshotModel snapshotModel) {
     try {
       String timeStamp = new SimpleDateFormat("yyyyMMddHHmmss").format(new java.util.Date());
       resourceApi.createDataRepoSnapshotReference(new CreateDataRepoSnapshotReferenceRequestBody()
-          .snapshot(
-              new DataRepoSnapshotAttributes()
-                  .instanceName(INSTANCE_NAME)
-                  .snapshot(snapshotModel.getId().toString()))
+              .snapshot(
```
yuliadub, line 33: I swear either github or intelliJ is playing game with me and adding more tabs without me knowing

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupService.java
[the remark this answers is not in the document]
yuliadub: after chatting through it, removing the concept of RequestEntry since what I was trying to track with this can be tracked in BackupEntry

Dockerfile
```diff
@@ -9,4 +9,4 @@ FROM us.gcr.io/broad-dsp-gcr-public/base/jre:17-debian
 RUN apt-get update
 
 # Add postgres client for pg_dump command
-RUN apt-get install postgresql-client-14 -y
+RUN apt-get install postgresql-client-14 -y
```
ashanhol, line 12: nit: is this change necessary?

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/BackupRestoreServiceFailureIntegrationTest.java
```diff
@@ -28,6 +28,9 @@ public class BackupRestoreServiceFailureIntegrationTest {
     @Autowired
     private BackupRestoreService backupRestoreService;
 
+    @Value("${twds.instance.workspace-id}")
```
ashanhol, line 31: What is this being used for? David's change seemed to indicate that just setting the host to empty was enough to make both backup and restore fail.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -40,34 +49,89 @@ public boolean isInCloneMode(String sourceWorkspaceId) {
...
                 UUID.fromString(sourceWorkspaceId);
             } catch (IllegalArgumentException e){
-                    LOGGER.warn("Source workspace id could not be parsed, unable to clone DB. Provided source workspace id: {}", sourceWorkspaceId);
+                    LOGGER.warn("Source workspace Id could not be parsed, unable to clone DB. Provided source workspace id: {}.", sourceWorkspaceId);
                     return false;
             }
+
+            if (sourceWorkspaceId.equals(workspaceId)) {
+                LOGGER.warn("Source workspace id and current workspace Id can't be the same.");
+                return false;
```
ashanhol, line 58: Thank you for checking for this. Not something to resolve now but do you think we might need to do more to bubble up this error to users? Can't think of a way this would trigger in the wild but right now it's failing silently.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -40,34 +49,89 @@ public boolean isInCloneMode(String sourceWorkspaceId) {
...
+            }
+
             try {
-                //TODO: this is a placeholder for checking that the db has already been cloned;
-                //In the future it could check the clone status,
-                //But for now assuming if we've created a workspace schema, work is done
+                // TODO at this stage of cloning work (where only backup is getting generated), just checking if an instance schema already exists is sufficient
+                // when the restore operation is added, it would be important to check if any record of restore state is present
+                // it is also possible to check if backup was initiated and completed (since if it did, we dont need to request it again)
```
ashanhol, line 64: Just getting the backup status = completed doesn't seem like the full story. Wouldn't we also want to trigger a backup if the last backup is out of date? (Do we have a way to know that?)

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -40,34 +49,89 @@ public boolean isInCloneMode(String sourceWorkspaceId) {
...
+
+            // check if our current workspace has already sent a request for backup
+            // if it did, no need to do it again
+            var backupFileName = "";
+            if (!backupDao.backupExistsForWorkspace(UUID.fromString(workspaceId))) {
+                // TODO since the backup api is not async, this will return once the backup finishes
+                var response = wdsDao.triggerBackup(startupToken, UUID.fromString(workspaceId));
+
+                // record the request this workspace made for backup
+                backupDao.createBackupEntry(UUID.fromString(response.getJobId()), new BackupRequest(UUID.fromString(workspaceId), "Track backup progress"));
```
ashanhol, line 105: nit: "Tracking backup progress triggered during startup" or something like that? distinguishes it in case we have a way to trigger it in other ways in the future

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresBackupDao.java
```diff
@@ -44,6 +44,7 @@ public Job<BackupResponse> getBackupStatus(UUID trackingId) {
             return null;
         } else {
             throw new RuntimeException("Unexpected error: %s rows found for backup status query".formatted(responses.size()));
+
```
ashanhol, line 47: nit: whitespace

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresInstanceDao.java
```diff
@@ -1,7 +1,6 @@
 package org.databiosphere.workspacedataservice.dao;
```
ashanhol, line 1: General comment- can we include descriptors on what classes do? It's not obvious to me what the difference between PostgresInstanceDao and PostgresBackupDao is and when they'd be used.

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupRestoreService.java
```diff
@@ -30,6 +30,7 @@
 public class BackupRestoreService {
     private final BackupDao backupDao;
     private final BackUpFileStorage storage;
+
```
ashanhol, line 33: nit: whitespace

Dockerfile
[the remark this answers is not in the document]
yuliadub: I am honestly not sure why github is marking this as a change

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/BackupRestoreServiceFailureIntegrationTest.java
yuliadub: this is a result of a merge fail - removing!

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
yuliadub: It's an interesting thought - if something like this happened, I am not sure what a user could do about it (and it could be confusing since user doesnt know how data tables work/etc in the background per say). It would likely be a more fundamental error somewhere in the terra infra, so I think at this point giving a user a blank WDS and saying cloning failed is likely an okay outcome. With that said, we can certainly talk this through with Beth as well to make sure the outcome makes sense from a product perspective. (i.e. would it make sense to surface this kind of error in the cloning status)

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
yuliadub: im my mind this is out of scope at this time (i.e. backup is up to date or not), if we took a backup and then WDS restarted so we are checking if one exists, we go with the one we already have. The access token we have only works for 1 hour anyways, so its unlikely that a new backup would be needed.

ashanhol (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresInstanceDao.java
yuliadub: added some comments, let me know if that helps

ashanhol: yeah I saw, I approved the PR :)

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -40,34 +49,89 @@ public boolean isInCloneMode(String sourceWorkspaceId) {
...
-        LOGGER.info("No source workspace id found, initializing default schema");
+        LOGGER.info("No source workspace id found, initializing default schema.");
         return false;
 
     }
 
+    /*
+    Cloning comes from the concept of copying an original (source) workspace data (from WDS data tables) into
+    a newly created (destination) workspace. WDS at start up will always have a current WorkspaceId, which in the
+    context of cloning will effectively be the destination. The SourceWorkspaceId will only be populated if the currently
```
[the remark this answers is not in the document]
yuliadub: not at this time - meant to bring this up in team topics. I will add this to the next one we have!

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: Note- I'm resolving this since I didn't understand we had moved the file from source to dest storage container.

service/src/test/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDaoTest.java
```diff
@@ -18,6 +18,7 @@
 import java.text.SimpleDateFormat;
 import java.util.*;
 
+
```
ashanhol, line 21: nit: unnecessary whitespace but don't bother if you're not making any other changes

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -12,62 +18,124 @@
...
+            // make the call to source wds to trigger back up, to do that set the source WDS dao endpoint
+            // url to the url retrieved from Leo
+            wdsDao.setWorkspaceDataServiceUrl(sourceWdsEndpoint);
+
+            // this fileName will be used for the restore
+            var backupFileName = "";
+            // check if the  current workspace has already sent a request for backup
+            // if it did, no need to do it again
+            if (!backupDao.backupExistsForWorkspace(UUID.fromString(workspaceId))) {
+                LOGGER.info("No backup exists, will iniated one.");
```
calypsomatic, line 102: nit: "iniated" -> "initiate"

davidangb (commented):
A few syntax-level comments inline, none of them big.

I'm struggling with reusing the `backup` database table here to track both the backup performed in the source and the request-for-backup in the destination. Looking forward to a future in which we support backup/restore outside of cloning, I think the row representing the request-to-source will be confusing here. I'd prefer to see a new `clone` table that can track the clone operation from request-for-backup all the way through restore.

That said, I'd be okay merging this PR as-is to make progress, then revamping it in a follow-on PR to move to a `clone` table. Happy to chat in person (and I know we've been back and forth on this issue a few times)

service/src/test/java/org/databiosphere/workspacedataservice/dao/MockBackupDao.java
```diff
@@ -0,0 +1,69 @@
...
+    private final Set<Job<BackupResponse> > backups = ConcurrentHashMap.newKeySet();
+
+    public MockBackupDao() {
+        super();
+    }
+
+    @Override
+    public boolean backupExistsForWorkspace(UUID workspaceId)  {
+        var data = backups.stream().filter(entry -> entry.getResult().requester().equals(workspaceId));
+        return data == null ? false : true ;
```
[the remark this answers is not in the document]
yuliadub: no thank you! Still learning the java gotchas compared to c# and this is so helpful!

yuliadub:
> A few syntax-level comments inline, none of them big.
> 
> I'm struggling with reusing the `backup` database table here to track both the backup performed in the source and the request-for-backup in the destination. Looking forward to a future in which we support backup/restore outside of cloning, I think the row representing the request-to-source will be confusing here. I'd prefer to see a new `clone` table that can track the clone operation from request-for-backup all the way through restore.
> 
> That said, I'd be okay merging this PR as-is to make progress, then revamping it in a follow-on PR to move to a `clone` table. Happy to chat in person (and I know we've been back and forth on this issue a few times)

No that makes total sense - I did not do a good job articulating why I wanted to have that second table. Having looked at the UI ticket more we will need a place where we can easily grab a status of all up backup/restore specifically for cloning and a separate table makes sense vs fishing that out of the corresponding backup/restore status ones.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresCloneDao.java
```diff
@@ -0,0 +1,52 @@
...
+        Timestamp now = Timestamp.from(Instant.now());
+        namedTemplate.getJdbcTemplate().update("insert into sys_wds.clone(id, status, createdtime, updatedtime, sourceworkspaceid, cloneStatus) " +
+                "values (?,?,?,?,?,?)", trackingId, JobStatus.QUEUED.name(), now, now, sourceWorkspaceId, CloneStatus.BACKUPQUEUED.name());
+    }
+
+    @Override
+    @WriteTransaction
+    public void updateCloneEntryStatus(UUID sourceWorkspaceId, CloneStatus status) {
+        namedTemplate.getJdbcTemplate().update("update sys_wds.clone SET cloneStatus = ?, status = ? where sourceWorkspaceId = ?",
+                status.name(), JobStatus.SUCCEEDED, sourceWorkspaceId);
```
ashanhol, line 49: Wait we're always succeeding if we're calling this function? Can we add successful into the name then?

yuliadub: still thinking this one through....the "job" in this flow is simply asking sql for things, so things could fail but it would be related to getting data/parsing the response. Maybe need to add a simple try/catch and mark as error if it catches an exception, otherwise it would always mark as success as long as it updated the status....will update as I decide

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresCloneDao.java
```diff
@@ -0,0 +1,67 @@
...
+    @WriteTransaction
+    public void createCloneEntry(UUID trackingId, UUID sourceWorkspaceId) {
+        Timestamp now = Timestamp.from(Instant.now());
+        namedTemplate.getJdbcTemplate().update("insert into sys_wds.clone(id, status, createdtime, updatedtime, sourceworkspaceid, clonestatus) " +
+                "values (?,?,?,?,?,?)", trackingId, JobStatus.QUEUED.name(), now, now, sourceWorkspaceId, CloneStatus.BACKUPQUEUED.name());
+    }
+
+    @Override
+    @WriteTransaction
+    public void updateCloneEntryStatus(UUID trackingId, CloneStatus status) {
```
yuliadub, line 48: I am not sure if it makes sense to have a separate status outside of Job status here - this operation is simple enough where Job status doesn't really indicate anything aside from that it was able to record the information, so we could overwrite it to indicated cloning status, but that would mean it having inconsistent values with other Jobs which I decided is likely not ideal and defeats the purpose of having that consistency. Hence why you see a "cloneStatus" alongside Job status here.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresCloneDao.java
```diff
@@ -0,0 +1,67 @@
...
+        try {
+            namedTemplate.getJdbcTemplate().update("update sys_wds.clone SET clonestatus = ?, status = ? where id = ?",
+                    status.name(), JobStatus.SUCCEEDED.name(), trackingId);
```
davidangb, line 51: this always sets `status` to `JobStatus.SUCCEEDED`, even in cases where `clonestatus` is being set to `BACKUPERROR`, is that intentional?

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresCloneDao.java
```diff
@@ -0,0 +1,67 @@
...
+            LOGGER.info("Clone status is now {}", status);
+        }
+        catch (Exception e){
+            namedTemplate.getJdbcTemplate().update("update sys_wds.clone SET status = ? where id = ?",
+                    JobStatus.ERROR.name(), trackingId);
```
davidangb, line 56: you could replace this `.update()` with:
```java
terminateBackupToError(trackingId, e.getMessage());
```

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresCloneDao.java
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+
+    @Override
+    @WriteTransaction
+    public void terminateBackupToError(UUID trackingId, String error) {
```
davidangb, line 62: nit: rename to `terminateCloneToError`?

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresCloneDao.java
davidangb: I like this - the main `status` gives a high-level view of where any long-running operation stands; the `cloneStatus` is a sub-status that gives a lower-level insight into the specifics of this particular operation.

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
```diff
@@ -52,30 +51,32 @@ public void createDataRepoSnapshotReference(SnapshotModel snapshotModel) {
...
+        // but currently there is no way to tell which one is the primary except for checking the actual container name
+        var storageUUID = extractResourceId(resourceList, storageWorkspaceId);
+        if (storageUUID != null) {
+          LOGGER.debug("Requesting SAS token-enabled storage url or workspace {} from Workspace Manager ...", workspaceUUID);
+          CreatedAzureStorageContainerSasToken sasBundle = azureResourceApi.createAzureStorageContainerSasToken(workspaceUUID, storageUUID, null, null, null, null);
+          return sasBundle.getUrl();
+        } else throw new ApiException("WorkspaceManagerDao: Can't locate a storage resource matching workspace Id. ");
+      } catch (ApiException e) {
+        if (++count == maxTries) throw new WorkspaceManagerException(e);
       }
```
davidangb, line 74: I think we can refactor this using Spring's `@Retryable` … but I suggest we do that in a follow-on PR and not block this one any more!

davidangb (commented):
so close!! One comment/question on `updateCloneEntryStatus` that needs an answer and possibly a change; otherwise looking good!

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresCloneDao.java
yuliadub: yeah I was thinking the job itself would be "successful" as in, nothing broke as it was retrieving the status - as in the overall job would only be marked as error if the operation it was trying to achieve didnt work. With that said, your earlier point on Job reflecting the overall state and the clone status reflecting a more specific state, maybe it does make sense to sync those two together. That will likely be less confusing!

davidangb (approved):
YAY!

PR #284, opened 2023-06-22
AJ-936: Implement Restore for Cloning API 

ashanhol:
This PR implements the API to restore a pg_dump to our database. It downloads a (hardcoded) file from Azure to local, and executes a psql -f command to restore.
I had to do a bit of finagling for the integration test. I had to call backup to mock the pg_dump file locally, then scrub the file and database of "workspace-id" to convert that guid to "source-workspace-id" since in lieu of spinning up a new version of wds.


service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
```diff
@@ -0,0 +1,85 @@
...
+    @Value("${twds.instance.workspace-id:}")
+    private String workspaceId;
+
+    @Value("${twds.instance.source-workspace-id:}")
+    private String sourceWorkspaceId;
+
+    private LocalFileStorage storage = new LocalFileStorage();
+    
+    @Test
+    void testRestoreAzureWDS() throws Exception {
```
ashanhol, line 48: So the direction I decided to take here was use the backup test to create a local pgdump file, which given that we're not spinning up a new instance of WDS, would make "source" and "destination" the same. So then I have to manually replace workspace-id in the file with source-workspace-id to give it a different value, and clean the database of those values. 
Alternates we could consider: 
- don't generate the pg_dump, hardcode the file to the test resources folder
- spin up a separate postgres db, which solves the db clean but would still need to find/replace values in pg_dump

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/BackupRestoreServiceFailureIntegrationTest.java
```diff
@@ -0,0 +1,31 @@
...
+public class BackupRestoreServiceFailureIntegrationTest {
+    @Autowired
+    private BackupRestoreService backupRestoreService;
+
+    private LocalFileStorage storage = new LocalFileStorage();
+
+    @Test
+    void testRestoreAzureWDSErrorHandling() {
+        BackupRestoreResponse response = backupRestoreService.restoreAzureWDS(storage, "v0.2");
+        assertFalse(response.backupRestoreStatus());
```
calypsomatic, line 23: These are failures because the host is empty?

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
calypsomatic: I'd be ok with a hardcoded file - that simulates how real life would have a file available by the time we get to restore, and there would be less code in the test that isn't actually testing the restore function.  But would we need to be concerned about updating the file if anything changed about how backup generated it?

service/src/integrationTest/java/org/databiosphere/workspacedataservice/storage/LocalFileStorage.java
```diff
@@ -21,4 +21,10 @@ public void streamOutputToBlobStorage(InputStream fromStream, String blobName) {
             throw new LaunchProcessException("Error streaming output during local test", ioEx);
         }
     }
+
+    public void downloadFromBlobStorage(String blobName) {
+        File backupFile = new File("backup.sql");
+        if(!backupFile.exists())
```
calypsomatic, line 27: Just trying to understand - is this basically just checking that the file exists?  Otherwise, once the `File` is created, it's not used for anything?

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupRestoreService.java
```diff
@@ -0,0 +1,173 @@
...
+            }
+
+            // rename workspace schema from source to dest
+            commandList.remove(commandList.size() - 1);
+            commandList.add(String.format("-c ALTER SCHEMA \"%s\" RENAME TO \"%s\"", sourceWorkspaceId, workspaceId));
+            localProcessLauncher.launchProcess(commandList, envVars);
+            error = checkForError(localProcessLauncher, LOGGER);
+            if (StringUtils.isNotBlank(error)) {
+                return new BackupRestoreResponse(false, error);
+            }
```
davidangb, line 112: This is executing one SQL statement. I think it's overkill to do that via `psql` - we should do this via a JDBC template instead. Search the codebase for `namedTemplate.getJdbcTemplate().update` for examples!

That said, we *also* need to update the `sys_wds.instance` table to track the postgres schemas. I think all the logic for renaming/tracking schemas should live inside `PostgresInstanceDao`, and `BackupRestoreService` can have an `instanceDao` and just call a method on that dao.

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupRestoreService.java
```diff
@@ -0,0 +1,173 @@
...
+            
+            // delete backup file
+            Files.deleteIfExists(Paths.get(BackupFileName));
```
davidangb, line 115: move the delete into a `finally` block, otherwise we won't delete it if we hit an error earlier in processing. An additional benefit of moving this to `finally` is that we can still return success to the end user even if we hit a problem during the deletion.

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupRestoreService.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    public String generateBackupFilename() {
+        LocalDateTime now = LocalDateTime.now();
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd_HH-mm-ss");
+        String timestamp = now.format(formatter);
+        return workspaceId + "-" + timestamp + ".sql";
+    }
+
+    private String checkForError(LocalProcessLauncher localProcessLauncher, Logger log) {
```
davidangb, line 164: the `LOGGER` variable is a class-level variable, no need to pass it in as an argument. You can remove `Logger log` from the argument list and use `LOGGER.error` instead of `log.error` in the body

service/src/main/java/org/databiosphere/workspacedataservice/storage/AzureBlobStorage.java
```diff
@@ -34,6 +36,19 @@ public void streamOutputToBlobStorage(InputStream fromStream, String blobName) {
...
+        // TODO: remove this once connection is switched to be done via SAS token
+        String storageConnectionString = System.getenv("STORAGE_CONNECTION_STRING");
+        BlobContainerClient blobContainerClient = constructBlockBlobClient(backUpContainerName, storageConnectionString);
+        
+        try{
+            blobContainerClient.getBlobClient(blobName).downloadToFile(blobName, true);
+        } catch(UncheckedIOException ioEx) {
+            throw new LaunchProcessException("Error downloading backup file from Azure Storage", ioEx);
+        }
+    }
```
davidangb, line 51: I know we're streaming the output of `pg_dump` to storage. In your opinion, is it worth streaming the .sql file from storage into `psql`, or are we better off downloading it and pointing `psql` at the file?

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -41,13 +41,34 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/BackupRestoreResponse'
+        400:
+          description: Something failed during restore
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/BackupRestoreResponse'
```
davidangb, line 71: should we even expose a `/restore` API? I know we'll need it eventually, but for now we are only using restore as part of cloning - maybe we should keep the functionality internal until we're ready to support it for end users.

service/src/main/java/org/databiosphere/workspacedataservice/controller/CloningController.java
```diff
@@ -10,17 +10,27 @@
...
+    public ResponseEntity<BackupRestoreResponse> createBackup(@PathVariable("version") String version) {
+        BackupRestoreResponse response = backupRestoreService.backupAzureWDS(storage, version);
+        if(response.backupRestoreStatus()) {
+            return new ResponseEntity<>(response, HttpStatus.OK);
+        }
+        
+        return new ResponseEntity<>(response, HttpStatus.INTERNAL_SERVER_ERROR);
+    }
+
+    @PostMapping("/restore/{version}")
```
davidangb, line 30: see my comment on openapi-docs.yaml - I wonder if we want this API at all.

service/src/test/java/org/databiosphere/workspacedataservice/service/RestoreServiceTest.java
```diff
@@ -0,0 +1,24 @@
...
+class RestoreServiceTest {
+    @Autowired
+    private BackupRestoreService backupRestoreService;
+
+    @Test
+    void CheckCommandLine() {
+        List<String> commandList = backupRestoreService.generateCommandList(false);
+        String command = String.join(" ", commandList);
+        assertThat(command).isEqualTo("/usr/bin/psql -h localhost -p 5432 -U wds -d wds -v -w");
+    }
```
davidangb, line 23: I don't think this needs a dedicated file for this one unit test; this can move into `BackupServiceTest` (feel free to rename to something like `BackupRestoreServiceTest` too)

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
davidangb: +1 for a hardcoded file, checked in somewhere under the test resources folder. This would also make it very easy to perform additional assertions in the test - you could easily test that the restore process renamed the source's instance UUID to the destination's instance UUID, for instance.

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
```diff
@@ -0,0 +1,85 @@
...
+        List<String> commandList = backupRestoreService.generateCommandList(false);
+        commandList.add(String.format("-c DROP SCHEMA IF EXISTS \"%s\" CASCADE; DROP SCHEMA IF EXISTS \"%s\" CASCADE;", workspaceId, "sys_wds"));
+        Map<String, String> envVars = Map.of("PGPASSWORD", "wds");
+        LocalProcessLauncher localProcessLauncher = new LocalProcessLauncher();
+        localProcessLauncher.launchProcess(commandList, envVars);
+        String error = localProcessLauncher.getOutputForProcess(LocalProcessLauncher.Output.ERROR);
+        int exitCode = localProcessLauncher.waitForTerminate();
+        if (exitCode != 0 && StringUtils.isNotBlank(error)) {
+            System.err.println("Error clearing database: " + error);
+        }
```
davidangb, line 82: I suggest executing SQL statements via `namedTemplate` instead of `psql` - no need for the extra complexity `psql` brings

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
davidangb, line 1: can this be a unit test instead of an integration test? Does that cause problems because we drop/restore sys_wds?

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
```diff
@@ -0,0 +1,85 @@
...
+            Files.write(path, content.getBytes(StandardCharsets.UTF_8));
+        }
+        catch(IOException ex) {
+            System.err.format("IOException in RestoreServiceIntegrationTest.testRestoreAzureWDS: %s%n", ex);
+        }
+    }
+
+    private void cleanDatabase() {
+        List<String> commandList = backupRestoreService.generateCommandList(false);
+        commandList.add(String.format("-c DROP SCHEMA IF EXISTS \"%s\" CASCADE; DROP SCHEMA IF EXISTS \"%s\" CASCADE;", workspaceId, "sys_wds"));
```
davidangb, line 74: what about the `public` schema? Is that also included in pg_dump files?

davidangb (commented):
first-pass comments inline!

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
```diff
@@ -0,0 +1,85 @@
...
+        Map<String, String> envVars = Map.of("PGPASSWORD", "wds");
```
yuliadub, line 75: since you are loading the config values, might as well use it here!

service/src/main/java/org/databiosphere/workspacedataservice/controller/CloningController.java
yuliadub: good point! I think originally I had it in the ticket since it made it easier to test things :)

service/src/main/resources/application.properties
```diff
@@ -46,6 +46,7 @@ twds.instance.source-workspace-id=${SOURCE_WORKSPACE_ID:}
 twds.startup-token=${STARTUP_TOKEN:}
 
 twds.pg_dump.path=${PGDUMP_PATH:/usr/bin/pg_dump}
+twds.pg_dump.psqlPath=${PSQL_PATH:/usr/bin/psql}
```
yuliadub, line 49: do we need two variables for this? or would it be easier to just have the path and hard code the commands? I keep going back and forth on this :D

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/BackupRestoreServiceFailureIntegrationTest.java
ashanhol: Yeah. That was the only way I could figure out a for sure way to trigger the failure state for backup. (It'd probably also fail if username or db was empty too)

service/src/integrationTest/java/org/databiosphere/workspacedataservice/storage/LocalFileStorage.java
ashanhol: This class is used for testing to replace azure functionality. My actual azure functionality downloads a specific file from a container. So I figured checking for an existing local file would be kind of the equivalent in this case. My test currently expects test backup to run before test restore but once I switch to a hardcoded file and get rid of that assumption I'll keep the file exists check to make sure we're grabbing the local file.

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupRestoreService.java
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        return new BackupRestoreResponse(true, "Backup successfully completed.");
+    }
+
+    public BackupRestoreResponse restoreAzureWDS(BackUpFileStorage storage, String version) {
+        validateVersion(version);
+        try {
+            // grab blob from storage
+            storage.downloadFromBlobStorage(BackupFileName);
```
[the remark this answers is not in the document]
ashanhol: From my experience we get failures if we try to restore a schema that's already in the database so I don't think there's any issues with overwriting. That said, I've seen some weird states where if your dump has duplicate _and_ new info for your database it'll throw a failure but then I'll look and pgAdmin shows new schemas created 🤷‍♀️I'll create a ticket to explore failure and edge cases, I think it needs some investigation outside the scope of this PR. 

It seems relative file path works as long as you're consistent. It downloads to same folder your code is calling from so in this case workspacedataservice/service/backup.sql, but I can modify it so it's `Path` if that'll be safer.

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
ashanhol: public is not a schema in pg_dump

service/src/main/resources/application.properties
ashanhol: I messaged you about that a while ago, we determined there could be a chance there could be a different path for install for the two commands so better to be safe than sorry

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
ashanhol: switched to using a hardcoded file. However, it had to be in main/resources not test/resources since LocalFileStorage.java (which we're using for mocking azure) is not in test :/

service/src/integrationTest/java/org/databiosphere/workspacedataservice/storage/LocalFileStorage.java
ashanhol: Edit: this is moot now, switched to using streams instead of files.

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupRestoreService.java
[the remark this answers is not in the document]
ashanhol: edit edit: file this is moot, switched to using streams

service/src/main/java/org/databiosphere/workspacedataservice/storage/AzureBlobStorage.java
ashanhol: I'm glad you asked, when I started out I thought file was the right way to go. And it definitely was for testing and making sure it all worked, but now thinking, it's probably better to switch to stream. Made that change.

service/src/test/java/org/databiosphere/workspacedataservice/service/RestoreServiceTest.java
ashanhol: switched it 👍

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
ashanhol: Made changes so it matches yulia's now (with streaming). @yuliadub is there a reason we can't run the functionality as a unit test?

service/src/main/resources/application.properties
yuliadub: aha fair enough. I do briefly remember us chatting about this.

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
ashanhol: done

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupRestoreService.java
ashanhol: Added a function to alter schema to PostgresInstanceDao (please confirm I did that right/modified the right tests?) 
I'm keeping the command line psql for executing the stream though if that's alright. I'm getting good detailed errors and I think it's easy to understand since it mirrors the backup function.

service/src/main/resources/static/swagger/openapi-docs.yaml
ashanhol: Removed. Made the service just return true or false for now and have it throw real exceptions once we hook it up.

client/build.gradle
```diff
@@ -3,7 +3,7 @@ plugins {
 	id 'java-library'
 	id 'maven-publish'
 
-	id 'com.jfrog.artifactory' version '4.18.2'
+	id 'com.jfrog.artifactory' version '4.32.0'
```
ashanhol, line 6: gdi looks like github is showing merge changes again (guess it wasn't unique to rebasing.) please ignore

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupRestoreService.java
```diff
@@ -76,6 +87,36 @@ public BackupResponse backupAzureWDS(BackUpFileStorage storage, String version)
...
+            LocalProcessLauncher localProcessLauncher = new LocalProcessLauncher();
+            localProcessLauncher.launchProcess(commandList, envVars);
+
+            // grab blob from storage
+            storage.streamInputFromBlobStorage(localProcessLauncher.getOutputStream(), BackupFileName);
+
+            String error = checkForError(localProcessLauncher);
+            if (StringUtils.isNotBlank(error)) {
+                return false;
+                //throw new LaunchProcessException(error);
```
calypsomatic, line 106: nit: remove commented out code

ashanhol: I left it in on purpose to illustrate how I believe we should be handling these errors once we hook it up if that's ok? I can add a todo comment if that'd be clearer?

service/src/main/resources/backup-test.sql
```diff
@@ -0,0 +1,236 @@
+--
+-- PostgreSQL database dump
```
ashanhol, line 2: gdi looks like github is showing merge changes again (guess it wasn't unique to rebasing.) please ignore

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupRestoreService.java
calypsomatic: Ah ok.  Yeah maybe TODO would be clearer, but if you and Yulia are aware of this are about to clean it up in the near future then that should be fine

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
yuliadub: the main reason I set this up to be a separate integration test really is the fact that we need pg_dump installed on the machine that is running the test, and this will effectively run a separate process on the machine that is executing the test (which didnt feel right to call unit test in my mind due to the dependecy:D) and I also at that moment thought that none of our unit tests interacted with a db directly and used a mock. However, as I have dug into my current PR work I know that is not necessarily true and some of our unit tests do indeed leverage the postgres db. 

I think in my mind anything that needs a special set up (like in this case installing something on the ubuntu's machine prior to running a test) and then verifying the outcome of that feels right to be an integration test. But I am open to merging this into unit tests if it feels like less things to manage :)

yuliadub (approved):

yuliadub:
looks good! thank you! There are a few things to adjust as my changes merge that may be a little annoying but we will get them figured out as we plug this into the e2e cloning flow! excited to get that working!

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+    private LocalFileStorage storage = new LocalFileStorage();
+    
+    @Test
+    void testRestoreAzureWDS() throws Exception {
+        // Clean database of workspace-id value
+        instanceDao.dropSchema(UUID.fromString(workspaceId));
+        var response = backupRestoreService.restoreAzureWDS(storage, "v0.2");
+        assertTrue(response);
+    }
```
davidangb, line 45: Two comments on this test … 

First, could you add a few more assertions to verify that it's actually doing what we expect it to, beyond just returning true? Something like:
```java
 @Test
    void testRestoreAzureWDS() throws Exception {
        UUID sourceInstance = UUID.fromString(sourceWorkspaceId);
        UUID destInstance = UUID.fromString(workspaceId);

        // neither source nor destination instance should exist in our list of schemas to start
        List<UUID> instancesBefore = instanceDao.listInstanceSchemas();
        assertThat(instancesBefore).doesNotContain(destInstance);
        assertThat(instancesBefore).doesNotContain(sourceInstance);

        // perform the restore
        var response = backupRestoreService.restoreAzureWDS(storage, "v0.2");
        assertTrue(response);

        // after restore, dest instance SHOULD exist in our list of schemas but source should not
        List<UUID> instancesAfter = instanceDao.listInstanceSchemas();
        assertThat(instancesAfter).contains(destInstance);
        assertThat(instancesAfter).doesNotContain(sourceInstance);

        // after restore, target instance should have one table named "test"
        List<RecordType> tables = recordDao.getAllRecordTypes(destInstance);
        assertThat(tables.size()).isEqualTo(1);
        assertThat(tables).contains(RecordType.valueOf("test"));
    }
```

Second, this unit test passes when run as part of all unit tests, but fails when run on its own. Is it unintentionally depending on some database state created in another test? Could you refactor it so it can be run independently?

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresInstanceDao.java
```diff
@@ -55,4 +55,10 @@ public void dropSchema(UUID instanceId) {
         namedTemplate.getJdbcTemplate().update("delete from sys_wds.instance where id = ?", instanceId);
     }
 
+    @Override
+    @WriteTransaction
+    @SuppressWarnings("squid:S2077") // since instanceId must be a UUID, it is safe to use inline
+    public void alterSchema(UUID sourceWorkspaceId, UUID workspaceId) {
+        namedTemplate.getJdbcTemplate().update("alter schema " + quote(sourceWorkspaceId.toString()) + " rename to " + quote(workspaceId.toString()));
+    }
```
davidangb, line 63: this will need to also update the corresponding row in the `sys_wds.instance` table. See the `createSchema()` and `dropSchema()` methods which insert those rows.

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupRestoreService.java
```diff
@@ -21,13 +23,18 @@
 import com.azure.identity.extensions.jdbc.postgresql.AzurePostgresqlAuthenticationPlugin;
 
 @Service
-public class BackupService {
-    private static final Logger LOGGER = LoggerFactory.getLogger(BackupService.class);
+public class BackupRestoreService {
+    private final InstanceDao instanceDao;
+    private static final Logger LOGGER = LoggerFactory.getLogger(BackupRestoreService.class);
+
+    private static final String BackupFileName = "backup.sql";
```
davidangb, line 30: this hardcoded "backup.sql" will get replaced by the actual filename from the source workspace once everything is hooked up, is that correct?

davidangb (changes requested):
couple changes needed, around renaming instances and allowing unit tests to pass independently

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/BackupRestoreServiceFailureIntegrationTest.java
```diff
@@ -0,0 +1,31 @@
...
+import org.databiosphere.workspacedataservice.storage.LocalFileStorage;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.TestPropertySource;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
+@SpringBootTest(properties = "spring.cache.type=NONE")
+@TestPropertySource(properties = {"twds.instance.workspace-id=123e4567-e89b-12d3-a456-426614174000", "twds.instance.source-workspace-id=123e4567-e89b-12d3-a456-426614174000", "twds.pg_dump.host="})
```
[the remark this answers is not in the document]
ashanhol: Yeah, it's a guaranteed sql error on the restore side. @yuliadub do you think this is something we'd want to check early on before we even back up?

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupRestoreService.java
ashanhol: Yeah, this will be 100% modified on the next bit of work that'll actually hook up restore (once yulia's PR hooking up backup goes in)

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/BackupRestoreServiceFailureIntegrationTest.java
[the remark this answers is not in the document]
yuliadub: yeah. thats a good idea! for some reason I thought we already did this but looks like we dont, would be an easy check to add in the initializer bean as you suggested.

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/BackupRestoreServiceFailureIntegrationTest.java
```diff
@@ -0,0 +1,31 @@
...
+import org.databiosphere.workspacedataservice.shared.model.BackupResponse;
+import org.databiosphere.workspacedataservice.storage.LocalFileStorage;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.TestPropertySource;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
+@SpringBootTest(properties = "spring.cache.type=NONE")
```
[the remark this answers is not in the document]
ashanhol: confirmed both were not needed. Maybe for one of the office hours I'll suggest going over some of the springboot stuff since a lot of it is opaque to me. I've been copying stuff but I'm probably at the point where I should understand it better 😁

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/BackupRestoreServiceFailureIntegrationTest.java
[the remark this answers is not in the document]
ashanhol: Added it as a note on the jira task that hooks up restore 👍

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/BackupServiceIntegrationTest.java
```diff
@@ -1,22 +1,25 @@
 package org.databiosphere.workspacedataservice.service;
 
+import org.databiosphere.workspacedataservice.shared.model.BackupResponse;
 import org.databiosphere.workspacedataservice.storage.LocalFileStorage;
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
 
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertFalse;
```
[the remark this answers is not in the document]
ashanhol: thank you

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
```diff
@@ -0,0 +1,47 @@
...
+@SpringBootTest(properties = "spring.cache.type=NONE")
+@TestPropertySource(
+    properties = {
+        "twds.instance.workspace-id=123e4567-e89b-12d3-a456-426614174000", 
+        "twds.instance.source-workspace-id=123e4567-e89b-12d3-a456-426614174001", 
+        "twds.pg_dump.port=5432",
+        "twds.pg_dump.user=wds",
+        "twds.pg_dump.dbName=wds",
+        "twds.pg_dump.password=wds",
+        "twds.pg_dump.host=localhost"
```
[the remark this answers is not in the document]
ashanhol: I think for some reason the default values weren't getting set right originally. They seem to be working now so I removed them.

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
ashanhol: So Autowire creates a schema of the workspace-id before any of our code in the test runs. My original test had a "clean" that removed that before running restore. I added that back in and it works fine. 
```  
// clean database of workspace-id value created from autowire
instanceDao.dropSchema(destInstance);
```

Thank you so much for the additional assertion suggestions!

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresInstanceDao.java
ashanhol: Added, thank you

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
davidangb: This is still failing for me when I run it on its own. I'm using `./gradlew integrationTest --tests '*RestoreServiceIntegrationTest'`.

> So Autowire creates a schema of the workspace-id before any of our code in the test runs

which autowire are you thinking of here? I don't see anything in the test that would cause it. I'm wondering if some *other* test creates that schema, which is why this test passes only if it's run in concert with other tests.

> My understanding is that schema creation is in the initializer bean so I'm not sure how we could bypass that

I'm guessing it's the TODO here … possibly just by addressing that TODO, it would help?

davidangb: digging even further, this doesn't pass for me when I run all the integration tests: `./gradlew integrationTest`

ashanhol:
Note: Github went down for a bit so one of the checks failed but all checks that could be rerun passed so merging.

PR #291, opened 2023-06-30
AJ-1091: Disallow any whitespace in TSV headers

ashanhol:
Problem: multiple trailing tabs in the header row was running into a "duplicate header" error. 
Resolution: after discussion we determined any whitespace in the headers is now considered malformed. Added an error and some appropriate tests. 

------------
Reminder:


service/src/main/java/org/databiosphere/workspacedataservice/service/TsvStreamWriteHandler.java
```diff
@@ -57,9 +57,13 @@ public TsvStreamWriteHandler(InputStream inputStream, ObjectReader tsvReader, Re
 			colNames = StreamSupport.stream(actualSchema.spliterator(), false)
 					.map(CsvSchema.Column::getName)
 					.toList();
-
-			if (colNames.stream().anyMatch(col -> Collections.frequency(colNames, col) > 1)) {
-				throw new InvalidTsvException("TSV contains duplicate column names."
+			// any blank headers or trailing tabs is considered malformed 
+			if(colNames.stream().anyMatch(col -> col.isBlank())) {
```
davidangb, line 61: nit: to avoid warnings in IntelliJ about "Lambda can be replaced with method reference":
```suggestion
			if(colNames.stream().anyMatch(String::isBlank)) {
```

davidangb (approved):
lgtm, thank you! One nit inline to avoid yellow squigglies in IntelliJ. Proactively giving this a thumb assuming that's all you change - I will re-review if you change other things.

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/service/TsvStreamWriteHandler.java
ashanhol: Oh cool I didn't realize you could do that.

PR #298, opened 2023-07-11
AJ-1146: activity logging for backup/restore

davidangb:
Activity logging (aka audit logging) for backup and restore.

If either the backup or restore code paths change, we may need to update the activity logging.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/activitylog/ActivityModels.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/activitylog/ActivityModels.java
@@ -6,7 +6,7 @@ public class ActivityModels {
         CREATE("created"), DELETE("deleted"),
         UPDATE("updated"), UPSERT("upserted"),
         MODIFY("modified"), LINK("linked"),
-        RESTORE("restore");
+        RESTORE("restored");
 
         private final String name;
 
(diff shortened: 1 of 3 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/activitylog/ActivityModels.java
```diff
@@ -5,7 +5,8 @@ public class ActivityModels {
     public enum Action {
         CREATE("created"), DELETE("deleted"),
         UPDATE("updated"), UPSERT("upserted"),
-        MODIFY("modified"), LINK("linked");
+        MODIFY("modified"), LINK("linked"),
+        RESTORE("restored");
```
yuliadub, line 9: do we want an Action for backup as well?

yuliadub: ah wait, I see, backup is a Thing vs an Action

yuliadub (approved):

service/src/main/java/org/databiosphere/workspacedataservice/activitylog/ActivityModels.java
davidangb: I'm reusing the "created" action here, which results in a log entry like:
```
2023-07-11 16:40:47.472  INFO [7Lm2bz2y] [http-nio-8080-exec-1] o.d.w.activitylog.ActivityLogger         : user 111137720113699110757 created 1 backup(s) with id(s) [wdsservice/cloning/backup/e137c203-c73d-4d81-a611-0e2e659a1c56-2023-07-11_16-40-41.sql]
```
do you think we need a separate action?

yuliadub: no I think that makes sense to me!

calypsomatic (approved):

PR #302, opened 2023-07-14
AJ-886- Trigger restore when backing up

ashanhol:
This PR: 
- Triggers restore in initializer bean if we've successfully backed up. Uses default initialization if we detect errors at any part.
- Expanded restore to return a job. Right now we're calling restore directly, not from an API call, but this will allow us to easily create one if needed and make the call async.
- Change pg_dump to only back up instances, not sys_wds.
- Expand functions to manually pass through an auth token to WorkspaceManagerClient since we call restore without an API call. 
- Restore saves state on cloning by updating the CloneDao in initialization.
- Created a restore table in sys_wds. 
- Expanded BackupDao interface to be more generic (now named BackupRestoreDao).
  - Made an abstract class that implements the interface, and two dao classes for backup (PostgresBackupDao) and restore (PostgresRestoreDao) that extend that abstract class in order to share logic.
  - I modified BackupRestoreService to take those two daos.
 
TODO: 
- How do I update the MockBackupDao class? Notes left on PR
-----------------------
Reminder:


service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -101,41 +103,56 @@ public void initCloneMode(){
...
+            if(cloneStatus.getResult().status() == CloneStatus.BACKUPSUCCEEDED) {
+                LOGGER.info("Restore from the following path on the source workspace storage container: {}", backupFileName);
+                cloneDao.updateCloneEntryStatus(cloneStatus.getJobId(), CloneStatus.RESTOREQUEUED);
+                var restoreResponse = restoreService.restoreAzureWDS("v0.2", backupFileName, cloneStatus.getJobId(), startupToken);
+                if(restoreResponse.getStatus() != JobStatus.SUCCEEDED) {
+                    LOGGER.error("Something went wrong with restore: {}. Starting with empty default instance schema.", restoreResponse.getErrorMessage());
+                    cloneDao.terminateCloneToError(cloneStatus.getJobId(), "Something went wrong with restore: " + restoreResponse.getErrorMessage(), false);
+                    initializeDefaultInstance();
+                }
+                LOGGER.info("Restore Successful");
```
calypsomatic, line 135: Does this need to be in an `else` (or otherwise rearranged) so that it doesn't log "Restore Successful" after the error and default initialization?

local-dev/run_postgres_destination.sh
```diff
@@ -0,0 +1,50 @@
...
+    # start up postgres
+    echo "starting up postgres container..."
+    BASEDIR=$(dirname "$0")
+    docker create --name $CONTAINER --rm -e POSTGRES_PASSWORD=password -e PGHOST=5434  -p "$POSTGRES_PORT:5432" postgres:$POSTGRES_VERSION -c listen_addresses='*'
+    docker cp $BASEDIR/local-postgres-init.sql $CONTAINER:/docker-entrypoint-initdb.d/docker_postgres_init.sql
+    docker start $CONTAINER
+
+    # validate postgres
+    echo "running postgres validation..."
+    docker exec $CONTAINER sh -c "$(cat $BASEDIR/sql_validate.sh)"
```
yuliadub, line 18: today this sql will fail with the new destination container. we either need to add a param to run command on a different port in this case or add a different validate script

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -101,41 +103,56 @@ public void initCloneMode(){
...
-            LOGGER.info("Restore from the following path on the source workspace storage container: {}", backupFileName);
+            
+            // TODO: re-evaluate running restore this way once backup becomes async. Will need to wait on it. 
+            // continue on to restore if backup has succeded. otherwise start with default instance schema.
+            var cloneStatus = cloneDao.getCloneStatus();
+            if(cloneStatus.getResult().status() == CloneStatus.BACKUPSUCCEEDED) {
+                LOGGER.info("Restore from the following path on the source workspace storage container: {}", backupFileName);
+                cloneDao.updateCloneEntryStatus(cloneStatus.getJobId(), CloneStatus.RESTOREQUEUED);
+                var restoreResponse = restoreService.restoreAzureWDS("v0.2", backupFileName, cloneStatus.getJobId(), startupToken);
+                if(restoreResponse.getStatus() != JobStatus.SUCCEEDED) {
```
yuliadub, line 130: code smells will enforce the use of `if(!restoreResponse.getStatus().equals(JobStatus.SUCCEEDED)`

local-dev/run_postgres_destination.sh
ashanhol: I actually didn't want to commit this, and my local git says I haven't, so I truly have no idea how this file made its way into the PR

yuliadub: feel free to delete for now (if you cant locally you do this in the github UI)

service/src/main/java/org/databiosphere/workspacedataservice/dao/CloneDao.java
```diff
@@ -11,5 +11,5 @@ public interface CloneDao {
     void updateCloneEntryStatus(UUID trackingId, CloneStatus status);
     boolean cloneExistsForWorkspace(UUID sourceWorkspaceId);
     Job<CloneResponse> getCloneStatus();
-    void terminateCloneToError(UUID trackingId, String error);
+    void terminateCloneToError(UUID trackingId, String error, Boolean isBackup);
```
yuliadub, line 14: I dont know if I love this value as a Boolean, I think I'd rather pass around an enum that would have backup and restore as values which could then also be used to set the tables names in each Dao (so each Dao own that value vs this being an arbitrary bool we have to remember what is true vs false)

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresBackupDao.java
```diff
@@ -22,22 +19,18 @@
...
-    private final NamedParameterJdbcTemplate namedTemplate;
+public class PostgresBackupDao extends AbstractBackupRestoreDao<BackupResponse> {
 
     /*
     PostgresBackupDao is used to interact with sys_wds backup table in postgres that tracks status of backups.
     This class will help add entries to the table, check if entries already exist and update them as necessary.
      */
     public PostgresBackupDao(NamedParameterJdbcTemplate namedTemplate) {
-        this.namedTemplate = namedTemplate;
+        super(namedTemplate, "backup");
```
yuliadub, line 29: lets make an enum that has backup and restore as values that we would use here instead of literal string

service/src/test/java/org/databiosphere/workspacedataservice/dao/MockBackupDao.java
```diff
@@ -12,53 +15,63 @@
...
-    public void createBackupEntry(UUID trackingId, BackupRequest request) {
+    public void createEntry(UUID trackingId, BackupRestoreRequest request) {
         Timestamp now = Timestamp.from(Instant.now());
-        var metadata = new BackupResponse("", request.requestingWorkspaceId(), request.description());
-        Job<BackupResponse> backup = new Job<>(trackingId, JobStatus.QUEUED, "", now.toLocalDateTime(), now.toLocalDateTime(), metadata);
-        backups.add(backup);
+        if(table.equals(CloneTable.BACKUP)) {
+            var metadata = new BackupResponse("", request.requestingWorkspaceId(), request.description());
+            Job<BackupResponse> backup = new Job<>(trackingId, JobStatus.QUEUED, "", now.toLocalDateTime(), now.toLocalDateTime(), metadata);
+            entries.add(backup);
```
ashanhol, line 43: So this doesn't work and I'm not entirely sure why. BackupResponse (and restore) extends JobResult and a Set<Parent> should allow set.add(Child).

service/src/test/java/org/databiosphere/workspacedataservice/dao/MockBackupDao.java
```diff
@@ -12,53 +15,63 @@
 import java.util.concurrent.ConcurrentHashMap;
 
 /**
- * Mock implementation of BackupDao that is in-memory instead of requiring Postgres
+ * Mock implementation of BackupRestoreDao that is in-memory instead of requiring Postgres
  */
-public class MockBackupDao implements BackupDao {
+public class MockBackupDao<T extends JobResult> implements BackupRestoreDao<T> {
```
ashanhol, line 20: Note: need to genericize the name here so it's not specific to backup

service/src/test/java/org/databiosphere/workspacedataservice/dao/MockBackupDaoConfig.java
```diff
@@ -11,7 +11,7 @@ public class MockBackupDaoConfig {
     @Bean
     @Profile("mock-backup-dao")
     @Primary
-    BackupDao mockBackupDao() {
+    BackupRestoreDao mockBackupDao() {
```
ashanhol, line 14: This also fails because we're not committing to a generic type here. This leads me to believe there isn't a way to make one mock class for BackupResponse and RestoreResponse 🤔

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -32,12 +37,12 @@ public class InstanceInitializerBean {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(InstanceInitializerBean.class);
 
-    public InstanceInitializerBean(InstanceDao instanceDao, BackupDao backupDao, LeonardoDao leoDao, WorkspaceDataServiceDao wdsDao, CloneDao cloneDao){
```
ashanhol, line 35: No need to write to this table here, BackupRestoreService takes care of that. state for this is tracked in the Clone table

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -114,28 +117,45 @@ public void initCloneMode(){
...
+                    cloneDao.terminateCloneToError(cloneStatus.getJobId(), restoreResponse.getErrorMessage(), CloneTable.RESTORE);
+                    initializeDefaultInstance();
+                } else {
+                    LOGGER.info("Restore Successful");
+                    cloneDao.updateCloneEntryStatus(cloneStatus.getJobId(), CloneStatus.RESTORESUCCEEDED);
+                }
+            } else {
+                LOGGER.error("Backup not successful, cannot restore. Will start with empty default instance schema.");
+                cloneDao.terminateCloneToError(cloneStatus.getJobId(), "Backup not successful, cannot restore.", CloneTable.RESTORE);
+                initializeDefaultInstance();
```
ashanhol, line 142: Added logic when we fail in backup and restore to initialize defaults.

service/src/main/java/org/databiosphere/workspacedataservice/dao/AbstractBackupRestoreDao.java
```diff
@@ -0,0 +1,66 @@
...
+import org.databiosphere.workspacedataservice.shared.model.job.JobStatus;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+
+import java.sql.Timestamp;
+import java.time.Instant;
+import java.util.UUID;
+
+public class AbstractBackupRestoreDao <T extends JobResult> implements BackupRestoreDao<T> {
```
ashanhol, line 18: This was my effort to reduce code dupe. Abstract the BackupDao to be BackupRestoreDao and allow us to have either a BackupResult or RestoreResult.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresBackupDao.java
```diff
@@ -53,38 +47,13 @@ public Job<BackupResponse> getBackupStatus(UUID trackingId) {
...
         Timestamp now = Timestamp.from(Instant.now());
         namedTemplate.getJdbcTemplate().update("insert into sys_wds.backup(id, status, createdtime, updatedtime, requester, description) " +
-                "values (?,?,?,?,?,?)", trackingId, JobStatus.QUEUED.name(), now, now, backupRequest.requestingWorkspaceId(), backupRequest.description());
+                "values (?,?,?,?,?,?)", trackingId, JobStatus.QUEUED.name(), now, now, BackupRestoreRequest.requestingWorkspaceId(), BackupRestoreRequest.description());
         LOGGER.info("Backup job {} is now {}", trackingId, JobStatus.QUEUED);
     }
 
-    @Override
-    @WriteTransaction
-    public void updateBackupStatus(UUID trackingId, JobStatus status) {
```
ashanhol, line 65: functions moved to the abstract class

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresRestoreDao.java
```diff
@@ -0,0 +1,80 @@
...
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Timestamp;
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.util.List;
+import java.util.UUID;
+
+@Repository
+public class PostgresRestoreDao extends AbstractBackupRestoreDao<RestoreResponse> {
```
ashanhol, line 23: Should basically be the same as PostgresBackupDao modified for one less column

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupRestoreService.java
```diff
@@ -113,62 +116,70 @@ public Job<BackupResponse> backupAzureWDS(String version, UUID trackingId, Backu
...
-            backupDao.terminateBackupToError(trackingId, ex.getMessage());
+            backupDao.terminateToError(trackingId, ex.getMessage());
         }
 
-        return backupDao.getBackupStatus(trackingId);
+        return backupDao.getStatus(trackingId);
     }
 
-    public boolean restoreAzureWDS(String version) {
+    public Job<RestoreResponse> restoreAzureWDS(String version, String backupFileName, UUID trackingId, String startupToken) {
```
ashanhol, line 144: I know it's not super necessary to have this be the return type at the moment but wanted to set us up for making the function async and possibly its own api call if we want to do so in the future

service/src/main/java/org/databiosphere/workspacedataservice/service/BackupRestoreService.java
```diff
@@ -183,7 +194,16 @@ private String determinePassword() throws PSQLException {
     // Same args, different command depending on whether we're doing backup or restore.
     public List<String> generateCommandList(boolean isBackup) {
         Map<String, String> command = new LinkedHashMap<>();
-        command.put(isBackup ? pgDumpPath : psqlPath, null);
+        if (isBackup) {
```
ashanhol, line 197: Logic to only backup instances and not sys_wds. Included extra flag for blobs (which is on by default but -n turns off), not sure if we actually store blobs but I figured more info retaining is better than less

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/CloneTable.java
```diff
@@ -0,0 +1,11 @@
+package org.databiosphere.workspacedataservice.shared.model;
+
+/**
+ * The various states encountered during cloning.
+ */
+
+public enum CloneTable {
```
ashanhol, line 7: Yulia intelligently suggested moving away from bool you have to remember to enum that's explicit  👍

service/src/main/java/org/databiosphere/workspacedataservice/storage/AzureBlobStorage.java
```diff
@@ -59,7 +55,8 @@ public void streamOutputToBlobStorage(InputStream fromStream, String blobName, S
     }
 
     @Override
-    public void streamInputFromBlobStorage(OutputStream toStream, String blobName, String workspaceId) {
+    public void streamInputFromBlobStorage(OutputStream toStream, String blobName, String workspaceId, String authToken) {
+        workspaceManagerDao.setAuthToken(authToken);
```
ashanhol, line 59: Since we don't call restore via an api call, we have to manually pass an auth token around. Made modifications to everything that bubbles up to workspacemanager to accept that.

service/src/main/resources/liquibase/changesets/20230714_restore_table.yml
```diff
@@ -0,0 +1,32 @@
...
+              - column:
+                  name: createdtime
+                  type: datetime
+              - column:
+                  name: updatedtime
+                  type: datetime
+              - column:
+                  name: requester
+                  type: uuid
+              - column:
```
ashanhol, line 30: Figured I didn't need to be saving the name of the file as a column like we do for backup. The other columns seemed valuable

service/src/test/java/org/databiosphere/workspacedataservice/InstanceInitializerBeanTest.java
```diff
@@ -20,12 +26,13 @@
...
 import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.*;
 
-@ActiveProfiles({"mock-instance-dao", "mock-backup-dao", "mock-clone-dao", "local"})
+@ActiveProfiles({"mock-instance-dao", "mock-backup-dao", "mock-clone-dao", "local,", "mock-sam"})
 @TestPropertySource(properties = {"twds.instance.workspace-id=90e1b179-9f83-4a6f-a8c2-db083df4cd03"})
 @DirtiesContext
-@SpringBootTest(classes = {InstanceInitializerConfig.class, MockInstanceDaoConfig.class, MockBackupDao.class, LeonardoConfig.class, WorkspaceDataServiceConfig.class, MockCloneDaoConfig.class})
+@SpringBootTest(classes = {InstanceInitializerConfig.class, MockInstanceDaoConfig.class, MockBackupDao.class, LeonardoConfig.class, WorkspaceDataServiceConfig.class, MockCloneDaoConfig.class, BackupRestoreService.class, AzureBlobStorage.class, WorkspaceManagerConfig.class, ActivityLoggerConfig.class, SamConfig.class, MockSamClientFactoryConfig.class})
```
ashanhol, line 35: since we added BackupRestoreService to the initializerbean this test was failing without all these extra classes

service/src/test/java/org/databiosphere/workspacedataservice/dao/MockBackupDao.java
yuliadub: it cant convert a specific type to T, but I *think* doing a cast here should handle it - since we will only add a specific type based on the CloneTable enum

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -78,68 +127,102 @@ a newly created (destination) workspace. WDS at start up will always have a curr
...
+
         }
-        catch(Exception e) {
-            LOGGER.error("An error occurred during clone mode. Will start with empty default instance schema. Error: {}", e.toString());
+        catch (Exception e) {
+            LOGGER.error("An error occurred during clone mode. Error: {}", e.toString());
+            try {
+                if (trackingId != null) {
+                    cloneDao.terminateCloneToError(trackingId, "Backup not successful, cannot restore.", CloneTable.RESTORE);
+                }
```
yuliadub, line 173: unlikely but i wonder if it makes sense to log a warning message if trackingId somehow ends up being null. The way it currently is, that condition could sort of get lost since it wont actually cause a failure

davidangb: I just pushed a commit that eliminates the possibility of `trackingId` being null - by adding some more validation of the status.

service/src/main/java/org/databiosphere/workspacedataservice/dao/AbstractBackupRestoreDao.java
```diff
@@ -0,0 +1,67 @@
...
+    /*
+    AbstractBackupRestoreDao abstracts functionality used to interact with sys_wds backup and restore tables to track progress of each.
+    Functionality for backup and restore specific content (such as overwriting dbName) will be implemented in PostgresBackupDao and PostgresRestoreDao.
+    */
+    public AbstractBackupRestoreDao(NamedParameterJdbcTemplate namedTemplate, CloneTable tableName) {
+        this.namedTemplate = namedTemplate;
+        this.tableName = tableName;
+    }
+
+    protected static final Logger LOGGER = LoggerFactory.getLogger(PostgresBackupDao.class);
```
calypsomatic, line 33: Is this intentionally `PostgresBackupDao` instead of `AbstractBackupRestoreDao`?

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresBackupDao.java
```diff
@@ -53,38 +47,13 @@ public Job<BackupResponse> getBackupStatus(UUID trackingId) {
 
     @Override
     @WriteTransaction
-    public void createBackupEntry(UUID trackingId, BackupRequest backupRequest) {
+    public void createEntry(UUID trackingId, BackupRestoreRequest BackupRestoreRequest) {
```
calypsomatic, line 50: nit: should the variable name start with a lowercase?

service/src/main/java/org/databiosphere/workspacedataservice/dao/AbstractBackupRestoreDao.java
davidangb: good catch, fixed

service/src/main/java/org/databiosphere/workspacedataservice/dao/AbstractBackupRestoreDao.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    protected static final Logger LOGGER = LoggerFactory.getLogger(AbstractBackupRestoreDao.class);
+
+
+    public Job<T> getStatus(UUID trackingId) {
+        return null;
+    }
+
+    public void createEntry(UUID trackingId, BackupRestoreRequest BackupRestoreRequest) {}
```
calypsomatic, line 40: another uppercased variable!

calypsomatic (approved):
hard to test so I only poked around locally and saw a few of the (graceful) failure modes.  otherwise looking at the code it seems good!

PR #310, opened 2023-07-18
AJ-961 - Insert instance record into sys_wds vs updating

yuliadub:
Slight change from ealier behavior where we were still copying sys-wds over - now that we are not, when we alter schema name, we need to insert schema record  (that UI uses to display tables) into sys_wds.instances. 

Tested in a BEE and cloning works!!! 

Reminder:


davidangb (approved):

davidangb (changes requested):
I think this test has issues because the `WDS-integrationTest-LocalFileStorage-input.sql` file we use for restores no longer represents what's in our backups. Specifically, it inserts a row into `sys_wds.instance`. This was on my list of things to fix … I'm gonna see if I can get that done sooner rather than later. Once that's in, I'm curious to see if this test is still a problem.

davidangb:
I've updated to a representative SQL dump in #311, and pulled in the fixes from here to my PR - I believe #311 supersedes this PR.

yuliadub:
thank you! I'll just close this one out!

PR #311, opened 2023-07-18
AJ-961 more test coverage for clones

davidangb:
Changes in this PR:
* update the exemplar dump file used by integration tests. The old file still contained `sys_wds` tables, which we no longer include in the dump.
* add an integration test for a successful clone
* ensure that after cloning, the `sys_wds.instance` table contains a row for the default instance. This supersedes #310.
* some refactoring of `RestoreServiceIntegrationTest` to be more precise in its assertions
* set the `CloneStatus.BACKUPQUEUED` status for clones when the clone is starting; previously, this status was never used


service/src/integrationTest/resources/WDS-integrationTest-LocalFileStorage-input.sql
```diff
@@ -19,218 +27,56 @@ SET client_min_messages = warning;
...
+-- Dependencies: 217
+-- Data for Name: thing; Type: TABLE DATA; Schema: 10000000-0000-0000-0000-000000000111; Owner: wds
 --
 
-CREATE TABLE sys_wds.instance (
-    id uuid NOT NULL
-);
-
-
-ALTER TABLE sys_wds.instance OWNER TO wds;
```
yuliadub, line 157: ah good call! didnt even think to look at the file we were restoring when I was fixing what I thought was a broken test :D

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -195,6 +195,7 @@ private Job<CloneResponse> currentCloneStatus(UUID trackingId) {
     private String requestRemoteBackup(UUID trackingId) {
         try {
             LOGGER.info("Requesting a backup file from the remote source WDS in workspace {}", sourceWorkspaceId);
+            cloneDao.updateCloneEntryStatus(trackingId, CloneStatus.BACKUPQUEUED);
```
yuliadub, line 198: good catch!

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresInstanceDao.java
```diff
@@ -59,8 +59,13 @@ public void dropSchema(UUID instanceId) {
...
     @WriteTransaction
     @SuppressWarnings("squid:S2077") // since instanceId must be a UUID, it is safe to use inline
-    public void alterSchema(UUID sourceWorkspaceId, UUID workspaceId) {
-        namedTemplate.getJdbcTemplate().update("alter schema " + quote(sourceWorkspaceId.toString()) + " rename to " + quote(workspaceId.toString()));
-        namedTemplate.getJdbcTemplate().update("update sys_wds.instance set id = ? where id = ?", workspaceId, sourceWorkspaceId);
+    public void alterSchema(UUID oldSchemaId, UUID newSchemaId) {
+        // rename the pg schema from old to new
+        namedTemplate.getJdbcTemplate().update("alter schema " + quote(oldSchemaId.toString()) + " rename to " + quote(newSchemaId.toString()));
+        // rename any rows in sys_wds.instance from old to new
+        namedTemplate.getJdbcTemplate().update("update sys_wds.instance set id = ? where id = ?", newSchemaId, oldSchemaId);
```
yuliadub, line 66: I am not sure I understand why we would want to try to update vs just create the record that the schema was just added - are you thinking if we ever do copy over sys_wds.instance we may want to preserve the instance history after restore?

yuliadub (approved):
thanks for adding the tests - definitely learning more about mockito/how to set tests up with dependencies using it, pretty cool!

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresInstanceDao.java
davidangb: Good question, and I admit this code is less than ideal. I am trying to be super defensive here because I think some of this will change (see AJ-1171), and I'm hoping to be resilient to any of that change.

PR #314, opened 2023-07-21
AJ-1069: JsonStreamWriteHandler features-enabled unit test

davidangb:
part of AJ-1069 but does not complete that ticket.

While evaluating AJ-1069 as a good ramp-up ticket for new developers, I found an inconsistency in our JSON batch-write API: the `JsonParser` used for batch-write does not fully inherit its configuration from the `ObjectMapper` definition in `JsonConfig`. This PR fixes that, so it inherits features correctly.

I am pushing this PR through to make AJ-1069 easier and more focused on ramp-up instead of digging through ugly Jackson internals.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/JsonStreamWriteHandler.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/JsonStreamWriteHandler.java
@@ -27,8 +27,7 @@ public class JsonStreamWriteHandler implements StreamingWriteHandler {
 
 	public JsonStreamWriteHandler(InputStream inputStream, ObjectMapper objectMapper) throws IOException {
 		this.inputStream = inputStream;
-		JsonFactory factory = new JsonFactory(objectMapper.copy());
-		parser = factory.createParser(inputStream);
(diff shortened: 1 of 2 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/service/JsonStreamWriteHandler.java
```diff
@@ -27,8 +27,7 @@ public class JsonStreamWriteHandler implements StreamingWriteHandler {
 
 	public JsonStreamWriteHandler(InputStream inputStream, ObjectMapper objectMapper) throws IOException {
 		this.inputStream = inputStream;
-		JsonFactory factory = new JsonFactory(objectMapper.copy());
-		parser = factory.createParser(inputStream);
+		parser = objectMapper.copy().tokenStreamFactory().createParser(inputStream);
```
davidangb, line 30: creating a `new JsonFactory` loses some of the `objectMapper`'s features; instead, we should get a `JsonFactory` from the `objectMapper` directly.

calypsomatic (approved):
LGTM

yuliadub (approved):

PR #315, opened 2023-07-24
AJ-1095: Add Java-PFB import and test

snf2ye:
As a part of the work to publish the java-pfb library, this PR adds a test to prove that the library is correctly published and that we can use the methods from java-pfb in WDS!

Review Request:
-> Does the location of the test make sense?


calypsomatic (approved):

yuliadub (approved):

davidangb (approved):
thank you!

PR #316, opened 2023-08-02
AJ-1078 Enable Liveness and Readiness

ashanhol:
This PR enables Liveness and Readiness built into SpringBoot to enable our app to communicate with K8. 
Listener and tests inspired by: https://github.com/DataBiosphere/jade-data-repo/pull/670
Confirmed this works in a BEE
Sister PR helm file change: https://github.com/broadinstitute/terra-helmfile/pull/4413

-----------------------
Reminder:

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/logging/AvailabilityListener.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/logging/AvailabilityListener.java
@@ -1,0 +1,42 @@
+package org.databiosphere.workspacedataservice.logging;
+
+import org.springframework.boot.availability.LivenessState;
+import org.springframework.boot.availability.ReadinessState;
+import org.springframework.boot.availability.AvailabilityChangeEvent;
+import org.springframework.context.event.EventListener;
+import org.springframework.stereotype.Component;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Component
+public class AvailabilityListener {
+    private static Logger logger = LoggerFactory.getLogger(AvailabilityListener.class);
+
+    @EventListener
+    public void onLivenessEvent(AvailabilityChangeEvent<LivenessState> event) {
+        switch (event.getState()) {
+            case BROKEN:
+                logger.error("LivenessState: {}", event.getState());
+                break;
+            case CORRECT:
+                logger.debug("LivenessState: {}", event.getState());
+                break;
+            default:
+                logger.error("LivenessState {} is not a known state", event.getState());
+        }
+    }
+
+    @EventListener
+    public void onReadinessEvent(AvailabilityChangeEvent<ReadinessState> event) {
+        switch (event.getState()) {
+            case REFUSING_TRAFFIC:
+                logger.error("ReadinessState: {}", event.getState());
+                break;
+            case ACCEPTING_TRAFFIC:
+                logger.debug("ReadinessState: {}", event.getState());
+                break;
+            default:
+                logger.error("ReadinessState {} is not a known state", event.getState());
+        }
+    }
+}
--- a/service/src/main/java/org/databiosphere/workspacedataservice/logging/AvailabilityListener.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/logging/AvailabilityListener.java
@@ -19,7 +19,7 @@ public class AvailabilityListener {
                 logger.error("LivenessState: {}", event.getState());
                 break;
             case CORRECT:
-                logger.debug("LivenessState: {}", event.getState());
+                logger.info("LivenessState: {}", event.getState());
                 break;
             default:
                 logger.error("LivenessState {} is not a known state", event.getState());
@@ -33,7 +33,7 @@ public class AvailabilityListener {
                 logger.error("ReadinessState: {}", event.getState());
                 break;
             case ACCEPTING_TRAFFIC:
-                logger.debug("ReadinessState: {}", event.getState());
+                logger.info("ReadinessState: {}", event.getState());
                 break;
             default:
                 logger.error("ReadinessState {} is not a known state", event.getState());
--- a/service/src/main/resources/application.properties
+++ b/service/src/main/resources/application.properties
@@ -17,8 +17,7 @@ management.endpoints.web.base-path=/
 management.endpoints.web.path-mapping.info=version
 management.endpoints.web.path-mapping.health=status
 management.endpoint.health.show-details=ALWAYS
-management.health.livenessstate.enabled=true
-management.health.readinessstate.enabled=true
+management.endpoint.health.group.liveness.include=livenessstate,db
 
 # additional keys to expose in the actuator info endpoint:
 management.info.env.enabled=true
--- a/service/src/main/resources/application.properties
+++ b/service/src/main/resources/application.properties
@@ -17,7 +17,7 @@ management.endpoints.web.base-path=/
 management.endpoints.web.path-mapping.info=version
 management.endpoints.web.path-mapping.health=status
 management.endpoint.health.show-details=ALWAYS
-management.endpoint.health.group.liveness.include=livenessstate,db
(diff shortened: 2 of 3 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/activitylog/AvailabilityTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.springframework.test.web.servlet.ResultActions;
+import org.springframework.context.ApplicationContext;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.*;
+
+@AutoConfigureMockMvc
+//@SpringBootTest(classes = {SpringRunner.class})
```
calypsomatic, line 21: nit: remove commented out code

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/logging/AvailabilityListener.java
```diff
@@ -0,0 +1,42 @@
...
+    private static Logger logger = LoggerFactory.getLogger(AvailabilityListener.class);
+
+    @EventListener
+    public void onLivenessEvent(AvailabilityChangeEvent<LivenessState> event) {
+        switch (event.getState()) {
+            case BROKEN:
+                logger.error("LivenessState: {}", event.getState());
+                break;
+            case CORRECT:
+                logger.debug("LivenessState: {}", event.getState());
```
rtitle, line 22: Should we do INFO level? I don't think we enable debug logging by default, and this is probably not too noisy.

rtitle (approved):
One nit about logging, otherwise LGTM.

I think this will also require a terra-helmfile PR to add `livenessProbe`, `readinessProbe`, `startupProbe` configs in the [deployment yaml](https://github.com/broadinstitute/terra-helmfile/blob/master/charts/wds/templates/deployment.yaml)?

service/src/main/java/org/databiosphere/workspacedataservice/logging/AvailabilityListener.java
```diff
@@ -0,0 +1,42 @@
...
+                logger.error("ReadinessState: {}", event.getState());
+                break;
+            case ACCEPTING_TRAFFIC:
+                logger.debug("ReadinessState: {}", event.getState());
+                break;
+            default:
+                logger.error("ReadinessState {} is not a known state", event.getState());
+        }
+    }
+}
```
davidangb, line 42: I see `LivenessState: CORRECT` and `ReadinessState: ACCEPTING_TRAFFIC` logged when WDS starts up. But, even after fussing with it I don't see `BROKEN` or `REFUSING_TRAFFIC` logged at any point. Is that working as expected?

service/src/test/java/org/databiosphere/workspacedataservice/activitylog/AvailabilityTest.java
davidangb, line 21: nit: remove commented-out line

service/src/main/resources/application.properties
```diff
@@ -11,11 +11,15 @@ env.wds.db.additionalUrlParams=${ADDITIONAL_JDBC_URL_PARAMS:}
...
 management.endpoint.info.enabled=true
 management.endpoint.health.enabled=true
+management.endpoint.health.probes.enabled=true
 management.endpoints.web.exposure.include=info,health
 management.endpoints.web.base-path=/
 management.endpoints.web.path-mapping.info=version
 management.endpoints.web.path-mapping.health=status
 management.endpoint.health.show-details=ALWAYS
+management.health.livenessstate.enabled=true
+management.health.readinessstate.enabled=true
```
davidangb, line 21: do these two lines do anything? When I set them both to `false`, it doesn't seem to change anything.

service/src/main/resources/application.properties
```diff
@@ -11,11 +11,15 @@ env.wds.db.additionalUrlParams=${ADDITIONAL_JDBC_URL_PARAMS:}
 management.endpoints.enabled-by-default=false
 management.endpoint.info.enabled=true
 management.endpoint.health.enabled=true
+management.endpoint.health.probes.enabled=true
```
davidangb, line 14: I think we really want to include db connectivity in the liveness probe. One of our main use cases is restarting the pod if the pod loses contact with the database. Hopefully this is easy enough to do via a boot/actuator property, something like:
```
management.endpoint.health.group.liveness.include=db
```
I tried starting postgres and WDS locally, saw the liveness probe return 200/UP. Then I shut down Postgres and tried the liveness probe again - it still returned UP. If you include the db check in the liveness probe it will instead return DOWN.

davidangb (commented):
Some comments inline. Also, this does change the response payload from the `/status` API; does this affect the Java client in a bad way? I think the changes are solely additive, but I haven't tried testing the Java client to see about any impact.

service/src/main/resources/application.properties
ashanhol: the baeldung tutorial had them included but yeah they seem extra, will remove.

service/src/test/java/org/databiosphere/workspacedataservice/activitylog/AvailabilityTest.java
ashanhol: computer exploded before i could, removing now

service/src/main/java/org/databiosphere/workspacedataservice/logging/AvailabilityListener.java
ashanhol: computer exploded before i could, removing now

service/src/main/java/org/databiosphere/workspacedataservice/logging/AvailabilityListener.java
ashanhol: I'm not entirely sure how to mess with it enough to break it without explicitly pushing an event. I think tying downtime to the DB is a good idea if only for that

service/src/main/resources/application.properties
ashanhol: great idea, i didn't want to tie any other parts to it until I got feedback from the team. are there any other pieces we want to tie to readiness or liveness? (i was only going to tie the db to liveness)

service/src/main/java/org/databiosphere/workspacedataservice/logging/AvailabilityListener.java
ashanhol: good point, fixed.

service/src/main/java/org/databiosphere/workspacedataservice/logging/AvailabilityListener.java
ashanhol: I'm going to try deploying in a bee tomorrow and maybe forcing a pod down will give us those logs

davidangb: > I'm not entirely sure how to mess with it enough to break it without explicitly pushing an event. I think tying downtime to the DB is a good idea if only for that

That's exactly what I did when testing - I included the db in the liveness check. Then, I started local Postgres and local WDS, saw everything come up correctly. Then I _stopped_ the local Postgres. When I hit http://localhost:8080/status/liveness, I could see that it reported DOWN … but I never saw any log messages about LivenessState.

service/src/main/resources/application.properties
davidangb: The db is the only dependency I can think of - without the db, almost everything is hosed. All other dependencies like WSM, TDR, Sam, etc. only affect _some_ of the APIs; plus … restarting WDS probably won't have any effect on WDS' connectivity to Sam et al.

PR #317, opened 2023-08-02
AJ-1187: WDS only needs one connection pool

davidangb:
Combine WDS's two connection pools into one, thereby halving the number of db connections WDS requires. Preserve the capability to stream TSV downloads from Postgres to the end user.

I verified that WDS still uses cursors/streams to query Postgres by setting `-c log_statement=all` on my local Postgres docker image and watching what queries were actually executed. I also verified that the autocommit value was set back to true by cranking up logging levels for `com.zaxxer.hikari.HikariConfig` and `com.zaxxer.hikari`.

I cannot figure out how to unit test the low-level cursor behavior of the JDBC driver. This PR turns on features that are implemented in libraries. I would love to hear ideas on how to test!

High-level changes in this PR:
* add Spring Batch for its `JdbcCursorItemReader`, which manages the Postgres cursor query
* remove the "streamingDs" datasource and connection pool
* reimplement `RecordDao.streamAllRecordsForType()` using Spring Batch

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -47,6 +47,7 @@ dependencies {
 	implementation 'org.springframework.boot:spring-boot-starter-jdbc'
 	implementation 'org.springframework.boot:spring-boot-starter-web'
 	implementation 'org.springframework.boot:spring-boot-starter-cache'
+	implementation 'org.springframework.boot:spring-boot-starter-batch'
 	implementation 'org.springframework.retry:spring-retry:1.3.4'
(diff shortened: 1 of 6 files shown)


ashanhol (approved):
Not a huge fan of overriding a function in a return statement but streamAllRecordsForType is very well documented

yuliadub (approved):
LGTM

calypsomatic (approved):

PR #318, opened 2023-08-03
AJ-1186 DB Connection Pool Size

calypsomatic:
See https://broadworkbench.atlassian.net/browse/AJ-1186
The default connection pool size for Spring/Hikari is 10.  It's not clear that WDS needs that many connections.  
Some informal testing with different pool sizes under load (greater load than we generally expect to have in real life, ~20 users all uploading files at once), I generally saw no more than 8 connections being used at a time.  (We seemed to hit limits in the relay listener before we hit limits in the DB).  Using only 6 connections seemed to result in only 1-2 pending at a time, leading me to select 7 as a rough estimate of a reasonable max pool size.

Reminder:

### DIFF
--- a/service/src/main/resources/application.properties
+++ b/service/src/main/resources/application.properties
@@ -24,6 +24,7 @@ info.app.image=${WDS_IMAGE:unknown}
 spring.datasource.jdbc-url=jdbc:postgresql://${env.wds.db.host}:${env.wds.db.port}/${env.wds.db.name}?reWriteBatchedInserts=true&${env.wds.db.additionalUrlParams}
 spring.datasource.username=${env.wds.db.user}
 spring.datasource.password=${env.wds.db.password}
+spring.datasource.maximumPoolSize=7
 
 streaming.query.jdbc-url=${spring.datasource.jdbc-url}
 streaming.query.username=${spring.datasource.username}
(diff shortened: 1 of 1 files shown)


service/src/main/resources/application.properties
```diff
@@ -24,6 +24,7 @@ info.app.image=${WDS_IMAGE:unknown}
 spring.datasource.jdbc-url=jdbc:postgresql://${env.wds.db.host}:${env.wds.db.port}/${env.wds.db.name}?reWriteBatchedInserts=true&${env.wds.db.additionalUrlParams}
 spring.datasource.username=${env.wds.db.user}
 spring.datasource.password=${env.wds.db.password}
+spring.datasource.maximumPoolSize=7
```
davidangb, line 27: do you also need to set `minimumIdle`? I don't know what happens if `minimumIdle` (default: 10) is greater than `maximumPoolSize`.

calypsomatic: Hmm, good point.  It didn't blow up when I had the max set less than 10 so presumably it'll still function, we'll just get some kind of performance loss.

calypsomatic: I tried giving it a minimum of 3 and it started off with 4.  Not sure what that's about but how about having a min of 2 then?

davidangb: I'd suggest setting min and max the same: both at 7. That matches the default behavior where both are 10; we're just reducing the number a bit

davidangb (approved):

nawatts (approved):

PR #313, opened 2023-07-21
AJ-1181: buffer writes to blob storage

davidangb:
When uploading to Azure blob storage, use a buffer for writes.

This is the recommended approach, per Azure documentation: https://learn.microsoft.com/en-us/java/api/com.azure.storage.blob.specialized.blockblobclient?view=azure-java-stable#com-azure-storage-blob-specialized-blockblobclient-getbloboutputstream():

> Note: We recommend you call write with reasonably sized buffers, you can do so by wrapping the BlobOutputStream obtained below with a [BufferedOutputStream](https://docs.oracle.com/javase/8/docs/api/java/io/BufferedOutputStream.html).

I added some custom logging to my locally-running WDS and called the backup API.

After this PR:
```
2023-07-21 08:23:21.091 DEBUG [3GqOoze1] [http-nio-8080-exec-3] o.d.w.w.WorkspaceManagerDao              : Finding storage resource for workspace 4237ed5b-b3d7-4fe7-9751-64b4ca33e479 from Workspace Manager ...
2023-07-21 08:23:21.417  INFO [] [scheduling-1] o.d.w.MemoryOutputter                    : ------ MEMORY: used, Mb: 54, max: 8589934592, total: 197132288, free: 140416376
2023-07-21 08:23:21.919  INFO [] [scheduling-1] o.d.w.MemoryOutputter                    : ------ MEMORY: used, Mb: 67, max: 8589934592, total: 197132288, free: 126312632
2023-07-21 08:23:21.956 DEBUG [3GqOoze1] [http-nio-8080-exec-3] o.d.w.w.WorkspaceManagerDao              : Requesting SAS token-enabled storage url or workspace 4237ed5b-b3d7-4fe7-9751-64b4ca33e479 from Workspace Manager ...
2023-07-21 08:23:22.424  INFO [] [scheduling-1] o.d.w.MemoryOutputter                    : ------ MEMORY: used, Mb: 69, max: 8589934592, total: 197132288, free: 124179224
2023-07-21 08:23:22.926  INFO [] [scheduling-1] o.d.w.MemoryOutputter                    : ------ MEMORY: used, Mb: 69, max: 8589934592, total: 197132288, free: 123990664
2023-07-21 08:23:23.430  INFO [] [scheduling-1] o.d.w.MemoryOutputter                    : ------ MEMORY: used, Mb: 69, max: 8589934592, total: 197132288, free: 123990664
[...]

PR #320, opened 2023-08-04
AJ-1187: use Hikari properties; bean initialization

davidangb:
A bit of cleanup after #317 and #313 ...

* Use `spring.datasource.hikari.*` properties instead of `spring.datasource.*` properties. This should have no impact to WDS runtime, but it makes inspection via IntelliJ easier since IntelliJ recognizes `spring.datasource.hikari.*`.
* Change initialization of the NamedParameterJdbcTemplate bean to use Spring-managed injection for its datasource, instead of bypassing injection and calling `mainDb()` directly.
* add logging properties for the Hikari connection pool, but set the properties to defaults; this is a developer convenience for anyone who needs to change them later

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/DataSourceConfig.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/DataSourceConfig.java
@@ -17,15 +17,15 @@ public class DataSourceConfig {
 
     @Bean
     @Primary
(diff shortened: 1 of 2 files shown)


ashanhol (approved):
If whatever calls namedParameterJdbcTemplate knows to pass it a DataSource now, lgtm 👍

calypsomatic (approved):
Thank you!

PR #321, opened 2023-08-04
AJ-1173: okhttp-based WDS client

davidangb:
Changes in this PR:
* switch from `swagger-codegen` to `openapi-generator` for creating the WDS client
* switch from `jersey2` to `okhttp-gson` for the client's [`library`](https://openapi-generator.tech/docs/generators/java/#config-options) value
* simultaneously publish two separate versions of the library:
    * `workspacedataservice-client-okhttp-javax`, which uses `javax.annotation.*` and `javax.ws.rs.core.GenericType`
    * `workspacedataservice-client-okhttp-jakarta`, which uses `jakarta.annotation.*` and `jakarta.ws.rs.core.GenericType`
* work around a [bug](https://github.com/OpenAPITools/openapi-generator/issues/10848) in `openapi-generator` related to our `RecordAttributes` model class. The workaround involves:
    * back up the generated .java model files to a separate directory
    * rewrite the .java model files to address the compilation part of the bug
    * specify a hand-written `RecordAttributes` .java file to use instead of what openapi-generator creates

See also https://github.com/DataBiosphere/cbas/pull/195, which is a POC of adopting this new client.

Note that this new client is almost a drop-in replacement for the old client. This PR optimizes for that ease-of-adoption. Are the steps we take to work around the openapi-generator bug worth it? The alternative is to remove our workarounds but cause more disruption for users of the client (known users are: ourselves, CBAS).

The minor changes to openapi-docs.yaml will also affect the Python client. Those changes look ok.

See also discussion in https://docs.google.com/document/d/1ESVrdwZ86bPtsDrdwL1ze-yqCOLvO_FzTDB5waG7deM/edit


client/ApiClient.mustache
davidangb, line 1: these changes were only needed for the Jersey-based client.

client/artifactory.gradle
```diff
@@ -19,7 +19,7 @@ java {
 publishing {
 	publications {
 		wdsClient(MavenPublication) {
-			artifactId = 'workspacedataservice-client'
+			artifactId = 'workspacedataservice-client-okhttp'
```
davidangb, line 22: the artifact name change here signifies the breaking change in functionality.

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -757,14 +757,6 @@ components:
...
           "arrayFile": ["drs://drs.example.org/file_id_1", "https://account_name.blob.core.windows.net/container-2/blob2"]
         }
-    RecordAttributeDefinition:
-      required:
-        - name
-      type: object
-      properties:
-        name:
-          type: string
-          description: Attribute name
```
davidangb, line 767: `RecordAttributeDefinition` was unused.

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -862,8 +854,8 @@ components:
       description: Search behavior for multiple filter terms
     SearchSortDirection:
       type: string
-      enum: [asc, desc]
-      default: asc
+      enum: [ASC, DESC]
+      default: ASC
```
davidangb, line 858: This capitalization change matches the actual values in the `SearchSortDirection` Java class. I guess the old client was forgiving about the difference; the new client is not. This change seems more correct.

service/src/test/java/org/databiosphere/workspacedataservice/GeneratedClientTests.java
```diff
@@ -77,7 +78,7 @@ void putRecordWithSpecifiedPk() throws ApiException {
         String recordId = "id1";
         String entityType = "FOO";
         String attributeName = "attr1";
-        RecordAttributes recordAttributes = new RecordAttributes();
+        Map<String, Object> recordAttributes = new HashMap<>();
```
davidangb, line 81: This is the only change I _noticed_ in the new client - it uses `Map<String, Object>` instead of `RecordAttributes`. However, these tests aren't comprehensive and there may be other changes.

service/src/integrationTest/java/org/databiosphere/workspacedataservice/InstanceInitializerCloneTest.java
```diff
@@ -140,7 +140,7 @@ void cloneSuccess() throws ApiException {
         BackupResponse sourceBackupResponse = new BackupResponse();
         sourceBackupResponse.setFilename("/fake/filename/for/unit/test");
         BackupJob sourceBackupJob = new BackupJob();
-        sourceBackupJob.setStatus(org.databiosphere.workspacedata.model.Job.StatusEnum.SUCCEEDED);
+        sourceBackupJob.setStatus(BackupJob.StatusEnum.SUCCEEDED);
```
davidangb, line 143: swagger-codegen created the `BackupJob` class with `extends Job`, but openapi-generator does not. So this code that used to work now needs a change.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -200,7 +200,7 @@ private String requestRemoteBackup(UUID trackingId) {
 
             // TODO when the wdsDao.triggerBackup is async, we will need a second call here to poll for/check its status
             // note that the Job class on the next line is from the generated WDS client:
-            if (backupResponse.getStatus().equals(org.databiosphere.workspacedata.model.Job.StatusEnum.SUCCEEDED)) {
+            if (backupResponse.getStatus().equals(org.databiosphere.workspacedata.model.BackupJob.StatusEnum.SUCCEEDED)) {
```
davidangb, line 203: swagger-codegen created the `BackupJob` class with `extends Job`, but openapi-generator does not. So this code that used to work now needs a change.

.github/workflows/publish-java-client.yml
```diff
@@ -8,7 +8,10 @@ on:
         required: true
 
 jobs:
-  publish-job:
+  jakarta_matrix:
+    strategy:
+      matrix:
+        jakarta: [ true, false ]
```
davidangb, line 14: we now generate and publish the client twice - for javax and jakarta.

yuliadub (approved):

client/swagger.gradle
```diff
@@ -1,51 +1,37 @@
...
-	additionalProperties = [
-			modelPackage  : "${artifactGroup}.model",
-			apiPackage    : "${artifactGroup}.api",
-			invokerPackage: "${artifactGroup}.client",
-			dateLibrary   : 'java11'
-	]
-
-	rawOptions = ["--ignore-file-override", "${projectDir}/.swagger-codegen-ignore"]
+openApiGenerate {
+	inputSpec = "$rootDir/service/src/main/resources/static/swagger/openapi-docs.yaml".toString()
```
calypsomatic, line 10: Ooh does this fix https://broadworkbench.atlassian.net/browse/AJ-1090 too?

calypsomatic (approved):
LGTM

client/swagger.gradle
jladieu: !!!!! if so 🎉

davidangb: No, it doesn't particularly fix that …. but …. #331 331

davidangb: yeah, on this branch it replaces the `java.net.MalformedURLException` error with a different error:
```
'scheme' not defined in the spec (2.0). Default to [http] for server URL [http://../]
```
so I declared it not fixed :)

client/build.gradle
```diff
@@ -4,13 +4,30 @@ plugins {
...
+
+// default to javax, since WDS is currently on Spring Boot 2. -Pjakarta=true overrides
+// the default and creates a Jakarta-based client.
+if (project.hasProperty("jakarta") && "true".equalsIgnoreCase(project.getProperty("jakarta"))) {
+	apply from: "dependencies-jakarta.gradle"
+} else {
+	apply from: "dependencies-javax.gradle"
+}
+
+logger.warn("***** building library ${libraryName}")
```
jladieu, line 29: Great messaging to help understand what's going on, and how to change it!  But does it need to be warn level?  I always assume `warn` == `something bad`

service/src/test/java/org/databiosphere/workspacedataservice/GeneratedClientTests.java
jladieu: SGTM to do this for now, but  it does seem like kind of a bummer in that we maybe will lose some type guarantees here. Might it be worth creating a ticket with pointers back to this change for when `generateAliasAsModel` matures?

jladieu (approved):
Whoosh...this PR shows me just how much I don't know about Gradle... nice work here!

Just a couple small questions/comments, but LGTM!

client/swagger.gradle
```diff
@@ -1,52 +1,87 @@
...
+tasks.register('patchSources', Copy) {
+	from layout.buildDirectory.dir("generated/src-original/main/java/org/databiosphere/workspacedata/model")
+	include "*.java"
+	into layout.buildDirectory.dir("generated/src/main/java/org/databiosphere/workspacedata/model")
+	filter { String line ->
+		line.replace(
+				"RecordAttributes attributes = new HashMap<>();",
+				"RecordAttributes attributes = new RecordAttributes();")
+	}
+}
```
davidangb, line 72: of note: these fixes mean this PR has *_no_* changes to `GeneratedClientTests` - that is, adopting this client has no syntax changes for our calling usages.

client/build.gradle
davidangb: I couldn't get info messages to show up, when using gradle defaults. TIL! Gradle has a custom log level `LIFECYCLE` which is the default log level … and it's between INFO and WARN. I have:
* changed these log messages to be `logger.lifecycle(…)`
* moved the log messages so we only log when we actually generate the client code, not on every gradle invocation

service/src/test/java/org/databiosphere/workspacedataservice/GeneratedClientTests.java
davidangb: This is now moot, with the latest changes in this PR

calypsomatic (approved):
Seems like a lot of hassle, but if it works!

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: Glad I read this PR, changes the way I was thinking about quartz stuff

ashanhol (approved):

PR #324, opened 2023-08-11
AJ-1219: Convert application.properties to YAML

jladieu:
In keeping with DSP standardizing on YAML, this converts the two props files in WDS to YML format.

The following two keys were pruned during conversion due to there not being any references:
* `twds.clientId`
* `sentry.pingTTL`


service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
```diff
@@ -17,7 +17,6 @@
 
 @Configuration
 @PropertySource("classpath:git.properties")
-@PropertySource("classpath:application.properties")
```
jladieu, line 16: I initially tried changing this line to just refer to the new YAML file with `"classpath:application.yml"` but this proved problematic because it was still being treated as a properties file and generating tons of duplicate keys.

Turns out `@PropertySource` can't be used with YAML files, per: https://docs.spring.io/spring-boot/docs/2.1.9.RELEASE/reference/html/boot-features-external-config.html#boot-features-external-config-yaml-shortcomings

However, just by having an application.yml on the classpath, the default Spring-boot loader precedence takes effect, recognizes it, and deserializes it correctly as YAML.  (Verified using the `DUMP_PROPERTIES` before/after snapshots).

ashanhol (approved):

service/src/main/java/org/databiosphere/workspacedataservice/logging/AppStartPropertiesPrinter.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.core.env.MapPropertySource;
+import org.springframework.stereotype.Component;
+
+import javax.annotation.PostConstruct;
+import java.util.Collection;
+import java.util.stream.Collectors;
+
+@Component
+public class AppStartPropertiesPrinter {
+    private static final Logger LOGGER = LoggerFactory.getLogger(AppStartPropertiesPrinter.class);
```
calypsomatic, line 17: This file and the associated dumpProperties value and behavior in SentryInitializer make sense for testing that you didn't make any unintended changes - but do we need to keep it in the codebase? Just as a convenience that we might want to check all the properties at some point in the future?

jladieu: I kept it around under the assumption that it'd be helpful for any time we need to fiddle with properties, but am not attached to it at all if we'd prefer to keep the code clean per YAGNI; same goes for the main props dump as well.

calypsomatic: Yeah I'm torn between 'oh it might help someday!' and YAGNI (which I had to look up!).  Please, someone with strong opinions, weigh in!

service/src/main/resources/application.yml
```diff
@@ -0,0 +1,127 @@
...
+      username: ${env.wds.db.user}
+      password: ${env.wds.db.password}
+      maximum-pool-size: 7
+      minimum-idle: 7
+  servlet:
+    multipart.max-request-size: 5GB
+    multipart.max-file-size: 5GB
+  sql:
+    init:
+      # Ensure data.sql is NOT run on startup
```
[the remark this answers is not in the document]
jladieu: Added 9fdcf39718ed5960ba40549ef0a33ece486a2f22 to prune this.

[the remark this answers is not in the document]
calypsomatic: Ok I can see that we [added](https://github.com/DataBiosphere/terra-workspace-data-service/pull/216/files) `sql.init.mode=never` at the same time we removed `data.sql`.  The comment made it seem like we shouldn't even need `sql.init.mode=never` if there's no `data.sql` - but I'm not sure.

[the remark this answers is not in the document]
jladieu: The impression I got from Googling `sql.init.mode` is that the default behavior is to do something here: 

"By default, Spring Boot automatically creates the schema of an embedded DataSource." [src]("By default, Spring Boot automatically creates the schema of an embedded DataSource.")

So having this overridden to `never` is still required if we don't want Spring Boot to do this?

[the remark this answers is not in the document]
calypsomatic: Ok - maybe instead of removing the comment, change it to explain but not reference data.sql?

service/src/main/java/org/databiosphere/workspacedataservice/logging/AppStartPropertiesPrinter.java
davidangb: I'm fine leaving this in, BUT think it's worth a code comment that this could potentially log sensitive information like the startup token, so devs should be careful not to turn it on in live environments. Or delete it to be safe :)

davidangb (approved):
lgtm, with the one comment inline about potentially logging sensitive info

service/src/main/java/org/databiosphere/workspacedataservice/logging/AppStartPropertiesPrinter.java
jladieu: Per https://broadinstitute.slack.com/archives/C02SJQ0FKHQ/p1692045383378929?thread_ts=1692029661.294659&cid=C02SJQ0FKHQ, to eliminate the chance of an accidental disclosure of sensitive props I'm going to append another commit to remove all the `DUMP_PROPERTIES` changes.  Hopefully they'll remain in the commit history if we ever need to temporarily resurrect them for more properties changes, but otherwise YAGNI =)

service/src/main/resources/application.yml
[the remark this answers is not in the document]
jladieu: So sorry I missed this comment before merging.  Followup in: https://github.com/DataBiosphere/terra-workspace-data-service/pull/326

PR #327, opened 2023-08-21
AJ-1143: Skip null relation values when determining relation columns.

jladieu:
* 33b8d450907e41263976e51b4103e5f4be50e8e9 - Formatting tweaks only to Improve test readability.
* be6595bd41d25820b1426236d965d00c300ae706 - Eliminate a `*` import and include some resources in the commit comment for why.
* 52ef9dfb27ea6db738fab0fc90a75d9d5e3d9a74 - Fix issue that caused null values in relation columns to crash schema infererence.
* 3ab4e01f3a4f8e689979c9c988e392bcaf94d76a - Apply similar fix for multi-relation values too.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
@@ -24,7 +24,20 @@ import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.*;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.ARRAY_OF_DATE;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.ARRAY_OF_DATE_TIME;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.ARRAY_OF_RELATION;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.ARRAY_OF_STRING;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.BOOLEAN;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.DATE;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.DATE_TIME;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.EMPTY_ARRAY;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.FILE;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.JSON;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.NULL;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.NUMBER;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.RELATION;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.STRING;
 
 public class DataTypeInferer {
 
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
@@ -8,6 +8,7 @@ import org.databiosphere.workspacedataservice.service.model.Relation;
 import org.databiosphere.workspacedataservice.service.model.RelationCollection;
 import org.databiosphere.workspacedataservice.shared.model.Record;
 import org.databiosphere.workspacedataservice.shared.model.RecordAttributes;
+import org.databiosphere.workspacedataservice.shared.model.RecordType;
 import org.springframework.util.CollectionUtils;
 
 import java.math.BigDecimal;
@@ -23,8 +24,11 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
-import java.util.stream.Collectors;
 
+import static java.util.stream.Collectors.toSet;
+import static org.databiosphere.workspacedataservice.service.RelationUtils.getTypeValue;
+import static org.databiosphere.workspacedataservice.service.RelationUtils.getTypeValueForArray;
+import static org.databiosphere.workspacedataservice.service.RelationUtils.getTypeValueForList;
 import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.ARRAY_OF_DATE;
 import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.ARRAY_OF_DATE_TIME;
 import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.ARRAY_OF_RELATION;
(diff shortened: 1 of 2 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/DataTypeInfererTest.java
```diff
@@ -16,135 +16,185 @@
...
+                                putAttribute("str", "hello")),
+                new Record("1", RecordType.valueOf("thing"),
+                        RecordAttributes.empty().
+                                putAttribute("rel", createRelationString(RecordType.valueOf("thing"), "1")).
+                                putAttribute("str", "world")));
+        Map<String, DataTypeMapping> schema = inferer.inferTypes(records);
+
+        Set<Relation> relations = inferer.findRelations(records, schema).relations();
+        assertThat(relations.size()).isEqualTo(1);
+        Relation relation = relations.stream().collect(onlyElement());
```
davidangb, line 173: TIL! `onlyElement()` looks like a very useful helper.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -257,7 +271,8 @@ public RelationCollection findRelations(List<Record> records, Map<String, DataTy
 			// find all scalar attributes for this record whose names are in relationAttributes
 			// and convert them to Relations, then save to the "relations" Set
 			Set<Relation> relationsForThisRecord = rec.attributeSet().stream()
-					.filter( entry -> relationAttributes.contains(entry.getKey()))
+					.filter(entry -> relationAttributes.contains(entry.getKey()))
+					.filter(entry -> Objects.nonNull(entry.getValue()))
```
davidangb, line 275: perhaps this is a micro-optimization, but how about combining the filters so you can short-circuit?
```java
.filter(entry -> relationAttributes.contains(entry.getKey()) && Objects.nonNull(entry.getValue()))
```

davidangb (approved):
optional optimization inline, thank you!

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
jladieu: I agree with this optimization, good call!  But please take a look at https://github.com/DataBiosphere/terra-workspace-data-service/pull/329 where I piggyback on this change to introduce some extra syntactic sugar using `BiStream`...

I think in _that_ context, the optimization has a higher cost in readability because it requires backing up a step from the higher-level predicate sugar.  I'll see if I can find a way to have optimal cake & eat it too though :)

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -271,29 +276,30 @@ public RelationCollection findRelations(List<Record> records, Map<String, DataTy
...
-					.collect(Collectors.toSet());
+					.filter(entry -> relationArrayAttributes.contains(entry.getKey()) && Objects.nonNull(entry.getValue()))
+					.map(entry -> new Relation(entry.getKey(), getMultiValueType(entry.getValue())))
+					.collect(toSet());
 			relationArrays.addAll(relationArraysForThisRecord);
 		}
 		return new RelationCollection(relations, relationArrays);
 	}
 
+	private RecordType getMultiValueType(Object value) {
```
jladieu, line 295: I also pulled in a little method extraction here which I thought helped improve readability while working on https://github.com/DataBiosphere/terra-workspace-data-service/pull/329 downstream.

jladieu:
I appended another commit with some followups based on the first round of feedback, PTAL :)

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
jladieu: Introduced the optimization, I'll deal with the readability in the followup.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -21,10 +22,28 @@
...
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.BOOLEAN;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.DATE;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.DATE_TIME;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.EMPTY_ARRAY;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.FILE;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.JSON;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.NULL;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.NUMBER;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.RELATION;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.STRING;
```
ashanhol, line 46: [philosophical] 
I know you said you don't like to use *, but is there an upper limit if you're importing a lot of things from the same place? i mean this is 1 -> 14 lines

calypsomatic:
> @calypsomatic Great testing to find that edge case! How did you know to try that out?

It was an accident!  I can't remember exactly what I was doing or trying to do, I think I was just trying to modify something else or simplify the tsv and this happened.

> To me the code looks like it's authored to actively prevent updating the relation value.

Hmm - I believe the intention of that is to not update the *type* of a relation or relation array column.  So when an attribute comes in with a null value, it doesn't recognize that it's a relation and thinks you're trying to change a relation column to something else.  It looks like we have a [NULL](https://github.com/DataBiosphere/terra-workspace-data-service/blob/c896ed4f84e251cb4b00c062e58d94f2a05ec0b2/service/src/main/java/org/databiosphere/workspacedataservice/service/model/DataTypeMapping.java#L9) data type so maybe we need logic that if incoming datatype is NULL and existing is RELATION, keep it as RELATION?

davidangb:
> maybe we need logic that if incoming datatype is NULL and existing is RELATION, keep it as RELATION?

Or to generalize it: if the incoming datatype is NULL, don't change the existing datatype, no matter what it is currently.

That said, we have a [slew of known issues/corner cases](https://broadworkbench.atlassian.net/browse/AJ-1134) around detecting and changing datatypes. I think it's ok to fix the targeted issue in this PR (avoid NPEs) and leave datatype detection/changes for a more holistic solution later.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
jladieu: Great question!  My short [philosophical] answer is nope.   I set my IDE to like 100 under the premise that I'll never breach that ceiling.  But yeah, I think the idea is that * imports can present enough of a burden that the shorthand to save lines isn't worth it.

In general I've developed a bias toward very explicit/non-glob(*) imports & references whenever possible.

That said, I realize is only feasible if your IDE and tooling can really help you not trip over it.  I think this highlights a dependency on the other conversation we're having about style guidelines & codeformatters.... if your imports don't follow a coherent, IDE-automated, and enforced order, then doing explicit imports can also be a lot more difficult because of all the churn they create in the codebase.

calypsomatic (approved):
👍 assuming the python test failure is transient

jladieu:
Just a followup to summarize outcomes on a couple of the questions posed so far:

1) I plan to just keep PR scoped to only fixing the NPE and will punt addressing some of the more interesting corner cases around providing a `NULL` column.  (These sound like fun to work on though, so happy to pick up this followup work when it makes sense!)

2) We had some interesting conversation about optimization w.r.t. streams and while this PR includes an optimization to combine two filters into one, I actually end up reversing that optimization in favor of readability in a followup PR.  I want to keep that transition in place though to highlight and memorialize the change, and more importantly the conversation in the PRs that resulted in the change. :)

service/src/test/java/org/databiosphere/workspacedataservice/DataTypeInfererTest.java
```diff
@@ -21,130 +21,158 @@
...
+import static java.util.Map.ofEntries;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.databiosphere.workspacedataservice.service.RelationUtils.createRelationString;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 @DirtiesContext
 @SpringBootTest(classes = {JsonConfig.class, DataTypeInfererConfig.class})
 class DataTypeInfererTest {
 
-	@Autowired
```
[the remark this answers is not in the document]
jladieu: Fixed whitespace.

davidangb (approved):
woot the python client test passed, chalking that up to a transient flake. :+1:

PR #328, opened 2023-08-21
[AJ-879] Update default user agent for client libraries

nawatts:
Currently, the WDS Java client's default user agent is `Swagger-Codegen/1.0.0/java`. The Python client's default user agent is a little better, since it includes the WDS version it was published from. For example, `OpenAPI-Generator/0.2.97/python`.

This updates the default user agents for both client libraries to `wds-client/<version>/<language>`. For example, `wds-client/0.2.97/java` for the Java client and `wds-client/0.2.97/python` for the Python client.

The generated client objects allow consumers to override the default user agent.

Note: will conflict with #321.

### DIFF
--- a/.github/workflows/release-python-client.yml
+++ b/.github/workflows/release-python-client.yml
@@ -55,7 +55,7 @@ jobs:
           -i service/src/main/resources/static/swagger/openapi-docs.yaml \
           -g python \
           -o wds-client \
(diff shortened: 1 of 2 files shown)


davidangb (approved):
lgtm, thank you! Did not test it myself.

nawatts:
I tested by looking at the generated code. The default user agent is set in the API Client's constructor.

calypsomatic (approved):
LGTM

PR #329, opened 2023-08-21
AJ-1286: Introduce `BiStream` idioms using `com.google.mug`

jladieu:
While working on `DataTypeInferer` I noticed these stanzas of code which cried out for `BiStream` which was designed exactly for this sort of streaming logic.  See https://github.com/google/mug#bistream-streams-pairs-of-objects for many examples.

This commit is pure refactoring and is intended to demonstrate how `BiStream` can be used to simplify complex iterations by using higher-level predicates.

Note, since `BiStream` inherits from the Stream API, it also picks up the `Autocloseable` interface, which triggers some extra warnings which you can safely ignore.  For Intellij, this can be done by adding `BiStream` and `BiStream.from` to the allowlist to ignore these warnings.

autocloseable-warnings

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
@@ -3,6 +3,7 @@ package org.databiosphere.workspacedataservice.service;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.mu.util.stream.BiStream;
 import org.databiosphere.workspacedataservice.service.model.DataTypeMapping;
 import org.databiosphere.workspacedataservice.service.model.Relation;
 import org.databiosphere.workspacedataservice.service.model.RelationCollection;
@@ -26,7 +27,6 @@ import java.util.Objects;
 import java.util.Set;
 
 import static java.util.stream.Collectors.toSet;
-import static org.databiosphere.workspacedataservice.service.RelationUtils.getTypeValue;
 import static org.databiosphere.workspacedataservice.service.RelationUtils.getTypeValueForArray;
 import static org.databiosphere.workspacedataservice.service.RelationUtils.getTypeValueForList;
 import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.ARRAY_OF_DATE;
@@ -264,8 +264,11 @@ public class DataTypeInferer {
 	 * @return Set of Relation for all referencing attributes
 	 */
 	public RelationCollection findRelations(List<Record> records, Map<String, DataTypeMapping> schema) {
-		List<String> relationAttributes = schema.entrySet().stream().filter(attr -> attr.getValue() == RELATION).map(Map.Entry::getKey).toList();
-		List<String> relationArrayAttributes = schema.entrySet().stream().filter(attr -> attr.getValue() == ARRAY_OF_RELATION).map(Map.Entry::getKey).toList();
+		List<String> relationAttributes = BiStream.from(schema)
+				.filterValues(mapping -> mapping == RELATION).keys().toList();
+		List<String> relationArrayAttributes = BiStream.from(schema)
+				.filterValues(mapping -> mapping == ARRAY_OF_RELATION).keys().toList();
+
 		Set<Relation> relations = new HashSet<>();
 		Set<Relation> relationArrays = new HashSet<>();
 		if (relationAttributes.isEmpty() && relationArrayAttributes.isEmpty()) {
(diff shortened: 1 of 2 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -264,33 +265,36 @@ public boolean isValidDate(String val) {
...
+                    .filter((attribute, value) -> relationAttributes.contains(attribute) && Objects.nonNull(value))
+                    .mapValues(RelationUtils::getTypeValue)
+                    .mapToObj(Relation::new).collect(toSet());
+            relations.addAll(relationsForThisRecord);
+
+            // find all array attributes for this record whose names are in relationArrayAttributes
+            // and convert them to Relations, then save to the "relationArrays" Set
+            Set<Relation> relationArraysForThisRecord = BiStream.from(rec.attributeSet())
+                    .filter((attribute, value) -> relationArrayAttributes.contains(attribute) && Objects.nonNull(value))
+                    .mapValues(this::getMultiValueType)
```
jladieu, line 292: @davidangb You'll see BiStream API favors more expressive uses of subsequent `mapValues` `mapKeys` but this also steers the code toward taking multiple passes through the same stream.  I think it comes into conflict with the [optimization suggestion](https://github.com/DataBiosphere/terra-workspace-data-service/pull/327#discussion_r1300570115) you made upstream.

In practice, I used `BiStream` extensively and expressively and never thought twice about multiple passes, and didn't encounter any problems.  However I wasn't dealing with enormous data sets.  As I think about this...I now wonder, might the streaming API actually designed to eliminate the concern of multiple passes?

Is it possible that [under the hood](https://developer.ibm.com/articles/j-java-streams-3-brian-goetz/) the implementation will obviates this concern and 'does the right thing'?

jladieu: After thinking about this a bit, my _hunch_ is that the stream API will 'do the right thing' here and protect another pass since it actually performs the operations on the stream on each entity processed.  I think this intentionally lets us use the Stream API expressively without having to consider the runtime characteristic of chained filters.  I tried to research this hunch a bit and came up with opposing arguments:

- Complex filters can be more performant as Set sizes get really big [link](https://www.baeldung.com/java-streams-multiple-filters-vs-condition)
- Don't worry about performance, using method references actually helps with performance [link](https://stackoverflow.com/a/24055107)

I think I bias to go with the 2nd argument though...favor readability over performance until performance is demonstrably an issue.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -265,8 +266,11 @@ public boolean isValidDate(String val) {
 	 * @return Set of Relation for all referencing attributes
 	 */
 	public RelationCollection findRelations(List<Record> records, Map<String, DataTypeMapping> schema) {
-		List<String> relationAttributes = schema.entrySet().stream().filter(attr -> attr.getValue() == RELATION).map(Map.Entry::getKey).toList();
-		List<String> relationArrayAttributes = schema.entrySet().stream().filter(attr -> attr.getValue() == ARRAY_OF_RELATION).map(Map.Entry::getKey).toList();
+		List<String> relationAttributes = BiStream.from(schema)
+				.filterValues(mapping -> mapping == RELATION).keys().toList();
+		List<String> relationArrayAttributes = BiStream.from(schema)
+				.filterValues(mapping -> mapping == ARRAY_OF_RELATION).keys().toList();
```
davidangb, line 272: hm - IntelliJ is giving me warnings/suggestions that `BiStream.from` is used outside of a try-with-resources. I didn't get that in the original code, even though `schema.entrySet().stream()` is also AutoCloseable.

Do you have any sense if this is an issue for `BiStream`?

davidangb (commented):
One question inline, will 👍 based on answers. I also agree with your research and think it's not worth over-optimizing the stream filtering, especially since it is probably already handled for us!

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
jladieu: Ahh, thanks for reminding me of this.  I encountered the same thing, puzzled over it a bit, and Googled around to see if it was a problem and then turned off the warning and forgot to mention it in the PR.

tl;dr: this is safe to ignore, and as part of pulling in this change I will include instructions to update IDE to allowlist this class

More details: It looks like this warning is an artifact of a subtle change from Java 7 -> Java 8.  in Java 7 `Autocloseable` implementors *must* use try-with-resources and in Java 8 they *can* (more [details](https://blog.jooq.org/a-subtle-autocloseable-contract-change-between-java-7-and-java-8/)).  The Stream API and its subclasses generally fall into the latter camp, and BiStream is no exception.

A better summary [answer](https://stackoverflow.com/questions/22125169/why-is-java-8-stream-class-autocloseable) from StackOverflow.

davidangb: aha! And that's why IntelliJ didn't warn about `Stream` - because the inspection explicitly lists `java.util.stream.Stream` et al as ignorable. Thank you!

davidangb (approved):

calypsomatic (approved):

PR #337, opened 2023-09-13
[AJ-1138] Fix restore multi-replica bug

ashanhol:
Problem: Cloning logic is currently: 
- On Initialization, check to see if we're already cloning. 
- If so, don't start in cloning mode, start regular mode.
   - Regular mode checks to see if default schema (destWorkspaceId) is created. If not, create it. 
   - Continue with normal initialization.
 - If not, start cloning mode. 
   -  Cloning mode will request a backup, and restore if that's successful.
   - Part of restore is running the pg_dump, then changing the schema name from sourceWorkspaceId to destWorkspaceId
   - CONFLICT/SQL ERROR- can't change the name because destWorkspaceId schema was created by another replica not starting in clone mode! 

Solution:
There are 2 ways to solve this, either have the other replicas wait until cloning is complete (not this PR and I don't recommend doing until cloning is Async). Or, drop the destWorkspaceId schema if it exists before renaming schema during restore. 

TODO: figure out a test for this

Reminder:

### DIFF
--- a/service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
+++ b/service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
@@ -78,4 +78,20 @@ public class RestoreServiceIntegrationTest {
         List<RecordType> actualTables = recordDao.getAllRecordTypes(destInstance);
         assertEquals(expectedTables, actualTables);
     }
+
+    @Test
+    void testRestoreWithDuplicateDestId() {
+        UUID destInstance = UUID.fromString(workspaceId);
+
+        // confirm neither source nor destination instance should exist in our list of schemas to start
+        List<UUID> instancesBefore = instanceDao.listInstanceSchemas();
+        assertThat(instancesBefore).isEmpty();
+
+        // now add the destination workspaceid to simulate a replica adding it first
+        instanceDao.createSchema(destInstance);
+
+        // perform the restore, confirm that it succeeds
+        var response = backupRestoreService.restoreAzureWDS("v0.2", "backup.sql", UUID.randomUUID(), "");
+        assertSame(JobStatus.SUCCEEDED, response.getStatus());
+    }
 }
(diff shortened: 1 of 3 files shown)


calypsomatic:
Would replicas potentially also make multiple requests to the source workspace to make a backup?  Is that something we're not worrying about since it won't hurt anything?

jladieu (approved):

ashanhol:
> Would replicas potentially also make multiple requests to the source workspace to make a backup? Is that something we're not worrying about since it won't hurt anything?

This scenario _shouldn't_ happen. We create an entry in the clone db once cloning is kicked off and part of "should we initiate cloning?" is to check that db for the source workspace id to see if there's already a clone in progress. 
That said, in the event that _does_ happen (race condition?), I think my change would actually fix any potential issues. 
The clone entry is made with a different random tracking ID and we track a particular clone with that random ID, not with the source workspace ID. Both replicas would request the pg_dump, files would get uploaded to the container with a unique names, then each restore would take each file and restore. My change would dump the first clone (schema) in favor of the second.

calypsomatic (approved):

davidangb:
Locking is hard.

This PR all boils down to "when starting multiple replicas for a clone, make sure only one actually does the cloning work". We've implemented our own version of this by checking `cloneDao.cloneExistsForWorkspace` and `instanceDao.instanceSchemaExists`; this PR adds our own custom logic to handle concurrency cases where one replica started in regular mode but the other started in clone mode.

I am thinking we should not be building our own custom logic for this. We could rely on Spring to implement a distributed lock: https://docs.spring.io/spring-integration/reference/distributed-locks.html; if starting with a `SOURCE_WORKSPACE_ID` env var present, attempt to get the lock. If you don't get the lock, do nothing; if you do get the lock, then you can be sure you're the only node performing a clone, and the only logic you have to implement is checking if a clone already ran at some point in the past.

Another *possible* implementation is using [ShedLock](https://github.com/lukas-krecan/ShedLock), as TDR did in https://github.com/DataBiosphere/jade-data-repo/pull/1421, but that's for scheduled tasks and cloning is a one-time task, so we'd be jumping through some hoops.

Quartz could also help us here, as it allows run-once tasks and includes support for only one replica to run the task.

ashanhol:
Closing for now, https://github.com/DataBiosphere/terra-workspace-data-service/pull/341 to replace

PR #338, opened 2023-09-15
AJ-1242: disable primary key cache

davidangb:
Disables the cache for the primary key lookup. See code comments inline.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/CachedQueryDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/CachedQueryDao.java
@@ -1,16 +1,11 @@
 package org.databiosphere.workspacedataservice.dao;
 
 import org.databiosphere.workspacedataservice.shared.model.RecordType;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.springframework.cache.annotation.Cacheable;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.stereotype.Repository;
 
 import java.util.UUID;
-
-import static org.databiosphere.workspacedataservice.service.model.ReservedNames.PRIMARY_KEY_COLUMN_CACHE;
 
 /**
  * In order for @Cacheable to function properly callers need to be outside the class where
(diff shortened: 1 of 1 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dao/CachedQueryDao.java
```diff
@@ -1,44 +1,51 @@
...
 import java.util.UUID;
 
-import static org.databiosphere.workspacedataservice.service.model.ReservedNames.PRIMARY_KEY_COLUMN_CACHE;
-
 /**
  * In order for @Cacheable to function properly callers need to be outside the class where
  * the @Cacheable method is specified.
  * <a href="https://stackoverflow.com/questions/16899604/spring-cache-cacheable-not-working-while-calling-from-another-method-of-the-s">https://stackoverflow.com/questions/16899604/spring-cache-cacheable-not-working-while-calling-from-another-method-of-the-s</a>
- * Many getPrimaryKeyColumn calls are in RecordDao thus that method can't belong to that class and still work properly with caching
+ * Many getPrimaryKeyColumn calls are in RecordDao thus that method can't belong to that class and still work properly with caching,
```
davidangb, line 14: fixes a grammar warning in IntelliJ. Maybe I should just turn those off …

service/src/main/java/org/databiosphere/workspacedataservice/dao/CachedQueryDao.java
```diff
@@ -1,44 +1,51 @@
...
+
+        // see https://wiki.postgresql.org/wiki/Retrieve_primary_key_columns for this query
+        // for commentary, see also:
+        // https://stackoverflow.com/questions/1214576/how-do-i-get-the-primary-keys-of-a-table-from-postgres-via-plpgsql
+        return namedTemplate.queryForObject("""
+                      SELECT a.attname
+                      FROM   pg_index i
+                      JOIN   pg_attribute a ON a.attrelid = i.indrelid AND a.attnum = ANY(i.indkey)
+                      WHERE  i.indrelid = :qTableName::regclass
+                      AND    i.indisprimary;""",
```
davidangb, line 48: The previous query was consistently throwing errors under load, specifically in the `ConcurrentDataTypeChangesTest` unit test. It was returning multiple rows, but `namedTemplate. queryForObject` expects only one. As I went to troubleshoot that, I saw that the best practice from the Postgres wiki is a totally different query, so I switched to that one, and it's now passing.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -173,12 +173,6 @@ public void createRelationJoinTable(UUID instanceId, String tableName, RecordTyp
 		}
 	}
 
-	private String getQualifiedTableName(RecordType recordType, UUID instanceId) {
-		// N.B. recordType is sql-validated in its constructor, so we don't need it here
-		return quote(instanceId.toString()) + "."
-				+ quote(SqlUtils.validateSqlString(recordType.getName(), RECORD_TYPE));
-	}
-
```
davidangb, line 181: had to move this to `SqlUtils` to allow it to be used from `CachedQueryDao`

service/src/main/java/org/databiosphere/workspacedataservice/dao/CachedQueryDao.java
davidangb, line 1: renamed to `PrimaryKeyDao`.

service/src/test/java/org/databiosphere/workspacedataservice/dao/RecordDaoCacheTest.java
davidangb, line 1: I mentioned at standup that some tests were failing. Deleting this test - which is no longer useful since we're not caching - fixed the issues.

This test contained a `@Config` that supplied a `CacheManager` `@Bean`. Since that `@Config` was in scope for other unit tests, they were using its cache definition at times they shouldn't.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PrimaryKeyDao.java
```diff
@@ -0,0 +1,59 @@
...
+        // see https://wiki.postgresql.org/wiki/Retrieve_primary_key_columns for this query
+        // for commentary, see also:
+        // https://stackoverflow.com/questions/1214576/how-do-i-get-the-primary-keys-of-a-table-from-postgres-via-plpgsql
+        return namedTemplate.queryForObject("""
+                      SELECT a.attname
+                      FROM   pg_index i
+                      JOIN   pg_attribute a ON a.attrelid = i.indrelid AND a.attnum = ANY(i.indkey)
+                      WHERE  i.indrelid = :qTableName::regclass
+                      AND    i.indisprimary;""",
+                params, String.class);
```
davidangb, line 57: The previous query was consistently throwing errors under load, specifically in the `ConcurrentDataTypeChangesTest` unit test. It was returning multiple rows, but `namedTemplate. queryForObject` expects only one. As I went to troubleshoot that, I saw that the best practice from the Postgres wiki is a totally different query, so I switched to that one, and it's now passing.

calypsomatic (approved):
Love all the helpful comments!

jladieu:
Agreed with Bria, I appreciate all the trailmarker comments!

jladieu (approved):

PR #339, opened 2023-09-18
AJ-1119 - Add ability to pull last updated Time for WDS records

yuliadub:
This PR adds the following: 
- a new sys wds table that will keep track of when each data table was last updated (timestamp) in a given instance 
- adds an API call to get a time stamp of when any of WDS tables were last updated (to be later used by cloning to know if we should or should not trigger a new backup)

Where I would appreciate your feedback: 
- my overall approach with this
- I think I got all the places where data gets updated or deleted to make sure sys wds table acts appropriately but would love another set of eyes to make sure I didnt miss anything
- anything else we should add that could be useful as part of this? (one idea I had, we could add a get updated time per table in WDS and use that api to display last updated time per data table in the UI on the data tab)

Reminder:


service/src/main/resources/liquibase/changesets/20230915_record_table.yaml
```diff
@@ -0,0 +1,20 @@
+databaseChangeLog:
+  - changeSet:
+      id: 20230426_syswds_schema
+      author: yuliadub
+      changes:
+        - createTable:
+            schemaName: sys_wds
+            tableName: record
```
calypsomatic, line 8: Can this table have a more specific name?  Something like record_updates, or if we think this might be expanded for other uses in the future, maybe record_metadata?

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -1012,11 +1039,33 @@ public void deleteRecordType(UUID instanceId, RecordType recordType) {
...
+  public void recordUpdatedTime(RecordType recordType) {
+    namedTemplate
+        .getJdbcTemplate()
+        .update(
+            "update sys_wds.record SET updatedtime = ? where id = ?",
+            Timestamp.from(Instant.now()),
+            recordType.getName());
+  }
+
+  public void deleteRecordFromSys(UUID instanceId) {
```
calypsomatic, line 1066: Is this used anywhere?  Or is it just here in case?

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -70,6 +71,14 @@ public ResponseEntity<RecordResponse> getSingleRecord(
     return new ResponseEntity<>(response, HttpStatus.OK);
   }
 
+  @GetMapping("/{instanceId}/records/{version}")
```
jladieu, line 74: Rather than claim the top-level `/records` path for this endpoint, consider adding some qualifier, either very specific: `lastUpdated`, or super generic for future stuff, like: `/records/{version}/metadata`.

If you are interested in the generic route, then you probably should also return a map of key/value pairs (starting with just one element) rather than just returning a Timestamp

service/src/main/resources/liquibase/changesets/20230915_record_table.yaml
```diff
@@ -0,0 +1,20 @@
+databaseChangeLog:
+  - changeSet:
+      id: 20230426_syswds_schema
+      author: yuliadub
+      changes:
+        - createTable:
```
jladieu, line 6: This might blow the scope up a bit, but for data provenance reasons, it might be worth considering modeling this as an append-only audit table rather than as a per-table snapshot.  Specifically, when any kind of edit occurs, create a new record with the type of transformation (CRUD), the record type, etc.

The fields would be roughly:
`record/type` -- foreign key to the record metadata table
`action` -- enumerable set of potential actions, eg. CREATE/UPDATE/DELETE
`recordedAt` -- timestamp of the activity
`<potential other stuff we'd wanna audit, like what fields changed, etc.>`

And then instead of doing an `UPDATE` every time any kind of edit occurs, you'd just insert a new record to the end of the table.

You would need to add an index on the recordedAt timestamp to make it speedy to search & sort.

calypsomatic: This is definitely blowing up the scope, but whether we do this or not, makes me think of the activity logger - maybe we could abstract out a function that's called with every change, that both logs with activity logger AND updates this table?  Or is that done at a different level (service vs dao) and we should keep it that way?

davidangb: The activity logger was specifically built with an eye towards supporting this - from #238 :

> The implementation in this PR is a bit extra and over-designed. I am looking ahead to a time when we are also saving activity log entries to a db

so I think this could be a great solution! I agree the activity log table could get big and unwieldy and sorting on it could be a pain; we could solve this with a trigger that updates a second table, perhaps.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -1012,11 +1039,33 @@ public void deleteRecordType(UUID instanceId, RecordType recordType) {
       namedTemplate
           .getJdbcTemplate()
           .update("drop table " + getQualifiedTableName(recordType, instanceId));
+
+      namedTemplate
```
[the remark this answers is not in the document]
yuliadub: ha great point! I think my mind was thinking - a table is gone, so why does it matter :D

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
yuliadub: not used - originally I was going to use this for cleaning up of tests, but decided it made more sense to couple that with instance deletion.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
yuliadub: I think for the scope of what this is, I'd like to keep it specific for now. In the future we can always add more  as we need it :)

service/src/main/resources/liquibase/changesets/20230915_record_table.yaml
yuliadub: I can totally see the overall benefit of this in theory but what is going to be the benefit of having that granular of records for user created content? Would this be something we could display to the user as history of their actions? would that help them in any way? are there any other uses we internally can benefit from having that granular data for record history? 

I thought about this approach and opted into not doing it due to the potential size of data activity getting large (think about a workflow updating records per task which is run on every row) and that is why I opted going his direction. 

I think my gut tells me to not expand the scope unless we have a specific reason we are going to use it for - I am intentionally keeping this scope small since I have a very specific use case. I think if we did want to expand in the future, we can always add a new table and slowly age this one out.

service/src/main/resources/liquibase/changesets/20230915_record_table.yaml
davidangb: I have the same thought, but we can always rename the table later, too

davidangb (commented):
A few thoughts for debate:
1. Do we need to expose a new API for this at all? Knowing that we can always add an API later, isn't the primary use case for tracking a last-updated timestamp to optimize backups? That's an internal optimization, no API needed.
2. Assuming that optimizing backups is the primary use case, we WILL need to update some timestamp when a record type/table is deleted. If a user deletes a table, any existing backups are no longer up-to-date … but the lastUpdated timestamp in this PR doesn't reflect that. Similarly, we need to bump on deleteInstance and probably createInstance.
3. I still feel that somehow hooking into the ActivityLogger is worthwhile to consider. We went through pains to ensure that all write operations included a call to the ActivityLogger … seems like you had to redo all that work to ensure that all write operations included a call to bump the last-updated timestamp. Future WDS features will also have to ensure that they include calls to _both_ the ActivityLogger and `recordUpdatedTime`; do we want that?

service/src/main/resources/liquibase/changesets/20230915_record_table.yaml
davidangb: Integration with the ActivityLogger doesn't necessarily mean we have to write one db row for each line of the activity logger. We could write each activity log line to a logfile, and then update a single timestamp in the db, or one timestamp per record type as you have it in this PR. See the review I just published for a bit more context on this!

yuliadub:
> A few thoughts for debate:
> 
> 1. Do we need to expose a new API for this at all? Knowing that we can always add an API later, isn't the primary use case for tracking a last-updated timestamp to optimize backups? That's an internal optimization, no API needed.
> 2. Assuming that optimizing backups is the primary use case, we WILL need to update some timestamp when a record type/table is deleted. If a user deletes a table, any existing backups are no longer up-to-date … but the lastUpdated timestamp in this PR doesn't reflect that. Similarly, we need to bump on deleteInstance and probably createInstance.
> 3. I still feel that somehow hooking into the ActivityLogger is worthwhile to consider. We went through pains to ensure that all write operations included a call to the ActivityLogger … seems like you had to redo all that work to ensure that all write operations included a call to bump the last-updated timestamp. Future WDS features will also have to ensure that they include calls to _both_ the ActivityLogger and `recordUpdatedTime`; do we want that?

1. Good point - likely makes sense to just take it out if we keep the scope to just back up. I left it in there since I thought showing a "last edited" timestamp in the UI could be useful but that is indeed increasing scope 

2. I thought I was capturing the delete record with this -- (https://github.com/DataBiosphere/terra-workspace-data-service/pull/339/files#diff-6fdffee8355da583e8db9f8b0995789baa890a7fc0b2ba3da10e381f60afee25R456) .

I will take a closer look at attributes - you are correct I do not capture that today.

Great point on instance schema drop - today I clean up everything when an instance is deleted. The way I viewed that is, no timestamp would be returned if that happened, which in the implementation for backup would mean there is no date present so a backup would need to be taken again. I like that implementation better vs trying to preserve a previous timestamp from the previous instance personally. 

3. This is my bad - I was not understanding what you were referring to when you mentioned "activity logger". Going to use it to achieve what is needed here :)

davidangb:
> I thought I was capturing the delete record with this …

I realize you said you're revamping to use the activity logger, but for clarity's sake: I am talking about deleting a _record type_ here, not a _record_ … in other words, deleting a whole table.

yuliadub:
this has been around long enough with no updates, going to close for now and we can revisit later on.

PR #341, opened 2023-09-19
[AJ-1138] Fix restore multi-replica bug with locks

ashanhol:
Problem: Cloning logic is not replica safe. This is due to the fact that currently if one replica initiates clone mode, another replica will see that and continue on to initialize the default schema. Problem is, cloning expects the default schema to not be created yet. 

Solution:
Implement Spring's distributed Lock system so only one replica can create the schema at a time (either through cloning or default initialization). 

TODO: figure out a test for this

Reminder:


service/src/test/java/org/databiosphere/workspacedataservice/InstanceInitializerBeanTest.java
```diff
@@ -59,10 +69,14 @@ class InstanceInitializerBeanTest {
     final UUID instanceID = UUID.fromString("90e1b179-9f83-4a6f-a8c2-db083df4cd03");
 
     @BeforeEach
-    void beforeEach() {
+    void beforeEach () throws InterruptedException {
         // clean up any instances left in the db
         List<UUID> allInstances = instanceDao.listInstanceSchemas();
         allInstances.forEach(instanceId -> instanceDao.dropSchema(instanceId));
+        Lock lock = mock(Lock.class);
```
ashanhol, line 76: Would love to know if there's a better way to to this

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -68,16 +73,25 @@ public void initializeInstance() {
...
-            LOGGER.info("Source workspace id loaded as {}.", sourceWorkspaceId);
-            boolean cloneSuccess = initCloneMode();
-            if (cloneSuccess) {
-                LOGGER.info("Cloning complete.");
-            } else {
+        Lock lock = lockRegistry.obtain(sourceWorkspaceId);
+        try {
+            // Make sure it's safe to start cloning (in case another replica is trying to clone)
+            boolean lockAquired = lock.tryLock(1, TimeUnit.SECONDS);
+            if (isInCloneMode && lockAquired) {
```
jladieu, line 80: Is it possible to pull the check for `isInCloneMode` out to before the lock is acquired?  The way I read this right now the lock will be acquired always, not just in clone situations.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -163,12 +176,11 @@ private boolean initCloneMode() {
             // backup succeeded; now attempt to restore
             restoreFromRemoteBackup(backupFileName, cloneStatus.getJobId());
 
-            // after the restore attempt, check the current clone status one more time
-            // and return the result
+            // after the restore attempt, check the current clone status one more time,
+            // unlock, and return the result
```
jladieu, line 180: I appreciate the comment here, but it's not clear from the immediate surrounding code where/who actually performs the unlock.  Should the unlock be pulled up into this method somehow, so it jives with the comment?  Alternately, should the method that's doing the unlock be renamed to reveal that responsibility?

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -255,18 +267,19 @@ private void restoreFromRemoteBackup(String backupFileName, UUID cloneJobId) {
...
-                LOGGER.debug("Default schema for workspaceId {} already exists; skipping creation.", workspaceId);
+                LOGGER.debug("Default schema for workspaceId {} already exists or is in progress; skipping creation.", workspaceId);
             }
 
         } catch (IllegalArgumentException e) {
             LOGGER.warn("Workspace id could not be parsed, a default schema won't be created. Provided id: {}.", workspaceId);
         } catch (DataAccessException e) {
             LOGGER.error("Failed to create default schema id for workspaceId {}.", workspaceId);
+        } catch (Exception e) {
+            LOGGER.error("Failed to create default schema id for workspaceId {}. {}", workspaceId, e.getMessage());
```
jladieu, line 282: I typically advocate to [fail fast](https://enterprisecraftsmanship.com/posts/fail-fast-principle/#:~:text=The%20fail%20fast%20principle%20stands,appear%20counter%2Dintuitive%20at%20first.) on  unhandled exceptions.  Is there something that was happening in particular that made you want to catch this class of exception rather than let it bubble up?  If possible, prefer catching and proceeding _only_ with the narrowest type possible.  If there isn't a particular exception type to work with, you can look for a regex in the exception message and rethrow anything that doesn't match.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -163,12 +176,11 @@ private boolean initCloneMode() {
...
-            // and return the result
+            // after the restore attempt, check the current clone status one more time,
+            // unlock, and return the result
             LOGGER.info("Re-checking clone job status after restore request");
             var finalCloneStatus = currentCloneStatus(trackingId);
             return finalCloneStatus.getStatus().equals(JobStatus.SUCCEEDED);
-
         }
         catch (Exception e) {
             LOGGER.error("An error occurred during clone mode. Error: {}", e.toString());
```
jladieu, line 186: Looks like you didn't touch this block, but I'm wondering what you think of failing fast here instead of catching & swallowing the error, per [comment](https://github.com/DataBiosphere/terra-workspace-data-service/pull/341/files#r1331662613)

service/src/test/java/org/databiosphere/workspacedataservice/InstanceInitializerBeanTest.java
jladieu: Is this extra stuff here just to get the locking mechanism out of the way so it doesn't interfere with the existing tests?  If so, maybe this indicates a good opportunity to make the locking logic optional/pluggable & disable it here.

service/src/test/java/org/databiosphere/workspacedataservice/InstanceInitializerBeanTest.java
jladieu, line 1: Re: suggestions on how to test.

I think mocking or faking the lock is probably the way to go to get test coverage for the locking mechanism.  I'm guessing you'll want to cover the following cases:

control group: non-cloning scenario, no locks acquired, confirm the code doesn't interact with the locking mechanism at all
non-cloning: locks present, but it doesn't matter since it's not cloning, confirm the code doesn't interact with the locking mechanism at all
cloning: locks absent, confirm code can grab a lock successfully and verify it does what you want
cloning: locks present, confirm code can't grab the lock and verify it doesn't do anything unexpected; since it won't have any side effects, you can verify expected log messages that indicate the no-op path is called
others ??? 

One question is: if you start writing these tests, do you find yourself tripping over the level of abstraction of the `Lock` class and mechanism or is it easy to stub/mock/spy as needed to cover the aforementioned test case?  If it's difficult to stub out and verify the scenarios you want, maybe the locking abstraction could be wrapped with our own simpler interface which is easier to test against, which would suggest doing a little refactoring to encapsulate the locking mechanism.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: thank you, this was actually a missed cleanup when I refactored some code. fixed.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: we don't necessarily want to fail fast in cloning, otherwise we wouldn't end up with any default schema. but i will look into fast fails elsewhere

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -68,16 +73,25 @@ public void initializeInstance() {
...
+                } else {
+                    initializeDefaultInstance();
+                }
+            } else if (lockAquired) {
                 initializeDefaultInstance();
             }
-        } else {
-            initializeDefaultInstance();
+        } catch (Exception e) {
+            LOGGER.error("Error with aquiring cloning Lock: {}", e.getMessage());
```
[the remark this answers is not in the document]
ashanhol: Took your suggestion but for some reason (wrapper class interference??) it's erroring if I try to `throw e;`

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -68,16 +73,25 @@ public void initializeInstance() {
...
+                    LOGGER.info("Cloning complete.");
+                } else {
+                    initializeDefaultInstance();
+                }
+            } else if (lockAquired) {
                 initializeDefaultInstance();
             }
-        } else {
-            initializeDefaultInstance();
+        } catch (InterruptedException e) {
```
ashanhol, line 91: @jladieu recommenting on updated code- for some reason (wrapper class interference?) this gets a build error if I try to `throw e;`

service/src/test/java/org/databiosphere/workspacedataservice/InstanceInitializerBeanTest.java
ashanhol: (copying and pasting from our slack cause its useful context on the PR)
I implemented your wrapping suggestion and made mock classes that make the lock always succeed or fail.
The way I've designed the code to be replica safe- we must acquire a lock in order to create the default schema OR clone. There's lock checks before doing both of these actions. The scenarios I imagine testing-
1. no clone, can acquire lock, assert default schema is created
2. no clone, can't acquire lock, assert default schema is not created
3. yes clone, can acquire lock, assert clone schema is created and NOT default schema.
4. yes clone, can't acquire lock for cloning, assert default schema is created.
Now, the difficulty is with the way I have the classes set up I dont think #4 is doable because the lock will fail every time, including when we try to create the default schema. I have no idea how to nuance-ly do that.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -73,6 +77,7 @@ public InstanceInitializerBean(
   public void initializeInstance() {
     LOGGER.info("Default workspace id loaded as {}.", workspaceId);
     boolean isInCloneMode = isInCloneMode(sourceWorkspaceId);
+    lock.obtainLock(sourceWorkspaceId);
```
davidangb, line 80: this can/should move right next to where `tryLock` runs - is there a need to do it up here?

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -151,8 +156,12 @@ a newly created (destination) workspace. WDS at start up will always have a curr
   private boolean initCloneMode() {
     LOGGER.info("Starting in clone mode...");
     UUID trackingId = UUID.randomUUID();
-
     try {
+      // Make sure it's safe to start cloning (in case another replica is trying to clone)
+      boolean lockAquired = lock.tryLock();
+      if (!lockAquired) {
+        return false;
+      }
```
davidangb, line 164: the call to `cloneDao.cloneExistsForWorkspace` should move out of `isInCloneMode` and instead be here. So, the order of operations is:
1. does the `SOURCE_WORKSPACE_ID` env var exist?
2. did I get a Spring lock?
3. does `cloneDao.cloneExistsForWorkspace` return empty?

If all of those are true, then proceed with cloning.

The reason I'm being a stickler about this is because the check for `cloneDao.cloneExistsForWorkspace` is subject to race conditions, so we want to ensure it is inside the Spring lock

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -288,14 +298,16 @@ private void restoreFromRemoteBackup(String backupFileName, UUID cloneJobId) {
   */
   private void initializeDefaultInstance() {
     try {
+      // Don't create schema if in the middle of cloning
+      boolean lockAquired = lock.tryLock();
```
davidangb, line 302: I don't think we need a lock here. The only case in which `initializeDefaultInstance` is called in a WDS where `SOURCE_WORKSPACE_ID` is populated is if the WDS attempted to restore a clone, but failed.

service/src/main/java/org/databiosphere/workspacedataservice/dao/DataSourceConfig.java
```diff
@@ -23,4 +26,17 @@ public DataSource mainDb() {
...
+    DefaultLockRepository defaultLockRepository = new DefaultLockRepository(dataSource);
+    defaultLockRepository.setPrefix("sys_wds.INT_");
+    defaultLockRepository.setTimeToLive(10 * 60 * 1000); // 10 minutes
+    return defaultLockRepository;
+  }
+
+  @Bean
+  public JdbcLockRegistry jdbcLockRegistry(LockRepository lockRepository) {
+    return new JdbcLockRegistry(lockRepository);
+  }
```
davidangb, line 41: I suggest moving these into a new `@Config` class inside the `distributed` package - something like `LockConfig`. I wouldn't think to look for these beans inside a class named `DataSourceConfig`

service/src/main/java/org/databiosphere/workspacedataservice/distributed/DefaultDistributedLock.java
```diff
@@ -0,0 +1,48 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(DefaultDistributedLock.class);
+
+  // A wrapper around a Lock for easier testing.
+  public DefaultDistributedLock(LockRegistry lockRegistry) {
+    this.lockRegistry = lockRegistry;
+  }
+
+  @Override
+  public void obtainLock(String lockId) {
+    lock = lockRegistry.obtain(lockId);
```
davidangb, line 25: consider moving the `Lock` object away from being a class-level variable and into method-local arguments. As a class-level variable, it's not threadsafe; if a first caller calls `obtainLock("foo")` and then a second caller calls `obtainLock("bar")`, when the first caller calls `tryLock()` it will be locking "bar", not "foo". So, you'd end up with signatures like:
```java
public Lock obtainLock(String lockId)

public Boolean tryLock(Lock lock)

public void unlock(Lock lock)
```

service/src/main/resources/liquibase/changesets/20230919_lock_table.yaml
```diff
@@ -0,0 +1,32 @@
+databaseChangeLog:
+  - changeSet:
+      id: 20230919_lock_table
+      author: adinas
+      changes:
+        - createTable:
+            schemaName: sys_wds
+            tableName: int_lock
```
davidangb, line 8: what does the "int" in the name signify? I know that's the default and used in examples; should we keep it?

davidangb (commented):
initial thoughts inline!

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: This is not the case with replicas. If one replica is in the middle of cloning, the second replica that starts after that but before the first is finished will have isInCloneMode return false and call initializeDefaultInstance. Even if I move the cloneDao.cloneExistsForWorkspace check to initCloneMode, we'd return false out of there and immediately call initializeDefaultInstance.

service/src/main/resources/liquibase/changesets/20230919_lock_table.yaml
ashanhol: I was under the impression that springs expected it, since [all their example schema creation](https://github.com/spring-projects/spring-integration/blob/v5.3.0.RELEASE/spring-integration-jdbc/src/main/resources/org/springframework/integration/jdbc/schema-postgresql.sql) has it. Since we're setting our own prefix I tried dropping the int from that and the liquibase yaml and ran into a vague issue where it doesn't think the current thread can unlock. Further investigation needed on whether it can done

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: It's one call one layer up instead of two calls with the same ID right before each trylock 🤷‍♀️
if you think it'd be clearer to just to the latter I can change it

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: Moved. 👍

service/src/main/java/org/databiosphere/workspacedataservice/dao/DataSourceConfig.java
ashanhol: Done. I put it there because it's where I saw datasource configured

service/src/main/java/org/databiosphere/workspacedataservice/distributed/DefaultDistributedLock.java
ashanhol: Fixed

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
davidangb: I know it's not the case currently - I really think we need to change that behavior, though. With `cloneDao.cloneExistsForWorkspace` moving out of `isInCloneMode`, it'll change what `isInCloneMode` returns.

https://lucid.app/lucidchart/fd59bf3a-da3b-4c35-bf5e-cac58c498cb2/edit?invitationId=inv_192a4441-f000-49b0-a97c-2d07889ba58e&page=0_0# is still how I think the overall flow should work!

davidangb: the TL;DR of how I think it should work:
| If | Then |
|--------|--------|
| `SOURCE_WORKSPACE_ID` undefined | create default instance; swallow errors if that instance already exists |
| `SOURCE_WORKSPACE_ID` defined; can't acquire lock | Do nothing |
| `SOURCE_WORKSPACE_ID` defined; lock acquired; row exists in clone table | Do nothing |
| `SOURCE_WORKSPACE_ID` defined; lock acquired; no row in clone table | Perform clone stuff |

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -287,15 +298,18 @@ private void restoreFromRemoteBackup(String backupFileName, UUID cloneJobId) {
      Create the default pg schema for this WDS. The pg schema name is the workspace id.
   */
   private void initializeDefaultInstance() {
+    Lock sourceLock = lock.obtainLock(sourceWorkspaceId);
     try {
+      // Don't create schema if in the middle of cloning
+      boolean lockAquired = lock.tryLock(sourceLock);
```
yuliadub, line 304: do we still need this lock here?

ashanhol: See my comment thread above with David. Tldr; i pulled the work he suggested into the new ticket.

ashanhol:
Just leaving a note here- wontfix that bug because sonarcloud doesn't know the lock class is wrapped. I did fix it so it unlocks if we hit the exception though.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: Noting here: This will be refactored in follow up work in this ticket https://broadworkbench.atlassian.net/browse/AJ-1363

service/src/test/java/org/databiosphere/workspacedataservice/InstanceInitializerBeanFailLockTest.java
```diff
@@ -0,0 +1,81 @@
...
+      LeonardoConfig.class,
+      WorkspaceDataServiceConfig.class,
+      MockCloneDaoConfig.class,
+      BackupRestoreService.class,
+      AzureBlobStorage.class,
+      WorkspaceManagerConfig.class,
+      ActivityLoggerConfig.class,
+      SamConfig.class,
+      MockSamClientFactoryConfig.class,
+      MockFailedDistributedLock.class
```
jladieu, line 53: I notice using class level dependency injection limits you to having one behavior for the distributed lock mock per test suite.  I think this also contributes to difficulty in standing up more nuanced or interesting test cases.  I don't actually know how to do this with `SpringBootTest` yet, but I am pretty sure we should be able to figure out a way to inject an instance of `DistributedLock` on a per-test (rather than per-suite) case basis, and that might be the hook we need to do more interesting unit tests.

It'd also let you collapse the two test case files into one `InstanceInitializerBeanTest`

I'm not suggesting you take this on in this PR, but I might merit followup in the refactoring/cleanup ticket, as part of helping improve the test coverage.

service/src/test/java/org/databiosphere/workspacedataservice/InstanceInitializerBeanFailLockTest.java
```diff
@@ -0,0 +1,81 @@
...
+class InstanceInitializerBeanFailLockTest {
+
+  @Autowired InstanceInitializerBean instanceInitializerBean;
+  @SpyBean InstanceDao instanceDao;
+
+  @Value("${twds.instance.workspace-id}")
+  String workspaceId;
+
+  // randomly generated UUID
+  final UUID instanceID = UUID.fromString("90e1b179-9f83-4a6f-a8c2-db083df4cd03");
```
jladieu, line 64: I'd suggest just using `randomUUID()` here instead of hardcoding the String, which doesn't really provide any more semantic value and then requires a comment explaining that it's just a random String anyway.  Is there anything that prevents you from doing that?

service/src/test/java/org/databiosphere/workspacedataservice/distributed/MockSuccessfulDistributedLock.java
```diff
@@ -0,0 +1,23 @@
+package org.databiosphere.workspacedataservice.distributed;
```
jladieu, line 1: This can go in the 'for future refactoring' bucket too (and I think it _must_ because my suggestion won't work unless you also pull that other work), alongside changes that let you have more finegrained control over _what_ gets injected into test cases, but I can't help but want to highlight that I think you can encapsulate these two mock classes into one with some nice inner class supports, as follows:

```
class MockDistributedLockFactory {
   static DistributedLock alwaysFailingLock() {
     return new AlwaysFailingLock();
   }
   
   static DistributedLock alwaysSucceedingLock() {
     return new AlwaysSucceedingLock();
   }

   private class AlwaysSucceedingLock implements DistributedLock {
     // implementation
   }

   private class AlwaysFailingLock implements DistributedLock {
     ///
   }
}
```

Then, later, when deciding which thing gets injected you could do something like:

```
// I have no idea if this syntax is even close to what's available in Springboot, so take with a grain of salt!!
@Inject(DistributedMock.class).with(FakeDistributedLockFactory.alwaysFailing())
```

jladieu (commented):
My feedback thus far has been focused on the unit tests, and I don't think I'm quite as read into @davidangb on the desired flow for the lock state.

I also realize you're trying to button this PR up to get to a good landing point, so I've just provided some comments on how I think we could further refine the tests in followup.

I'd like to defer my approval to @davidangb's here, but happy to tack my thumb on after you can get his :)

service/src/test/java/org/databiosphere/workspacedataservice/InstanceInitializerBeanFailLockTest.java
ashanhol: This is copied from another test suite. I can, but I didn't think it made a difference

service/src/test/java/org/databiosphere/workspacedataservice/InstanceInitializerBeanFailLockTest.java
ashanhol: I added a note to revisit this in the ticket. Honestly it might be moot once I refactor the logic to use the lock only once but I think finding a way to replace a class per test is still a useful endeavor.

service/src/test/java/org/databiosphere/workspacedataservice/distributed/MockSuccessfulDistributedLock.java
jladieu: Experimentation draft PR is [here](https://github.com/DataBiosphere/terra-workspace-data-service/pull/351)

Again, no need to do anything with this as part of this PR, but wanted to share my thinking made more concrete.

service/src/test/java/org/databiosphere/workspacedataservice/InstanceInitializerBeanFailLockTest.java
davidangb: I know you want to get this PR merged … but if you remove the distributed lock from `initializeDefaultInstance`, I agree you can delete this whole file as well as `InstanceInitializerBeanSuccesfulLockCloneTest`. If you delete these tests, you can also delete `MockFailedDistributedLock`.

service/src/test/java/org/databiosphere/workspacedataservice/InstanceInitializerBeanTest.java
```diff
@@ -51,7 +52,8 @@
       WorkspaceManagerConfig.class,
       ActivityLoggerConfig.class,
       SamConfig.class,
-      MockSamClientFactoryConfig.class
+      MockSamClientFactoryConfig.class,
+      MockSuccessfulDistributedLock.class
```
davidangb, line 56: and, if you used mocks here you could get rid of `MockSuccessfulDistributedLock` entirely too!

```java
class InstanceInitializerBeanTest {

  @Autowired InstanceInitializerBean instanceInitializerBean;
  @SpyBean InstanceDao instanceDao;

  @MockBean
  DistributedLock mockDistributedLock;
  Lock mockLock = Mockito.mock(Lock.class);

  …

  @BeforeEach
  void beforeEach() {
    // clean up any instances left in the db
    List<UUID> allInstances = instanceDao.listInstanceSchemas();
    allInstances.forEach(instanceId -> instanceDao.dropSchema(instanceId));

    // always return a valid Lock object and always return true for acquiring that lock
    when(mockDistributedLock.obtainLock(anyString())).thenReturn(mockLock);
    when(mockDistributedLock.tryLock(any(Lock.class))).thenReturn(true);
  }

davidangb (commented):
Some comments inline on a hopefully small change which would in turn allow you to delete a bunch of these files and make the entire PR smaller!

I know you want to finish this work in a follow-on PR. For clarity's sake, the remainder of the work that I see is:
* change `isInCloneMode` to only check if `SOURCE_WORKSPACE_ID` is a non-blank valid UUID, nothing else
* remove the distributed lock from `initializeDefaultInstance`
* ensure that `initializeDefaultInstance` doesn't throw an error if it runs into a conflict creating the instance. I suggest it should always call `InstanceDao.createSchema()` and swallow any conflicts, instead of checking to see if the instance exists first (because multi-replicas can have a race condition between checking and creating)
* add test coverage

service/src/test/java/org/databiosphere/workspacedataservice/InstanceInitializerBeanFailLockTest.java
ashanhol: The lock in initializeDefaultInstance keeps cloning replica safe at the moment (removing it makes the replicas interfere with each other). I'm assuming you're suggesting removing it so we don't have to add and remove these tests, which makes sense. We're not turning replicas on until after we complete the follow up ticket anyway so if we're cool with cloning w/ replicas being broken until then, yeah I can just remove it all.

ashanhol:
Punting this work to https://github.com/DataBiosphere/terra-workspace-data-service/pull/352. Follow up work documented in next ticket.

PR #346, opened 2023-09-28
AJ-1357: restrict to HTTP 1.1 for OkHttp clients

davidangb:
For all our OkHttp-based client libraries (Sam, Leo, WDS), restrict to HTTP 1.1 and disallow HTTP 2.

In scale testing, we have seen intermittent/hard-to-reproduce errors like `okhttp3.internal.http2.StreamResetException: stream was reset: CANCEL`. Previous investigation hinted at HTTP 2 being a culprit here.

We don't need HTTP 2, and I don't anticipate that restricting to HTTP 1.1 will have any repercussions.

See:
* https://github.com/broadinstitute/rawls/pull/2154
* https://github.com/DataBiosphere/cbas/pull/208
* https://broadinstitute.slack.com/archives/C05BYVBM2HE/p1695916806679619


yuliadub (approved):

jladieu (approved):
Weird!  But cool.  Seems legit :)

PR #331, opened 2023-08-22
AJ-1090: swagger definition and UI tweaks

davidangb:
In this PR:
* eliminate the warning about `../` url when generating the swagger client (which also happens at bootRun). The `../` url is a leftover from when WDS was bundled with CBAS and is no longer necessary.

Screenshot 2023-08-22 at 5 29 01 PM

I piggybacked on that fix with a few other swagger UI tweaks. These are my personal opinions; feel free to opine if you disagree!

start with each major section collapsed, making it easier to navigate quickly to the section you care about:
Screenshot 8-22-2023 at 05 30 PM

add a filter bar to quickly filter to the main section (aka tag) you care about:
Screenshot 8-22-2023 at 05 31 PM

display request durations for each API invocation:
Screenshot 8-22-2023 at 05 30 PM (1)

display the `operationId` for each API; these ids become method names in the generated Java and Python clients:
Screenshot 8-22-2023 at 05 31 PM (2)

PR #347, opened 2023-09-29
AJ-1011: openapi cleanup

davidangb:
Prefactoring for AJ-1011. In `service/src/main/resources/static/swagger/openapi-docs.yaml`:

* clean up some formatting
* add some code comments
* sort paths first by their tag (Records, Schema, Instances, etc) and second alphabetically
* sort all components alphabetically

A lot of lines of code have changed, but there should be zero functional changes; it's all reordering and formatting. You can view commit-by-commit to see it in smaller chunks.

### DIFF
--- a/service/src/main/resources/static/swagger/openapi-docs.yaml
+++ b/service/src/main/resources/static/swagger/openapi-docs.yaml
@@ -609,7 +609,7 @@ components:
           description: name of this attribute.
         datatype:
           type: string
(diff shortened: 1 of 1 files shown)


calypsomatic (commented):
While you're at it, can you put the API name over on the right side, like here? 
image

davidangb:
> While you're at it, can you put the API name over on the right side, like here? image

@calypsomatic I did that back in #331! Note that you'll only see it when hitting the UI from a running WDS; you won't see it when viewing the spec via https://petstore.swagger.io/ - because the change is specific to the HTML/JS that interprets the spec.

jladieu (approved):
Lookin' good!

calypsomatic:
> @calypsomatic I did that back in #331! Note that you'll only see it when hitting the UI from a running WDS; you won't see it when viewing the spec via https://petstore.swagger.io/ - because the change is specific to the HTML/JS that interprets the spec.
Huh!

calypsomatic (approved):

PR #348, opened 2023-09-29
AJ-1011: add openapi validation to build

davidangb:
More prefactoring work for AJ-1011. This adds validation of our OpenAPI spec to the gradle build.

If the spec is invalid, we'll get a build failure:
Screenshot 2023-09-29 at 3 36 55 PM

Note that we also get non-fatal recommendations on how to clean up the spec!


calypsomatic (approved):

jladieu (approved):
Love this!

PR #349, opened 2023-10-02
AJ-1011: setup for async snapshot api

davidangb:
This PR introduces a bunch of concepts to WDS; I'm putting this out for debate and discussion.

Note that this PR does not actually implement a new/async snapshot API yet. I will do that in follow-on PRs; I wanted to get this out for discussion before then.

In this PR:

## We can and should bump the version for individual APIs
Or, bump groups of APIs together. We don't have to version all WDS APIs at once. This PR adds a new `importV1` API as a successor to the v0.2 `importSnapshot` API.

## We can deprecate old APIs
In this PR, I've marked the `importSnapshot` API as deprecated and given it a 3-month time to live. Admittedly, this is an easy deprecation since we're not using that API anywhere yet.

## We should generate API interface definitions programmatically
And then hand-code implementation classes on top of those interfaces. This forces attention to detail on our OpenAPI spec. Since that spec is used to create our Java and Python clients, it's really important to get the spec right and name things in a friendly manner. By dogfooding our server-side API interfaces, we improve our clients as well. This PR introduces a `generateApiInterfaces` gradle task which generates the `ImportApi` interface at build time; `ImportController` then implements that interface. It also generates the `ImportRequest` model class.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,10 +257,8 @@ their implementations are hand-coded.
 APIs at `v1` or later use code generation. To create or update an API, you must:
 
 1. Define the API using OpenAPI syntax
-   at `service/src/main/resources/static/swagger/openapi-docs.yaml`
-2. Update the `generateApiInterfaces` Gradle task in `service/build.gradle` as necessary.
-  1. Add your APIs to `apiFilesConstrainedTo`
-  2. Add your API's models to `modelFilesConstrainedTo`
+   at `service/src/main/resources/static/swagger/apis-v1.yaml`
+2. Include the API, via `$ref:`, in `service/src/main/resources/static/swagger/openapi-docs.yaml`
 3. Run the `generateApiInterfaces` Gradle task
 4. Note the new/updated Java interfaces and models under the `o.d.w.generated` package
 5. Create or update a controller class that implements the new Java interface
--- a/README.md
+++ b/README.md
@@ -1,15 +1,23 @@
 # Workspace Data Service
+
 [![Quality Gate Status](https://sonarcloud.io/api/project_badges/measure?project=DataBiosphere_workspace-data-service&metric=alert_status)](https://sonarcloud.io/summary/new_code?id=DataBiosphere_workspace-data-service)
 
 Workspace Data Service provides an API to manage records in Terra.
 
-Design docs and other references may be found [here](https://broadworkbench.atlassian.net/wiki/spaces/AS/pages/2437742606/Workspace+Data+Service).
+Design docs and other references may be
+found [here](https://broadworkbench.atlassian.net/wiki/spaces/AS/pages/2437742606/Workspace+Data+Service).
 
 ## Setup
-### Prerequisites
-Make sure you have Java 17 installed.  Our favorite way to do this is with [SDKMAN](https://sdkman.io/jdks).  Once it is installed , use `sdk list java` to see available versions, and `sdk install java 17.0.2-tem` to install, for example, the Temurin version of Java 17.
 
-[jenv](https://www.jenv.be/) is also helpful for managing active versions.  On a Mac,jenv can be installed and used like so:
+### Prerequisites
+
+Make sure you have Java 17 installed. Our favorite way to do this is
+with [SDKMAN](https://sdkman.io/jdks). Once it is installed , use `sdk list java` to see available
+versions, and `sdk install java 17.0.2-tem` to install, for example, the Temurin version of Java 17.
+
+[jenv](https://www.jenv.be/) is also helpful for managing active versions. On a Mac,jenv can be
+installed and used like so:
+
 ```
 brew install jenv
 # follow postinstall instructions to activate jenv...
@@ -24,7 +32,7 @@ brew install homebrew/cask-versions/temurin17
 jenv add /Library/Java/JavaVirtualMachines/temurin-17.jdk/Contents/Home
 ```
 
-We are using Postgres 14.  You do not need to have postgres installed on your system.
+We are using Postgres 14. You do not need to have postgres installed on your system.
 Instead use the `run_postgres.sh` script to set up a docker container running postgres (see below).
 
 #### Environment Variables
@@ -37,6 +45,7 @@ You may want to add them to your `~/.zshrc` or similar shell profile.
 
 WDS contacts Sam for permission checks. You will need to configure Sam's URL by setting an
 environment variable, such as:
+
 ```
 export SAM_URL=https://sam.dsde-dev.broadinstitute.org/
 ```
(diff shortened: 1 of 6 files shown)


README.md
davidangb, line 1: bah, since I made changes to this file, Spotless and/or IntelliJ decided to reformat a lot of it. Sorry for the noise!

service/src/main/java/org/databiosphere/workspacedataservice/controller/ImportController.java
```diff
@@ -0,0 +1,20 @@
...
+
+import java.util.UUID;
+import org.databiosphere.workspacedataservice.generated.api.ImportApi;
+import org.databiosphere.workspacedataservice.generated.model.ImportRequest;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class ImportController implements ImportApi {
```
davidangb, line 11: The `implements ImportApi` is the big deal here; `ImportApi` is generated code but is git-ignored so you don't see it in this PR.

README.md
jladieu: Oh weird, that's kind of a surprise...I was thinking it'd only run on `.java` files!  Was this annoying enough that you'd want to prevent it going forward?

service/build.gradle
```diff
@@ -95,6 +102,24 @@ dependencies {
...
 
+tasks.register('generateApiInterfaces', GenerateTask) {
+	generatorName = "spring"
+	library = "spring-boot"
+	inputSpec = "$rootDir/service/src/main/resources/static/swagger/openapi-docs.yaml"
+	outputDir = "$rootDir/service"
+	cleanupOutput = false // else would delete the /service directory
+	apiPackage = "org.databiosphere.workspacedataservice.generated.api"
+	modelPackage = "org.databiosphere.workspacedataservice.generated.model"
+	apiFilesConstrainedTo.set(["Import"]) // seems to respect the tag value instead of the filename
```
jladieu, line 113: I wonder if it'd be possible to instead _ignore_ the legacy stuff, so that we don't have to touch this configuration every time we add something new?

I the open API generator suggests configuring an `ignoreFileOverride` and I wonder if that would have enough flexibility to accomplish this?

service/src/main/java/org/databiosphere/workspacedataservice/controller/ImportController.java
```diff
@@ -0,0 +1,20 @@
...
+
+  @Override
+  public ResponseEntity<Void> importV1(UUID instanceUuid, ImportRequest importRequest) {
+    // TODO: implementation for imports
+    // return TOO_EARLY here as a proof-of-concept that we can override the implementation
+    // in ImportApi (which returns NOT_IMPLEMENTED)
+    return new ResponseEntity<>(HttpStatus.TOO_EARLY);
```
jladieu, line 18: TIL HTTP 425

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -328,6 +330,24 @@ paths:
...
+  ##############################
+  /{instanceUuid}/import/v1:
+    post:
+      summary: Import from a file
+      description: Imports records from the specified URL.
+      operationId: importV1
+      tags:
+        - Import
+      parameters:
+        - $ref: '#/components/parameters/instanceUuidPathParam'
```
jladieu, line 344: Would you consider inlining `instanceUuidPathParam` and `ImportRequestBody` here?

I think inlining might have a couple advantages:
- the definition for the method is all in one place and I don't have to bounce around the file
- don't have to come up with new names for everything, or have to think about and enforce a naming convention

Possible drawbacks:
- less reuse between method definitions
- plenty of stuff I'm not aware of implied by the current organization =)

jladieu (approved):

README.md
davidangb: Eh, I don't think we need to change anything, these should become less and less common as we find any stragglers

README.md
```diff
@@ -172,16 +227,45 @@ repositories {
...
+
+WDS's `v0.2` APIs are documented in `service/src/main/resources/static/swagger/openapi-docs.yaml`
+and
+their implementations are hand-coded.
+
+APIs at `v1` or later use code generation. To create or update an API, you must:
+
+1. Define the API using OpenAPI syntax
+   at `service/src/main/resources/static/swagger/openapi-docs.yaml`
+2. Update the `generateApiInterfaces` Gradle task in `service/build.gradle` as necessary.
```
calypsomatic, line 261: For dummies like me, can you be a bit more explicit what "as necessary" entails?

calypsomatic: Oh wait that's what the next two steps are.  NM.

calypsomatic (approved):

README.md
davidangb: In hindsight, these changes may also be because I (finally) configured my local IntelliJ to format on save!

davidangb:
@jladieu (and @calypsomatic ) - I reset your approvals because I added a pretty major update based on Josh's feedback. In this update:
* separate YAML files for the v0.2 and v1 API definitions, which allows us to remove ignores/includes from the `generateApiInterfaces` gradle tasks. In turn, this makes it more future-proof for when we start adding more v1 apis.
* add a "ServerModel" suffix to the generated models, to disambiguate these server models from similarly-named classes in the client generated code

service/src/main/resources/static/swagger/apis-v1.yaml
```diff
@@ -0,0 +1,79 @@
...
+  securitySchemes:
+    bearerAuth:
+      type: http
+      scheme: bearer
+
+  parameters:
+    # instanceUuidPathParam is very close to duplicating instanceIdPathParam from v0.2,
+    # but this one specifies format: uuid, which creates Java UUID objects in generated
+    # code instead of a String.
+    instanceUuidPathParam:
```
jladieu, line 71: Could `instanceUuidPathParam` be a candidate to inline as well (at least and only until we have another endpoint or two using it)?

calypsomatic (approved):

service/src/main/resources/static/swagger/apis-v1.yaml
davidangb: It is valid to inline, but as soon as we have another API it'll need to be extracted; I opted to leave it external as a proactive measure. If you feel strongly about it I can inline it!

service/build.gradle
```diff
@@ -1,122 +1,127 @@
...
+    generatorName = "spring"
+    library = "spring-boot"
+    inputSpec = "$rootDir/service/src/main/resources/static/swagger/apis-v1.yaml"
+    outputDir = "$rootDir/service"
+    cleanupOutput = false // else would delete the /service directory
+    apiPackage = "org.databiosphere.workspacedataservice.generated.api"
+    // apiNameSuffix does not work; see https://github.com/OpenAPITools/openapi-generator/issues/8822
+    // apiNameSuffix = "Server" // disambiguate from classes in the generated client
+    modelPackage = "org.databiosphere.workspacedataservice.generated.model"
+    modelNameSuffix = "ServerModel"   // disambiguate from classes in the generated client
```
jladieu, line 117: I think I understand what you're doing here but not entirely sure the exact implication of what we're disambiguating against.   Does every model object get a corresponding class generated in the client library?  Did you encounter collisions somehow?  I'm wondering where the client and server code would co-mingle in a way that this would be problematic.

Rather than adding a suffix (which I think will add some extra boilerplate broadly across all classnames in server code), I'm wondering if disambiguation could be achieved (or maybe already is?) using the package namespace?

service/src/main/resources/static/swagger/apis-v1.yaml
jladieu: I'd feel a lot less strongly about it if you tell me it's easy to wiggle around these configs after the fact and it won't have knock-on effects.  In which case, no biggie at all!

However, if it is a bit binding to start with this, I do think it's slightly more complicated than the blunt approach of just having it all in one spot, and we'll probably copy the patterns set up here when adding new stuff, so it could proliferate quickly.

I tend to like to defer factoring things for reuse until you have _at least_ two, (but sometimes even three) instances of duplication to justify that extra bump in complexity.

service/src/main/resources/static/swagger/apis-v1.yaml
```diff
@@ -0,0 +1,79 @@
+# This file contains only the v1 WDS APIs.
```
jladieu, line 1: I like it!

service/build.gradle
davidangb: Yes, every model in the API gets a corresponding class in the client library. With this PR, we are now generating duplicate models for the server.

We do already disambiguate by package. If we didn't specify a value for "modelNameSuffix", we would generate:
org.databiosphere.workspacedata.model.ImportRequest.ImportRequest (client)
org.databiosphere.workspacedataservice.generated.model.ImportRequest (server)

We have tricky dependencies, not least because the WDS server imports the WDS client library in order to connect to a remote WDS for cloning. So, both packages are in scope.

As a developer, I have multiple times run into a problem where I mistakenly imported the wrong object from the wrong package. I thought that by renaming our server models, I'd make it easier on us as developers to avoid that!

service/build.gradle
```diff
@@ -1,100 +1,130 @@
...
+    apiFilesConstrainedTo.set([""])   // empty string means generate all apis
+    modelFilesConstrainedTo.set([""]) // empty string means generate all models
+    supportingFilesConstrainedTo.set(["ApiUtil.java"])
```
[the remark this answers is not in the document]
davidangb: It's generated code, so it's git-ignored and not visible in the PR. It adds a "setExampleResponse" method which … we don't really need! I've removed ApiUtil.java from being generated.

service/build.gradle
jladieu: I do think the packages, as named, are quiet confusing if co-mingled in the same space, so if the packages are gonna stay the same, I definitely approve of adding the suffix.

But!  I think maybe we could have better package names.  I don't know how much of a can of worms this opens up with compatibility/churn on existing clients though, so please push back if this sounds like a pain.  Maybe it's a convention that we could adopt going forward with just this new thing?

// is it possible to drop the additional `ImportRequest.ImportRequest` and add `client` to the namespace?
org.databiosphere.workspacedata.client.ImportRequest (client)

// I care less about this because it's internal and we can in theory wiggle this around as much as we want
org.databiosphere.workspacedataservice.generated.ImportRequest (server)

davidangb: Oh whoops, copy/paste error, the client class doesn't have a duplicate "ImportRequest", it's just org.databiosphere.workspacedata.model.ImportRequest.

I consider it a breaking change to rename any of the client classes, so I'm really hesitant to do that, at least without a bunch of warnings and deprecation flags.

I am totally fine collapsing `o.d.w.generated.api` and `o.d.w.generated.api` into just `o.d.w.generated`. Here's the current state:
Screenshot 2023-10-03 at 11 40 26 AM

Would you like to see those two packages collapsed into one?

jladieu: I think that might be an improvement over having `model` in the namespace on the server-side code, which makes it look even more confusingly similar to the client namespace.  SGTM!

davidangb: done in `63bbd5ba`

jladieu (approved):
One last thought...after thinking a lot about the client code namespace...

`can/should` we add the apiVersion to the namespace of the generated client code going forward?  I could imagine client code wanting to access both the old and new versions of a thing during functional migrations, and having that disambiguating element may be helpful.

This is also unintentionally a backdoor way of convincing you to change the namespace to include `client` since you'd be changing it anyway 😆 

Approved as is either way!

service/src/main/resources/static/swagger/apis-v1.yaml
davidangb: It is very easy to wiggle these around after the fact :)

PR #350, opened 2023-10-02
Tidy up setup and teardown methods.

jladieu:
Tidy up `setUp` and `tearDown` methods and improve test isolation.

* Rename generally to `setUp` and `tearDown`.  This is just for consistency.
* Where possible, prefer `@BeforeEach` and `@AfterEach` over `@BeforeAll` and `@AfterAll`.  This is to improve test isolation within test suites.
* Prefer to do cleanup in `tearDown` rather than pre-emptively in `setUp`.  Tests should be responsible for cleaning up after themselves, rather than other tests.
* Drop the `static` modifier from `private` but non-`final` instance variables in test cases.  This is to make each test manage its own instance state rather than share across invocations within the test file.

### DIFF
--- a/service/src/test/java/org/databiosphere/workspacedataservice/controller/FullStackRecordControllerTest.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/controller/FullStackRecordControllerTest.java
@@ -31,7 +31,6 @@ import org.databiosphere.workspacedataservice.shared.model.RecordType;
 import org.databiosphere.workspacedataservice.shared.model.SearchRequest;
 import org.databiosphere.workspacedataservice.shared.model.SortDirection;
 import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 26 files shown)


nawatts (approved):

service/src/test/java/org/databiosphere/workspacedataservice/controller/FullStackRecordControllerTest.java
```diff
@@ -62,20 +61,17 @@
...
 class FullStackRecordControllerTest {
   @Autowired private TestRestTemplate restTemplate;
-  private static HttpHeaders headers;
-  private static UUID instanceId;
+  private HttpHeaders headers;
+  private UUID instanceId;
   private static final String versionId = "v0.2";
   @Autowired private ObjectMapper mapper;
 
-  @BeforeAll
```
ashanhol, line 70: Was there a particular reason All was used here instead of Each? It looks like there's plenty of tests that do use Each so I'm wondering if there was some logic for this here

ashanhol (approved):

service/src/test/java/org/databiosphere/workspacedataservice/controller/FullStackRecordControllerTest.java
jladieu: I couldn't convince myself there was a good reason for it, and in general recommend not using `@BeforeAll` unless it's for a very specific reason (with an accompanying explanatory comment ;)

PR #352, opened 2023-10-03
[AJ-1243] Introduce Lock in Cloning

ashanhol:
Currently cloning logic is not replica safe. This is due to the fact that currently if one replica initiates clone mode, another replica will see that and continue on to initialize the default schema. Problem is, cloning expects the default schema to not be created yet. 
This PR takes the first step in cloning replica safety by introducing a Lock that will prevent more than one replica from initiating cloning. Follow up work (in ticket AJ-1363) will refactor the initialization logic so a second replica won't try and create the default schema if cloning is happening. This PR **does not** mean cloning is replica safe yet. 

----------------
Reminder:

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
@@ -1,6 +1,8 @@
 package org.databiosphere.workspacedataservice;
 
 import java.util.UUID;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Lock;
 import org.apache.commons.lang3.StringUtils;
 import org.databiosphere.workspacedata.client.ApiException;
 import org.databiosphere.workspacedataservice.dao.CloneDao;
@@ -20,6 +22,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.event.ContextRefreshedEvent;
 import org.springframework.dao.DataAccessException;
+import org.springframework.integration.support.locks.LockRegistry;
 
 public class InstanceInitializerBean {
 
@@ -27,6 +30,7 @@ public class InstanceInitializerBean {
   private final LeonardoDao leoDao;
   private final WorkspaceDataServiceDao wdsDao;
   private final CloneDao cloneDao;
+  private final LockRegistry lockRegistry;
 
   private final BackupRestoreService restoreService;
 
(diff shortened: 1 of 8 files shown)


davidangb (approved):
thank you for extracting this out!

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -187,6 +201,10 @@ private boolean initCloneMode() {
       var finalCloneStatus = currentCloneStatus(trackingId);
       return finalCloneStatus.getStatus().equals(JobStatus.SUCCEEDED);
 
+    } catch (InterruptedException e) {
+      LOGGER.error("Error with aquiring cloning/schema initialization Lock: {}", e.getMessage());
```
jladieu, line 205: typonit: aquiring -> acquiring :)

Also, does this catch clause actually need to wrap _all_ this code at this point, or can it be tightened up to just be around the `tryLock` call now?   Just curious what else throws `InterruptedException` out of this stanza.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -198,6 +216,8 @@ private boolean initCloneMode() {
             inner.toString());
       }
       return false;
+    } finally {
```
jladieu, line 219: Noice!  _Finally_ got that `finally` in there

service/src/test/java/org/databiosphere/workspacedataservice/InstanceInitializerNoWorkspaceIdTest.java
```diff
@@ -50,9 +57,17 @@
...
   @Autowired InstanceInitializerBean instanceInitializerBean;
-
+  @MockBean JdbcLockRegistry registry;
   @SpyBean InstanceDao instanceDao;
 
+  Lock mockLock = mock(Lock.class);
+
+  @BeforeEach
+  void setUp() {
+    when(mockLock.tryLock()).thenReturn(true);
```
jladieu, line 67: ... so little code 

Glad to see this return as a mock

jladieu (commented):
One comment/question about the scope of the `try { } catch { }` but otherwise GTG!

jladieu (approved):

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: ty, will fix typo 

you're right, i can put the lock in an inner try/catch, has to be an inner one though cause the `finally` has to release after the rest of the clonings done

PR #353, opened 2023-10-04
AJ-1011: Quartz integration, shell of async snapshot API

davidangb:
In this PR:
* Add a new v1 API to retrieve import status to complement the create-import API from #349
* Add the Quartz starter dependency for Spring Boot
* Create the Quartz db tables via liquibase. This makes table creation multi-replica safe, and allows us to create the Quartz tables in the `sys_wds` schema.
* Create a new `sys_wds.import` db table, with defaults and triggers to automatically populate created and updated timestamps (looking for feedback on this approach)
* Create a simple "ImportQuartzJob" as a spike/proxy for the real PFB or snapshot import. This job sleeps for 5-15 seconds and then calls itself done
* Enable "ImportQuartzJob" jobs to be created via API, run in Quartz, and then retrieved via API

Still TODO, _lots_ of stuff!:
* permission checking and validation
* error handling and persisting error messages to the db so they can be retrieved when viewing failed jobs
* implementation of the data imports itself
* code cleanup and polish
* tests
* probably renaming `o.d.w.shared.model.job.Job` to avoid conflict with Quartz's `Job` class

Open Questions:
1. I don't like how the generated `ImportJobStatusServerModel` class, and the `ImportStatusResponse` that extends it, duplicate the `o.d.w.shared.model.job.Job` and `o.d.w.shared.model.job.JobStatus` classes. I'd like to keep the auto-generated classes; open to any ideas on how to align these.
2. I'm using `hashids` to generate shorter ids for import jobs, instead of using a UUID (might wanna upgrade to https://sqids.org/). How do you feel about this instead of using UUIDs?


service/build.gradle
```diff
@@ -119,6 +120,7 @@ tasks.register('generateApiInterfaces', GenerateTask) {
     apiFilesConstrainedTo.set([""])   // empty string means generate all apis
     modelFilesConstrainedTo.set([""]) // empty string means generate all models
     supportingFilesConstrainedTo.set([]) // empty array means generate none
+    skipOperationExample = true // example responses require the ApiUtil.java supporting file
```
davidangb, line 123: I realized we need this flag as I started to use the auto-generated api interfaces in earnest

service/src/main/resources/application.yml
```diff
@@ -78,6 +106,7 @@ server:
   error:
     include-stacktrace: never
     include-message: always
+  port: 8080
```
davidangb, line 109: not necessary as part of this PR, but I found I kept adding this (and changing the default) to test multi-replica behavior so I figured I'd put it here to make it easier to change

service/src/main/resources/static/swagger/apis-v1.yaml
```diff
@@ -77,3 +107,29 @@ components:
         type: string
         format: uuid
 
+  schemas:
+    ImportJobStatus:
```
jladieu, line 111: While working on the TDR snapshot stuff, I noticed that the TDR API just has a generic top level `jobs` endpoint:

image

What you're specifying here looks pretty generic, and I wonder if it'd lend itself to a generalizable implementation, rather than crafting a handful of APIs _for each_ category of job.

I vaguely recall in a past life working with systems who followed a similar model...where all the async jobs were piped through a generic polling API, but the only difference is they were called `operations` :)

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresImportDao.java
```diff
@@ -0,0 +1,99 @@
...
+            "Error serializing options to jsonb for import job {}: {}", jobId, e.getMessage());
+      }
+    }
+
+    // insert the import request to the db. Note that the created and updated
+    // columns in the db are automatically handled by Postgres.
+    namedTemplate
+        .getJdbcTemplate()
+        .update(
+            "insert into sys_wds.import(id, type, status, url, options) "
```
yuliadub, line 56: does created/updated auto default to a now timestamp at creation?

service/src/main/java/org/databiosphere/workspacedataservice/controller/ImportController.java
```diff
@@ -1,20 +1,54 @@
...
+    // TODO: validate importRequest, e.g. does it contain a valid URL
+
+    // save this import request to the WDS db and queue up its async execution in Quartz
+    ImportStatusResponse jobStatus = importService.queueJob(importRequest);
+
+    return new ResponseEntity<>(jobStatus, HttpStatus.ACCEPTED);
+  }
+
+  @Override
+  public ResponseEntity<ImportJobStatusServerModel> importStatusV1(
```
calypsomatic, line 37: I find it pretty confusing how the `ImportJobStatusServerModel` only exists in non-tracked generated code.  Is this inevitable or standard practice?  I know the interface is generated but can/should we write our own models?

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresImportDao.java
```diff
@@ -0,0 +1,99 @@
...
+  @Override
+  @ReadTransaction
+  public ImportStatusResponse getImport(String jobId) {
+    return namedTemplate.queryForObject(
+        "select id, type, status, url, created, updated, "
+            + "options, result, error, stacktrace "
+            + "from sys_wds.import "
+            + "where id = :jobId",
+        new MapSqlParameterSource("jobId", jobId),
+        new AsyncJobRowMapper());
```
yuliadub, line 84: async job mapper is missing some of the mapping columns in the insert - dont we want to include all of them in the mapper?

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresImportDao.java
yuliadub: answered my own question lower down :)

service/src/main/resources/static/swagger/apis-v1.yaml
yuliadub: +1 for generic jobs endpoint for all jobs

jladieu:
> Could we have a generic quartz table & API that handles all jobs, and reduce the work of standing up a new job to implementing the backing business logic only?

service/src/main/java/org/databiosphere/workspacedataservice/controller/ImportController.java
yuliadub: I also find this confusing. is there any way we can keep our Job classes and use them instead? or merge them and actually have them as code checked into the repo

davidangb: I see a few paths forward here … 
1. current state; interfaces and models are generated and git-ignored, you find this confusing
2. continue generating code, but track the results in git so we can see any changes. Will require removing generation timestamps from the generated code so we don't constantly have updates. Potential optimization here to only re-generate code if the YAML file has actually changed.
3. your suggestion: generate interfaces but don't generate the models used by those interfaces. My fear here is that the YAML and the hand-coded models can get out of sync; we lose the benefit of generation.
4. don't generate anything for the server side; revert to how we handled the v0.2 APIs

Do any of these sound right?

calypsomatic: For #2 - do we need the timestamps?  Where do they appear?

service/src/main/resources/static/swagger/apis-v1.yaml
davidangb: I had this same debate in my mind, too. I worry about creating an API that is generic enough to support all our async operations, including ones we would need in the future. The TDR API has dedicated APIs for _creating_ jobs and then a generic API for _monitoring_ them - is that what you're suggesting?

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresImportDao.java
davidangb: This insert statement omits `created` and `updated`, which are managed by Postgres; it also omits `errorMessage`, `result`, and `stacktrace`, all of which should be null upon first insert.

I definitely have TODOs to update a job with `errorMessage`, `result`, and `stacktrace` as appropriate, post-insert!

service/src/main/java/org/databiosphere/workspacedataservice/controller/ImportController.java
davidangb: @yuliadub you had a similar comment, would love your opinion on this^^ list too

davidangb:
On hashids vs uuids: no technical reason, I find their terseness easier to read. I find a URL of `jobs/v1/SumQjUk8` preferable to `jobs/v1/4bfde0f1-a53a-4c40-a7d3-fde859e016e9`. I will probably just roll back to using UUIDs.

On genericizing a solution: I think a solution with dedicated APIs for creating jobs (i.e. keep both `POST /{instanceUuid}/import/v1` and `POST /backup/{v}`) and a generic backend table and generic API to monitor jobs would work. We'd still have the type safety of dedicated inbound requests, but gain reuse in the back end. I can work on implementing that.

service/src/main/java/org/databiosphere/workspacedataservice/controller/ImportController.java
davidangb: > For https://github.com/DataBiosphere/terra-workspace-data-service/pull/2 - do we need the timestamps? Where do they appear?

By default, openapi-generator creates classes that look like:

```java
@Generated(value = "org.openapitools.codegen.languages.SpringCodegen", date = "2023-10-05T08:58:38.760147-04:00[America/New_York]")
public class ImportJobStatusServerModel {
```

which is kinda nice because we can see at a glance when they were generated, but I bet we could do without them.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresImportDao.java
davidangb: oh sorry, I misunderstood your first comment - yeah, I have outstanding work to update the rowmapper to include other fields, I've updated to add a TODO comment

jladieu:
> We'd still have the type safety of dedicated inbound requests, but gain reuse in the back end. I can work on implementing that.

Just to push on this a bit more...I completely agree with wanting some type safety on the way in, but I wonder if we don't actually need to specify a dedicated job 'intake' pattern at all.  Couldn't creation of a job just be a side effect of a more general-purpose `POST /widget` or `DELETE /widget`?

Where, if widget can't be created synchronously with a pointer to the newly created resource returned in the response, we instead return 202 (ACCEPTED) with details needed to poll the /jobs endpoint?

Then, only the backend implementing code needs to know how to enqueue stuff to the async pipeline, and the client code only needs to know how to monitor for completion and doesn't need to be concerned with knowing whether to create something as synchronous vs. async?

I imagine we'd end up with a set of restful resources, some of which are sync and some of which are not, but all largely still following the same set of HTTP verbs.

jladieu:
(A more complicated possible use case is unlocked by the aforementioned suggestion, which I'm not advocating for here, but which I think is useful as a thought exercise):

What if a given RESTful endpoint could be both synchronous or asynchronous?   For small requests, a POST returns synchronously, for big requests, the POST returns 202/pointer to job -- if clients could get used to this pattern you could unlock a whole host of fast vs. slow features.

I could see this coming in to play as we migrate WDS the LZ, and making the Workspace creation synchronous vs. async/poll.  It'd be fancy if the UI was able to gracefully switch from the poll model to the sync model, wouldn't it?

davidangb:
> Just to push on this a bit more...I completely agree with wanting some type safety on the way in, but I wonder if we don't actually need to specify a dedicated job 'intake' pattern at all. Couldn't creation of a job just be a side effect of a more general-purpose POST /widget or DELETE /widget?

It sounds like we're in agreement, that's exactly what I was trying to describe. To make it concrete, I'm picturing we have:
* `POST /{instanceUuid}/import/v1` to create a data import, which accepts a data-import-specific request payload, enqueues an async job under the covers, and returns a jobId to the caller
* `GET /{instanceUuid}/job/v1/{jobId}` to poll for status of any job, including data-import jobs

I like the concept of our create/delete APIs returning synchronously when possible - but I'm unlikely to implement that for this MVP, that could be follow-on work

jladieu:
Just leaving a breadcrumb trail to the (now old, but still prescient) design [doc](https://docs.google.com/document/d/1PTd4xvmV9xnEkWaIgFc6d3VUyJwwymQTFdhpKrdsUKw/edit) @davidangb shared for future discoverability.

jladieu:
A possible bookkeeping suggestion as part of landing this PR...given some of the conversation around this PR hinges on what's per-job vs. what's initial setup, it might be worth carving out a discrete PR that gets the one-time infrastructure in place (eg all the quartz stuff, config, etc.), and pulling the import-specific stuff out as a separate PR.  This would leave a nice reference implementation that we could follow for future endpoints.

davidangb:
Closing; this has been implemented instead by #357, #358, #361, #367

PR #354, opened 2023-10-05
AJ-1011: generated server

davidangb:
Based on feedback from #353, this PR adds our generated server-side code to git, so we have it in the codebase and we can track changes to it.

In this PR, the following files were autogenerated (so you can choose to skip review of them if desired):
* ImportApi
* ImportRequestServerModel


jladieu:
I wonder if we should mark the files as generated (rather than ignored) so GitHub can suppress them on initial render, but make them available for optional review, as described in https://medium.com/@clarkbw/managing-generated-files-in-github-1f1989c09dfd and https://medium.com/@clarkbw/managing-generated-files-in-github-1f1989c09dfd

Note...I've never done this, so no idea how usable the outcome would be.

calypsomatic (approved):

service/build.gradle
```diff
@@ -120,8 +120,9 @@ tasks.register('generateApiInterfaces', GenerateTask) {
     modelFilesConstrainedTo.set([""]) // empty string means generate all models
     supportingFilesConstrainedTo.set([]) // empty array means generate none
     configOptions.set([
-        interfaceOnly: "true", // Java interfaces only; no controllers or implementation classes
-        useTags      : "true" // use the OpenAPI tag to classify apis, not the first segment of the api path
+        interfaceOnly          : "true", // Java interfaces only; no controllers or implementation classes
+        useTags                : "true", // use the OpenAPI tag to classify apis, not the first segment of the api path
+        hideGenerationTimestamp: "true" // omit timestamp from generated code, so we can track it via git
```
jladieu, line 125: What's the impact of toggling this to true?  I'm assuming a timestamp would get bumped _on every PR_ without this?

jladieu: (If not, as in, if it only bumps timestamp _when there is a change_ I almost think having a timestamp get bumped in generated code might be just fine...)

davidangb: I am having a hard time nailing down when that timestamp gets updated. I am sure I saw it getting updated when I didn't make a change, but now I can't reproduce that; it's only updating when the contents themselves change. So, I am putting the timestamp back in.

davidangb:
@jladieu I've updated to 1) add back the timestamp, and 2) mark those files as generated in `.gitattributes`, good call on that.

.gitattributes
```diff
@@ -0,0 +1,2 @@
+service/src/main/java/org/databiosphere/workspacedataservice/generated/*          linguist-generated=true
+service/src/main/java/org/databiosphere/workspacedataservice/generated/README.md  linguist-generated=false
```
davidangb, line 2: note this will only affect github display; local `git diff` will still show changes. That's my preference - I want to be aware of when and how they changed before I commit. But open to feedback.

jladieu (approved):
Cool!  It even auto-collapsed the generated stuff in this PR.  Thanks for trying that out!

service/build.gradle
jladieu: Put up a PR for this: https://github.com/DataBiosphere/terra-workspace-data-service/pull/365

PR #355, opened 2023-10-06
AJ-1590: cluster-aware and restart-tolerant Quartz in WDS

davidangb:
This switches Quartz to be cluster-aware and restart-tolerant by using Postgres as its backing store instead of holding jobs in memory. This requires:
* Create the Quartz db tables via liquibase. This makes table creation multi-replica safe, and allows us to create the Quartz tables in the `sys_wds` schema.
* Configure Quartz, including being multi-replica aware, via `application.yml`


service/src/main/resources/liquibase/changesets/20231003_quartz_tables.sql
```diff
@@ -0,0 +1,215 @@
+-- this SQL copied from the Quartz library at org/quartz/impl/jdbcjobstore/tables_postgres.sql
+-- and then modified by David An to:
+--     * not drop tables before creating tables
+--     * create tables within the sys_wds schema instead of the public schema
+--     * separate out the CREATE and DROP statements to allow Liquibase rollback
```
jladieu, line 5: Can you clarify the third comment by indicating that the DROP statements are now in a separate file?

service/src/main/resources/liquibase/changesets/20231003_quartz_tables.sql
```diff
@@ -0,0 +1,215 @@
...
+
+
+
+-- Thanks to Patrick Lightbody for submitting this...
+--
+-- In your Quartz properties file, you'll need to set
+-- org.quartz.jobStore.driverDelegateClass = org.quartz.impl.jdbcjobstore.PostgreSQLDelegate
+
+-- DROP TABLE IF EXISTS QRTZ_FIRED_TRIGGERS;
```
jladieu, line 14: Maybe just eliminate this commented stuff and include a pointer to the DROP file?

calypsomatic (approved):

service/src/main/resources/liquibase/changesets/20231003_quartz_tables_rollback.sql
```diff
@@ -0,0 +1,27 @@
...
+--     * separate out the CREATE and DROP statements to allow Liquibase rollback
+
+
+
+-- Thanks to Patrick Lightbody for submitting this...
+--
+-- In your Quartz properties file, you'll need to set
+-- org.quartz.jobStore.driverDelegateClass = org.quartz.impl.jdbcjobstore.PostgreSQLDelegate
+
+DROP TABLE IF EXISTS sys_wds.QRTZ_FIRED_TRIGGERS;
```
jladieu, line 14: Supernit: consider reordering these so that they are the perfect _reverse_ of the `CREATE TABLE` statements, that way the rollback is closer to being pure in its ordering to exactly resetting things the way they were.  I don't know if this will ever matter, but I feel like maybe establishing this pattern makes it easier to troubleshoot failed rollbacks in the future.

jladieu (approved):
Approved with small comments, thanks for splitting this out!

yuliadub (approved):

service/src/main/resources/liquibase/changesets/20231003_quartz_tables_rollback.sql
davidangb: The order as written here is the required order, due to foreign key references. For instance, it's necessary to drop `QRTZ_BLOB_TRIGGERS` before  `QRTZ_TRIGGERS` because the former has a foreign key to the latter. Plus, since this is a copy-paste from the sql that Quartz ships, I'd like to modify it as little as possible.

service/src/main/resources/liquibase/changesets/20231003_quartz_tables.sql
davidangb: updated!

service/src/main/resources/liquibase/changesets/20231003_quartz_tables.sql
davidangb: service/src/main/resources/liquibase/changesets/20231003_quartz_tables_rollback.sql

jladieu (approved):

service/src/main/resources/application.yml
```diff
@@ -63,6 +63,33 @@ spring:
...
+        jobStore:
+          driverDelegateClass: org.quartz.impl.jdbcjobstore.PostgreSQLDelegate
+          # support multiple horizontal WDS replicas all using the same quartz tables
+          isClustered: true
+          # Set the frequency (in milliseconds) at which this instance “checks-in”* with the other instances of the cluster.
+          # Affects the quickness of detecting failed instances
+          # default is 15000; we may want to change this later
+          # clusterCheckinInterval: 15000
+          # look for Quartz tables in the sys_wds schema. This requires the tables to be created by
+          # Liquibase and custo mized to be inside sys_wds; Quartz does not support creating the
```
ashanhol, line 83: nit: remove space in "customized"

ashanhol (approved):

PR #356, opened 2023-10-09
[ AJ-1363] Make Cloning/startup multi-replica safe

ashanhol:
This PR makes cloning replica safe by refactoring our initialization logic to follow the flowchart below:

image

Ultimately, in code `initCloneMode()` returning false will have us create the default schema and returning true has us "do nothing" (ie just start WDS). 

Here is what the logs will look like. 
**Non-Cloning Replica**: 
```
2023-10-11 18:40:51.780  INFO [] [main] o.d.w.InstanceInitializerBean            : Default workspace id loaded as e28ba126-6d7a-4046-993f-a0e73512078b.
2023-10-11 18:40:51.785  INFO [] [main] o.d.w.InstanceInitializerBean            : SourceWorkspaceId found, checking validity
2023-10-11 18:40:51.785  INFO [] [main] o.d.w.InstanceInitializerBean            : Cloning mode enabled, attempting to clone from 5d8e8691-75a1-4c76-99ad-8263d29c58c1 into e28ba126-6d7a-4046-993f-a0e73512078b.
2023-10-11 18:40:51.785  INFO [] [main] o.d.w.InstanceInitializerBean            : Starting in clone mode...
2023-10-11 18:40:51.897  INFO [] [main] o.d.w.InstanceInitializerBean            : clone entry in database: true
2023-10-11 18:40:51.906  INFO [] [main] o.d.w.InstanceInitializerBean            : Previous clone entry found. Instance schema exists: true.
2023-10-11 18:40:51.912  INFO [] [main] o.d.w.InstanceInitializerBean            : Cloning complete.
2023-10-11 18:40:51.918  INFO [] [main] o.d.w.WorkspaceDataServiceApplication    : Started WorkspaceDataServiceApplication in 76.778 seconds (JVM running for 95.266)
2023-10-11 18:40:51.985  INFO [] [main] o.d.w.logging.AvailabilityListener       : LivenessState: CORRECT
2023-10-11 18:40:51.987  INFO [] [main] o.d.w.logging.AvailabilityListener       : ReadinessState: ACCEPTING_TRAFFIC
```
**Cloning Replica** (filtered): 
```
2023-10-11 18:40:25.379  INFO [] [main] o.d.w.InstanceInitializerBean            : Default workspace id loaded as e28ba126-6d7a-4046-993f-a0e73512078b.
2023-10-11 18:40:25.394  INFO [] [main] o.d.w.InstanceInitializerBean            : SourceWorkspaceId found, checking validity
2023-10-11 18:40:25.394  INFO [] [main] o.d.w.InstanceInitializerBean            : Cloning mode enabled, attempting to clone from 5d8e8691-75a1-4c76-99ad-8263d29c58c1 into e28ba126-6d7a-4046-993f-a0e73512078b.
2023-10-11 18:40:25.394  INFO [] [main] o.d.w.InstanceInitializerBean            : Starting in clone mode...
2023-10-11 18:40:25.592  INFO [] [main] o.d.w.InstanceInitializerBean            : clone entry in database: false
2023-10-11 18:40:25.592  INFO [] [main] o.d.w.InstanceInitializerBean            : Creating entry to track cloning process.
2023-10-11 18:40:26.794  INFO [] [main] o.d.w.InstanceInitializerBean            : Retrieved source wds endpoint url https://lz73d73b0c51c86d3f3f84cf157194cb1167472d4e243f3a56.servicebus.windows.net/wds-5d8e8691-75a1-4c76-99ad-8263d29c58c1-5d8e8691-75a1-4c76-99ad-8263d29c58c1/
2023-10-11 18:40:26.802  INFO [] [main] o.d.w.InstanceInitializerBean            : Requesting a backup file from the remote source WDS in workspace 5d8e8691-75a1-4c76-99ad-8263d29c58c1
2023-10-11 18:40:26.830  INFO [] [main] o.d.w.dao.PostgresCloneDao               : Clone status is now BACKUPQUEUED.
2023-10-11 18:40:28.183  INFO [] [main] o.d.w.dao.PostgresCloneDao               : Clone status is now BACKUPSUCCEEDED.
2023-10-11 18:40:28.195  INFO [] [main] o.d.w.InstanceInitializerBean            : Re-checking clone job status after backup request
2023-10-11 18:40:28.213  INFO [] [main] o.d.w.InstanceInitializerBean            : Restore from the following path on the source workspace storage container: wdsservice/cloning/backup/5d8e8691-75a1-4c76-99ad-8263d29c58c1-2023-10-11_18-40-27.sql
2023-10-11 18:40:28.235  INFO [] [main] o.d.w.dao.PostgresCloneDao               : Clone status is now RESTOREQUEUED.
2023-10-11 18:40:28.243  INFO [] [main] o.d.w.service.BackupRestoreService       : Starting restore.
2023-10-11 18:40:28.271  INFO [] [main] o.d.w.dao.AbstractBackupRestoreDao       : Restore job abdd42fd-e033-418d-a907-6ac35732ab6c is now QUEUED
2023-10-11 18:40:28.707  INFO [] [main] o.d.w.dao.AbstractBackupRestoreDao       : Job abdd42fd-e033-418d-a907-6ac35732ab6c is now RUNNING
2023-10-11 18:40:28.779  INFO [] [main] o.d.w.service.BackupRestoreService       : Grabbing data from the backup file.
2023-10-11 18:40:32.311  INFO [] [main] o.d.w.dao.AbstractBackupRestoreDao       : Job abdd42fd-e033-418d-a907-6ac35732ab6c is now SUCCEEDED
2023-10-11 18:40:32.691  INFO [] [main] o.d.w.InstanceInitializerBean            : Restore Successful
2023-10-11 18:40:32.703  INFO [] [main] o.d.w.dao.PostgresCloneDao               : Clone status is now RESTORESUCCEEDED.
2023-10-11 18:40:32.711  INFO [] [main] o.d.w.InstanceInitializerBean            : Re-checking clone job status after restore request
2023-10-11 18:40:32.723  INFO [] [main] o.d.w.InstanceInitializerBean            : Cloning complete.
2023-10-11 18:40:32.729  INFO [] [main] o.d.w.WorkspaceDataServiceApplication    : Started WorkspaceDataServiceApplication in 60.199 seconds (JVM running for 77.275)
```

-----------------------
Reminder:

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
@@ -78,63 +78,71 @@ public class InstanceInitializerBean {
    */
   public void initializeInstance() {
     LOGGER.info("Default workspace id loaded as {}.", workspaceId);
-    boolean isInCloneMode = isInCloneMode(sourceWorkspaceId);
-    LOGGER.info("isInCloneMode={}.", isInCloneMode);
-    if (isInCloneMode) {
+    // Enter clone mode if sourceWorkspaceId is specified
+    if (StringUtils.isNotBlank(sourceWorkspaceId)) {
       LOGGER.info("Source workspace id loaded as {}.", sourceWorkspaceId);
-      boolean cloneSuccess = initCloneMode();
-      if (cloneSuccess) {
-        LOGGER.info("Cloning complete.");
+      boolean cloneExitSuccessfuly = initCloneMode();
+      if (cloneExitSuccessfuly) {
+        LOGGER.info("Exited cloning.");
       } else {
+        LOGGER.info("Cloning failed, initializing default schema.");
         initializeDefaultInstance();
       }
     } else {
+      LOGGER.info("No SourceWorkspaceId found, initializing default schema.");
       initializeDefaultInstance();
     }
   }
 
   /**
-   * Determine if this WDS is starting as a clone of some other WDS. If a valid {@code
-   * SOURCE_WORKSPACE_ID} env var is provided to this WDS, it will start in clone mode.
+   * Determine whether we're in the midst of a clone already or if a clone has completed.
    *
    * @param sourceWorkspaceId value of {@code SOURCE_WORKSPACE_ID}; provided as an argument to
    *     assist with unit tests.
-   * @return whether this WDS is a clone.
+   * @return true if we should not proceed with cloning.
    */
-  protected boolean isInCloneMode(String sourceWorkspaceId) {
-    if (StringUtils.isNotBlank(sourceWorkspaceId)) {
-      LOGGER.info("SourceWorkspaceId found, checking database");
-      try {
-        UUID.fromString(sourceWorkspaceId);
-      } catch (IllegalArgumentException e) {
-        LOGGER.warn(
-            "SourceWorkspaceId could not be parsed, unable to clone DB. Provided SourceWorkspaceId: {}.",
-            sourceWorkspaceId);
-        return false;
-      }
+  protected boolean cloningStartedOrCompleted(UUID sourceWorkspaceId) {
+    LOGGER.info("Checking database for cloning and existing instances");
+    // is a clone operation already running? This can happen when WDS is running with multiple
+    // replicas,
+    // and another replica started first and has initiated the clone. It can also happen in a
+    // corner case
+    // where this replica restarted during the clone operation.
+    boolean cloneAlreadyRunning = cloneDao.cloneExistsForWorkspace(sourceWorkspaceId);
+    // does the default pg schema already exist for this workspace? This should not happen,
+    // but we want to protect against overwriting it.
+    boolean instanceAlreadyExists = instanceDao.instanceSchemaExists(UUID.fromString(workspaceId));
+    LOGGER.info(
+        "isInCloneMode(): cloneAlreadyRunning={}, instanceAlreadyExists={}",
+        cloneAlreadyRunning,
+        instanceAlreadyExists);
+    // TODO handle the case where a clone already ran, but failed; should we retry?
+    return cloneAlreadyRunning || instanceAlreadyExists;
+  }
 
-      if (sourceWorkspaceId.equals(workspaceId)) {
-        LOGGER.warn("SourceWorkspaceId and current WorkspaceId can't be the same.");
-        return false;
-      }
-
-      try {
-        // does the default pg schema already exist for this workspace? This should not happen,
-        // but we want to protect against overwriting it.
-        boolean instanceAlreadyExists =
-            instanceDao.instanceSchemaExists(UUID.fromString(workspaceId));
-        LOGGER.info("isInCloneMode(): instanceAlreadyExists={}", instanceAlreadyExists);
-        // TODO handle the case where a clone already ran, but failed; should we retry?
-        return !instanceAlreadyExists;
-      } catch (IllegalArgumentException e) {
-        LOGGER.warn(
-            "WorkspaceId could not be parsed, unable to clone DB. Provided default WorkspaceId: {}.",
-            workspaceId);
-        return false;
-      }
+  /**
+   * Determine if the arguments needed for cloning (in this case, a valid {@code
+   * SOURCE_WORKSPACE_ID} env var) allow us to proceed with cloning.
+   *
+   * @param sourceWorkspaceId value of {@code SOURCE_WORKSPACE_ID}; provided as an argument to
+   *     assist with unit tests.
+   * @return whether {@code SOURCE_WORKSPACE_ID} is valid.
+   */
+  protected boolean cloningArgumentsSafe(String sourceWorkspaceId) {
+    try {
+      UUID.fromString(sourceWorkspaceId);
+    } catch (IllegalArgumentException e) {
+      LOGGER.warn(
+          "SourceWorkspaceId could not be parsed, unable to clone DB. Provided SourceWorkspaceId: {}.",
+          sourceWorkspaceId);
+      return false;
     }
-    LOGGER.info("No SourceWorkspaceId found, initializing default schema.");
-    return false;
+
+    if (sourceWorkspaceId.equals(workspaceId)) {
+      LOGGER.warn("SourceWorkspaceId and current WorkspaceId can't be the same.");
+      return false;
+    }
+    return true;
   }
 
   /*
@@ -142,29 +150,28 @@ public class InstanceInitializerBean {
   a newly created (destination) workspace. WDS at start up will always have a current WorkspaceId, which in the
   context of cloning will effectively be the destination. The SourceWorkspaceId will only be populated if the currently
   starting WDS was initiated via a clone operation and will contain the WorkspaceId of the original workspace where the cloning
-  was triggered.
+  was triggered. initCloneMode() returns false only if we want to initialize a default schema.
   */
   private boolean initCloneMode() {
     LOGGER.info("Starting in clone mode...");
     UUID trackingId = UUID.randomUUID();
     Lock lock = lockRegistry.obtain(sourceWorkspaceId);
+    // Check if we have a clone safe sourceWorkspaceId. Create default schema if not
+    if (!cloningArgumentsSafe(sourceWorkspaceId)) {
+      return false;
+    }
+
     try {
       // Make sure it's safe to start cloning (in case another replica is trying to clone)
       boolean lockAquired = lock.tryLock(1, TimeUnit.SECONDS);
       if (!lockAquired) {
-        LOGGER.info("Failed to acquire lock in initCloneMode");
-        return false;
+        LOGGER.info("Failed to acquire lock in initCloneMode. Exiting clone mode.");
+        return true;
       }
 
-      // is a clone operation already running? This can happen when WDS is running with
-      // multiple replicas and another replica started first and has initiated the clone.
-      // It can also happen in a corner case where this replica restarted during the clone
-      // operation.
-      UUID sourceWorkspaceUuid = UUID.fromString(sourceWorkspaceId);
-      boolean cloneAlreadyRunning = cloneDao.cloneExistsForWorkspace(sourceWorkspaceUuid);
-      if (cloneAlreadyRunning) {
-        LOGGER.info("Clone already running, terminating initCloneMode");
-        return false;
+      // If another replica has started or completed cloning, exit safely
+      if (cloningStartedOrCompleted(UUID.fromString(sourceWorkspaceId))) {
+        return true;
       }
 
       // First, create an entry in the clone table to mark cloning has started
@@ -205,7 +212,7 @@ public class InstanceInitializerBean {
       LOGGER.error("An error occurred during clone mode. Error: {}", e.toString());
       // handle the interrupt if lock was interrupted
       if (e instanceof InterruptedException) {
-        LOGGER.error("Error with acquiring cloning/schema initialization Lock: {}", e.getMessage());
+        LOGGER.error("Error with acquiring cloning Lock: {}", e.getMessage());
         Thread.currentThread().interrupt();
       }
       try {
--- a/service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
@@ -79,45 +79,37 @@ public class InstanceInitializerBean {
   public void initializeInstance() {
     LOGGER.info("Default workspace id loaded as {}.", workspaceId);
     // Enter clone mode if sourceWorkspaceId is specified
-    if (StringUtils.isNotBlank(sourceWorkspaceId)) {
-      LOGGER.info("Source workspace id loaded as {}.", sourceWorkspaceId);
-      boolean cloneExitSuccessfuly = initCloneMode();
-      if (cloneExitSuccessfuly) {
+    if (isInCloneMode(sourceWorkspaceId)) {
+      LOGGER.info(
+          "Cloning mode enabled, attempting to clone from {} into {}.",
+          sourceWorkspaceId,
+          workspaceId);
+      // Initialize default schema if initCloneMode() returns false.
+      if (initCloneMode()) {
         LOGGER.info("Exited cloning.");
-      } else {
-        LOGGER.info("Cloning failed, initializing default schema.");
-        initializeDefaultInstance();
+        return;
       }
-    } else {
-      LOGGER.info("No SourceWorkspaceId found, initializing default schema.");
-      initializeDefaultInstance();
+      LOGGER.info("Cloning failed, falling back to initialize default schema.");
     }
+
+    initializeDefaultInstance();
   }
 
   /**
-   * Determine whether we're in the midst of a clone already or if a clone has completed.
+   * Determine whether we've completed, failed, or are in the midst of a clone already.
    *
    * @param sourceWorkspaceId value of {@code SOURCE_WORKSPACE_ID}; provided as an argument to
    *     assist with unit tests.
-   * @return true if we should not proceed with cloning.
+   * @return true if there's a clone entry in the database.
    */
-  protected boolean cloningStartedOrCompleted(UUID sourceWorkspaceId) {
-    LOGGER.info("Checking database for cloning and existing instances");
+  protected boolean cloningStarted(UUID sourceWorkspaceId) {
     // is a clone operation already running? This can happen when WDS is running with multiple
-    // replicas,
-    // and another replica started first and has initiated the clone. It can also happen in a
-    // corner case
-    // where this replica restarted during the clone operation.
+    // replicas, and another replica started first and has initiated the clone. It can also happen
+    // in a corner case where this replica restarted during the clone operation.
     boolean cloneAlreadyRunning = cloneDao.cloneExistsForWorkspace(sourceWorkspaceId);
-    // does the default pg schema already exist for this workspace? This should not happen,
-    // but we want to protect against overwriting it.
-    boolean instanceAlreadyExists = instanceDao.instanceSchemaExists(UUID.fromString(workspaceId));
-    LOGGER.info(
-        "isInCloneMode(): cloneAlreadyRunning={}, instanceAlreadyExists={}",
-        cloneAlreadyRunning,
-        instanceAlreadyExists);
-    // TODO handle the case where a clone already ran, but failed; should we retry?
-    return cloneAlreadyRunning || instanceAlreadyExists;
+
+    LOGGER.info("clone entry in database: {}", cloneAlreadyRunning);
+    return cloneAlreadyRunning;
   }
 
   /**
@@ -128,21 +120,26 @@ public class InstanceInitializerBean {
    *     assist with unit tests.
    * @return whether {@code SOURCE_WORKSPACE_ID} is valid.
    */
-  protected boolean cloningArgumentsSafe(String sourceWorkspaceId) {
-    try {
-      UUID.fromString(sourceWorkspaceId);
-    } catch (IllegalArgumentException e) {
-      LOGGER.warn(
-          "SourceWorkspaceId could not be parsed, unable to clone DB. Provided SourceWorkspaceId: {}.",
-          sourceWorkspaceId);
-      return false;
-    }
+  protected boolean isInCloneMode(String sourceWorkspaceId) {
+    if (StringUtils.isNotBlank(sourceWorkspaceId)) {
+      LOGGER.info("SourceWorkspaceId found, checking validity");
+      try {
+        UUID.fromString(sourceWorkspaceId);
+      } catch (IllegalArgumentException e) {
+        LOGGER.warn(
+            "SourceWorkspaceId could not be parsed, unable to clone DB. Provided SourceWorkspaceId: {}.",
+            sourceWorkspaceId);
+        return false;
+      }
 
-    if (sourceWorkspaceId.equals(workspaceId)) {
-      LOGGER.warn("SourceWorkspaceId and current WorkspaceId can't be the same.");
-      return false;
+      if (sourceWorkspaceId.equals(workspaceId)) {
+        LOGGER.warn("SourceWorkspaceId and current WorkspaceId can't be the same.");
+        return false;
+      }
+      return true;
     }
-    return true;
+    LOGGER.info("No SourceWorkspaceId found, initializing default schema.");
+    return false;
   }
 
   /*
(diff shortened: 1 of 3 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -78,93 +78,100 @@ public InstanceInitializerBean(
...
-    boolean isInCloneMode = isInCloneMode(sourceWorkspaceId);
-    LOGGER.info("isInCloneMode={}.", isInCloneMode);
-    if (isInCloneMode) {
+    // Enter clone mode if sourceWorkspaceId is specified
+    if (StringUtils.isNotBlank(sourceWorkspaceId)) {
       LOGGER.info("Source workspace id loaded as {}.", sourceWorkspaceId);
-      boolean cloneSuccess = initCloneMode();
-      if (cloneSuccess) {
-        LOGGER.info("Cloning complete.");
+      boolean cloneExitSuccessfuly = initCloneMode();
```
yuliadub, line 84: typo: Successfully

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -78,93 +78,100 @@ public InstanceInitializerBean(
    */
   public void initializeInstance() {
     LOGGER.info("Default workspace id loaded as {}.", workspaceId);
-    boolean isInCloneMode = isInCloneMode(sourceWorkspaceId);
-    LOGGER.info("isInCloneMode={}.", isInCloneMode);
-    if (isInCloneMode) {
+    // Enter clone mode if sourceWorkspaceId is specified
+    if (StringUtils.isNotBlank(sourceWorkspaceId)) {
```
yuliadub, line 82: I think I like the previous pattern a bit better (i.e. having a isCloneMode bool vs keeping this as a check if the source workspace is null). It used to be more obvious what we were checking for and why)

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -78,93 +78,100 @@ public InstanceInitializerBean(
...
       LOGGER.info("Source workspace id loaded as {}.", sourceWorkspaceId);
-      boolean cloneSuccess = initCloneMode();
-      if (cloneSuccess) {
-        LOGGER.info("Cloning complete.");
+      boolean cloneExitSuccessfuly = initCloneMode();
+      if (cloneExitSuccessfuly) {
+        LOGGER.info("Exited cloning.");
       } else {
+        LOGGER.info("Cloning failed, initializing default schema.");
         initializeDefaultInstance();
```
yuliadub, line 89: from a multiple replica perspective, how possible is it for 2 replicas to collide while creating default schema? I assume it is extremely unlikely since it is so fast to create an empty schema, but if for some reason this took longer in the future, could that cause an edge case?

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -78,93 +78,100 @@ public InstanceInitializerBean(
...
-    if (isInCloneMode) {
+    // Enter clone mode if sourceWorkspaceId is specified
+    if (StringUtils.isNotBlank(sourceWorkspaceId)) {
       LOGGER.info("Source workspace id loaded as {}.", sourceWorkspaceId);
-      boolean cloneSuccess = initCloneMode();
-      if (cloneSuccess) {
-        LOGGER.info("Cloning complete.");
+      boolean cloneExitSuccessfuly = initCloneMode();
+      if (cloneExitSuccessfuly) {
+        LOGGER.info("Exited cloning.");
```
yuliadub, line 86: nit pick: I think the previous log message makes more sense.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -78,93 +78,100 @@ public InstanceInitializerBean(
...
+    LOGGER.info("Checking database for cloning and existing instances");
+    // is a clone operation already running? This can happen when WDS is running with multiple
+    // replicas,
+    // and another replica started first and has initiated the clone. It can also happen in a
+    // corner case
+    // where this replica restarted during the clone operation.
+    boolean cloneAlreadyRunning = cloneDao.cloneExistsForWorkspace(sourceWorkspaceId);
+    // does the default pg schema already exist for this workspace? This should not happen,
+    // but we want to protect against overwriting it.
+    boolean instanceAlreadyExists = instanceDao.instanceSchemaExists(UUID.fromString(workspaceId));
```
yuliadub, line 114: do we need to check for this if the previous boolean return true? seems like we could just return

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -78,93 +78,100 @@ public InstanceInitializerBean(
...
-      }
+  /**
+   * Determine if the arguments needed for cloning (in this case, a valid {@code
+   * SOURCE_WORKSPACE_ID} env var) allow us to proceed with cloning.
+   *
+   * @param sourceWorkspaceId value of {@code SOURCE_WORKSPACE_ID}; provided as an argument to
+   *     assist with unit tests.
+   * @return whether {@code SOURCE_WORKSPACE_ID} is valid.
+   */
+  protected boolean cloningArgumentsSafe(String sourceWorkspaceId) {
```
yuliadub, line 131: I think this could become IsInCloneMode and be used in initializeInstance()

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
yuliadub: unless exit and complete are different operations which I am guessing they are - then it would be helpful to have the "cloning complete" message somewhere else.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -78,93 +78,100 @@ public InstanceInitializerBean(
...
   a newly created (destination) workspace. WDS at start up will always have a current WorkspaceId, which in the
   context of cloning will effectively be the destination. The SourceWorkspaceId will only be populated if the currently
   starting WDS was initiated via a clone operation and will contain the WorkspaceId of the original workspace where the cloning
-  was triggered.
+  was triggered. initCloneMode() returns false only if we want to initialize a default schema.
   */
   private boolean initCloneMode() {
     LOGGER.info("Starting in clone mode...");
     UUID trackingId = UUID.randomUUID();
     Lock lock = lockRegistry.obtain(sourceWorkspaceId);
```
jladieu, line 158: Consider waiting to grab the reference to the lock until _after_ the validation of `sourceWorkspaceId, and do it right before the `try` block where it gets used.  The goal here being to keep the declaration as close to the usage as possible, which makes it a bit easier to follow the flow.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -78,93 +78,100 @@ public InstanceInitializerBean(
...
+    // Check if we have a clone safe sourceWorkspaceId. Create default schema if not
+    if (!cloningArgumentsSafe(sourceWorkspaceId)) {
+      return false;
+    }
+
     try {
       // Make sure it's safe to start cloning (in case another replica is trying to clone)
       boolean lockAquired = lock.tryLock(1, TimeUnit.SECONDS);
       if (!lockAquired) {
```
jladieu, line 167: typonit: `lockAcquired`

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
yuliadub: looking at the WDS replica doc, it is listed as a non issue :) https://docs.google.com/document/d/1-s2-jPXFJLbrWM3f8113FTRZalecKR7acJhDjKWz20o/edit#heading=h.i4a5yga4e21e

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -78,93 +78,100 @@ public InstanceInitializerBean(
...
   starting WDS was initiated via a clone operation and will contain the WorkspaceId of the original workspace where the cloning
-  was triggered.
+  was triggered. initCloneMode() returns false only if we want to initialize a default schema.
   */
   private boolean initCloneMode() {
     LOGGER.info("Starting in clone mode...");
     UUID trackingId = UUID.randomUUID();
     Lock lock = lockRegistry.obtain(sourceWorkspaceId);
+    // Check if we have a clone safe sourceWorkspaceId. Create default schema if not
+    if (!cloningArgumentsSafe(sourceWorkspaceId)) {
```
jladieu, line 160: I agree with @yuliadub 's suggestion that that this could become `isInCloneMode`.

I also think that means that this extra check happening within `initCloneMode` can happen there too, and doesn't need to be a responsibility of `initCloneMode()`; per the [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single-responsibility_principle), I think this method's job is to get the init completed, not to validate whether or not it's safe to do so.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
jladieu: I agree with @yuliadub's suggestion here and would recommend putting this `isNotBlank` check in as the first check within what is currently named `cloningArgumentsSafe`, but which I also think makes sense to be renamed to `isCloneMode` (or `isInCloneMode`).

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -78,93 +78,100 @@ public InstanceInitializerBean(
    */
   public void initializeInstance() {
```
jladieu, line 79: I think if you shuffle around `isInCloneMode` as suggested in the other comments, this method would then benefit from using an early return after successful cloning to reduce the span of the multi-nested conditional.

```suggestion
  public void initializeInstance() {
    if (isInCloneMode(sourceWorkspaceId)) {
      LOGGER.info("Cloning mode enabled, attempting to clone from {} into {}.", sourceWorkspaceId, workspaceId);

      if (initCloneMode()) {
        LOGGER.info("Cloning complete.");
        return;
      }
      
      LOGGER.info("Cloning failed, falling back to initialize default schema.");
    }
    
    initializeDefaultInstance();
  }
```

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -78,93 +78,100 @@ public InstanceInitializerBean(
...
+    if (StringUtils.isNotBlank(sourceWorkspaceId)) {
       LOGGER.info("Source workspace id loaded as {}.", sourceWorkspaceId);
-      boolean cloneSuccess = initCloneMode();
-      if (cloneSuccess) {
-        LOGGER.info("Cloning complete.");
+      boolean cloneExitSuccessfuly = initCloneMode();
+      if (cloneExitSuccessfuly) {
+        LOGGER.info("Exited cloning.");
       } else {
+        LOGGER.info("Cloning failed, initializing default schema.");
```
jladieu, line 88: Combined with my suggested change to rejigger the logic of this method I'd recommend eliminating the LOGGER lines here saying that you're initializing the default schema, and instead let `initializeDefaultInstance()` method log something suitably generic out as the very thing it does, eg: `LOGGER.info("initializing default schema...")` -- or, maybe even better for mapping it to code: `LOGGER.info("initializeDefaultInstance() invoked");`

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -78,93 +78,100 @@ public InstanceInitializerBean(
...
 
   /**
-   * Determine if this WDS is starting as a clone of some other WDS. If a valid {@code
-   * SOURCE_WORKSPACE_ID} env var is provided to this WDS, it will start in clone mode.
+   * Determine whether we're in the midst of a clone already or if a clone has completed.
    *
    * @param sourceWorkspaceId value of {@code SOURCE_WORKSPACE_ID}; provided as an argument to
    *     assist with unit tests.
-   * @return whether this WDS is a clone.
+   * @return true if we should not proceed with cloning.
```
jladieu, line 102: I think this `@return` comment knows too much about how it's being called; I think it'd be better for this method to document its own behavior (returns true when somethingsomething), and the calling code can decide what to do with this info; if explanation is required in the calling code, this would make a perfect comment there instead of here

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -78,93 +78,100 @@ public InstanceInitializerBean(
...
+    return true;
   }
 
   /*
   Cloning comes from the concept of copying an original (source) workspace data (from WDS data tables) into
   a newly created (destination) workspace. WDS at start up will always have a current WorkspaceId, which in the
   context of cloning will effectively be the destination. The SourceWorkspaceId will only be populated if the currently
   starting WDS was initiated via a clone operation and will contain the WorkspaceId of the original workspace where the cloning
-  was triggered.
+  was triggered. initCloneMode() returns false only if we want to initialize a default schema.
```
jladieu, line 153: Similar to my comment earlier, I think talking about what _should happen_ after this method gets called means this method knows too much about its caller.  Instead return what true or false means, and let the calling code decide what to do with this information (and document that behavior at the call site rather than here).

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: absolutely beautiful code from someone who has not been staring at this too long. thank you ❤️

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -78,93 +78,100 @@ public InstanceInitializerBean(
...
   */
   private boolean initCloneMode() {
     LOGGER.info("Starting in clone mode...");
     UUID trackingId = UUID.randomUUID();
```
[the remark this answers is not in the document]
ashanhol: We reference `trackingId` in the catch block though. 
Can't define it in try even though that's closer to where we use it. :(

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: youre right, exit is different. I added a cloning complete message when we finish restore and mark the job as successful. I know it's not an ideal placement, but it's the only place I felt comfortable knowing that's true.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: refactored `isInCloneMode()` back into place

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: fixed, ty

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: Legit, bad refactor. Put it back into place.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: I refactored this logic a bit. Since returning true and false from `initCloneMode` determine whether or not we initialize the default schema, I considered the scenario where a replica starts clone mode, then errors or crashes in a way where the default schema is never created. The previous logic wouldn't create the default schema in that case because it would see an entry in the clone DB with no nuance. The new logic now checks if there's a clone entry, and if so creates the default schema if it doesn't exist.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: yeah, fixed.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: fixed, ty

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: yeah I put isInCloneMode back

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: Fair, hadn't considered that pattern but it makes total sense. Rewrote my return comments for the various methods

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: Actually I take it back, I should log the reason for exiting clone mode when we return early and change it back to cloning complete. ill fix that.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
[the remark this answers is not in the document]
jladieu: Dang!  Got it.  Thanks for the explain :)

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -128,50 +120,52 @@ protected boolean cloningStartedOrCompleted(UUID sourceWorkspaceId) {
...
-      return false;
-    }
+  protected boolean isInCloneMode(String sourceWorkspaceId) {
+    if (StringUtils.isNotBlank(sourceWorkspaceId)) {
+      LOGGER.info("SourceWorkspaceId found, checking validity");
+      try {
+        UUID.fromString(sourceWorkspaceId);
+      } catch (IllegalArgumentException e) {
+        LOGGER.warn(
+            "SourceWorkspaceId could not be parsed, unable to clone DB. Provided SourceWorkspaceId: {}.",
```
jladieu: Just a suggested rephrase, `Source workspaceId \"{}\" could not be parsed`

The reason I suggest this is when I read "Provided XYZ" I interpreted it at first as "I provided XYZ instead", rather than "I was provided XYZ".

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -128,50 +120,52 @@ protected boolean cloningStartedOrCompleted(UUID sourceWorkspaceId) {
...
-      LOGGER.warn("SourceWorkspaceId and current WorkspaceId can't be the same.");
-      return false;
+      if (sourceWorkspaceId.equals(workspaceId)) {
+        LOGGER.warn("SourceWorkspaceId and current WorkspaceId can't be the same.");
+        return false;
+      }
+      return true;
     }
-    return true;
+    LOGGER.info("No SourceWorkspaceId found, initializing default schema.");
```
jladieu, line 141: Instead of saying "initializing default schema here" (which maybe will happen but might change if the callsite does something different), can you say "unable to proceed with cloning"

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -307,6 +301,7 @@ private void restoreFromRemoteBackup(String backupFileName, UUID cloneJobId) {
           cloneJobId, restoreResponse.getErrorMessage(), CloneTable.RESTORE);
     } else {
       LOGGER.info("Restore Successful");
+      LOGGER.info("Cloning Complete");
```
jladieu, line 306: Can you combine these two log statements into one?

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: true, corrected it in another place too

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: corrected

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
ashanhol: ty, actually meant to delete that one. fixed

davidangb (commented):
See comments inline. At a high level, I see two places where the logic doesn't seem quite right:
1. We made it all the way to actually attempting the backup/restore part of cloning, but either the backup or the restore failed. In this case, we should initialize the default schema (as empty)
2. The code as-is always calls initializeDefaultInstance() even when the logic dictates it should do nothing

At the code level, it seems like we're jumping through some hoops to return a boolean from `initCloneMode()`. If you change that to return void, does it simplify the implementation?

It might be helpful to write (or identify pre-existing) unit tests for the various use cases and see how they turn out:
* SOURCE_WORKSPACE_ID not defined. 1 call to instanceDao.createSchema.
* SOURCE_WORKSPACE_ID is defined, but couldn't get a lock. No call to instanceDao.createSchema.
* SOURCE_WORKSPACE_ID is defined, got a lock, but entry already exists in clone table. No call to instanceDao.createSchema.
* SOURCE_WORKSPACE_ID is defined, got a lock, no entry in clone table. Calls to add row to clone table and attempt cloning.
* SOURCE_WORKSPACE_ID is defined, got a lock, no entry in clone table, but cloning failed. 1 call to instanceDao.createSchema.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -78,62 +78,66 @@ public InstanceInitializerBean(
...
+      LOGGER.info(
+          "Cloning mode enabled, attempting to clone from {} into {}.",
+          sourceWorkspaceId,
+          workspaceId);
+      // Initialize default schema if initCloneMode() returns false.
+      if (initCloneMode()) {
         LOGGER.info("Cloning complete.");
-      } else {
-        initializeDefaultInstance();
+        return;
```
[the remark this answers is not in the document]
davidangb: after discussing in person, I now understand what's happening here and it looks good to me, thank you for explaining!

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -78,62 +78,66 @@ public InstanceInitializerBean(
...
       }
-    } else {
-      initializeDefaultInstance();
+      LOGGER.info("Cloning failed, falling back to initialize default schema.");
     }
+
+    initializeDefaultInstance();
```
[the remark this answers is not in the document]
davidangb: after discussing in person, I now understand what's happening here and it looks good to me, thank you for explaining!

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -78,62 +78,66 @@ public InstanceInitializerBean(
...
+   * Determine whether we've completed, failed, or are in the midst of a clone already.
+   *
+   * @param sourceWorkspaceId value of {@code SOURCE_WORKSPACE_ID}; provided as an argument to
+   *     assist with unit tests.
+   * @return true if there's a clone entry in the database.
+   */
+  protected boolean cloningStarted(UUID sourceWorkspaceId) {
+    // is a clone operation already running? This can happen when WDS is running with multiple
+    // replicas, and another replica started first and has initiated the clone. It can also happen
+    // in a corner case where this replica restarted during the clone operation.
```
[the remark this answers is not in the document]
davidangb: discussed in person - my comment above is just informative, I am not requesting any changes in this PR

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerBean.java
```diff
@@ -142,29 +146,27 @@ Cloning comes from the concept of copying an original (source) workspace data (f
...
-      if (cloneAlreadyRunning) {
-        LOGGER.info("Clone already running, terminating initCloneMode");
-        return false;
+      // Acquiring the lock means other replicas have not started or have finished cloning.
+      // If there's a clone entry and no default schema, another replica errored before completing.
+      // If there's a clone entry and a default schema there's nothing for us to do here.
+      if (cloningStarted(UUID.fromString(sourceWorkspaceId))) {
+        boolean instanceExists = instanceDao.instanceSchemaExists(UUID.fromString(workspaceId));
+        LOGGER.info("Previous clone entry found. Instance schema exists: {}.", instanceExists);
+        return instanceExists;
```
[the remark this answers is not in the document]
davidangb: discussed in person, this is here to check for cases where a previous clone attempt ran but failed.

davidangb (approved):
After chatting about this PR, I am comfortable with the implementation, thank you for persisting on it and explaining it all!

I'd love the confidence more unit tests would give us. If those are troublesome I'd be ok merging this and tackling a unit-test related refactor as a follow-on PR.

With all the deep attention on this flow I have concerns about races within `initializeDefaultInstance()`. I think these are easily solvable and I will create a follow-on ticket to explain and track them.

PR #358, opened 2023-10-10
AJ-1011: Job refactor

davidangb:
WIP draft, do not review yet.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/shared/model/job/EmptyJobInput.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/shared/model/job/EmptyJobInput.java
@@ -1,0 +1,3 @@
+package org.databiosphere.workspacedataservice.shared.model.job;
+
+public class EmptyJobInput implements JobInput {}
--- a/service/src/main/java/org/databiosphere/workspacedataservice/shared/model/job/Job.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/shared/model/job/Job.java
@@ -9,16 +9,18 @@ import java.util.UUID;
  * before completing. Different jobs can specify different JobResult implementations to satisfy
  * their own response payloads.
  *
- * @param <T> the response payload for this job.
+ * @param <T> the input arguments for this job.
+ * @param <U> the response payload for this job.
  */
 @JsonInclude(JsonInclude.Include.NON_NULL)
-public class Job<T extends JobResult> {
+public class Job<T extends JobInput, U extends JobResult> {
 
   private UUID jobId;
   private JobStatus status;
   private String errorMessage;
   private LocalDateTime created, updated;
-  private T result;
+  private final T input;
+  private U result;
 
   public Job(
       UUID jobId,
@@ -26,12 +28,14 @@ public class Job<T extends JobResult> {
       String errorMessage,
       LocalDateTime created,
       LocalDateTime updated,
-      T result) {
+      T input,
+      U result) {
     this.jobId = jobId;
     this.status = status;
     this.errorMessage = errorMessage;
     this.created = created;
     this.updated = updated;
+    this.input = input;
     this.result = result;
   }
 
@@ -77,11 +81,15 @@ public class Job<T extends JobResult> {
     this.updated = updated;
   }
 
-  public T getResult() {
+  public T getInput() {
+    return input;
+  }
+
+  public U getResult() {
     return result;
   }
 
-  public void setResult(T result) {
+  public void setResult(U result) {
     this.result = result;
   }
 }
--- a/service/src/test/java/org/databiosphere/workspacedataservice/dao/MockBackupRestoreDao.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/dao/MockBackupRestoreDao.java
@@ -9,6 +9,7 @@ import org.databiosphere.workspacedataservice.shared.model.BackupResponse;
 import org.databiosphere.workspacedataservice.shared.model.BackupRestoreRequest;
 import org.databiosphere.workspacedataservice.shared.model.CloneTable;
 import org.databiosphere.workspacedataservice.shared.model.RestoreResponse;
+import org.databiosphere.workspacedataservice.shared.model.job.EmptyJobInput;
 import org.databiosphere.workspacedataservice.shared.model.job.Job;
 import org.databiosphere.workspacedataservice.shared.model.job.JobResult;
 import org.databiosphere.workspacedataservice.shared.model.job.JobStatus;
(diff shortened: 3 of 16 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/shared/model/job/EmptyJobInput.java
```diff
@@ -0,0 +1,4 @@
+package org.databiosphere.workspacedataservice.shared.model.job;
+
+/** used for Jobs that have no input arguments. */
+public class EmptyJobInput implements JobInput {}
```
yuliadub, line 4: seems like all of our jobs are this way - when would they not be? (maybe I am just stuck on clone land with my head and not thinking of the right scenarios)

davidangb: The soon-to-be-coming data import jobs will use input arguments - so, this will be used soon! (Hard to separate out the prefactor from the actual implementation … )

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -762,6 +765,9 @@ components:
         - $ref: '#/components/schemas/Job'
         - type: object
           properties:
+            input:
+              type: object
+              description: Input arguments; expected to be empty.
```
davidangb, line 770: future APIs, like the data import API, will specify input arguments and have a different value here!

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -890,8 +899,6 @@ components:
           type: string
         errorMessage:
           type: string
-        exception:
-          type: string
```
davidangb, line 894: Technically this is a breaking change - removing a field from our clients. BUT, this `exception` definition in the OpenAPI spec was extraneous - there is no corresponding "exception" in the business logic. So it feels safe to clean up, nobody could ever have used this for anything.

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/job/JobType.java
```diff
@@ -0,0 +1,9 @@
+package org.databiosphere.workspacedataservice.shared.model.job;
+
+/** The various types of Jobs that can be executed. */
+public enum JobType {
```
jladieu, line 4: Might you underscore these?  EG: `DATA_IMPORT`, `SYNC_BACKUP`, etc.?

(or is there a reason not to?)

service/src/test/java/org/databiosphere/workspacedataservice/dao/MockBackupRestoreDao.java
```diff
@@ -38,26 +42,30 @@ public void createEntry(UUID trackingId, BackupRestoreRequest request) {
...
-      Job<RestoreResponse> restore =
+      Job<EmptyJobInput, RestoreResponse> restore =
           new Job<>(
               trackingId,
+              SYNCRESTORE,
               JobStatus.QUEUED,
               "",
               now.toLocalDateTime(),
               now.toLocalDateTime(),
+              new EmptyJobInput(),
```
jladieu, line 66: Instead of `newing` this up everywhere, maybe you could provide a static instance and/or factory method on the interface/class like: `JobInput.EMPTY` `JobInput.empty()` - that way if things ever get tricky with instantiation you'll only have to do it in one place.

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/job/JobType.java
```diff
@@ -0,0 +1,9 @@
...
+  DATAIMPORT, // async data import jobs
+  SYNCBACKUP, // legacy synchronous backups
```
calypsomatic, line 6: "legacy" - so the current synchronous clones still use a job even though they're not async, will these enums go away/change to non-legacy when clone becomes async?  Or will they live here forever and we'll add new async jobs?

davidangb: I plan to add new job types for the new async jobs, and mark the current sync APIs as deprecated. Moving an API from sync to async is a significant breaking change imho and so we should treat it as a new API.

davidangb: And you're also calling out one of the things I struggled with in this refactor - should I create a brand new set of model classes for async jobs? Or, should I reuse this `Job` model and its friends, even though it spans both sync and async jobs? I went for the latter, but open to feedback that it's the wrong choice.

jladieu (approved):
I had some vague feedback about `EmptyJobInput`'s name/namespace but I can't point exactly to _why_ so feel free to take with a grain of salt.

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/job/Job.java
```diff
@@ -9,29 +9,37 @@
...
 @JsonInclude(JsonInclude.Include.NON_NULL)
-public class Job<T extends JobResult> {
+public class Job<T extends JobInput, U extends JobResult> {
 
-  private UUID jobId;
+  private final UUID jobId;
+  private final JobType jobType;
   private JobStatus status;
   private String errorMessage;
+  private StackTraceElement[] stackTrace;
```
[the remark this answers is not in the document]
jladieu: (Oops, just seeing now you explained this in the PR description, but might be useful to include here too)

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/job/JobType.java
davidangb: updated with underscores.

service/src/test/java/org/databiosphere/workspacedataservice/dao/MockBackupRestoreDao.java
davidangb: I like this, and have implemented `JobInput.empty()` everywhere!

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/job/Job.java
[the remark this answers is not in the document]
davidangb: Updated all the fields of this class with doc.

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/job/Job.java
```diff
@@ -15,13 +15,38 @@
...
+  /** id for this job. */
   private final UUID jobId;
+
+  /** type of this job, e.g. PFB import vs. database backup; {@link JobType} */
   private final JobType jobType;
+
+  /** status of the job; {@link JobStatus} */
   private JobStatus status;
+
+  /** Short error message, if this job hit an exception. Designed to displayed to end users. */
```
jladieu, line 27: nit: 'to be displayed'

jladieu (approved):
Looking good!

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/job/Job.java
jladieu: hah did you find it at the same time as me?

PR #361, opened 2023-10-11
AJ-1011: async snapshot api skeleton

davidangb:
In this PR:
* create a new `sys_wds.job` Postgres table
    * with triggers to automatically track created and last-updated timestamps
    * new `JobDao` and `PostgresJobDao` to interact with this table
* define a new `jobStatusV1` API to retrieve the status of any async job
    * new `JobController` and `JobService` to power this API
* tweak the response payload and security scheme for the `importV1` API
* skeleton/proof-of-concept implementation of an async snapshot-import job
    * definition of a Quartz job in `ImportQuartzJob`
    * `SchedulerDao` and `QuartzSchedulerDao` to interact with Quartz
* lots of changes to the generated interfaces and models

Still outstanding, for follow-on PRs:
* replacing the skeleton/proof-of-concept async job with the real snapshot import
* definitions of `JobInput` payload(s) for the data-import API
* addressing outstanding TODOs
* paying attention to JSON serialization in API responses, e.g. omitting nulls

Open questions/areas for feedback:
1. I'm strugging with the overlap/conflict between our hand-written models (e.g. `Job`) and the generated models (e.g. `GenericJobServerModel`). I believe this conflict exists because we're transitioning from hand-written models used by our v0.2 APIs and the generated models in our v1 APIs, and I'm trying to reuse functionality between the two. If I were writing this from scratch, I'd just use the generated models and maybe add some subclasses where necessary. What's your reaction?


service/build.gradle
```diff
@@ -120,6 +120,7 @@ tasks.register('generateApiInterfaces', GenerateTask) {
     apiFilesConstrainedTo.set([""])   // empty string means generate all apis
     modelFilesConstrainedTo.set([""]) // empty string means generate all models
     supportingFilesConstrainedTo.set([]) // empty array means generate none
+    skipOperationExample = true // example responses require the ApiUtil.java supporting file
```
davidangb, line 123: I realized we need this flag as I started to use the auto-generated api interfaces in earnest

service/src/main/java/org/databiosphere/workspacedataservice/controller/ImportController.java
```diff
@@ -1,20 +1,26 @@
...
 
   @Override
-  public ResponseEntity<Void> importV1(UUID instanceUuid, ImportRequestServerModel importRequest) {
-    // TODO: implementation for imports
-    // return TOO_EARLY here as a proof-of-concept that we can override the implementation
-    // in ImportApi (which returns NOT_IMPLEMENTED)
-    return new ResponseEntity<>(HttpStatus.TOO_EARLY);
+  public ResponseEntity<GenericJobServerModel> importV1(
+      UUID instanceUuid, ImportRequestServerModel importRequest) {
+    GenericJobServerModel importJob = importService.createImport(instanceUuid, importRequest);
```
davidangb, line 23: use a subclass to control json serialization

service/src/main/java/org/databiosphere/workspacedataservice/controller/JobController.java
```diff
@@ -0,0 +1,34 @@
...
+
+  JobService jobService;
+
+  public JobController(JobService jobService) {
+    this.jobService = jobService;
+  }
+
+  @Override
+  public ResponseEntity<GenericJobServerModel> jobStatusV1(UUID instanceUuid, UUID jobId) {
+    GenericJobServerModel job = jobService.getJob(instanceUuid, jobId);
```
davidangb, line 23: subclass here too

service/src/main/java/org/databiosphere/workspacedataservice/controller/JobController.java
```diff
@@ -0,0 +1,34 @@
...
+
+    // return job status, 200 if job is completed and 202 if job is still running
+    HttpStatus responseCode = HttpStatus.ACCEPTED;
+    if (job.getStatus() == GenericJobServerModel.StatusEnum.SUCCEEDED
+        || job.getStatus() == GenericJobServerModel.StatusEnum.ERROR) {
```
davidangb, line 28: subclass could have a `isFinished` convenience to check status

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
```diff
@@ -0,0 +1,171 @@
...
+            + "where id = :jobId",
+        new MapSqlParameterSource("jobId", jobId.toString()),
+        new AsyncJobRowMapper());
+  }
+
+  // rowmapper for retrieving Job objects from the db
+  private static class AsyncJobRowMapper implements RowMapper<GenericJobServerModel> {
+    @Override
+    public GenericJobServerModel mapRow(ResultSet rs, int rowNum) throws SQLException {
+      UUID jobId = UUID.fromString(rs.getString("id"));
```
calypsomatic, line 141: Will we ever have to worry about this UUID not parsing?  Seems like it shouldn't happen but we have try/catch over all the other values

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
```diff
@@ -0,0 +1,69 @@
...
+    // validate write permission
+    boolean hasWriteInstancePermission = samDao.hasWriteInstancePermission();
+    logger.debug("hasWriteInstancePermission? {}", hasWriteInstancePermission);
+    if (!hasWriteInstancePermission) {
+      throw new AuthorizationException("Caller does not have permission to write to instance.");
+    }
+
+    // TODO: translate the ImportRequestServerModel into a Job
+    // for now, just make an example job
+    Job<JobInput, JobResult> job = Job.newJob(JobType.DATA_IMPORT, JobInput.empty());
```
calypsomatic, line 55: It seems like every job will go through the steps create/schedule/updateStatus.  Should that collection of actions be pulled out into a job-specific class or will there be meaningful differences between them depending on the job?

calypsomatic (approved):
Great, thank you for this!  I took at glance at the 63 code smells and it seems like most of them were due to this being a skeleton, but there were a handful of unused imports or whatnot, so you could clean that up a bit if you care to.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
```diff
@@ -0,0 +1,171 @@
...
+   * update the `updated` column's value.
+   *
+   * @param jobId id of the job to update
+   * @param status the status to which the job will be updated
+   * @param errorMessage a short error message, if the job is in error
+   * @param stackTrace a full stack trace for debugging, if the job is in error
+   * @return the updated job
+   */
+  @Override
+  public GenericJobServerModel updateStatus(
```
jladieu, line 86: You might be able to make this method private, with a handful of overloads with strongly typed `@NonNull` params if you take the approach of exposing state-machine style verbs.

This might let you cut down on the null checks within the method body, and simplify / eliminate the caveats from the Javadoc describing the params.

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
```diff
@@ -0,0 +1,69 @@
...
+
+    // persist the job
+    GenericJobServerModel createdJob = jobDao.createJob(job);
+
+    // schedule the job. after successfully scheduling, mark the job as queued
+    schedulerDao.schedule(job);
```
jladieu, line 61: Could the `schedulerDao` take care of talking to the `jobDao` instead?  It seems like you're having to go to two places to get one task done here by having the `ImportService` do both.

jladieu (commented):
Looking really good so far!

I took a pass through and have two fairly major/impactful suggestions which you may or may not want to pursue:

1) Consider tightening up JobDao's API by making it more of a state machine.
2) Consider giving SchedulerDao more responsibility to actually create & manage state of the jobs.

With these two in mind, I didn't dig deep on doing code level inspection yet.  LMK your thoughts, also happy to just go through the code if you don't want to shuffle things around that much right now.

jladieu:
BTW, I didn't exactly know what to look for with regards to your struggling with the generated model vs. hand-built model.  I'd welcome some trailmarker comments showing me what to look at here if you'd like my input on it.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
davidangb: With our business logic, a parsing failure should never happen … that said, if a parsing failure DOES happen I think it should be fatal, so I'm omitting any try/catch here and just letting the exception happen.

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
davidangb: I'm honestly not sure. I suspect the lifecycle will be the same for all types of jobs, but maybe not? I'm leaving that refactoring for whenever we have more visibility into what each job will do.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
```diff
@@ -0,0 +1,171 @@
...
+      UUID jobId,
+      GenericJobServerModel.StatusEnum status,
+      String errorMessage,
+      StackTraceElement[] stackTrace) {
+
+    // start our sql statement and map of params
+    StringBuilder sb = new StringBuilder("update sys_wds.job set status = :status");
+    MapSqlParameterSource params = new MapSqlParameterSource("jobId", jobId.toString());
```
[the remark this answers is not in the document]
davidangb: heh I had the same thought when writing the code, but that approach gets tripped up by the presence of `jobId` in the param map. Of course, I could have special handling for `jobId`, or I could build the "set x = y" part of the SQL string first, then add jobId into the param map, then concat the SQL together … it didn't seem worth it at that point. If you feel strongly about this I can refactor!

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
davidangb: I also wondered about this, but I don't like DAOs depending on each other; I'd rather have a Service depend on two DAOs. I agree this is an area to keep an eye on; if the writes to the `job` table and to `Quartz` are always in sync, it might make sense to encapsulate it all into a single DAO.

service/src/main/resources/static/swagger/apis-v1.yaml
```diff
@@ -56,6 +60,41 @@ paths:
...
+        - Job
+      parameters:
+        - $ref: '#/components/parameters/instanceUuidPathParam'
+        - name: jobId
+          in: path
+          required: true
+          schema:
+            type: string
+            format: uuid
+      responses:
```
ashanhol, line 85: Would it ever be appropriate to define a 401 not authorized if no bearerAuth?

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/job/JobStatus.java
```diff
@@ -1,11 +1,50 @@
...
+  UNKNOWN(HttpStatus.INTERNAL_SERVER_ERROR, false, false); // note neither running nor terminated
+
+  private final HttpStatus code;
+  private final boolean isTerminated;
+  private final boolean isRunning;
+
+  JobStatus(HttpStatus code, boolean isTerminated, boolean isRunning) {
+    this.code = code;
+    this.isTerminated = isTerminated;
+    this.isRunning = isRunning;
```
ashanhol, line 23: Trying to understand why there's two bools for this, is this noting a situation where we send a termination request to a running job, but the job is still mid run?

davidangb: We have the `UNKNOWN` status, used as a fallback if we can't interpret the value stored in the db, and `UNKNOWN` is neither running nor terminated. So, I have two separate booleans so `UNKNOWN` can be false for both.

service/src/main/resources/static/swagger/apis-v1.yaml
davidangb: We'd want to define that for every API … or leave it off for every API, and the latter seemed better. I think there's a way to define common responses for all APIs, we might want to look into that but I don't want to do it just for one or two.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportQuartzJob.java
```diff
@@ -0,0 +1,57 @@
...
+  public ImportQuartzJob(JobDao jobDao) {
+    this.jobDao = jobDao;
+  }
+
+  @Override
+  public void execute(JobExecutionContext context) throws JobExecutionException {
+    UUID jobId = UUID.fromString(context.getJobDetail().getKey().getName());
+
+    // as the first step when this kicks off, set status to running
+    jobDao.updateStatus(jobId, GenericJobServerModel.StatusEnum.RUNNING);
```
ashanhol, line 35: Why do we have to use GenericJobServerModel's status instead of JobStatus's status? It's hard to follow when we use each

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportQuartzJob.java
```diff
@@ -0,0 +1,57 @@
...
+import org.springframework.stereotype.Component;
+
+/**
+ * Quartz-executable job definition for data imports.
+ *
+ * <p>As of this writing, the job can be queued and executed by Quartz, but the execution itself
+ * will do nothing but sleep for a random time between 5 and 15 seconds
+ */
+@Component
+public class ImportQuartzJob implements Job {
```
ashanhol, line 21: I also find this confusing since we've opted to keep our own Job.java model. It'd be really really nice if we could name this QuartzJob to distinguish it, but since we cant I strongly recommend we change the name of our Job to something more descriptive of what it really is

davidangb:
> 1. Consider tightening up JobDao's API by making it more of a state machine.
> 2. Consider giving SchedulerDao more responsibility to actually create & manage state of the jobs.

@jladieu I've refactored to address the first comment, and am pushing back softly on the second (see [comment](https://github.com/DataBiosphere/terra-workspace-data-service/pull/361#discussion_r1357049638)). I've also added a TODO to `JobDao` as a trailmarker re: the competing models; see also Adina's comments on models.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportQuartzJob.java
davidangb: I agree, this is awkward :(

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportQuartzJob.java
davidangb: added a code comment to make this clearer

service/src/test/java/org/databiosphere/workspacedataservice/shared/model/job/JobStatusTest.java
```diff
@@ -27,23 +25,15 @@ void equivalentToGeneratedModel() {
     assertEquals(generatedNames, names);
   }
 
-  private static Stream<Arguments> provideJobStatuses() {
-    return Arrays.stream(JobStatus.values()).map(Arguments::of);
-  }
-
   @ParameterizedTest(name = "JobStatus {0} should translate to GenericJobServerModel.StatusEnum")
-  @MethodSource("provideJobStatuses")
```
jladieu: I like it!

service/src/test/java/org/databiosphere/workspacedataservice/dao/PostgresJobDaoTest.java
```diff
@@ -1,5 +1,6 @@
 package org.databiosphere.workspacedataservice.dao;
 
+import static org.databiosphere.workspacedataservice.generated.GenericJobServerModel.*;
```
jladieu, line 3: Can you do a non-splat import of all the stuff you need?

service/src/test/java/org/databiosphere/workspacedataservice/shared/model/job/JobStatusTest.java
```diff
@@ -1,5 +1,6 @@
 package org.databiosphere.workspacedataservice.shared.model.job;
 
+import static org.databiosphere.workspacedataservice.generated.GenericJobServerModel.*;
```
jladieu, line 3: Can you non-splat this too?

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
jladieu: Sounds good.  I almost wonder if there's a generic thing that `ImportService` is currently doing which we might find factors out nicely into a standalone actor (which coordinates with both DAOs) as we start adding on more use cases, but ok to wait until we've got actual duplication that lends itself to reuse.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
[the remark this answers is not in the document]
jladieu: Sounds fine!  I don't feel strongly but do have a hunch this might be something we can tighten up, ok to wait on that and maybe take it as a rainyday task to try out my theory later on.  I think the code as is is easy to follow and what I'm envisioning is just extra optional polish that might just make things harder to read anyway ;)

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/job/Job.java
```diff
@@ -68,6 +68,19 @@ public Job(
...
+   * Convenience for creating a new Job object, using a random UUID as jobId and defaults for job
+   * status
+   *
+   * @param jobType type of the job
+   * @param input input arguments for the job
+   * @return the job, after creation
+   */
+  public static Job<JobInput, JobResult> newJob(JobType jobType, JobInput input) {
+    LocalDateTime now = LocalDateTime.now();
+    return new Job<>(UUID.randomUUID(), jobType, JobStatus.CREATED, null, now, now, input, null);
```
jladieu, line 81: Small readability thing I like to do with large param lists, in the absence of named parameters: add an explainer comment for any arg that doesn't have a self-documenting name.  (This suggestion might need some re-wrapping to look nice)

```suggestion
    return new Job<>(UUID.randomUUID(), jobType, JobStatus.CREATED, /* errorMessage=*/ null, /* created=*/ now, /* updated=*/ now, input, /* result=*/ null);
```

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/job/JobStatus.java
```diff
@@ -1,11 +1,50 @@
...
+
 /** The various states a Job can move through. */
 public enum JobStatus {
-  QUEUED, // job has been created but not yet started
-  RUNNING, // job has started
-  SUCCEEDED, // job completed as expected
-  ERROR, // job failed
-  CANCELLED, // job was interrupted or intentionally not started
-  UNKNOWN
+  CREATED(HttpStatus.ACCEPTED, false, true), // job has been created but not yet scheduled to run
```
jladieu, line 8: These enums would benefit from the explainer comments I mentioned earlier too:

```suggestion
  CREATED(HttpStatus.ACCEPTED, /* isTerminated= */ false, /* isRunning=*/ true), // job has been created but not yet scheduled to run
```

service/src/test/java/org/databiosphere/workspacedataservice/controller/JobControllerMockMvcTest.java
```diff
@@ -0,0 +1,68 @@
...
+            .perform(get("/{instanceUuid}/job/v1/{jobId}", instanceId, expected.getJobId()))
+            .andExpect(status().isAccepted())
+            .andReturn();
+
+    // and the API response should be a valid GenericJobServerModel
+    GenericJobServerModel actual =
+        mapper.readValue(mvcResult.getResponse().getContentAsString(), GenericJobServerModel.class);
+
+    // which is equal to the expected job
+    // TODO: this is failing due to comparisons on created/updated times, due to timezone issue
```
jladieu, line 65: Aww man, I hate timezones...

I am used to using protobufs as opposed to POJOs and there were really nice constructs for doing fluent comparison of expected vs. actual Protobuf records (https://truth.dev/api/0.42/com/google/common/truth/extensions/proto/ProtoFluentAssertion.html)

It'd let you explicitly compare just fields you set on the expected value during creation (with `compareExpectedFieldsOnly()`), or exclude fields you knew to be unreliable (like timestamps).

Have you ever encountered a library like that for regular Java objects?  I find myself reaching toward something like that to try to help with assertion false failures like this one, but don't have a bead on what exists.

service/src/test/java/org/databiosphere/workspacedataservice/dao/PostgresJobDaoTest.java
```diff
@@ -0,0 +1,199 @@
...
+                    + "and input = '{}'::jsonb "
+                    + "and result is null and stacktrace is null",
+                params,
+                String.class),
+        "should properly update the job with an error message");
+  }
+
+  // update status with an error message and stacktrace
+  @Test
+  void updateWithErrorMessageAndStackTrace() throws JsonProcessingException {
```
jladieu, line 158:
```suggestion
  void failWithErrorMessageAndStackTrace() throws JsonProcessingException {
```

service/src/test/java/org/databiosphere/workspacedataservice/dao/PostgresJobDaoTest.java
```diff
@@ -0,0 +1,199 @@
...
+                    + "and input = '{}'::jsonb "
+                    + "and result is null and error is null and stacktrace is null",
+                params,
+                String.class),
+        "should properly update the job");
+  }
+
+  // update status with an error message
+  @Test
+  void updateWithErrorMessage() {
```
jladieu, line 129:
```suggestion
  void failWithErrorMessage() {
```

service/src/test/java/org/databiosphere/workspacedataservice/dao/PostgresJobDaoTest.java
```diff
@@ -0,0 +1,199 @@
...
+    // after updating the job, there should be exactly one row with:
+    // this jobId and type, the new status, the new error message, and the new stack trace
+    // and updated timestamp greater than the created timestamp.
+    // input should still be {}, and result should still be null
+    var params = new MapSqlParameterSource("jobId", testJob.getJobId().toString());
+    params.addValue("type", jobType.name());
+    params.addValue("status", StatusEnum.ERROR.name());
+    params.addValue("error", errorMessage);
+    params.addValue("stacktrace", mapper.writeValueAsString(stackTrace));
+    assertDoesNotThrow(
```
jladieu, line 175: I think these tests will be brittle if we ever get in the business of renaming/refactoring the actual table under test.

Instead of crafting a query in each test case that will throw when not all predicates match, could you just provide a generic `findById` helper method and then check the results on the object that gets returned?  This might let you create failure messages which do a better job of explaining what attribute is missing/different/unexpected, and centralize the predicate generation in one place so you don't have to play whackamole if the columns ever change.

service/src/test/java/org/databiosphere/workspacedataservice/controller/JobControllerMockMvcTest.java
```diff
@@ -0,0 +1,68 @@
...
+@SpringBootTest
+@AutoConfigureMockMvc
+class JobControllerMockMvcTest {
+  @Autowired private MockMvc mockMvc;
+  @Autowired private ObjectMapper mapper;
+  @Autowired private InstanceDao instanceDao;
+  @MockBean private JobDao jobDao;
+
+  @Test
+  void smokeTestCreateImport() throws Exception {
```
jladieu, line 36:
```suggestion
  void smokeTestGetJob() throws Exception {
```

jladieu (approved):
Looks great.  I have some non-blocking suggestions, but ship it when ready!

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/job/JobStatus.java
davidangb: Huh - I get this for free for booleans in IntelliJ, I assumed it was the default:
Screenshot 2023-10-12 at 3 12 24 PM

service/src/test/java/org/databiosphere/workspacedataservice/controller/JobControllerMockMvcTest.java
davidangb: oh lol this TODO was a note-to-self as I was developing, I resolved the issue and it was faulty code on my part! that said, to your question: I am not aware of a library for this, but I wonder if it exists. I might just write a new `equals()` function or a new `Comparator` or something.

service/src/test/java/org/databiosphere/workspacedataservice/dao/PostgresJobDaoTest.java
davidangb: I struggled with this - I wanted to push as much logic to the db as possible to avoid false-positives or false-negatives because I wrote a bad `findById` method. I will watch for fragility here.

ashanhol (approved):
I'm fine slowly refactoring our Job.java out as we make the APIs that use it (cloning, etc) move to async/quartz. It's definitely confusing though so any comment clarity we can provide will make the experience better

PR #364, opened 2023-10-11
AJ-1074 - Have WDS manage certain config variables on its own vs getting them from Leo

yuliadub:
I choose to use Spring Profiles since it is built in functionality that provides easy configuration per profile. The main change that we will need in Leo is for Leo to provide the profile name using the following env variable "spring_profiles_active". Based on that value WDS will first load the default application.yaml and then the specific profile (specific profile will overwrite any values application already defined, for example in local profile I am setting workspace_id) 

I created 4 new profiles:
- local (should be nice since now we only need to set 1 env variable to run locally, this profile will care of the rest)
- dev
- prod
- bee

Will need to make change in Leo to exclude all the values moved to profile and instead set the spring_profiles_active for WDS and provide a BEE name when one is present. 

Reminder:


README.md
```diff
@@ -37,9 +37,22 @@ Instead use the `run_postgres.sh` script to set up a docker container running po
...
+```
+
+Other profiles that are available are: 
+- dev
+- prod
+- bee
+
+You are unlikely to use prod and bee when running locally, those profiles are leveraged when WDS is deployed in Terra. 
+
+If you would like to not use a profile, you can set the following enviorment variables manually. The variables that need to be set are described below.
```
jladieu, line 55:
```suggestion
If you would like to not use a profile, you can set the following environment variables manually. The variables that need to be set are described below. 
```

README.md
```diff
@@ -37,9 +37,22 @@ Instead use the `run_postgres.sh` script to set up a docker container running po
...
 
-For WDS to work properly, several environment variables are expected.
-These should be configured before running WDS.
-You may want to add them to your `~/.zshrc` or similar shell profile.
+For WDS to work properly, a spring profile needs to be set so that several environment variables are populated correctly.
+
+To run WDS locally and therefore use the local application properties, set the profile the following way: 
+
+```bash
+export spring_profiles_active=local
```
jladieu, line 45: Does `spring_profiles_active` need to be downcased to work correctly?  If not, it might be nice to use `SCREAMING_SNAKE_CASE` for consistency with how most of our other custom environment variables are set.

service/src/main/java/org/databiosphere/workspacedataservice/generated/ImportRequestServerModel.java
```diff
@@ -24,7 +24,7 @@
  */
 
 @JsonTypeName("ImportRequest")
-@Generated(value = "org.openapitools.codegen.languages.SpringCodegen", date = "2023-10-05T15:39:06.900914-04:00[America/New_York]")
+@Generated(value = "org.openapitools.codegen.languages.SpringCodegen", date = "2023-10-11T15:46:55.015375076-07:00[America/Los_Angeles]")
```
jladieu, line 27: @yuliadub no action needed on this comment, just leaving it here to use as a pointer from another PR to highlight that the generated timestamp diff occurs even in the absence of an API change.

Original context: https://github.com/DataBiosphere/terra-workspace-data-service/pull/354#discussion_r1347917553

README.md
calypsomatic: TIL: SCREAMING SNAKE CASE 😱 🐍

service/src/main/resources/application-bee.yml
```diff
@@ -0,0 +1,8 @@
+beename: ${BEE_NAME}
+
+sentry:
+  samurl: https://sam.${beename}.broadinstitute.org/
+
+datarepourl: https://jade.${beename}.bee.envs-terra.bio
```
calypsomatic, line 6: Isn't it `https://data.${beename}.bee.envs-terra.bio`?  Or am I confusing that with something else?

service/src/main/resources/application-prod.yml
```diff
@@ -0,0 +1,5 @@
+samurl: https://sam.dsde-prod.broadinstitute.org/
+
+datarepourl: https://jade.datarepo-prod.broadinstitute.org/
+workspacemanagerurl: https://workspace.dsde-prod.broadinstitute.org/
+leoUrl: https://leonardo.dsde-prod.broadinstitute.org/
```
jladieu, line 5: Eeeee I wanted to make a suggestion to have all your stuff cased the same, but I realized that's not reasonable because we use so many different types of naming convention in our config files.  Just leaving this here to point out I created a ticket to converge our configurables onto some...any naming pattern: https://broadworkbench.atlassian.net/browse/AJ-1376

jladieu:
This is gonna be so great!  Thanks for working on this.  I like the changes in general but agree with @calypsomatic's observation that the TDR URL might need to be updated.  Approving under the assumption that you'll work that out :)

jladieu (approved):

README.md
```diff
@@ -37,9 +37,22 @@ Instead use the `run_postgres.sh` script to set up a docker container running po
...
+```
+
+Other profiles that are available are: 
+- dev
+- prod
+- bee
```
davidangb, line 51: we'll need staging and alpha profiles too (at least until alpha goes away), otherwise WDS won't work in those environments, right?

README.md
davidangb: agreed on using caps, that's what Spring recommends - I can try to dig up the doc for that if necessary

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamConfig.java
```diff
@@ -14,7 +14,7 @@
 @Configuration
 public class SamConfig {
 
-  @Value("${SAM_URL:}")
+  @Value("${sentry.samurl:}")
```
davidangb, line 17: I don't think the sam url should be nested under "sentry" - Sam and Sentry are separate systems/concepts

README.md
yuliadub: ah good point, since I assume both staging and alpha have different control pane configs

README.md
yuliadub: 🤦

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamConfig.java
yuliadub: I kept it this way since that is how it was in the config originally, assuming there was a specific reason for it. will move!

service/src/main/resources/application-bee.yml
yuliadub: I started to doubt myself but this tells me this looks right :) https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/2838429697/Terra+Services+URLs

yuliadub:
> A few comments inline - overall I like this approach and am so glad to see it happening.
> 
> We should review [everything Leo sends](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/WdsAppInstall.scala) and see if anything else can move over - for instance, the `instrumentationEnabled` value.
> 
> A kinda pedantic note: I don't think that Leo should know anything about WDS's Spring profiles. Leo should just send an "env" specification of prod, bee, dev, etc. …. then it's up to WDS and WDS's Helm chart to translate the "env" into a profile. This gives use more freedom in the future to interpret Leo's "env" in different ways.

do we actually use instrumentationEnabled? I cant seem to find it referenced anywhere in terra-helm.

davidangb:
> do we actually use instrumentationEnabled? I cant seem to find it referenced anywhere in terra-helm.

I think you're right. From memory, we had problems coordinating delivery of the `instrumentationEnabled` out of Leo, so we bypassed it and did our own thing, detecting the environment based on the Sam url and turning Sentry on/off condition on the environment: https://github.com/DataBiosphere/terra-workspace-data-service/blob/ff233679c956e4d35ebd24c9ee875903d017efb5/service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java#L52

So, we could revisit `instrumentationEnabled` and use that instead of our environment-detection, or just remove it

README.md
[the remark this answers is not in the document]
yuliadub: really good point. Spent time experimenting with this and reading about use of @profile in code vs letting spring decide based on the env variable. Its all very interesting. For now, since the profiles we do have all have to do with local testing and we only use those in code, I changed the name to represent specifically what they are related to and to ensure they dont clash with each other. 

However, I am going to make a follow up ticket to figure out how we can more the @profile values in code into the config files, plus investigate more around some of the claim the internet makes about avoiding notations like @profile("!local").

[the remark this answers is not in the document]
yuliadub: created this ticket to five more into this: https://broadworkbench.atlassian.net/browse/AJ-1413

README.md
```diff
@@ -37,9 +37,24 @@ Instead use the `run_postgres.sh` script to set up a docker container running po
 
 #### Environment Variables
 
-For WDS to work properly, several environment variables are expected.
-These should be configured before running WDS.
-You may want to add them to your `~/.zshrc` or similar shell profile.
+For WDS to work properly, a spring profile needs to be set so that several environment variables are populated correctly.
+
+To run WDS locally and therefore use the local application properties, set the profile the following way:
```
ashanhol, line 42: I personally found the line about adding to a shell profile a useful reminder, just a plain export call disappears when the shell window closes.

ashanhol (approved):

davidangb (approved):
:+1: for syntax.

Before you merge - what is the rollout plan for this? In other words, what is the order of operations for merging PRs in Leo, terra-helmfile, WDS and releasing those to production? We need to make sure that we don't break older versions of WDS (see also: what's our plan for updating existing WDS deployments in prod), or break the app-upgrade process.

service/src/main/resources/application-local.yml
```diff
@@ -0,0 +1,9 @@
+samurl: http://localhost:9889
+datarepourl: https://jade.datarepo-dev.broadinstitute.org/
+workspacemanagerurl: https://workspace.dsde-dev.broadinstitute.org/
+leoUrl: https://leonardo.dsde-dev.broadinstitute.org/
+
+twds:
+  instance:
+    # Workspace Id for launching instance
+    workspace-id: 123e4567-e89b-12d3-a456-426614174000
```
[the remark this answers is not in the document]
yuliadub: in local setting it really doesnt matter since there is nothing to collide with except your own stuff - be default according to our instruction that is what we recommend everyone set it to when running locally. 

setting it here saves a step of having to remember to manually set it yourself.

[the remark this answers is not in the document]
nawatts: Doesn't this need to be set to a workspace that the developer has access to though?

From what I understand, there's no harm if it can be easily overridden by environment variables, but it doesn't seem like a default provides a lot of value since it's not likely to be usable by many developers.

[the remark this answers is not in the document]
yuliadub: yeah, that could be it - when I run wds locally I start it from the command line and I have to pass all the variables to it before I can run it for things to work. Having those things already set for me saves me a few steps. 

you do bring a good point though - when you run Terra UI locally with WDS how do you set the workspace ID? Do you set it the same way for Terra-UI (i.e. via env variable or does terra-ui generate one somehow?)? Could we just default local Terra-UI/WDS to always be the same workspace ID or is that not reasonable?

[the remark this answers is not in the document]
nawatts: Terra UI loads the workspaces list from Rawls. So I've been setting WDS' workspace ID to match one of the workspaces I have access to.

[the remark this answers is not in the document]
yuliadub: chatted with David a bit more about this - added a change to use env variable unless it is empty! let me know what you think of that :)

jladieu (approved):

service/src/main/resources/application-local.yml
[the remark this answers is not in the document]
nawatts: Nice! That sounds good to me.

service/src/main/resources/application-local.yml
```diff
@@ -0,0 +1,9 @@
+samurl: http://localhost:9889
```
davidangb, line 1: I expected this to be `https://sam.dsde-dev.broadinstitute.org/`, for real Sam permission checks instead of bypassing Sam and requiring a local nginx mock

service/src/main/resources/application-prod.yml
```diff
@@ -0,0 +1,4 @@
+samurl: https://sam.dsde-prod.broadinstitute.org/
+datarepourl: https://data.terra.bio/
+workspacemanagerurl: https://workspace.dsde-prod.broadinstitute.org/
+leoUrl: https://leonardo.dsde-prod.broadinstitute.org/
```
davidangb, line 4: Do you know what we use in production today? Is it `https://leonardo.dsde-prod.broadinstitute.org/` or `https://notebooks.firecloud.org/`?

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializer.java
```diff
@@ -7,7 +7,7 @@
 import org.springframework.stereotype.Component;
 
 @Component
-@Profile({"!local"})
+@Profile({"!local-cors"})
```
davidangb, line 10: Should this one remain as `"!local"`? Or even remove it entirely? I'm not even sure of the intent here … when running locally, don't create the default instance automatically?

davidangb (commented):
A few more comments inline!

service/src/main/resources/application-prod.yml
yuliadub: referencing what leo uses looks like it could be either (https://github.com/broadinstitute/terra-helmfile/blob/master/values/app/leonardo/live/prod.yaml) but the param that we pass for Leo url into WDS today is "https://leonardo.dsde-prod.broadinstitute.org/"

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializer.java
yuliadub: I am honestly not sure why it is where and I only changed it to !local-cors so it follows what the definition of this was before the config changes (when the local profile was just the cors stuff), I created this follow up ticket to decide if we should remove or not in a separate PR. https://broadworkbench.atlassian.net/browse/AJ-1413

service/src/main/resources/application-local.yml
```diff
@@ -0,0 +1,9 @@
+samurl: https://sam.dsde-dev.broadinstitute.org/
```
nawatts, line 1: Does this still allow using a local mock Sam?

Or does it need to do something like `twds.instance.workspace-id`, where it can override the default with an environment variable?

yuliadub: if not suing profiles, you can set all the values vie env variables, but it is a good point to keep this flexible as well! let me adjust

nawatts: Alternatively/additionally, we could add like a `local-sam` profile that sets this. Then WDS could be run with `SPRING_PROFILES_ACTIVE=local,local-sam`.

README.md
```diff
@@ -109,7 +125,7 @@ environment variable WDS will use on the next startup to connect to Sam.
 ```bash
 # start the SAM mock as a docker container in detached mode
 docker run -v `pwd`/service/src/test/resources/nginx.conf:/etc/nginx/nginx.conf -p 9889:80 -d nginx:1.23.3
-export SAM_URL=http://localhost:9889
+export SAM_URL=http://localhost:9889 # this is automatically set if you set the profile to "local"
```
nawatts, line 128: After https://github.com/DataBiosphere/terra-workspace-data-service/pull/364/commits/b78c13cd91612d26a3ea891de3441ce2b87d7412, this comment is no longer accurate.

test.txt
nawatts, line 1: Was this intended to be committed?

yuliadub: ha definitely not, thanks for catching that.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializer.java
davidangb: this was my only outstanding question and why I self-requested another review on this. I definitely don't think that a cors-related profile should control whether or not WDS creates a default instance at startup. This was added in https://github.com/DataBiosphere/terra-workspace-data-service/pull/190, and reading back to that PR I believe we added the `@Profile` here because when running WDS locally, we have no `startup-token` and therefore would fail to create a resource in Sam during instance initialization. I think this line should switch back to `@Profile({"!local"})`.

yuliadub: switched!

davidangb (approved):

PR #365, opened 2023-10-12
Re-enable `hideGenerationTimestamp`

jladieu:
This is to eliminate diff noise for unrelated changes.

Ideally we could use the timestamp, but we need to solve the problem of having it only run when there are relevant changes first.


jladieu:
This is a followup to https://github.com/DataBiosphere/terra-workspace-data-service/pull/354#discussion_r1356780712 after finding timestamps creeping into unrelated changes, example: https://github.com/DataBiosphere/terra-workspace-data-service/pull/364/files#r1356775131

davidangb (approved):

calypsomatic (approved):

PR #367, opened 2023-10-12
AJ-1011: snapshot import via Quartz

davidangb:
connects the dots and performs async snapshot import via Quartz. Well, as of this PR the import fails because we're not properly passing a token. But you can see that it's trying, and token work will be a follow-on PR ([Slack thread](https://broadinstitute.slack.com/archives/C02SJQ0FKHQ/p1697205316686489))

To test this locally:
1. Export a snapshot from TDR via the [exportSnapshot API](https://jade.datarepo-dev.broadinstitute.org/swagger-ui.html#/snapshots/exportSnapshot) and copy the url to the exported manifest
2. Call WDS's `importV1` API with a payload of `{"type": "TDRMANIFEST", "url": "url-to-manifest-from-previous-step"}` and make note of the jobId it returns
3. Call WDS's `jobStatusV1` API, passing the jobId from the previous step
4. View the WDS logs to see progress of the job

For future PRs:
* Properly pass an auth token from the Quartz job to the WSM and TDR API calls
* return some kind of response payload from the import, such as the number of rows affected
* Update the OpenAPI spec to define the expected response payloads
* always can use more test coverage

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -78,7 +78,7 @@ dependencies {
     implementation group: 'org.broadinstitute.dsde.workbench', name: 'sam-client_2.13', version: '0.1-08b6588'
     implementation group: 'org.broadinstitute.dsde.workbench', name: 'leonardo-client_2.13', version: '1.3.6-22ee00b'
     implementation 'com.squareup.okhttp3:okhttp:4.11.0' // required by Sam client
-    implementation "bio.terra:datarepo-client:1.476.0-SNAPSHOT"
+    implementation "bio.terra:datarepo-client:1.537.0-SNAPSHOT"
     implementation "bio.terra:workspace-manager-client:0.254.717-SNAPSHOT"
     implementation "bio.terra:java-pfb-library:0.5.0"
     implementation project(path: ':client')
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/QuartzSchedulerDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/QuartzSchedulerDao.java
@@ -23,7 +23,7 @@ public class QuartzSchedulerDao implements SchedulerDao {
   }
 
   public void schedule(Schedulable schedulable) {
-    JobKey jobKey = new JobKey(schedulable.getName(), schedulable.getGroup());
+    JobKey jobKey = new JobKey(schedulable.getId(), schedulable.getGroup());
 
     // TODO: read JobData from the Job's inputs, instead of hardcoding it here
     // TODO: choose the implementing class based on the job's type, instead of hardcoding
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/QuartzSchedulerDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/QuartzSchedulerDao.java
@@ -25,9 +25,6 @@ public class QuartzSchedulerDao implements SchedulerDao {
   public void schedule(Schedulable schedulable) {
     JobKey jobKey = new JobKey(schedulable.getId(), schedulable.getGroup());
 
-    // TODO: read JobData from the Job's inputs, instead of hardcoding it here
-    // TODO: choose the implementing class based on the job's type, instead of hardcoding
-    // ImportQuartzJob.class
     JobDetail jobDetail =
         JobBuilder.newJob()
             .ofType(schedulable.getImplementation())
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/QuartzSchedulerDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/QuartzSchedulerDao.java
@@ -2,10 +2,7 @@ package org.databiosphere.workspacedataservice.dao;
 
 import static org.quartz.TriggerBuilder.newTrigger;
 
-import org.databiosphere.workspacedataservice.dataimport.ImportQuartzJob;
-import org.databiosphere.workspacedataservice.shared.model.job.Job;
-import org.databiosphere.workspacedataservice.shared.model.job.JobInput;
-import org.databiosphere.workspacedataservice.shared.model.job.JobResult;
+import org.databiosphere.workspacedataservice.shared.model.Schedulable;
 import org.quartz.JobBuilder;
 import org.quartz.JobDetail;
 import org.quartz.JobKey;
@@ -25,19 +22,19 @@ public class QuartzSchedulerDao implements SchedulerDao {
     this.scheduler = scheduler;
   }
 
-  public void schedule(Job<JobInput, JobResult> job) {
-    JobKey jobKey = new JobKey(job.getJobId().toString(), job.getJobType().name());
+  public void schedule(Schedulable schedulable) {
+    JobKey jobKey = new JobKey(schedulable.getName(), schedulable.getGroup());
 
     // TODO: read JobData from the Job's inputs, instead of hardcoding it here
     // TODO: choose the implementing class based on the job's type, instead of hardcoding
     // ImportQuartzJob.class
     JobDetail jobDetail =
         JobBuilder.newJob()
-            .ofType(ImportQuartzJob.class)
+            .ofType(schedulable.getImplementation())
             .withIdentity(jobKey)
-            .usingJobData("token", "proof-of-concept-token")
+            .setJobData(schedulable.getArgumentsAsJobDataMap())
             .storeDurably(false) // delete from the quartz table after the job finishes
-            .withDescription("Import for job " + job.getJobId().toString())
+            .withDescription(schedulable.getDescription())
             .build();
 
     // tell Quartz to run the job: run only once, start immediately
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportQuartzJob.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportQuartzJob.java
@@ -1,58 +1,8 @@
 package org.databiosphere.workspacedataservice.dataimport;
 
-import java.util.UUID;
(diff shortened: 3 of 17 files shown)


service/build.gradle
```diff
@@ -78,7 +78,7 @@ dependencies {
     implementation group: 'org.broadinstitute.dsde.workbench', name: 'sam-client_2.13', version: '0.1-08b6588'
     implementation group: 'org.broadinstitute.dsde.workbench', name: 'leonardo-client_2.13', version: '1.3.6-22ee00b'
     implementation 'com.squareup.okhttp3:okhttp:4.11.0' // required by Sam client
-    implementation "bio.terra:datarepo-client:1.476.0-SNAPSHOT"
+    implementation "bio.terra:datarepo-client:1.537.0-SNAPSHOT"
```
davidangb, line 81: required to read the TDR manifest; the manifest has fields that are not recognized by `1.476.0-SNAPSHOT`

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/Schedulable.java
```diff
@@ -0,0 +1,63 @@
+package org.databiosphere.workspacedataservice.shared.model;
```
jladieu, line 1: Love the thoughtful interface design here :)

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceTest.java
```diff
@@ -0,0 +1,67 @@
...
+    UUID jobId = UUID.randomUUID();
+    Map<String, Serializable> arguments = Map.of("foo", "bar", "twenty-three", 23);
+
+    Schedulable actual = importService.createSchedulable(importType, jobId, arguments);
+
+    JobDataMap expectedJobDataMap = new JobDataMap();
+    expectedJobDataMap.put("foo", "bar");
+    expectedJobDataMap.put("twenty-three", 23);
+
+    assertEquals(jobId.toString(), actual.getName());
```
jladieu, line 41: As I read this, right above the other assertion, which is looking at the `importType.name()`, I am wondering if it'd be better to just rename `Schedule.getName()` to `Schedulable.getId()`... thoughts?  Having to hold two different notions of `name` in my head at the same time is making my brain hurt :D

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
```diff
@@ -20,6 +22,10 @@ public GenericJobServerModel getJob(UUID instanceUuid, UUID jobId) {
     // validate instance exists
     instanceService.validateInstance(instanceUuid);
 
-    return jobDao.getJob(jobId);
+    try {
+      return jobDao.getJob(jobId);
+    } catch (EmptyResultDataAccessException e) {
+      throw new MissingObjectException("Job");
```
jladieu, line 28: might `e` contain any useful information that we lose here?  Should it be wrapped within `MissingObjectException`?

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
```diff
@@ -50,20 +54,45 @@ public GenericJobServerModel createImport(
...
   }
+
+  protected Schedulable createSchedulable(
+      ImportRequestServerModel.TypeEnum importType,
+      UUID jobId,
+      Map<String, Serializable> arguments) {
+    return switch (importType) {
+      case TDRMANIFEST -> new TdrManifestSchedulable(
+          jobId.toString(), "TDR manifest import", arguments);
+      case PFB -> new PfbSchedulable(jobId.toString(), "TODO: PFB import", arguments);
```
jladieu, line 95: I've relied previously on linters telling me to always add in a `default` clause, or provide a comprehensive accounting of enum values in my switch statements.

Does our tooling / Java / give any guarantee right now that this switch is comprehensive?  If not, I'm wondering if you should put in a catchall default that crashes if reached.

service/src/main/java/org/databiosphere/workspacedataservice/dao/QuartzSchedulerDao.java
```diff
@@ -25,19 +22,19 @@ public QuartzSchedulerDao(Scheduler scheduler) {
     this.scheduler = scheduler;
   }
 
-  public void schedule(Job<JobInput, JobResult> job) {
-    JobKey jobKey = new JobKey(job.getJobId().toString(), job.getJobType().name());
+  public void schedule(Schedulable schedulable) {
+    JobKey jobKey = new JobKey(schedulable.getName(), schedulable.getGroup());
 
     // TODO: read JobData from the Job's inputs, instead of hardcoding it here
     // TODO: choose the implementing class based on the job's type, instead of hardcoding
```
calypsomatic, line 29: these TODOs look done to me

service/src/main/java/org/databiosphere/workspacedataservice/dao/QuartzSchedulerDao.java
```diff
@@ -25,19 +22,19 @@ public QuartzSchedulerDao(Scheduler scheduler) {
     this.scheduler = scheduler;
   }
 
-  public void schedule(Job<JobInput, JobResult> job) {
-    JobKey jobKey = new JobKey(job.getJobId().toString(), job.getJobType().name());
+  public void schedule(Schedulable schedulable) {
+    JobKey jobKey = new JobKey(schedulable.getName(), schedulable.getGroup());
 
     // TODO: read JobData from the Job's inputs, instead of hardcoding it here
```
jladieu, line 28: I think you took care of these TODOs quite nicely!

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportQuartzJob.java
```diff
@@ -1,58 +1,8 @@
...
-      jobDao.updateStatus(jobId, GenericJobServerModel.StatusEnum.SUCCEEDED);
-    } catch (InterruptedException e) {
-      // ensure the thread is actually interrupted
-      Thread.currentThread().interrupt();
-      jobDao.fail(jobId, e.getMessage());
-      throw new JobExecutionException(e);
-    }
-  }
+/** Base class for data-import Quartz jobs, containing constants and utilities for data imports. */
+public abstract class ImportQuartzJob extends QuartzJob {
```
jladieu, line 6: Can you get by without this extra abstract class in between the implementation and `QuartzJob`?  That it is just a home for a single constant (for the moment) makes me think it might be unneeded.

jladieu (approved):
Beautiful work!

Few minor questions is all.

calypsomatic (commented):
In addition to the documentation I've demanded around how to write an API/model, I would like to also demand some documentation on the different classes and general flow of an async job - just looking at the code (even if it's beautifully commented) I start to lose track of what is a Job vs a Schedulable and who does what when.  Can you throw a pretty little diagram or something into the wiki or somewhere?  Please and thank you!

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportQuartzJob.java
yuliadub: also wondering the same thing

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
davidangb: In this particular case, there's nothing meaty in there. `EmptyResultDataAccessException` is:
"Data access exception thrown when a result was expected to have at least one row (or element) but zero rows (or elements) were actually returned."

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
davidangb: Java fails compilation in this case, and IntelliJ shows a red squiggly. From `./gradlew compileJava`:
```
> Task :service:compileJava FAILED
/Users/davidan/work/src/terra-workspace-data-service/service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java:91: error: the switch expression does not cover all possible input values
    return switch (importType) {
           ^
1 error

FAILURE: Build failed with an exception.
```

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceTest.java
davidangb: done!

service/src/main/java/org/databiosphere/workspacedataservice/dao/QuartzSchedulerDao.java
davidangb: removed

service/src/main/java/org/databiosphere/workspacedataservice/dao/QuartzSchedulerDao.java
davidangb: removed

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportQuartzJob.java
davidangb: yeah, this is premature optimization, removed.

davidangb:
> I would like to also demand some documentation on the different classes and general flow of an async job

@calypsomatic yes, absolutely, I will do that right after I get this PR buttoned up

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
jladieu: Thank you!

jladieu (approved):

PR #323, opened 2023-08-10
AJ-1263 Fix `IllegalStateException` on startup while polling SAM status.

jladieu:
AJ-1263 Fix `IllegalStateException` on startup while polling SAM status.

When starting WDS locally, the startup process attempts to poll SAM status outside the context of a user request.  Spring throws an error when `RequestContextHolder.currentRequestAttributes` is called outside the context of a web request, which caused the below crash indicating SAM was down.

This fix adds a check to only use `currentRequestAttributes` in the context of a request.  Otherwise the reqeust will be treated as unauthenticated via an access token.

```
2023-08-10 11:32:16.535  INFO [] [restartedMain] o.d.w.logging.AvailabilityListener       : ReadinessState: ACCEPTING_TRAFFIC
2023-08-10 11:32:16.946 DEBUG [] [RMI TCP Connection(2)-10.1.7.98] o.d.w.sam.HttpSamClientSupport           : Sending getSystemStatus request to Sam ...
2023-08-10 11:32:16.951 ERROR [] [RMI TCP Connection(2)-10.1.7.98] o.d.w.sam.HttpSamClientSupport           : getSystemStatus Sam request resulted in No thread-bound request found: Are you referring to request attributes outside of an actual web request, or processing a request outside of the originally receiving thread? If you are actually operating within a web request and still receive this message, your code is probably running outside of DispatcherServlet: In this case, use RequestContextListener or RequestContextFilter to expose the current request.

java.lang.IllegalStateException: No thread-bound request found: Are you referring to request attributes outside of an actual web request, or processing a request outside of the originally receiving thread? If you are actually operating within a web request and still receive this message, your code is probably running outside of DispatcherServlet: In this case, use RequestContextListener or RequestContextFilter to expose the current request.
	at org.springframework.web.context.request.RequestContextHolder.currentRequestAttributes(RequestContextHolder.java:131)
	at org.databiosphere.workspacedataservice.sam.HttpSamClientFactory.getApiClient(HttpSamClientFactory.java:50)
	at org.databiosphere.workspacedataservice.sam.HttpSamClientFactory.getStatusApi(HttpSamClientFactory.java:85)
	at org.databiosphere.workspacedataservice.sam.HttpSamDao.lambda$getSystemStatus$2(HttpSamDao.java:111)
	at org.databiosphere.workspacedataservice.sam.HttpSamClientSupport.withRetryAndErrorHandling(HttpSamClientSupport.java:57)
[...]

PR #370, opened 2023-10-16
AJ-1011: tokens in quartz jobs

davidangb:
Reviewer: this PR has Java threading and `ThreadLocal` fun, please review that carefully as it is easy to get wrong!

In this PR:
* Add `JobContextHolder`, modeled after Spring's `RequestContextHolder`, to allow Quartz jobs to stash a token and allow that token to be retrieved from DAOs, just like we do for http requests
* Add `TokenContextUtil`, to centralize retrieval of auth tokens from either request context or job context
* Modify all REST DAOs, as well as ActivityEventBuilder, to use `TokenContextUtil` instead of their mostly copy/pasted custom logic
* Modify `ImportService` to use the current user's token for use in data import jobs (this will change to be a pet token in later PRs)
* Modify `QuartzJob` to read a token from the job's JobDataMap, stash it in job context, then clear job context after the job finishes

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
@@ -1,5 +1,8 @@
 package org.databiosphere.workspacedataservice.jobexec;
 
+import static org.databiosphere.workspacedataservice.sam.BearerTokenFilter.ATTRIBUTE_NAME_TOKEN;
+import static org.databiosphere.workspacedataservice.shared.model.Schedulable.ARG_TOKEN;
+
 import java.net.URL;
 import java.util.UUID;
 import org.databiosphere.workspacedataservice.dao.JobDao;
@@ -37,6 +40,13 @@ public abstract class QuartzJob implements Job {
     try {
       // mark this job as running
       getJobDao().updateStatus(jobId, GenericJobServerModel.StatusEnum.RUNNING);
+      // look for an auth token in the Quartz JobDataMap
+      String authToken = getJobDataString(context.getMergedJobDataMap(), ARG_TOKEN);
+      // and stash the auth token into job context
+      if (authToken != null) {
+        JobContextHolder.init();
+        JobContextHolder.setAttribute(ATTRIBUTE_NAME_TOKEN, authToken);
+      }
       // execute the specifics of this job
       executeInternal(jobId, context);
       // if we reached here, mark this job as successful
@@ -44,6 +54,8 @@ public abstract class QuartzJob implements Job {
     } catch (Exception e) {
       // on any otherwise-unhandled exception, mark the job as failed
       getJobDao().fail(jobId, e.getMessage(), e.getStackTrace());
+    } finally {
+      JobContextHolder.destroy();
     }
   }
 
--- a/service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamClientFactory.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamClientFactory.java
@@ -1,8 +1,5 @@
 package org.databiosphere.workspacedataservice.sam;
 
-import static org.databiosphere.workspacedataservice.sam.BearerTokenFilter.ATTRIBUTE_NAME_TOKEN;
-import static org.springframework.web.context.request.RequestAttributes.SCOPE_REQUEST;
-
 import java.util.List;
 import java.util.Objects;
 import okhttp3.OkHttpClient;
@@ -14,7 +11,6 @@ import org.broadinstitute.dsde.workbench.client.sam.api.StatusApi;
 import org.broadinstitute.dsde.workbench.client.sam.api.UsersApi;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.web.context.request.RequestContextHolder;
 
 /**
  * Implementation of SamClientFactory that creates a Sam ApiClient, initializes that client with the
@@ -38,7 +34,7 @@ public class HttpSamClientFactory implements SamClientFactory {
     // this requires we import terra-common-lib
   }
 
-  private ApiClient getApiClient(String accessToken) {
+  private ApiClient getApiClient(String authToken) {
     // create a new Sam client
     ApiClient apiClient = new ApiClient();
     apiClient.setHttpClient(commonHttpClient);
@@ -46,21 +42,18 @@ public class HttpSamClientFactory implements SamClientFactory {
     if (StringUtils.isNotBlank(samUrl)) {
       apiClient.setBasePath(samUrl);
     }
-    if (accessToken == null) {
-      // grab the current user's bearer token (see BearerTokenFilter)
-      Object token =
-          RequestContextHolder.currentRequestAttributes()
-              .getAttribute(ATTRIBUTE_NAME_TOKEN, SCOPE_REQUEST);
-      // add the user's bearer token to the client
-      if (!Objects.isNull(token)) {
-        LOGGER.debug("setting access token for Sam request");
-        apiClient.setAccessToken(token.toString());
-      } else {
-        LOGGER.warn("No access token found for Sam request.");
-      }
+
+    // grab the current user's bearer token (see BearerTokenFilter) or use parameter value
+    String token = TokenContextUtil.getToken(authToken, () -> null);
+
+    // add the user's bearer token to the client
+    if (!Objects.isNull(token)) {
+      LOGGER.debug("setting access token for Sam request");
+      apiClient.setAccessToken(token);
     } else {
-      apiClient.setAccessToken(accessToken);
+      LOGGER.warn("No access token found for Sam request.");
     }
+
     // return the client
     return apiClient;
   }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamClientFactory.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamClientFactory.java
@@ -44,7 +44,7 @@ public class HttpSamClientFactory implements SamClientFactory {
     }
 
     // grab the current user's bearer token (see BearerTokenFilter) or use parameter value
-    String token = TokenContextUtil.getToken(authToken, () -> null);
+    String token = TokenContextUtil.getToken(authToken);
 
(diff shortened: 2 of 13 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamClientFactory.java
```diff
@@ -38,29 +34,26 @@ public HttpSamClientFactory(String samUrl) {
     // this requires we import terra-common-lib
   }
 
-  private ApiClient getApiClient(String accessToken) {
+  private ApiClient getApiClient(String authToken) {
```
jladieu, line 37: I'm thinking more about what it might take to sidestep the `ThreadLocal` complexity without having to do a bunch of property drilling.

To make the thought exercise slightly more concrete:

What if `HttpSamClientFactory` (and its siblings) took the authToken as a 2nd param to their constructor.

Then, instead of providing the factory as an injected `@Bean`, you'd provide something like a `ClientFactoryProvider` (although I realize the name is getting silly at this point) whose main signature was something like:

`ClientFactoryProvider.gimmeClientFactoryForAuthentication(token)`, which would serve up a a `ClientFactory` that already has the token injected.

You could also provide a convenience method like `ClientFactoryProvider.gimmeClientFactoryForJobContext(...)` that knows how to figure out the auth token from the JobContext (or RequestContext, or whatever?).

Or even, 'ClientFactoryProvider.gimmeUnauthenticatedFactory(...)` if that has any sensible path.  This might allow us to hide all the auth token stuff from a bunch of business logic and keep it restricted to a few key chokepoints where we have the auth token and context in hand at the same time.

service/src/main/java/org/databiosphere/workspacedataservice/sam/TokenContextUtil.java
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  /**
+   * Look in RequestContextHolder and then JobContextHolder, in that order, for a non-null String
+   * ATTRIBUTE_NAME_TOKEN. If none found, return the value of orElse.
+   *
+   * @param orElse the value to return if the token was not found otherwise
+   * @return the final value
+   */
+  public static String getToken(Supplier<String> orElse) {
```
jladieu, line 51: Love providing the `orElse` offramp for the client code to decide what to do here!

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/JobContextHolderTest.java
```diff
@@ -0,0 +1,95 @@
...
+    // semaphore to track when all threads are done
+    CountDownLatch latch = new CountDownLatch(numThreads);
+
+    // generate {numThreads} random String values; these are the values we'll eventually assert on
+    List<String> randomStrings =
+        IntStream.rangeClosed(1, numThreads)
+            .mapToObj(i -> RandomStringUtils.randomAlphabetic(8))
+            .toList();
+
+    /* ----- inner class: the thread to execute during the test ----- */
```
jladieu, line 33: This is just a 🤯 level of awesome coverage.  Even if I end up convincing you to avoid ThreadLocal, I feel like this unit test deserves to be boosted & memorialized for how to actually test thread aware code for safety.

service/src/main/java/org/databiosphere/workspacedataservice/sam/TokenContextUtil.java
```diff
@@ -0,0 +1,118 @@
...
+    String foundToken = getToken();
+    if (foundToken != null) {
```
jladieu, line 53: should we also check for non-empty?

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/HttpWorkspaceManagerClientFactory.java
```diff
@@ -37,16 +33,12 @@ private ApiClient getApiClient(String authToken) {
     }
 
     // grab the current user's bearer token (see BearerTokenFilter) or use parameter value
-    Object token =
-        StringUtils.isNotBlank(authToken)
-            ? authToken
-            : RequestContextHolder.currentRequestAttributes()
-                .getAttribute(ATTRIBUTE_NAME_TOKEN, SCOPE_REQUEST);
+    String token = TokenContextUtil.getToken(authToken, () -> null);
```
jladieu, line 36: I think returning `null` doesn't do a great job of describing the behavior that happens as a result of doing so.  My other suggestion to more strongly type the authToken might help with this, but I think using a constant with an intention-revealing name might also do the trick:

This is a snippet from a [PR](https://github.com/DataBiosphere/terra-workspace-data-service/pull/323/files#r1361225123) I abandoned while trying this out:
```
    public static final String USE_BEARER_TOKEN_IF_PRESENT = null;
```

service/src/main/java/org/databiosphere/workspacedataservice/sam/TokenContextUtil.java
yuliadub: also check the same for line 39?

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
```diff
@@ -37,13 +40,22 @@ public void execute(JobExecutionContext context) throws org.quartz.JobExecutionE
     try {
       // mark this job as running
       getJobDao().updateStatus(jobId, GenericJobServerModel.StatusEnum.RUNNING);
+      // look for an auth token in the Quartz JobDataMap
+      String authToken = getJobDataString(context.getMergedJobDataMap(), ARG_TOKEN);
+      // and stash the auth token into job context
+      if (authToken != null) {
+        JobContextHolder.init();
+        JobContextHolder.setAttribute(ATTRIBUTE_NAME_TOKEN, authToken);
```
yuliadub, line 48: random thought for discussion: when we set the auth Token, do we want to set the time it expires? I know we talked about the life of this token being very short, would it be in any way helpful to know exactly when we got it so we know exactly when it will expire? Or would this be a safer property to not store due to security reasons? 

basically im trying to think if having the expiration time to check against can help make some token related errors more specific and easier to parse through in the future. but I am not actually sure it will be that useful, just a thought that came to mind as I was looking at this.

service/src/main/java/org/databiosphere/workspacedataservice/sam/TokenContextUtil.java
davidangb: I waffled on this, and decided not to - it feels like it's bordering on business logic of what's a valid token and what's not. If the user sends a request with `Authorization: Bearer   ` then we'll just accept that and let Sam/other services yell at us about 401s, and we'll bubble up that 401. I could be convinced otherwise, though.

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
davidangb: We don't have the expire time, unless we parse the token itself. I think it's fine to parse the token if we have a reason to do so, but up to now we've been able to treat it as an opaque value - that is, somebody sends WDS a token and then WDS passes that token on to another service without caring about its contents.

If we really needed the expire time, let's parse it - but until we have that use case I think it would just add complexity.

service/src/main/java/org/databiosphere/workspacedataservice/sam/TokenContextUtil.java
```diff
@@ -0,0 +1,116 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.web.context.request.RequestAttributes;
+import org.springframework.web.context.request.RequestContextHolder;
+
+public class TokenContextUtil {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(TokenContextUtil.class);
+
+  private TokenContextUtil() {}
```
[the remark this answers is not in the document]
davidangb: I've updated the logging! I want to be careful about logging too many lines, but I think this is better? LMK what you think.

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/HttpWorkspaceManagerClientFactory.java
davidangb: fyi I misread this comment at first and thought you were talking about the `() -> null` part, and I updated `TokenContextUtil` to not require that - hopefully it makes an incrementally cleaner API, even if it doesn't address your actual comment

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamClientFactory.java
davidangb: As shown in #371 - you still have to get the auth token from the controller layer through to the service layer before you can pass that token to any factory (whether it's a dao factory or a dao client factory or whatever). That's my main takeaway: there will still be a bunch of property drilling - and we'll end up with half drilling, half factory - unless we refactor both the service layer and the dao layer. At that point it's a really big refactor.

We could definitely do something with request-scoped Spring beans - but in my mind that's equivalent to using `RequestContextHolder` as both solutions delegate to spring for creation/deletion of objects in sync with the servlet request lifecycle - and request-scoped Spring beans won't help with async jobs or anything else that isn't an http request.

jladieu (approved):
Go for it!  I think the refactoring into factories is something worth exploring still but doesn't need to hold this up.  Let's park that idea for now.

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamClientFactory.java
jladieu: I agree with not wanting to have half drilling & half factory, and also agree that it could be a pretty big (but potentially worthy) refactor to get all the underlying layers injected with auth context.  Totally ok to punt on this and proceed as is.  Thanks for trying it out to see where the rough edges were.

One thing of note, not blocking at all for this PR: I tend to have an early bias toward the most stateless possible implementation as possible, which I think means I favor request scoped Spring beans over singletons or any bean with objects that may potentially persist between requests, even at the cost of all that extra object creation & teardown (until signs point to that being problematic/costly).  Does Spring (or how it runs at scale) push us in one direction or another?

PR #371, opened 2023-10-16
AJ-1011: POC: pass tokens via dao factories

davidangb:
An alternative approach to #370. This POC only deals with the DataRepoDao.

1. In `DataRepoController`, look at the HTTP request and retrieve any auth token it finds.
2. Pass the auth token by argument into `DataRepoService`
3. In `DataRepoService`, pass the auth token by argument into the new `DataRepoDaoFactory.getDao()` method. This returns a dedicated DataRepoDao instance, configured just for this token.

Under the covers, this also includes a pretty significant refactor of our DAO structure.

Previously, we had:
* DataRepoDao - singleton Spring bean, accepts a DataRepoDaoClientFactory
* DataRepoDaoClientFactory - singleton Spring bean, creates the underlying ApiClient

Now, we have:
* DataRepoDao - POJO, accepts a DataRepoDao.ClientFactory
* DataRepoDaoFactory - singleton Spring bean, creates a DataRepoDao.ClientFactory which creates the underlying ApiClient, then creates a DataRepoDao POJO using that ClientFactory


service/src/main/java/org/databiosphere/workspacedataservice/auth/TokenExtractor.java
davidangb, line 1: I broke this logic out of BearerTokenFilter so it could be easily reused

service/src/main/java/org/databiosphere/workspacedataservice/controller/DataRepoController.java
```diff
@@ -28,10 +30,12 @@ public DataRepoController(DataRepoService dataRepoService, InstanceService insta
   public ResponseEntity<Void> importSnapshot(
       @PathVariable("instanceId") UUID instanceId,
       @PathVariable("version") String version,
-      @PathVariable("snapshotId") UUID snapshotId) {
+      @PathVariable("snapshotId") UUID snapshotId,
+      HttpServletRequest request) {
```
davidangb, line 34: I added the `HttpServletRequest request` argument here, so the controller has access to read a token out of it.

I wonder if this will cause problems for v1 APIs, because this method signature does not match generated signatures - we don't specify `request` arguments in our OpenAPI spec.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataRepoService.java
```diff
@@ -33,19 +33,19 @@ public class DataRepoService {
...
     this.workspaceManagerDao = workspaceManagerDao;
     this.activityLogger = activityLogger;
     this.recordDao = recordDao;
   }
 
-  public void importSnapshot(UUID instanceId, UUID snapshotId) {
+  public void importSnapshot(UUID instanceId, UUID snapshotId, String authToken) {
     // getSnapshot will throw exception if caller does not have access
-    SnapshotModel snapshot = dataRepoDao.getSnapshot(snapshotId);
+    SnapshotModel snapshot = dataRepoDaoFactory.getDao(authToken).getSnapshot(snapshotId);
```
davidangb, line 48: We'll have to update all our `*Service` classes to accept tokens as arguments and then create the appropriate DAOs. Alternately, we could ALSO change our service classes to use factories, and create dedicated service instances in our controllers.

service/src/main/java/org/databiosphere/workspacedataservice/controller/DataRepoController.java
jladieu: Interesting, does the framework pass the `HttpServletRequest` in automatically?  Is it an option to get it injected via constructor, so the method signatures can remain stable?

service/src/main/java/org/databiosphere/workspacedataservice/auth/TokenExtractor.java
```diff
@@ -0,0 +1,31 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpHeaders;
+
+public class TokenExtractor {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(TokenExtractor.class);
+
+  private TokenExtractor() {}
+
+  public static String getToken(ServletRequest request) {
```
jladieu, line 18: Nice!

service/src/main/java/org/databiosphere/workspacedataservice/service/DataRepoService.java
jladieu: I have a hunch that it's a good goal to strive to hide away the auth token from all the business logic stuff _as much as_ possible.

I don't have a good handle on how significant a change it is to have the services use factories, but I think it'd be preferable to having to add the parameters throughout.  Looking to you for your instincts on whether it's too much churn to do all this refactoring.

service/src/test/java/org/databiosphere/workspacedataservice/activitylog/LogStatementTest.java
```diff
@@ -61,7 +62,8 @@ public class LogStatementTest {
       Mockito.mock(ReferencedGcpResourceApi.class);
 
   // mocking for data repo
-  @MockBean DataRepoClientFactory mockDataRepoClientFactory;
+  @MockBean DataRepoDaoFactory mockDataRepoDaoFactory;
```
jladieu, line 65: It looks like one dependency has become two here...is it possible that both of these things could be provided by the same class, and we could depend_just on_ that class?

davidangb: Possibly - it would require some more poking around at this. We separate out the DAO and the ClientFactory to allow for targeted unit testing/mocking:
* the DAO can contain retries and error-handling, so we want to be able to mock/test at its level
* the ClientFactory produces the underling ApiClient, which we want to be able to mock and simulate low-level HTTP exceptions

to be honest, I had a hard time trying to factor all this out, I am also not super pleased with how it turned out

service/src/main/java/org/databiosphere/workspacedataservice/service/DataRepoService.java
yuliadub: as I think of this more, I am not the biggest fan of having all services take tokens as params or wrapping tokens as a concept for a service level variable. I like the simplicity of the current approach of keeping tokens alongside the client creation -- in my opinion it is easier to debug when a given call is failing due to an auth issue or similar. I currently I am not seeing the major benefit of the services using factories approach, but maybe I just need to talk through it a bit more...?

service/src/main/java/org/databiosphere/workspacedataservice/controller/DataRepoController.java
davidangb: hm - potentially, if we changed this controller from being a singleton to being a request-scoped bean, we could set `request` via constructor. That would of course mean a lot more object instantiation.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataRepoService.java
calypsomatic: I'm also not a fan of passing tokens as params.  I prefer the context method - but if you can wrangle something something factories that might work.

service/src/main/java/org/databiosphere/workspacedataservice/controller/DataRepoController.java
jladieu: I tend to have an early bias toward the most stateless possible implementation as possible, which I think means I favor request scoped Spring beans over singletons or any bean with objects that may potentially persist between requests, even at the cost of all that extra object creation & teardown (until signs point to that being problematic/costly).  Does Spring (or how it runs at scale) push us in one direction or another?

service/src/main/java/org/databiosphere/workspacedataservice/service/DataRepoService.java
jladieu: I think what I am advocating for (not necessarily now, but perhaps for a rainy day) is to have the AuthContext become a completely injected dependency for all service/dao/business objects that need to run in an authenticated context.  The runtime/environment where they are running should be aware of the authentication context and provide it as one of the first dependencies.  That would mean going all the way with the inversion of control, and making it so all service objects, DAOs, clients, etc. had the AuthContext provided at construction time.

I think {Request,Job}ContextHolder we _accomplish_ having that AuthContext be _available to_ all these objects, but the objects have to essentially go out to get them, by asking the ContextHolder for that info from its global state, rather than having it just provided as part of creation.

PR #373, opened 2023-10-17
BearerToken model

davidangb:
Incorporate a `BearerToken` class as a safer and more encapsulated option instead of using raw `String`s.


service/src/main/java/org/databiosphere/workspacedataservice/shared/model/BearerToken.java
```diff
@@ -0,0 +1,25 @@
...
+   * Validates that a given String is a valid access token. For the sake of WDS's validation, all we
+   * care about is that the String is non-null; we do not call out to an OAuth provider to validate
+   * expiration or authenticity.
+   *
+   * @param input the value to validate
+   * @return true if the value is valid
+   */
+  public static boolean isValid(String input) {
+    return input != null;
+  }
```
jladieu, line 24: Now that you have the `BearerToken` class, I think you also have a place to formalize what it means for the actual token to be `null`, give that thing a name, and encapsulate that concern instead of having to do null-checks anywhere.

I don't know if I have the name right here...but maybe it's as simple as `Unauthenticated`? :shrug:

```suggestion
  public static BearerToken unauthenticated() {
     return new BearerToken(null);
  }
  
  public boolean isAuthenticated() {
     return value != null;
  }
```

service/src/main/java/org/databiosphere/workspacedataservice/sam/TokenContextUtil.java
```diff
@@ -33,8 +34,8 @@ private TokenContextUtil() {}
    * @param initialValue the first value to check for a non-null token
    * @return the final value
    */
-  public static String getToken(String initialValue) {
-    return getToken(initialValue, () -> null);
+  public static BearerToken getToken(String initialValue) {
+    return getToken(new BearerToken(initialValue), () -> null);
```
jladieu, line 38: Here you could potentially use `() => BearerToken.unauthenticated()`

service/src/main/java/org/databiosphere/workspacedataservice/sam/TokenContextUtil.java
```diff
@@ -54,8 +55,8 @@ public static String getToken(String initialValue) {
    * @param orElse the value to return if the token was not found otherwise
    * @return the final value
    */
-  public static String getToken(String initialValue, Supplier<String> orElse) {
-    if (initialValue != null) {
+  public static BearerToken getToken(BearerToken initialValue, Supplier<BearerToken> orElse) {
+    if (initialValue != null && initialValue.value() != null) {
```
jladieu, line 59: Here you could potentially use `initialValue.isAuthenticated()` instead of checking `initialValue.value() != null` (subject to that being a reasonable name/concept to formalize)

jladieu (approved):
Yes, this is pretty much what I was suggesting...with a little extra encapsulation for the nully state, which I've added some commentary for.

Curious what you think of it too.

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/BearerToken.java
davidangb: Semantically, I want to be pretty clear about what we're actually providing here. What the current implementation offers is: "WDS found an `Authentication: Bearer … ` header, and saved its contents for later use". What I am trying to avoid is: "WDS has an OAuth token that is currently valid for use in Terra".

The latter implies that the token isn't expired, decodes as a JWT, has the right scopes, was issued by the right OAuth issuer, etc. I'm trying to keep WDS out of this business, and instead delegate to Sam and Azure b2c for this work.

The former makes very few promises. This keeps WDS simple: if it finds some opaque value in the http header, it will then pass that same http header onwards when it makes outbound requests.

Does this stance 1) make sense, and 2) change anything about the model you're suggesting?

Perhaps it's just a terminology thing, I could be satisfied with simply changing `BearerToken.isAuthenticated()` to `BearerToken.isEmpty()`.

jladieu: Totally agree with your separation of concerns here, keeping WDS lean and mean wrt to the assertions it makes about this token.  I'm mainly motivated to drive `null`s out the interaction here, and completely incapable of coming up with a good name for what you describe.  If you're good with `isEmpty` I am too!

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/BearerToken.java
```diff
@@ -0,0 +1,52 @@
...
+    return value;
+  }
+
+  /** indicates if this BearerToken is empty; that is, has a null `value` */
+  public boolean isEmpty() {
+    return value == null;
+  }
+
+  /** indicates if this BearerToken is not empty; that is, has a non-null `value` */
+  public boolean nonEmpty() {
```
nawatts, line 40: The double negative here isn't the easiest to read, but I don't have a better alternative.

nawatts (approved):

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/BearerToken.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  /** returns an empty BearerToken; that is, a BearerToken whose `value` is null */
+  public static BearerToken empty() {
+    return new BearerToken(null);
+  }
+
+  /** returns a BearerToken with the specified `value` */
+  public static BearerToken of(String val) {
+    return new BearerToken(val);
```
yuliadub, line 27: only thought I had is - would it be useful to check if the token actually contains "Bearer" (or contains more then one) and if it doesnt throw a specific error that format is not correct or something like that? Maybe a very rare occurrence but when I was testing cloning I had Bearer in the token twice and it took an embarrassing amount of time to figure out :D

yuliadub (approved):
thank you!

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/BearerToken.java
davidangb: See my previous reply to Josh: https://github.com/DataBiosphere/terra-workspace-data-service/pull/373#discussion_r1362548564 … I am hoping to stay away from anything beyond the most basic validation. That said, if I'm in the minority here, we can add it …

PR #374, opened 2023-10-18
AJ-1011: remove instanceId from get-job API

davidangb:
Cleanup/followup from previous async-job PRs.

This removes the instanceId from the `jobStatusV1` API, changing it from:
```
/{instanceUuid}/job/v1/{jobId}
```
to:
```
/job/v1/{jobId}
```

Async jobs aren't necessarily associated with an instance. Consider, for instance, if we made backups asynchronous.


calypsomatic (commented):
Do we need to add `instanceId` to any other jobs or request or response models to make up for this?  Or does nothing use/rely on it yet?

davidangb:
> Do we need to add instanceId to any other jobs or request or response models to make up for this? Or does nothing use/rely on it yet?

I think we do - e.g. instance is an input argument for data imports. I have broken out that work as AJ-1401 and AJ-1402 (which I haven't yet filled in details for, so you just have to take my word for it)

calypsomatic (approved):

nawatts (approved):

PR #375, opened 2023-10-18
AJ-1011: higher-level methods on JobDao, like running()

davidangb:
Responding to https://github.com/DataBiosphere/terra-workspace-data-service/pull/361#discussion_r1356973572

Add a few convenience methods to `JobDao`/`PostgresJobDao`. This makes it a bit easier on callers, who can now call e.g.
```java
jobDao.running(myJobId);
```
instead of 
```java
jobDao.updateStatus(myJobId, GenericJobServerModel.StatusEnum.RUNNING);
```


calypsomatic (approved):

nawatts (approved):

jladieu (commented):
Thanks for taking this on @davidangb!

Are there any external callers to `updateStatus` left?  I wonder if that method could be removed from the public interface now, letting the `StatusEnum` become a hidden implementation detail and/or readonly attribute.

PR #377, opened 2023-10-18
[AJ-1069] Reject records containing duplicate keys

nawatts:
Currently, WDS' batchWriteRecords API accepts records containing duplicate keys.

For example, this request
```json
[
  {
    "operation": "upsert",
    "record": {
      "id": "1",
      "type": "thing",
      "attributes": {
        "key": "value1",
        "key": "value2"
      }
    }
  }
]
```

Results in a record:
```json
{
  "id": "1",
  "type": "thing",
  "attributes": {
    "sys_name": "1",
    "key": "value2",
    "stringAttr": null
  }
}
```

The first value for the duplicated key is silently dropped.

With this change, that same request returns a 400 response with a "Duplicate field 'key'" message.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/JsonConfig.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/JsonConfig.java
@@ -1,6 +1,7 @@
 package org.databiosphere.workspacedataservice.service;
 
 import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -21,6 +22,7 @@ public class JsonConfig {
             .enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS)
             .configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false)
             .configure(JsonGenerator.Feature.WRITE_BIGDECIMAL_AS_PLAIN, true)
+            .configure(JsonParser.Feature.STRICT_DUPLICATE_DETECTION, true)
             .configure(DeserializationFeature.USE_BIG_DECIMAL_FOR_FLOATS, true)
             .configure(DeserializationFeature.USE_BIG_INTEGER_FOR_INTS, true)
             .findAndAddModules()
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/model/exception/BadStreamingWriteRequestException.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/model/exception/BadStreamingWriteRequestException.java
@@ -1,5 +1,6 @@
 package org.databiosphere.workspacedataservice.service.model.exception;
 
+import com.fasterxml.jackson.databind.JsonMappingException;
 import java.io.IOException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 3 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/service/model/exception/BadStreamingWriteRequestException.java
```diff
@@ -14,8 +15,12 @@ public class BadStreamingWriteRequestException extends IllegalArgumentException
...
   public BadStreamingWriteRequestException(IOException ex) {
     super(
-        "The server doesn't understand the request. Please verify you are using "
-            + "the proper format.");
-    LOGGER.error("Error parsing request stream as json ", ex);
+        // If the original exception was a JsonMappingException, forward the original message
+        // describing the issue to the user.
+        // Otherwise, send a generic error message.
+        (ex instanceof JsonMappingException)
+            ? ((JsonMappingException) ex).getOriginalMessage()
```
nawatts, line 22: Since BadStreamingWriteRequestException accepts an IOException, this seemed like the least invasive way to change the message.

service/src/main/java/org/databiosphere/workspacedataservice/service/model/exception/BadStreamingWriteRequestException.java
```diff
@@ -14,8 +15,12 @@ public class BadStreamingWriteRequestException extends IllegalArgumentException
 
   public BadStreamingWriteRequestException(IOException ex) {
     super(
-        "The server doesn't understand the request. Please verify you are using "
-            + "the proper format.");
-    LOGGER.error("Error parsing request stream as json ", ex);
```
nawatts, line 19: Made this more generic (dropped the "as json") since this exception is also used for TSV.

service/src/main/java/org/databiosphere/workspacedataservice/service/JsonConfig.java
```diff
@@ -21,6 +22,7 @@ public ObjectMapper objectMapper() {
             .enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS)
             .configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false)
             .configure(JsonGenerator.Feature.WRITE_BIGDECIMAL_AS_PLAIN, true)
+            .configure(JsonParser.Feature.STRICT_DUPLICATE_DETECTION, true)
```
nawatts, line 25: Thanks @davidangb for the pointer to this setting. I would have _never_ known about this.

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/service/JsonConfig.java
jladieu: Nice!

jladieu (approved):

service/src/test/java/org/databiosphere/workspacedataservice/service/BatchWriteServiceTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+  @AfterEach
+  void tearDown() {
+    instanceDao.dropSchema(INSTANCE);
+  }
+
+  @Test
+  void testRejectsDuplicateKeys() throws IOException {
+    String streamContents =
+        "[{\"operation\": \"upsert\", \"record\": {\"id\": \"1\", \"type\": \"thing\", \"attributes\": {\"key\": \"value1\", \"key\": \"value2\"}}}]";
```
davidangb, line 46: fwiw, you can use a Java triple-quote "[text block](https://www.baeldung.com/java-text-blocks)" here to avoid all the escaping, like:
```java
    String streamContents =
        """
      [
         {
            "operation":"upsert",
            "record":{
               "id":"1",
               "type":"thing",
               "attributes":{
                  "key":"value1",
                  "key":"value2"
               }
            }
         }
      ]
    """;
```

nawatts: Nice! That would definitely have been an improvement.

PR #372, opened 2023-10-16
AJ-1234: Add `WsmPactTest`

jladieu:
This is the first of several test suites verifying interaction with WSM; this one focuses specifically on `createDataRepoSnapshotReference`.

Is paired with https://github.com/DataBiosphere/terra-workspace-manager/pull/1511

PR #378, opened 2023-10-19
AJ-1188 Add wds consumer tests against datarepo provider

calypsomatic:
https://broadworkbench.atlassian.net/browse/AJ-1188

This PR add consumer-side contract tests against TDR, which are verified in https://github.com/DataBiosphere/jade-data-repo/pull/1528.  See also the pact verification at https://pact-broker.dsp-eng-tools.broadinstitute.org/dashboard/provider/datarepo/consumer/terra-workspace-data-service.


.github/workflows/publish-pacts.yml
```diff
@@ -53,7 +53,7 @@ jobs:
 
     steps:
       - name: Checkout current code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
```
calypsomatic, line 56: This was giving a node-deprecation warning.

.github/workflows/publish-pacts.yml
```diff
@@ -65,7 +65,7 @@ jobs:
       - name: Output consumer contract as non-breaking base64 string
         id: encode-pact
         run: |
-          NON_BREAKING_B64=$(cat service/build/pacts/wds-consumer-sam-provider.json | base64 -w 0)
+          NON_BREAKING_B64=$(cat service/build/pacts/terra-workspace-data-service-datarepo.json | base64 -w 0)
```
calypsomatic, line 68: TODO: Publish to more than one provider.

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
```diff
@@ -0,0 +1,124 @@
...
+    return builder
+        .given("snapshot doesn't exist")
+        .uponReceiving("a snapshot request")
+        .path("/api/repository/v1/snapshots/12345678-abc9-012d-3456-e7fab89cd01e")
+        .query("include=TABLES")
+        .method("GET")
+        .willRespondWith()
+        .status(404)
+        .body(
+            "{\"message\": \"Snapshot not found - id: 12345678-abc9-012d-3456-e7fab89cd01e\",\"errorDetail\": []}")
```
okotsopoulos, line 53: Question: is verifying the body contents critical to your Pact?  How does WDS use this message?

I am inclined to simplify the companion provider state to only throw a simple `SnapshotNotFoundException` rather than constructing a message.  My instinct is that it's the code here that matters.

calypsomatic: You're right, that's not important, I can ignore the body

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
```diff
@@ -0,0 +1,124 @@
...
+
+  @BeforeEach
+  void setUp() {
+    // Without this setup, the HttpClient throws a "No thread-bound request found" error
+    MockHttpServletRequest request = new MockHttpServletRequest();
+    // Set the mock request as the current request context
+    RequestContextHolder.setRequestAttributes(new ServletRequestAttributes(request));
+  }
+
+  static final String dummySnapshotId = "12345678-abc9-012d-3456-e7fab89cd01e";
```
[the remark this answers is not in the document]
okotsopoulos: :tada: @calypsomatic , I added this [TDR commit](https://github.com/DataBiosphere/jade-data-repo/pull/1528/commits/5ca6609427f9bf19474a025de349f4a5a1af86c5) so that TDR provider states expect the snapshot ID to be provided as a parameter.

Locally off of your feature branch here, I modified your Pact builders to pass in your dummy ID, and I saw that TDR's Pact verification ran successfully off of a local Pact broker. I'll leave a specific example in a comment below.

(I am still running into trouble off of my TDR PR not being able to find any Pacts from the live Pact Broker, but that's another can of fish.)

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+  @Pact(consumer = "wds", provider = "datarepo")
+  public RequestResponsePact noSnapshotPact(PactDslWithProvider builder) {
+    return builder
+        .given("snapshot doesn't exist")
```
okotsopoulos, line 45: To successfully pass in the snapshot ID to the TDR provider state, this and other instances would be altered like so:

```
.given("snapshot doesn't exist", Map.of("id", dummySnapshotId))
```

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
```diff
@@ -0,0 +1,122 @@
...
+        .uponReceiving("a snapshot request")
+        .path("/api/repository/v1/snapshots/12345678-abc9-012d-3456-e7fab89cd01e")
```
okotsopoulos, line 47: This path specification works, but now that things are stabilizing I'd recommend that anywhere you've hardcoded the same dummy snapshot ID you instead reuse the reference, maybe like this:

```
.path("/api/repository/v1/snapshots/" + dummySnapshotId)
```

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
```diff
@@ -0,0 +1,122 @@
...
+  @Test
+  @PactTestFor(pactMethod = "userHasAccessToSnapshotPact", pactVersion = PactSpecVersion.V3)
+  void testUserHasAccessToSnapshot(MockServer mockServer) {
+    DataRepoClientFactory clientFactory = new HttpDataRepoClientFactory(mockServer.getUrl());
+    DataRepoDao dataRepoDao = new DataRepoDao(clientFactory);
+
+    SnapshotModel snapshot =
+        dataRepoDao.getSnapshot(UUID.fromString("12345678-abc9-012d-3456-e7fab89cd01e"));
+    assertNotNull(snapshot, "Snapshot request should return a snapshot");
+    assertNotNull(snapshot.getId(), "Snapshot response should have an id");
```
okotsopoulos, line 119: IMO, it would be fair to expect TDR gives you back the ID you've passed in.

okotsopoulos (commented):
In the home stretch!

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
calypsomatic: Yeah I tried to do it fancy and was running into issues... let me try again

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
```diff
@@ -0,0 +1,125 @@
...
+        .method("GET")
+        .willRespondWith()
+        .status(404)
+        .toPact();
+  }
+
+  @Pact(consumer = "wds", provider = "datarepo")
+  public RequestResponsePact noAccessToSnapshotPact(PactDslWithProvider builder) {
+    return builder
+        .given("user does not have access to snapshot")
```
jladieu, line 60: Piggybacking on @okotsopoulos ' suggestion, I'd recommend providing the snapshot ID in each given, like you do on line 47, so the provider can stub/mock out specifically for the behavior that should be simulated for that ID.

I started using a convention "with the given *id*" as a readable call to action for the provider implementor to look in the params hash for a variable named *id*.

jladieu, line 60:
```suggestion
        .given("user does not have access to a snapshot with the given id", Map.of("id", dummySnapshotId))
```

Piggybacking on @okotsopoulos ' suggestion, I'd recommend providing the snapshot ID in each given, like you do on line 47, so the provider can stub/mock out specifically for the behavior that should be simulated for that ID.

I started using a convention "with the given *id*" as a readable call to action for the provider implementor to look in the params hash for a variable named *id*.

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  @Pact(consumer = "wds", provider = "datarepo")
+  public RequestResponsePact userHasAccessToSnapshotPact(PactDslWithProvider builder) {
+    // Current use of datarepo snapshot only relies on id and name
+    // Future development should update expected response shape to include fields we need
+    var snapshotResponseShape =
+        new PactDslJsonBody().stringValue("id", dummySnapshotId).stringType("name");
+    return builder
+        .given("user has access to snapshot")
```
jladieu, line 77:
```suggestion
        .given("user has access to a snapshot with the given id", Map.of("id", dummySnapshotId))
```

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+  @BeforeEach
+  void setUp() {
+    // Without this setup, the HttpClient throws a "No thread-bound request found" error
+    MockHttpServletRequest request = new MockHttpServletRequest();
+    // Set the mock request as the current request context
+    RequestContextHolder.setRequestAttributes(new ServletRequestAttributes(request));
+  }
+
+  static final String dummySnapshotId = "12345678-abc9-012d-3456-e7fab89cd01e";
```
jladieu, line 42: This is nitpicky, but I recommend storing a `UUID` instance variable, and `toString()`ing it later on when you need to pass it anywhere as a String.  This maintains its strong typeyness for longer.

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+  @Test
+  @PactTestFor(pactMethod = "noAccessToSnapshotPact", pactVersion = PactSpecVersion.V3)
+  void testNoAccessToSnapshot(MockServer mockServer) {
+    DataRepoClientFactory clientFactory = new HttpDataRepoClientFactory(mockServer.getUrl());
+    DataRepoDao dataRepoDao = new DataRepoDao(clientFactory);
+
+    assertThrows(
+        DataRepoException.class,
+        () -> dataRepoDao.getSnapshot(UUID.fromString("12345678-abc9-012d-3456-e7fab89cd01e")),
```
jladieu, line 108: I'd make sure to refer to the same UUID variable (eg: `dummySnapshotId`) where you set up the `given` of the `pactMethod`

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
jladieu, line 42: I found https://gist.github.com/pesterhazy/120e1926dfeff603c264268fdf52adfb which had some easy to remember UUIDs (I'm partial to `facade`, if you'd like to lower the strain of identifying a hardcoded UUID).

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
```diff
@@ -0,0 +1,125 @@
...
+        new PactDslJsonBody().stringValue("id", dummySnapshotId).stringType("name");
+    return builder
+        .given("user has access to snapshot")
+        .uponReceiving("a snapshot request")
+        .path("/api/repository/v1/snapshots/" + dummySnapshotId)
+        .query("include=TABLES")
+        .method("GET")
+        .willRespondWith()
+        .status(200)
+        .body(snapshotResponseShape)
```
jladieu, line 84: For a simple response shape like this, inlining the definition and using the fancy new DSL can make the contract more readable.  You also get to avoid coming up with a name like `snapshotResponseShape`!

```suggestion
        .body(newJsonBody(snapshot -> {
           // toString() only needed if you apply my suggestion to strongly type dummyId
           snapshot.stringValue("id", dummySnapshotId.toString());
           snapshot.stringType("name");
        }).build())
```

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
jladieu: +1

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
```diff
@@ -0,0 +1,125 @@
...
+  @Pact(consumer = "wds", provider = "datarepo")
+  public RequestResponsePact noAccessToSnapshotPact(PactDslWithProvider builder) {
+    return builder
+        .given("user does not have access to snapshot")
+        .uponReceiving("a snapshot request")
+        .path("/api/repository/v1/snapshots/" + dummySnapshotId)
+        .query("include=TABLES")
+        .method("GET")
+        .willRespondWith()
+        .status(403)
```
jladieu, line 66: Just doublechecking...is there anything in the body of the 403 response that would indicate the specifics of _why_ it's a 403?

I ask because I had a similar test looking for a 409 conflict, and it was actually initially responding with a different error message than I was expecting, for a different error reason that also happened to produce a 409.  Ref: https://github.com/DataBiosphere/terra-workspace-data-service/pull/372/files#r1380313724

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
jladieu: Also found this https://docs.pact.io/getting_started/matching#random-data---avoid-it suggesting to avoid random data, and give more weight to your decision to include a hardcoded string.

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
```diff
@@ -0,0 +1,125 @@
...
+  public RequestResponsePact userHasAccessToSnapshotPact(PactDslWithProvider builder) {
+    // Current use of datarepo snapshot only relies on id and name
+    // Future development should update expected response shape to include fields we need
+    var snapshotResponseShape =
+        new PactDslJsonBody().stringValue("id", dummySnapshotId).stringType("name");
+    return builder
+        .given("user has access to snapshot")
+        .uponReceiving("a snapshot request")
+        .path("/api/repository/v1/snapshots/" + dummySnapshotId)
+        .query("include=TABLES")
```
jladieu, line 80: Just curious, how essential is the `include=TABLES` to the provider behaving the way you expect?  Is it strictly required to add that param in in order to get the response shape you're looking for?

If not, omit it.

If so, I might suggest making the matcher more permissive by using `.matchQuery("include", "TABLES")`... I *believe* this will make it so it doesn't start failing if you add more query params to the implementation later on, or the order changes, etc.

calypsomatic: It is completely essential!  It took me a while to figure that out, without it FAILURE

calypsomatic: I'll try `matchQuery` although I can't promise I didn't already try it before and had it mysteriously not work.  I don't remember, only remember that this was pretty tricky.

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
calypsomatic: I think this was another case where it was mysteriously failing when I used a variable and I just re-wrote the string out of frustration.  Maybe enough else has changed that it'll work this time!

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
okotsopoulos: @jladieu ah, I had previously commented on this too, but in the other direction :) 

https://github.com/DataBiosphere/terra-workspace-data-service/pull/378#discussion_r1379127364

My belief was that if WDS is dictating its behavior based off of the HTTP response code and otherwise ignoring the message, then the Provider state should only define the bare minimum of what's used for maintainability.  I'm not sure what the best practice is here.

.github/workflows/publish-pacts.yml
```diff
@@ -7,12 +7,14 @@ on:
   push:
     branches:
       - main
+      - aj-1188-tdr-consumer
```
okotsopoulos, line 10: Note -- I think this is added to allow Pact verification to run while this and the companion TDR PR are still in flight, but it shouldn't be checked in.

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
okotsopoulos: Ah, sorry, I misunderstood the domain here.

I now see in the WDS source code that the snapshot's `tables` field is used like so:
https://github.com/DataBiosphere/terra-workspace-data-service/blob/e4906e9f4cc812549c0d32f838ada38a2e70ff59/service/src/main/java/org/databiosphere/workspacedataservice/service/DataRepoService.java#L80-L88

In that case, I'd suggest expanding `testUserHasAccessToSnapshot` to also check that the snapshot response has tables in the bare minimum format that you expect.  That will also require a change to the TDR provider state response, which I am happy to help with.

As for `matchQuery`, I am somewhat inclined towards your existing exact matching of your request, per the recommendations from Pact docs:

https://docs.pact.io/getting_started/matching#request-matching

> As a rule of thumb, you generally want to use exact matching when you're setting up the expectations for a request (upon_receiving(...).with(...)) because you're under control of the data at this stage, and according to Postel's Law, we want to be "strict" with what we send out. Note that the request matching does not allow "unexpected" values to be present in JSON request bodies or query strings. (It does however allow extra headers, because we found that writing expectations that included the headers added by the various frameworks that might be used resulted in tests that were very fiddly to maintain.)

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
okotsopoulos: Neat, I see that there's a new assertion added below to make sure that the IDs match.  In that case, this assertion could be removed if desired as it's now redundant.

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
```diff
@@ -0,0 +1,125 @@
...
+  @Pact(consumer = "wds", provider = "datarepo")
+  public RequestResponsePact noSnapshotPact(PactDslWithProvider builder) {
+    return builder
+        .given("snapshot doesn't exist", Map.of("id", dummySnapshotId))
+        .uponReceiving("a snapshot request")
+        .path("/api/repository/v1/snapshots/" + dummySnapshotId)
+        .query("include=TABLES")
+        .method("GET")
+        .willRespondWith()
+        .status(404)
```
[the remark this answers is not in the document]
jladieu: Looking at this particular route again, and after seeing @okotsopoulos's [feedback](https://github.com/DataBiosphere/terra-workspace-data-service/pull/378#discussion_r1379127364) to leave out the message validation, I think I'm fine with leaving this as is and just look at the 404.

My reasoning for wanting to get specific about the error message for 404s are for cases where you're dealing with a nested RESTful resource (eg:  /parent/1/child/2) which could have two different causes for a 404 which I'd want to make sure we were observing.  Since this is just a top level resource, the only two reasonable scenarios for a 404 are:

1) Route not configured correctly
2) Resource doesn't exist

Since we've got a positive test looking for an existing snapshot, I think that completely covers (1), and this test adequately covers (2).

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
jladieu: BTW for the 404 case, I think it's reasonable to leave as is (just checking http status code), see [comment](https://github.com/DataBiosphere/terra-workspace-data-service/pull/378/files/b071e41bd0b8d6c72805e51b3c4656393f523e1f#r1386793030)

For this error scenario, I'm on the fence and deferring to @okotsopoulos if the logic for a 403 is simple enough that we don't need to verify the _reason_ for the failure.  If there's really only one reason a 403 might happen, then probably fine to leave as is.

For the example I linked where I felt compelled to add the extra regex check, I explicitly encountered the _wrong_ 409 while setting up the test and felt the need to close that gap.

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
jladieu: I just pushed up bd0f748ebbfb5dabb207d2c6b686d51e633ffb9d to check for the table `name`... @okotsopoulos  thanks for tracking down that usage in WDS and highlighting it!

jladieu (approved):

.github/workflows/publish-pacts.yml
jladieu: I'll yank this out before I merge :)

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
```diff
@@ -0,0 +1,146 @@
...
+  @Pact(consumer = "wds", provider = "datarepo")
+  public RequestResponsePact userHasAccessToSnapshotPact(PactDslWithProvider builder) {
+    var snapshotResponseShape =
+        new PactDslJsonBody().stringValue("id", dummySnapshotId.toString()).stringType("name");
+    return builder
+        .given(
+            "user has access to snapshot with given id", Map.of("id", dummySnapshotId.toString()))
+        .given(
+            "snapshot with given id has at least one table",
+            Map.of("id", dummySnapshotId.toString()))
```
okotsopoulos, line 83: @jladieu , how would you feel about removing this provider state assignment?

The reason being, a TDR snapshot is guaranteed to have at least one table.  Since your request has "include=TABLES" as its query parameter, that's what will instruct TDR to return tables on retrieval.

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
okotsopoulos: @jladieu and I chatted more about this on Slack!

Summary --

In TDR, we were trying to make it a true unit test of the controller only, and mocking all of its dependencies.  So we are manually creating the exception that would be thrown, and adding it to the mock.  But the way that we're currently mocking in TDR provider states is missing assurance that the mocks and real app aren't drifting away from each other

I'm inclined to stick with TDR's current provider state set-up so that we can get this first iteration of contract tests live.  With that, I'd prefer if WDS for now asserts only on the 403 error code rather than also checking for a particular message, since right now we're creating that exception as part of the mock set-up rather than throwing it from a real component.

But, I'll be ticketing further work on mock-sharing / increasing our confidence in those interlocking boundaries in our tests, so we could absolutely expand the consumer verification along with that work.

okotsopoulos (approved):
Thanks very much Bria and Josh for your work on this!

@jladieu , I have one request regarding the removal of a TDR provider state specification, but otherwise looks good to merge.  I was able to massage things locally in TDR and see the published Pacts pass verification, so I know how I'll be updating my provider PR in flight.

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
jladieu: Sounds fine to me, I'll get this incorporated!

jladieu: Also, this is a perfect example of how contracts help you discover assumptions you are making (or should make) about collaborator services.

jladieu: Fixed in 7e451128932b628655630d6ec937e5efb7cb5fa9

.github/workflows/publish-pacts.yml
jladieu: Removed in d6239895217ce57d1b81bf884953661288127976

service/src/test/java/org/databiosphere/workspacedataservice/pact/TDRPactTest.java
jladieu: Fixed in 4749f4bbfb8b5ba5d03042adb1d7301a04c4d049

.github/workflows/publish-pacts.yml
jladieu: Fixed in b071e41bd0b8d6c72805e51b3c4656393f523e1f

PR #379, opened 2023-10-19
AJ-1396: update swagger UI version

davidangb:
* acknowledge that `webjars-locator-core` is managed by the [Spring Dependency Management Plugin](https://docs.spring.io/dependency-management-plugin/docs/current/reference/html/) and not by WDS directly
* update `webjars-locator-core` to latest via the Spring Dependency Management Plugin
* update `swagger-ui-dist` to latest
* review our customized HTML and JS in swagger-ui.html against the newer version of `swagger-ui-dist` (no modifications required)
* make comments in swagger-ui.html very explicit about where we have customized


build.gradle
```diff
@@ -39,6 +39,7 @@ subprojects {
         dependencies {
             dependency 'org.yaml:snakeyaml:2.0'
             dependency 'io.netty:netty-handler:4.1.94.Final'
+            dependency 'org.webjars:webjars-locator-core:0.53'
```
davidangb, line 42: moved the version here from service/build.gradle, since this is a managed dependency

service/src/main/resources/static/swagger/swagger-ui.html
davidangb, line 1: spotless reformatted this file a bunch, you may want to ignore whitespace when viewing the diff

jladieu (approved):
Nice housekeeping!

calypsomatic (approved):

PR #382, opened 2023-10-24
AJ-1370 parse snapshotIds from PFB

calypsomatic:
[AJ-1370](https://broadworkbench.atlassian.net/browse/AJ-1370)

This PR imports `java-pfb` and uses it to parse the snapshotIds from a PFB as part of the `PfbQuartzJob`.  It assumes that snapshot ids are under they key `source_datarepo_snapshot_id`.  In order to test that the snapshot Ids have been obtained, the beginnings of https://broadworkbench.atlassian.net/browse/AJ-1371 have been added - a call to WSM - but that ticket still needs to pass the pet token and verify behavior.

Reminder:

[AJ-1370]: https://broadworkbench.atlassian.net/browse/AJ-1370?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -27,7 +43,36 @@ protected JobDao getJobDao() {
 
   @Override
   protected void executeInternal(UUID jobId, JobExecutionContext context) {
-    // TODO: AJ-1227? implement PFB import.
+    JobDataMap jobDataMap = context.getMergedJobDataMap();
+    URL url = getJobDataUrl(jobDataMap, ARG_URL);
+    Set<String> snapshotIds = new HashSet();
```
davidangb, line 48: nit: use `<>` to avoid squigglies about "Raw use of parameterized class"
```suggestion
    Set<String> snapshotIds = new HashSet<>();
```

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -27,7 +43,36 @@ protected JobDao getJobDao() {
...
+    try (DataFileStream<GenericRecord> dataStream =
+        PfbReader.getGenericRecordsStream(url.toString())) {
+      while (dataStream.hasNext()) {
+        GenericRecord record = dataStream.next();
+        // Records in a pfb are stored under the key "object"
+        GenericRecord recordObject = (GenericRecord) record.get("object");
+        if (recordObject == null) {
+          // TODO what type of exception/what message
+          throw new RuntimeException("Record in pfb missing data");
+        }
```
davidangb, line 58: instead of a potentially-unsafe cast and an explicit check for null, you could use `instanceof`:
```
if (record.get("object") instanceof GenericRecord genericRecord) {
    // … keep processing inside here …      
} else {
    // TODO what type of exception/what message
    throw new RuntimeException("Record in pfb missing data");
}
```

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -27,7 +43,36 @@ protected JobDao getJobDao() {
...
+        if (recordObject.hasField(SNAPSHOT_ID_IDENTIFIER)) {
+          Object snapId = recordObject.get(SNAPSHOT_ID_IDENTIFIER);
+          if (snapId != null) {
+            snapshotIds.add(snapId.toString());
+          }
+        }
```
davidangb, line 64: I think it's worth checking here to see if the snapshotId is a valid UUID, something like:
```java
        try {
          if (recordObject.get(SNAPSHOT_ID_IDENTIFIER) instanceof String snapshotStr) {
            snapshotIds.add(UUID.fromString(snapshotStr));
          }
        } catch (Exception e) {
          // log a message? just skip silently?
        }
```

davidangb (commented):
Couple comments inline. I am totally fine with this approach, it's straightforward and understandable.

Your call - you could also approach this using Java streams, which is a bit more functional-programming land, something like:

```java
      // translate the Avro DataFileStream into a Java stream
      Stream<GenericRecord> recordStream =
          StreamSupport.stream(
              Spliterators.spliteratorUnknownSize(dataStream.iterator(), Spliterator.ORDERED),
              false);

      // process the stream into a list of unique snapshotIds
      List<String> uniqueSnapshotIds = recordStream
          .map(rec -> rec.get("object")) // Records in a pfb are stored under the key "object"
          .filter(obj -> obj instanceof GenericRecord) // which we expect to be a GenericRecord
          .map(obj -> ((GenericRecord) obj).get(SNAPSHOT_ID_IDENTIFIER).toString()) // within the GenericRecord, find the source_datarepo_snapshot_id
          .filter(Objects::nonNull) // expect source_datarepo_snapshot_id to be non-null
          .distinct() // find only the unique snapshotids
          .toList();
```
note that this approach doesn't include validating the snapshotid is a uuid, which I commented on inline.

service/src/test/java/org/databiosphere/workspacedataservice/pfb/PfbJobTest.java
```diff
@@ -0,0 +1,97 @@
...
+@SpringBootTest
+class PfbJobTest {
+
+  @MockBean JobDao jobDao;
+  @MockBean WorkspaceManagerDao wsmDao;
+
+  @Test
+  void doNotFailOnMissingSnapshotId() throws JobExecutionException {
+    JobExecutionContext mockContext = mock(JobExecutionContext.class);
+    // TODO is there a smarter/better way to get the correct actual path to this file
```
calypsomatic, line 40: I struggled to reliably locate the file and this was what I came up with.  Is there a better way?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -27,7 +47,50 @@ protected JobDao getJobDao() {
...
+              .filter(Objects::nonNull) // expect source_datarepo_snapshot_id to be non-null
+              .distinct() // find only the unique snapshotids
+              .toList();
+
+      // TODO AJ-1371 pass snapshotIds to WSM
+      for (String id : snapshotIds) {
+        try {
+          wsmDao.createDataRepoSnapshotReference(new SnapshotModel().id(UUID.fromString(id)));
+        } catch (Exception e) {
+          // TODO what if it's not a valid UUID?
```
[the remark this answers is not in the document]
davidangb: My opinion, I think Beth should chime in too:
* if the `source_datarepo_snapshot_id` is blank/empty, just skip it. This allows source systems to have empty values in a column.
* if the `source_datarepo_snapshot_id` is nonempty but not a uuid, it's an error and we should fail the import. This suggests the source system _tried_ to link a snapshot but something went wrong.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -27,7 +47,50 @@ protected JobDao getJobDao() {
...
+        }
+      }
+
+    } catch (IOException e) {
+      // TODO what type of error
+      throw new RuntimeException("Error processing PFB");
```
davidangb, line 90: You could use `JobExecutionException` here but that's pretty generic; I'd suggest making a dedicated `PfbImportException` (or whatever you want to name it)

service/src/test/java/org/databiosphere/workspacedataservice/pfb/PfbJobTest.java
davidangb: I think this is the right way to do it! You could also get all Spring-y about it and do something like:
```java
@Value("classpath:minimal_data.avro")
private Resource minimalData;
…
// do something with:
minimalData.getURL();
```

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -27,7 +48,49 @@ protected JobDao getJobDao() {
...
+      for (String id : snapshotIds) {
+        try {
+          wsmDao.createDataRepoSnapshotReference(new SnapshotModel().id(UUID.fromString(id)));
+        } catch (Exception e) {
+          throw new PfbParsingException("Error processing PFB: Invalid snapshot UUID");
+        }
+      }
+
+    } catch (IOException e) {
+      throw new PfbParsingException("Error processing PFB");
```
jladieu, line 90: When catching/re-raising I usually try to serve up the original exception too, otherwise the caller has to go look into the sourcecode to figure out what happened.

Can you make `PfbParsingException` allow a wrapped exception and then use it like so?

```suggestion
      throw new PfbParsingException("Error processing PFB", e);
```

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -27,7 +48,49 @@ protected JobDao getJobDao() {
...
+          StreamSupport.stream(
+              Spliterators.spliteratorUnknownSize(dataStream.iterator(), Spliterator.ORDERED),
+              false);
+
+      // process the stream into a list of unique snapshotIds
+      List<String> snapshotIds =
+          recordStream
+              .map(rec -> rec.get("object")) // Records in a pfb are stored under the key "object"
+              .filter(GenericRecord.class::isInstance) // which we expect to be a GenericRecord
+              .filter(
```
jladieu, line 66: To avoid having to cast twice on lines 68 & 72, could you do an single, intermediate cast with a map?

```map(obj -> (GenericRecod) obj)```

File: service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java [hunk 9]
```diff
@@ -27,7 +48,49 @@ protected JobDao getJobDao() {
...
+              .filter(Objects::nonNull) // expect source_datarepo_snapshot_id to be non-null
+              .distinct() // find only the unique snapshotids
+              .toList();
+
+      // TODO AJ-1371 pass snapshotIds to WSM
+      for (String id : snapshotIds) {
+        try {
+          wsmDao.createDataRepoSnapshotReference(new SnapshotModel().id(UUID.fromString(id)));
+        } catch (Exception e) {
+          throw new PfbParsingException("Error processing PFB: Invalid snapshot UUID");
```
jladieu, line 85: Same suggestion as earlier, can you rewrap `e`?

jladieu (approved):
Few minor suggestions but nothing critical -- feel free to ship as is!

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -27,7 +48,48 @@ protected JobDao getJobDao() {
...
+              .map(obj -> (GenericRecord) obj)
+              .filter(
+                  obj ->
+                      obj.hasField(SNAPSHOT_ID_IDENTIFIER)) // avoid exception if field nonexistent
+              .map(
+                  obj ->
+                      obj.get(SNAPSHOT_ID_IDENTIFIER)
+                          .toString()) // within the GenericRecord, find the
+              // source_datarepo_snapshot_id
+              .filter(Objects::nonNull) // expect source_datarepo_snapshot_id to be non-null
```
davidangb, line 75: with the `obj.hasField` check above, I think `Objects::nonNull` is redundant

davidangb (approved):
one minor comment inline. I agree with Josh's suggestion about propagating the original exceptions when wrapping in `PfbParsingException`, too. Giving this a proactive :+1:

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
calypsomatic: Hmm, but couldn't it have the field and then have nothing in it?

service/build.gradle
```diff
@@ -69,6 +69,7 @@ dependencies {
     implementation 'org.hashids:hashids:1.0.3'
     implementation 'javax.ws.rs:javax.ws.rs-api:2.1.1'
     implementation 'com.google.mug:mug:6.6'
+    implementation "org.apache.avro:avro:1.11.2"
```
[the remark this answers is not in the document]
davidangb: created https://github.com/DataBiosphere/java-pfb/pull/13 to address this

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
davidangb: ah - I could be wrong. I assumed that the underlying object would return false for `obj.hasField` if the value was null - but that's a faulty assumption. That said, if it IS null, then `obj.get(SNAPSHOT_ID_IDENTIFIER).toString()` will throw an NPE anyway …

calypsomatic: ok i moved the `toString` to after the nonNull filter

service/build.gradle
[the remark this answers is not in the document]
davidangb: `bio.terra:java-pfb-library:0.12.0` is now available, and it allows you to delete the `org.apache.avro:avro` dependency

PR #386, opened 2023-10-27
AJ-1371: link snapshots during AnVIL pfb import

davidangb:
When importing a PFB:
* query WSM to find all pre-existing snapshot references for this workspace. This is a paginated query so it requires some extra logic.
* compare the list of snapshot ids mentioned in the pfb to the list of pre-existing snapshots, and identify the list of not-yet-linked snapshots
* create references from the workspace to those new snapshots


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -90,4 +93,41 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
...
+    List<UUID> existingSnapshotIds = pfbQuartzJobSupport.existingPolicySnapshotIds(50);
+    // find the snapshots in this PFB that are not already linked to this workspace
+    List<UUID> newSnapshotIds =
+        snapshotIds.stream().filter(id -> !existingSnapshotIds.contains(id)).toList();
+
+    logger.info(
+        "PFB contains {} snapshot ids. {} of these are already linked to the workspace; {} new links will be created.",
+        snapshotIds.size(),
+        snapshotIds.size() - newSnapshotIds.size(),
+        snapshotIds.size());
```
calypsomatic, line 110: Should this be `newSnapshotIds.size()`?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -90,4 +93,41 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
     // TODO: AJ-1227 implement PFB import.
     logger.info("TODO: implement PFB import.");
   }
+
+  protected void linkSnapshots(List<UUID> snapshotIds) {
+    // list existing snapshots linked to this workspace
+    PfbQuartzJobSupport pfbQuartzJobSupport =
+        new PfbQuartzJobSupport(workspaceId, wsmDao, restClientRetry);
+    List<UUID> existingSnapshotIds = pfbQuartzJobSupport.existingPolicySnapshotIds(50);
```
yuliadub, line 101: what does this "50" number represent?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
davidangb: fixed, good catch

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
davidangb: I added a code comment

yuliadub (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -90,4 +93,42 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
     // TODO: AJ-1227 implement PFB import.
     logger.info("TODO: implement PFB import.");
   }
+
+  protected void linkSnapshots(List<UUID> snapshotIds) {
+    // list existing snapshots linked to this workspace
+    PfbQuartzJobSupport pfbQuartzJobSupport =
+        new PfbQuartzJobSupport(workspaceId, wsmDao, restClientRetry);
+    List<UUID> existingSnapshotIds =
```
calypsomatic, line 101: Are we assuming this will never get above 50?  Seems reasonable but should that assumption be built into the code?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -36,9 +38,15 @@ public class PfbQuartzJob extends QuartzJob {
   private final JobDao jobDao;
   private final WorkspaceManagerDao wsmDao;
 
-  public PfbQuartzJob(JobDao jobDao, WorkspaceManagerDao wsmDao) {
+  @Value("${twds.instance.workspace-id}")
```
jladieu, line 41: Any chance this could be injected as a constructor parameter too and made into a `private final`?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
calypsomatic: Nm I see that's only how many we get at a time.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -90,4 +93,42 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
...
+            (() -> wsmDao.createDataRepoSnapshotReference(new SnapshotModel().id(uuid)));
+        restClientRetry.withRetryAndErrorHandling(
+            voidRestCall, "WSM.createDataRepoSnapshotReference");
+      } catch (Exception e) {
+        throw new PfbParsingException("Error processing PFB: Invalid snapshot UUID", e);
+      }
+    }
+  }
+
+  private UUID maybeUuid(String input) {
```
jladieu, line 126: Possibly a place for an `Optional` instead of a `null`? (_disclosure: I generally steer clear of nulls in Java if possible_)

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -90,4 +93,42 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
     // TODO: AJ-1227 implement PFB import.
     logger.info("TODO: implement PFB import.");
   }
+
+  protected void linkSnapshots(List<UUID> snapshotIds) {
+    // list existing snapshots linked to this workspace
+    PfbQuartzJobSupport pfbQuartzJobSupport =
+        new PfbQuartzJobSupport(workspaceId, wsmDao, restClientRetry);
+    List<UUID> existingSnapshotIds =
+        pfbQuartzJobSupport.existingPolicySnapshotIds(/* pageSize= */ 50);
```
jladieu, line 102: Thanks for the trailmarker comment :heart:

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJobSupport.java
```diff
@@ -0,0 +1,113 @@
...
+import bio.terra.workspace.model.ResourceList;
+import java.util.List;
+import java.util.Objects;
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.databiosphere.workspacedataservice.retry.RestClientRetry;
+import org.databiosphere.workspacedataservice.service.model.exception.PfbImportException;
+import org.databiosphere.workspacedataservice.workspacemanager.WorkspaceManagerDao;
+
+public class PfbQuartzJobSupport {
```
calypsomatic, line 15: No strong feelings about this at all, but even though this is used in `PfbQuartzJob`, it looks like everything it does has to do with snapshots and policies, so should it be named `SnapshotPolicySupport` (or something better that references that)?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJobSupport.java
```diff
@@ -0,0 +1,113 @@
...
+  protected UUID safeGetSnapshotId(ResourceDescription resourceDescription) {
+    ResourceAttributesUnion resourceAttributes = resourceDescription.getResourceAttributes();
+    if (resourceAttributes != null) {
+      DataRepoSnapshotAttributes dataRepoSnapshot = resourceAttributes.getGcpDataRepoSnapshot();
+      if (dataRepoSnapshot != null) {
+        String snapshotIdStr = dataRepoSnapshot.getSnapshot();
+        try {
+          return UUID.fromString(snapshotIdStr);
+        } catch (Exception e) {
+          // noop; this will return null
```
jladieu, line 107: Would it be worth it to print something here in this case, similar to how you do in `PfbQuartzJob`?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJobSupport.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+      }
+    }
+    return null;
```
jladieu, line 111: Another possible place for an `Optional`?

service/src/main/java/org/databiosphere/workspacedataservice/service/model/exception/PfbImportException.java
```diff
@@ -0,0 +1,16 @@
+package org.databiosphere.workspacedataservice.service.model.exception;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(code = HttpStatus.INTERNAL_SERVER_ERROR)
```
jladieu, line 6: The way `PfbImportException` gets thrown (for blowing the hard limit), and the fact that it subclasses `IllegalArgumentException` make it seem more like some kind of client error than an internal error to me.  Should this be a different HTTP status?

That said, I think we'd want to _know about_ these early on as we learn how this feature gets used, so if it makes more sense that it's an 'us' problem than a client problem, then you can keep it as `INTERNAL_SERVER_ERROR`...I just wonder if it should subclass a different kind of exception in that case.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJobSupport.java
```diff
@@ -0,0 +1,113 @@
...
+      if (thisPage.getResources().size() < pageSize) {
+        // fewer results from WSM than we requested; this is the last page of results
+        return finalList;
+      } else {
+        // bump our offset and request another page of results
+        offset.addAndGet(pageSize);
+      }
+    }
+
+    throw new PfbImportException(
```
jladieu, line 87: Is there any sensible way to fail faster if we'll exceed the hard limit?   Maybe by pushing the hard limit enforcement down to WSM?

Feels like this approach ends up doing a whole bunch of wasted work and would potentially fail slowly and cause resource waste/exhaustion when paired with some form of retries.

jladieu (approved):
I had a few style questions and one more involved thought about how to deal with the hard limit enforcement.   But LGTM overall!

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJobSupport.java
```diff
@@ -0,0 +1,113 @@
...
+
+  /**
+   * Get the full list of all snapshot references for the current workspace. WSM returns these
+   * results paginated; this method retrieves pages from WSM and aggregates the results.
+   *
+   * @param pageSize number of results to return from WSM at once
+   * @return the full list of all snapshot references for the workspace.
+   */
+  protected ResourceList listAllSnapshots(int pageSize) {
+    final AtomicInteger offset = new AtomicInteger(0);
```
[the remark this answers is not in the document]
davidangb: If I use a plain `int` here, the compiler complains about its use at line 69-70, saying "Variable used in lambda expression should be final or effectively final". So, I used a `final AtomicInteger` to get around that. In practice, I think few workspaces will end up with > 50 snapshot references anyway, so it shouldn't be a significant performance hit.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
davidangb: moved to constructor.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
davidangb: I'm gonna push back here - I think that's sound advice to use `Optional` instead of `null` for public methods, but this is a private method used only within the class. If it returned `Optional`, it would require additional steps inside of `executeInternal()`. Whereas now we can use one step `.filter(Objects::nonNull)`, we'd need to change to `.filter(Optional::isPresent).map(Optional::get)`.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJobSupport.java
davidangb: Ah, fair point. I had been thinking about its name as "I provide support to the PfbQuartzJob class", whereas you're seeing it as "I provide support for snapshots and policies" … you think it's worth changing?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJobSupport.java
davidangb: added logging here, good call.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJobSupport.java
calypsomatic: It seems plausible that these methods are re-usable in other snapshot situations that don't come from PFBs, in which case probably yes.  But if you think they're pretty specific to PfbQuartz then it can stay the same.  But again, not a strong opinion.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJobSupport.java
davidangb: The only other approach I can think of is to pick a number (maybe not as high as 10000), and make a single request to WSM for that many resources - don't attempt to paginate. Do you think that would be better?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJobSupport.java
davidangb: renamed!

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJobSupport.java
davidangb: same pushback here - as a non-public method (it's only protected instead of private so I can unit-test it), I'm optimizing for performance instead of a good API - this one is also used directly as a stream mapping function

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
jladieu: Sounds fine, particularly making the distinction between public and private APIs.  I have a hunch that there'd be some optional-friendly predicate that could stand in for (and have equivalent terseness to) what's already there, but don't know off the top of my head what it'd be so carry on!

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJobSupport.java
jladieu: I think if by your guess this is a really really unlikely scenario that it's maybe not something to worry about and this might be the perfect safeguard...wait till it happens, then worry about it.

I don't think I understand well enough how these snapshots get written to WSM, but the only thing that I think might be better is if we could somehow head them off at write time (into WSM) rather than here at read time.

jladieu (approved):

PR #387, opened 2023-10-27
AJ-1227: parse incoming PFBs into tables

calypsomatic:
WIP

Reminder:


service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -61,12 +66,12 @@ public BatchWriteService(
    * @param primaryKey PK for the record type
    * @return the number of records written
    */
-  public int consumeWriteStream(
+  public Map<RecordType, Integer> consumeWriteStream(
```
calypsomatic, line 69: I changed the response type here so that activitylogger (in `PfbQuartzJob`) knew what record types were being updated.  I propagated this change as little as possible, which requires awkward `.get(recordType)` calls on the response in other methods that use this one.  Am very happy to take suggestions for a better design!

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -75,19 +80,38 @@ public int consumeWriteStream(
...
           info = streamingWriteHandler.readRecords(batchSize)) {
         List<Record> records = info.getRecords();
-        if (firstUpsertBatch && info.getOperationType() == OperationType.UPSERT) {
-          schema = inferer.inferTypes(records);
-          createOrModifyRecordType(
-              instanceId, recordType, schema, records, primaryKey.orElse(RECORD_ID));
-          firstUpsertBatch = false;
+        // TODO is a recordType == null check the best way to do this?
+        if (recordType == null) {
+          Map<RecordType, List<Record>> sortedRecords =
```
calypsomatic, line 85: I tried making a helper method so that these lines of code are not duplicated between the two branches, but because both `schema` and `firstUpsertBatch` need to be updated within that helper method, I ran into a lot of trouble with pass-by-value-v-reference, updates not happening, etc, and finally gave up and left it with the duplicate code you see here.  Is it worth it to create a response object that contains the schema and firstUpsertBatch?  Is there a better way?

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -75,19 +80,38 @@ public int consumeWriteStream(
...
+              records.stream().collect(Collectors.groupingBy(Record::getRecordType));
+          for (Map.Entry<RecordType, List<Record>> recList : sortedRecords.entrySet()) {
+            RecordType recType = recList.getKey();
+            List<Record> rList = recList.getValue();
+            schema = inferer.inferTypes(records);
```
calypsomatic, line 90: With a single record type, we only check the record type for the first batch, afterwards I suppose we fail if later records don't match the initial schema.  Since a new record type could show up in any batch, I infer the type every time.  This could be simplified by checking against the `result` map and not re-inferring, or potentially using java-pfb to figure out the schema ahead of time.  Opinions?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -94,8 +105,29 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
       throw new PfbParsingException("Error processing PFB", e);
     }
```
yuliadub, line 106: what happens (or should happen) if "id" is missing? Also potentially a very dumb question, but a bit confused about what "id" this is, is it the snapshot Id or something else?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -94,8 +105,29 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
...
 
-    // TODO: AJ-1227 implement PFB import.
-    logger.info("TODO: implement PFB import.");
+    logger.info("Parsing PFB.");
+
+    try (DataFileStream<GenericRecord> dataStream =
+        PfbReader.getGenericRecordsStream(url.toString())) {
+      // TODO should "id" be a static variable
+      Map<RecordType, Integer> result =
+          batchWriteService.batchWritePfbStream(dataStream, workspaceId, Optional.of("id"));
```
yuliadub, line 114: what happens (or should happen) if "id" is missing? Also potentially a very dumb question, but a bit confused about what "id" this is, is it the snapshot Id or something else?

yuliadub: ah wait, I think get it, this is the ID of the specific record that is being parsed.

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
yuliadub: I am a bit confused - since we pass "recordType" as a param into this function, dont we already know what it is or are those somehow different? Or this just how it has to be so that the activity logged can see this?

like here: https://github.com/DataBiosphere/terra-workspace-data-service/blob/f37479456ad6ca71c02de112d4335487ee4ee09a/service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java#L124C68-L124C78

we pass "recordType" as a parameter to the consumeWriteStream function and then use the same recordType to get the number of records written.

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -75,19 +80,39 @@ public int consumeWriteStream(
...
+        } else {
+          result.putIfAbsent(recordType, 0);
+          if (firstUpsertBatch && info.getOperationType() == OperationType.UPSERT) {
+            schema = inferer.inferTypes(records);
+            createOrModifyRecordType(
+                instanceId, recordType, schema, records, primaryKey.orElse(RECORD_ID));
+            firstUpsertBatch = false;
+          }
+          writeBatch(instanceId, recordType, schema, info.getOperationType(), records, primaryKey);
+          result.compute(
```
yuliadub, line 107: this would overwrite the result computer from line 85 above (the if statement part) - is that by design?

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -75,19 +80,39 @@ public int consumeWriteStream(
...
+            Map<String, DataTypeMapping> inferredSchema = inferer.inferTypes(rList);
+            createOrModifyRecordType(
+                instanceId, recType, inferredSchema, rList, primaryKey.orElse(RECORD_ID));
+            writeBatch(
+                instanceId, recType, inferredSchema, info.getOperationType(), rList, primaryKey);
+            result.compute(
+                recType, (key, value) -> (value == null) ? rList.size() : value + rList.size());
+          }
+        } else {
+          result.putIfAbsent(recordType, 0);
```
yuliadub, line 99: wouldn't recordType be "null" the way this is currently implemented? when would that happen and why? I am a bit confused about recordType and what it is I think

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
yuliadub: ah wait, as I am looking at this more, is this because there may be other record types listed within the record itself? and we'd still want to record those? although in the end we kind of end up ignoring any types that are not the recordType that was passed in so I am still confused :) sounds like I need another walkthrough of this process e2e.

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -75,19 +80,39 @@ public int consumeWriteStream(
...
+          if (firstUpsertBatch && info.getOperationType() == OperationType.UPSERT) {
+            schema = inferer.inferTypes(records);
+            createOrModifyRecordType(
+                instanceId, recordType, schema, records, primaryKey.orElse(RECORD_ID));
+            firstUpsertBatch = false;
+          }
+          writeBatch(instanceId, recordType, schema, info.getOperationType(), records, primaryKey);
```
yuliadub, line 106: could we ever end up in a position where schema ends up being null cause we never got an Upsert that marked firstUpsertBatch to false?

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
yuliadub: tried to think through a few variations here and didnt come up with anything yet. I dont think this is a whole log of duplication though and in some ways the duplication makes this a bit easier to track (in my case anyways)

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
calypsomatic: The thing that changes here is that in a pfb, there may be multiple record types all mixed together, whereas the existing import streams only take one record type at a time.  In the case of pfb, I've passed in 'null' as the record type, which indicates that there are possibly multiple and they need to be determined from the stream.  I could do something more elegant but that would require changing the signature type and that would propagate largely throughout the code and seemed dangerous.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
calypsomatic: In PFBs, the primary key of each record (as far as I can tell) is always named "id".  My TODO here is trying to indicate that I'm not sure if I should hardcode that, if so if this is the best way to do so, or if there's a possibility that isn't the case, etc.

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
calypsomatic: It should only reach this else in the case that record type is NOT null, in which case, if I haven't initialized the result map, do that.

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
calypsomatic: MAYBE but the hope is that I haven't changed this part so it should work as it has already been which I think takes care of that?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
davidangb: At some point - as part of this ticket or as a follow-on - let's write down all our assumptions, like:
* primary key is named "id"
* snapshot ids are in "source_datarepo_snapshot_id"
* datetimes are formatted as "2022-06-01T00:00:00.000000Z" (or wherever we end up)
* all records of the same type are contiguous in the PFB
* etc

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -94,8 +105,29 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
...
+              entry -> {
+                RecordType recordType = entry.getKey();
+                int quantity = entry.getValue();
+                activityLogger.saveEventForCurrentUser(
+                    user ->
+                        user.upserted().record().withRecordType(recordType).ofQuantity(quantity));
+              });
+
+    } catch (IOException e) {
+      // TODO where do we identify that there may have been a partial write?
```
davidangb, line 128: I'm curious if we can write the entire PFB inside a single transaction, and thus either commit or rollback the whole thing and avoid partial writes

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
davidangb: Instead of returning `Map<RecordType, Integer>`, could you make a new class `BatchWriteResult` (or whatever you want to name it) which includes, as a member, the map of record type->count? Then, you could add a `BatchWriteResult.getUpdatedCount(RecordType recordType)` method - keep all the logic encapsulated inside `BatchWriteResult` and don't require callers to know how to interpret the structure.

service/src/main/java/org/databiosphere/workspacedataservice/service/PfbStreamWriteHandler.java
```diff
@@ -0,0 +1,66 @@
...
+                Spliterators.spliteratorUnknownSize(inputStream.iterator(), Spliterator.ORDERED),
+                false)
+            .limit(numRecords);
+
+    List<Record> records = recordStream.map(this::genericRecordToRecord).toList();
+
+    return new WriteStreamInfo(records, OperationType.UPSERT);
+  }
+
+  // TODO how to not write the whole path
```
davidangb, line 39: with the 
```java
import org.databiosphere.workspacedataservice.shared.model.Record;
```
in place at line 13,
```java
private Record genericRecordToRecord(...
```
works just fine for me

service/src/main/java/org/databiosphere/workspacedataservice/service/PfbStreamWriteHandler.java
```diff
@@ -0,0 +1,66 @@
...
+  private org.databiosphere.workspacedataservice.shared.model.Record genericRecordToRecord(
+      GenericRecord genRec) {
+    // TODO ok to use plain strings here so should they be static variables or what?
+    org.databiosphere.workspacedataservice.shared.model.Record converted =
+        new org.databiosphere.workspacedataservice.shared.model.Record(
+            genRec.get("id").toString(),
+            RecordType.valueOf(genRec.get("name").toString()),
+            RecordAttributes.empty());
+    GenericRecord objectAttributes = (GenericRecord) genRec.get("object"); // contains attributes
+    Schema schema = objectAttributes.getSchema();
```
davidangb, line 49: is the `Schema` dynamic per `GenericRecord`? I'm wondering if there's any way to get the schema once per record _type_ instead of once per _record_, potentially optimize the flow a bit.

service/src/main/java/org/databiosphere/workspacedataservice/service/PfbStreamWriteHandler.java
```diff
@@ -0,0 +1,66 @@
...
+    List<Schema.Field> fields = schema.getFields();
+    RecordAttributes attributes = RecordAttributes.empty();
+    for (Schema.Field field : fields) {
+      String fieldName = field.name();
+      Object value =
+          objectAttributes.get(fieldName) == null ? null : objectAttributes.get(fieldName);
+      attributes.putAttribute(fieldName, value);
```
davidangb, line 56: this is where we'd need to convert the native PFB types into the types that `Record` expects, such as datetimes and `BigDecimals`, right?

calypsomatic: It's figured out during `BatchWriteService.consumeWriteStream`, which calls `inferTypes`, but maybe if we did something here we could make that more reliable - or avoid calling `inferTypes` for PFBs altogether

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
yuliadub: is it safe to assume that "id" in this context will always be unique?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
yuliadub: we could just fail if its patrial since likely partial data is not useful and creates more issues when trying to re-import?

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
yuliadub: I agree with David's suggestion. It wasnt straightforward to track that you were indeed passing null in this context and that can become confusing in the future if that detail is missed

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
yuliadub: ah I see, I think I had this backwards. This is the case when you know there are not multiple records vs the other case is when there are, which means you dont have one specific record type. I got confused with that logic :/

calypsomatic: I'll add helpful comments as I fix it up :)

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
calypsomatic: Looks like `batchwriteservice` already actually prevents partial writes

service/src/test/java/org/databiosphere/workspacedataservice/service/BatchWriteServiceTest.java
```diff
@@ -55,4 +76,316 @@ void testRejectsDuplicateKeys() throws IOException {
...
+                  "released",
+                  false));
+      RecordType expectedRecordType = RecordType.valueOf("data_release");
+      Record expectedRecord =
+          new Record(
+              "data_release.3511bcae-8725-53f1-b632-d06a9697baa5.1",
+              expectedRecordType,
+              expectedAttributes);
+      // TODO deal with data types, setting them to what actually happens for now
+      Map<String, DataTypeMapping> expectedSchema =
```
calypsomatic, line 155: I'm torn on whether I need to verify schema and attributes here or not - the attributes seem like they should be checked in `PfbStreamWriteHandlerTest`, so that checking them here too would be redundant, but it also kind of makes sense to check what's getting upserting.  Thoughts?

davidangb (commented):
a few interim comments, as you continue to work on this

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
davidangb: I was able to consolidate this, please give it a look!

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
davidangb: I added some code around this to only change the schema the first time a given record type is seen. I also am pushing a lot of this work off to https://broadworkbench.atlassian.net/browse/AJ-1452 (I still need to add details to the Jira ticket)

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
davidangb: it is not safe to assume that "id" in this context will always be unique … but if it is not, it will be treated like other upserts. Subsequent rows with the same id value will upsert over the earlier rows.

service/src/main/java/org/databiosphere/workspacedataservice/service/PfbStreamWriteHandler.java
davidangb: pushing this off to https://broadworkbench.atlassian.net/browse/AJ-1452 .

service/src/main/java/org/databiosphere/workspacedataservice/service/PfbStreamWriteHandler.java
davidangb: pushing this off to https://broadworkbench.atlassian.net/browse/AJ-1452 .

service/src/test/java/org/databiosphere/workspacedataservice/service/BatchWriteServiceTest.java
davidangb: I removed a bunch of the schema/attribute testing. Since https://broadworkbench.atlassian.net/browse/AJ-1452 will do the bulk of the work with schemas and attributes, I think we should add more test coverage then.

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -75,19 +79,37 @@ public int consumeWriteStream(
...
+        // If no recordType is given, assume there may be multiple types and sort by type
+        if (recordType == null) {
+          Map<RecordType, List<Record>> sortedRecords =
+              records.stream().collect(Collectors.groupingBy(Record::getRecordType));
+          for (Map.Entry<RecordType, List<Record>> recList : sortedRecords.entrySet()) {
+            RecordType recType = recList.getKey();
+            List<Record> rList = recList.getValue();
+            Map<String, DataTypeMapping> inferredSchema = inferer.inferTypes(rList);
+            createOrModifyRecordType(
+                instanceId, recType, inferredSchema, rList, primaryKey.orElse(RECORD_ID));
```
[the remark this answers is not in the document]
davidangb: I was able to implement this - we now only do schema work the first time we see a given record type.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -60,44 +70,109 @@ protected JobDao getJobDao() {
...
+    UUID targetInstance = getJobDataUUID(jobDataMap, ARG_INSTANCE);
+
+    // Find all the snapshot ids in the PFB, then create or verify references from the
+    // workspace to the snapshot for each of those snapshot ids.
+    // This will throw an exception if there are policy conflicts between the workspace
+    // and the snapshots.
+    // TODO AJ-1452: can this pass also identify all schemas/datatypes?
+    // TODO AJ-1452: can this pass also check if record types are contiguous?
+    logger.info("Finding snapshots in this PFB...");
+    List<UUID> snapshotIds = withPfbStream(url, this::findSnapshots);
```
jladieu, line 85: Are there ever repeated entries of snapshot IDs?  If so, might `snapshotIds` benefit from being a `Set`?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -60,44 +70,109 @@ protected JobDao getJobDao() {
...
+
+  /**
+   * definition for some function that consumes a PFB stream (as a DataFileStream<GenericRecord>)
+   */
+  @FunctionalInterface
+  public interface PfbStreamConsumer<T> {
+    T run(DataFileStream<GenericRecord> dataStream);
+  }
+
+  /**
```
jladieu, line 105: Love this :)

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -60,44 +70,109 @@ protected JobDao getJobDao() {
...
+   * Given a DataFileStream representing a PFB, import all the tables and rows inside that PFB.
+   *
+   * @param dataStream stream representing the PFB.
+   */
+  BatchWriteResult importTables(DataFileStream<GenericRecord> dataStream, UUID targetInstance) {
+    BatchWriteResult result =
+        batchWriteService.batchWritePfbStream(dataStream, targetInstance, Optional.of(ID_FIELD));
+
+    result
+        .entrySet()
```
jladieu, line 131: I'm not sure what activityLogger does downstream, but assuming it might be expensive, I wonder if this could be done as a rollup of _all_ entries rather than one per entry?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -60,44 +70,109 @@ protected JobDao getJobDao() {
...
+        StreamSupport.stream(
+            Spliterators.spliteratorUnknownSize(dataStream.iterator(), Spliterator.ORDERED), false);
+
+    // process the stream into a list of unique snapshotIds
+    return recordStream
+        .map(rec -> rec.get("object")) // Records in a pfb are stored under the key "object"
+        .filter(GenericRecord.class::isInstance) // which we expect to be a GenericRecord
+        .map(GenericRecord.class::cast)
+        .filter(obj -> obj.hasField(SNAPSHOT_ID_IDENTIFIER)) // avoid exception if field nonexistent
+        .map(obj -> obj.get(SNAPSHOT_ID_IDENTIFIER)) // within the GenericRecord, find the
```
jladieu, line 161: I noticed the inline comment wrapped here; maybe you can drop the extra text about 'within the GenericRecord' by renaming `obj` to `record` and get it to fit on one line again?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -122,7 +197,7 @@ protected void linkSnapshots(List<UUID> snapshotIds) {
         restClientRetry.withRetryAndErrorHandling(
             voidRestCall, "WSM.createDataRepoSnapshotReference");
       } catch (Exception e) {
```
jladieu, line 199: Is there potentially a narrower exception that you could catch here or is it really super important that _all_ exceptions be handled in this way?

service/src/main/java/org/databiosphere/workspacedataservice/service/PfbStreamWriteHandler.java
```diff
@@ -0,0 +1,41 @@
...
+    Stream<GenericRecord> pfbBatch =
+        StreamSupport.stream(
+                Spliterators.spliteratorUnknownSize(inputStream, Spliterator.ORDERED), false)
+            .limit(numRecords);
+
+    // convert the PFB GenericRecord objects into WDS Record objects
+    PfbRecordConverter pfbRecordConverter = new PfbRecordConverter();
+    List<Record> records = pfbBatch.map(pfbRecordConverter::genericRecordToRecord).toList();
+
+    return new WriteStreamInfo(records, OperationType.UPSERT);
```
jladieu, line 34: I didn't really understand why setting the operation type is a concern of this class.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJobTest.java
```diff
@@ -113,6 +136,95 @@ void linkSomeWhenSomeExist() throws ApiException {
                 "should not have created reference to pre-existing snapshot"));
   }
 
+  @Test
+  void doNotFailOnMissingSnapshotId() throws JobExecutionException {
+    JobExecutionContext mockContext = mock(JobExecutionContext.class);
+    // This uses a non-TDR file so does not have the snapshotId
+    URL resourceUrl = getClass().getResource("/minimal_data.avro");
```
jladieu, line 143: `getClass().getResource()` would make a good private method (eg: `getUrlToTestData`)

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJobTest.java
```diff
@@ -45,7 +65,8 @@ void linkAllNewSnapshots() throws ApiException {
 
     // call linkSnapshots
     PfbQuartzJob pfbQuartzJob =
-        new PfbQuartzJob(jobDao, wsmDao, restClientRetry, UUID.randomUUID());
+        new PfbQuartzJob(
```
jladieu, line 68: It looks like a lot goes into building a `PfbQuartzJob` which leads me to other questions...but could this be encapsulated by a private helper method?

Then if we ever want to change the constructor we only have to change it in one spot in this file.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJobTest.java
```diff
@@ -113,6 +136,95 @@ void linkSomeWhenSomeExist() throws ApiException {
...
+    // Should not call wsm dao
+    verify(wsmDao, times(0)).linkSnapshotForPolicy(any());
+    // But job should succeed
+    verify(jobDao).succeeded(jobId);
+  }
+
+  @Test
+  void snapshotIdsAreParsed() throws JobExecutionException {
+    JobExecutionContext mockContext = mock(JobExecutionContext.class);
+    URL resourceUrl = getClass().getResource("/test.avro");
```
jladieu, line 181: This is a lot of setup code that might lend itself to factoring out into a method with an intention revealing name to get something like:

```
// declare here since we need it later on to verify success, otherwise I'd let stub take care of it
UUID jobId = UUID.randomUUID(); 
JobExecutionContext = stubJobContext("/test.avro", jobId);
```

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJobTest.java
```diff
@@ -113,6 +136,95 @@ void linkSomeWhenSomeExist() throws ApiException {
...
+
+    new PfbQuartzJob(
+            jobDao, wsmDao, restClientRetry, batchWriteService, activityLogger, UUID.randomUUID())
+        .execute(mockContext);
+
+    // This is the snapshotId given in the test pfb
+    verify(wsmDao)
+        .linkSnapshotForPolicy(
+            ArgumentMatchers.argThat(
+                new SnapshotModelMatcher(UUID.fromString("790795c4-49b1-4ac8-a060-207b92ea08c5"))));
```
jladieu, line 214: This hardcoded UUID seems like it comes out of nowhere...is it defined anywhere else such that we could refer to a constant/variable here?

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -61,33 +67,72 @@ public BatchWriteService(
...
-          createOrModifyRecordType(
-              instanceId, recordType, schema, records, primaryKey.orElse(RECORD_ID));
-          firstUpsertBatch = false;
+
+        // Group the incoming records by their record types. PFB inputs expect to have multiple
+        // types within the same input stream. TSV and JSON are expected to have a single record
+        // type, so this will result in a grouping of 1.
+        // TSV and JSON inputs are validated against the recordType argument. PFB inputs pass
+        // a null recordType argument so there is nothing to validate.
+        Map<RecordType, List<Record>> sortedRecords =
```
jladieu, line 94: Is this variable incorrectly named?  I don't see any sorting happening anywhere...

jladieu, line 94: How do you feel about Guava data structures?  This (and several other complex data structures in this PR) looks like it might lend itself well to being represented as a higher level data structure.  In this case, I wonder if `Multimap` (ref: https://guava.dev/releases/19.0/api/docs/com/google/common/collect/Multimap.html) might simplify things.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverterTest.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericData;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.commons.lang3.RandomStringUtils;
+import org.databiosphere.workspacedataservice.shared.model.Record;
+import org.junit.jupiter.api.Test;
+
+class PfbRecordConverterTest {
```
jladieu, line 17: Nice tests & explanatory comments!

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -109,10 +155,21 @@ public int batchWriteTsvStream(
...
+              recordType,
+              Optional.of(streamingWriteHandler.getResolvedPrimaryKey()))
+          .getUpdatedCount(recordType);
+    } catch (IOException e) {
+      throw new BadStreamingWriteRequestException(e);
+    }
+  }
+
+  @WriteTransaction
+  public BatchWriteResult batchWritePfbStream(
```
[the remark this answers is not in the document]
davidangb: I agree the `primaryKey` argument here is a little strange; I wondered about the same thing. The benefits of having it here are:
* it keeps the `batchWritePfbStream` signature mostly parallel to `batchWriteJsonStream` and `batchWriteTsvStream`
* it keeps `BatchWriteService` generic, and keeps knowledge of the expected "id" column name encapsulated inside `PfbQuartzJob`. This may be important as we import PFBs from providers beyond AnVIL.

I've added a comment to explain!

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
davidangb: Good call, not sure why I used a `List` here, updated.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
davidangb: right now, the activity logger is just a facade to logging; at some future point it may also start writing to the db or performing other calculations - see [this thread](https://github.com/DataBiosphere/terra-workspace-data-service/pull/339#discussion_r1334326097), for instance.

I'd vote for leaving this as-is for now, until/unless we see it is too noisy in practice.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
davidangb: reworded to fit on one line.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
davidangb: updated to catch `RestException` instead.

service/src/test/java/org/databiosphere/workspacedataservice/service/BatchWriteServiceTest.java
```diff
@@ -55,4 +86,133 @@ void testRejectsDuplicateKeys() throws IOException {
...
+          .batchUpsertWithErrorCapture(
+              eq(INSTANCE), eq(RecordType.valueOf("item")), any(), any(), eq(primaryKey.get()));
+      verify(recordService, times(8))
+          .batchUpsertWithErrorCapture(
+              eq(INSTANCE), eq(RecordType.valueOf("widget")), any(), any(), eq(primaryKey.get()));
+
+      // but we should only have inferred schemas three times - once for each record Type
+      @SuppressWarnings("unchecked")
+      ArgumentCaptor<List<Record>> argumentCaptor = ArgumentCaptor.forClass(List.class);
+      verify(inferer, times(3)).inferTypes(argumentCaptor.capture());
```
jladieu, line 163: Do we know any more specifics about _when_ the type inference should have happened?  I'm just wondering if instead of verifying that it only happened once overall for each type, if instead we might check more specifically (it should have only inferred "thing" after the first call, "item" after the second call, and "widget" after the third call...

And then maybe test that types aren't inferred redundantly by following up with subsequent batch upserts with those same types and confirming that `inferTypes` isn't called at all?

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -146,14 +203,14 @@ private void writeBatch(
       UUID instanceId,
       RecordType recordType,
       Map<String, DataTypeMapping> schema,
-      StreamingWriteHandler.WriteStreamInfo info,
+      OperationType opType,
```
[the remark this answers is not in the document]
davidangb: `writeBatch` is called as a result of `batchWriteJsonStream`, `batchWriteTsvStream`, or (as of this PR) `batchWritePfbStream`. We allow both UPSERT and DELETE `OperationType`s via `batchWriteJsonStream`, so this is a result of centralizing the flow for all three of those import types through the same code path.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJobTest.java
davidangb: I moved these to Spring-managed injection, via the `@Value` annotation

service/src/main/java/org/databiosphere/workspacedataservice/service/PfbStreamWriteHandler.java
davidangb: see previous comment; this is an artifact of how `writeBatch` is common across PFB, TSV, and JSON API calls

jladieu (commented):
Looking good!  I had a couple comments/questions throughout.

The suggestions about using Guava stuff is completely optional and happy to punt that but it might be an opportunity to try some stuff out.

I am most interested in better understanding `opType` and the interactions it has with the code that has to know about it.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJobTest.java
davidangb: moved to a helper method.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJobTest.java
davidangb: moved to a helper method.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJobTest.java
davidangb: I've updated the code comment just before this `verify` to hopefully explain it better

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
davidangb: switched to a multimap. I'm all for it, as long as it's readable to our engineers who may or may not have experience with Guava. I think it's a good call though and worth the exposure.

service/src/test/java/org/databiosphere/workspacedataservice/service/BatchWriteServiceTest.java
davidangb: I think we should defer this question until AJ-1452, because it may change significantly. I've added a TODO comment to that effect.

jladieu (approved):

service/src/main/java/org/databiosphere/workspacedataservice/service/model/BatchWriteResult.java
```diff
@@ -0,0 +1,40 @@
...
+  public static BatchWriteResult empty() {
+    return new BatchWriteResult(new HashMap<>());
+  }
+
+  public Integer getUpdatedCount(RecordType recordType) {
+    return resultCounts.get(recordType);
+  }
+
+  public void increaseCount(RecordType recordType, int count) {
+    Preconditions.checkArgument(count >= 0, "Count cannot be negative");
```
yuliadub, line 29: would we want to throw or log an error here since having it ever be negative maybe is worth looking into?

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverterTest.java
```diff
@@ -0,0 +1,87 @@
...
+            "namespace",
+            false,
+            List.of(
+                new Schema.Field("marco", Schema.create(Schema.Type.STRING)),
+                new Schema.Field("pi", Schema.create(Schema.Type.LONG)),
+                new Schema.Field("afile", Schema.create(Schema.Type.STRING))));
+
+    GenericData.Record objectAttributes = new GenericData.Record(myObjSchema);
+    objectAttributes.put("marco", "polo");
+    objectAttributes.put("pi", 3.14159);
```
yuliadub, line 68: thinking of this bug (https://broadworkbench.atlassian.net/browse/AJ-1292) should we test with more decimal places? maybe this is something to add in the future but since it came to my mind wanted to call this out

yuliadub (approved):
@davidangb Thank you for all the work on this between you and Bria, this is starting to make more sense in my brain!

service/src/main/java/org/databiosphere/workspacedataservice/service/model/BatchWriteResult.java
davidangb: this will throw a `IllegalArgumentException` if it is negative!

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverterTest.java
davidangb: I think we should defer that to AJ-1452 - in this PR, I'm not trying to handle datatypes at all, but when we tackle datatype stuff we should add tests as necessary.

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -105,7 +105,7 @@ private BatchWriteResult consumeWriteStream(
         // time we've seen this type, calculate a schema from its records and update the record type
         // as necessary. Then, write the records into the table.
         // TODO AJ-1452: for PFB imports, get schema from Avro, not from attribute values inference
-        for (RecordType recType : groupedRecords.keys()) {
+        for (RecordType recType : groupedRecords.keySet()) {
```
jladieu, line 108: Just linking the [thread](https://broadinstitute.slack.com/archives/C02SJQ0FKHQ/p1699970783134969) talking about your odyssey to get to this change which is just brilliant!

jladieu (approved):

PR #389, opened 2023-10-30
AJ-1396: refix bearer token

davidangb:
#370 mistakenly reverted the fix from #335.

This PR unreverts that, so WDS cloning should work again. Also includes a unit test.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/sourcewds/HttpWorkspaceDataServiceClientFactory.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/sourcewds/HttpWorkspaceDataServiceClientFactory.java
@@ -23,7 +23,7 @@ public class HttpWorkspaceDataServiceClientFactory implements WorkspaceDataServi
         new ApiClient().getHttpClient().newBuilder().protocols(List.of(Protocol.HTTP_1_1)).build();
   }
 
-  protected ApiClient getApiClient(String authToken, String workspaceDataServiceUrl) {
+  ApiClient getApiClient(String authToken, String workspaceDataServiceUrl) {
     // create a new client
     ApiClient apiClient = new ApiClient();
     apiClient.setHttpClient(commonHttpClient);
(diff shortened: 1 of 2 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/sourcewds/HttpWorkspaceDataServiceClientFactoryTest.java
davidangb, line 1: I verified that this test fails when `HttpWorkspaceDataServiceClientFactory` calls `apiClient.setAccessToken`, but passes when it calls `apiClient.setBearerToken`.

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/sourcewds/HttpWorkspaceDataServiceClientFactory.java
```diff
@@ -23,7 +23,7 @@ public HttpWorkspaceDataServiceClientFactory() {
         new ApiClient().getHttpClient().newBuilder().protocols(List.of(Protocol.HTTP_1_1)).build();
   }
 
-  private ApiClient getApiClient(String authToken, String workspaceDataServiceUrl) {
+  protected ApiClient getApiClient(String authToken, String workspaceDataServiceUrl) {
```
jladieu, line 26: Could this be slightly less permissive at package protected instead?  Or do you want to explicitly make sure subclasses can see it too?

jladieu (approved):
One question but lgtm!

service/src/main/java/org/databiosphere/workspacedataservice/sourcewds/HttpWorkspaceDataServiceClientFactory.java
davidangb: switched to default visibility.

PR #390, opened 2023-10-30
AJ-1396: allow cloning earlier versions of WDS

davidangb:
Resolves an error when cloning earlier versions of WDS. The error is `BACKUPERROR (The required field `jobType` is not found in the JSON string`:
Screenshot 10-30-2023 at 04 04 PM

And is caused by https://github.com/DataBiosphere/terra-workspace-data-service/pull/358, which added `jobType` as a required field in the `BackupJob` model.

This PR makes `jobType` not required.


yuliadub (approved):
thank you!

service/src/test/java/org/databiosphere/workspacedataservice/sourcewds/CloneOlderWdsTest.java
davidangb, line 1: I verified this test fails when` jobType` is required

nawatts (approved):

ashanhol (approved):

PR #391, opened 2023-10-30
AJ-1409: add property to policy-only snapshot references

davidangb:


service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
```diff
@@ -23,8 +32,20 @@ public WorkspaceManagerDao(
...
+      policyProperty.setKey(PROP_PURPOSE);
+      policyProperty.setValue(PURPOSE_POLICY);
+      properties = new Properties();
+      properties.add(policyProperty);
+    }
+    createDataRepoSnapshotReference(snapshotModel, properties);
+  }
+
   /** Creates a snapshot reference in workspaces manager and creates policy linkages. */
-  public void createDataRepoSnapshotReference(SnapshotModel snapshotModel) {
```
calypsomatic, line 27: Xtreme nit that I see is not even from this PR: "workspaces manager" should be "workspace manager"

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
```diff
@@ -23,8 +32,20 @@ public WorkspaceManagerDao(
     this.workspaceId = workspaceId;
   }
 
+  public void createDataRepoSnapshotReference(SnapshotModel snapshotModel, Boolean policyOnly) {
```
calypsomatic, line 35: Are we only ever going to care about the policy property, or is there any reason to make the use of properties more flexible than just policy yes/no in this method?

nawatts: If nothing else, is there a more readable alternative to a boolean argument?

https://alexkondov.com/should-you-pass-boolean-to-functions/

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
davidangb: fixed :)

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
davidangb: I realized that (contrary to the commits you reviewed) today in WDS, we *only* create snapshot references for the purpose of policy linking. So, I removed the boolean argument and renamed the method, while still leaving the internal private method in place for future expansion.

calypsomatic (approved):

nawatts (approved):

PR #393, opened 2023-10-31
AJ-1429 clone failure

davidangb:
*sigh* #386 broke cloning by extracting a reusable function `WorkspaceManagerDao.enumerateResources()`.

This PR fixes that regression by ensuring the auth token is passed properly through `enumerateResources()`, plus unit tests to verify.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
@@ -50,7 +50,12 @@ public class WorkspaceManagerDao {
       throws ApiException {
     // get a page of results from WSM
     return enumerateResources(
-        workspaceId, offset, limit, ResourceType.DATA_REPO_SNAPSHOT, StewardshipType.REFERENCED);
+        workspaceId,
+        offset,
+        limit,
+        ResourceType.DATA_REPO_SNAPSHOT,
+        StewardshipType.REFERENCED,
+        null);
   }
 
   /** Retrieves the azure storage container url and sas token for a given workspace. */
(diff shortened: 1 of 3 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/sam/TokenContextUtilTest.java
```diff
@@ -18,6 +18,29 @@
...
+      JobContextHolder.init();
+      JobContextHolder.setAttribute(ATTRIBUTE_NAME_TOKEN, "dummy job value");
+
+      // call getToken with a non-null initialValue
+      BearerToken actual = TokenContextUtil.getToken(expected);
+      assertEquals(BearerToken.of(expected), actual);
+    } finally {
+      JobContextHolder.destroy();
+    }
+  }
```
davidangb, line 42: This isn't technically part of the PR, but during investigation of this issu I noticed that `TokenContextUtil.getToken(String initialValue)` was missing a unit test, so I added the test.

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
```diff
@@ -50,7 +50,12 @@ public ResourceList enumerateDataRepoSnapshotReferences(UUID workspaceId, int of
       throws ApiException {
     // get a page of results from WSM
     return enumerateResources(
-        workspaceId, offset, limit, ResourceType.DATA_REPO_SNAPSHOT, StewardshipType.REFERENCED);
+        workspaceId,
+        offset,
+        limit,
+        ResourceType.DATA_REPO_SNAPSHOT,
+        StewardshipType.REFERENCED,
+        null);
```
yuliadub, line 58: does this not need a token since the token will come from the async job context?

davidangb: correct!

yuliadub (commented):
my bad to not have caught this in the async changes PR. looks good!

yuliadub (approved):

PR #396, opened 2023-11-01
[AJ-1160] Add Retries to WSM and Leo 

ashanhol:
Initial PR #385 added a reusable REST function retry library `RestClientRetry` and hooked up SAM. This PR hooks up all Leo calls in LeoDao, all WSM calls in WorkspaceManagerDao, and a single call to `createBackup` in WorkspaceDataServiceDao using `RestClientRetry.withRetryAndErrorHandling()`. There are try/catches around every withRetryAndErrorHandling catching `RestException`s

----------------------
Reminder:


service/src/test/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDaoTest.java
```diff
@@ -50,7 +52,7 @@ void testWdsUrlNotReturned() throws ApiException {
             new org.broadinstitute.dsde.workbench.client.leonardo.ApiException(
```
ashanhol, line 52: So this test is failing, but not for the reason you'd expect. For some reason even though it's specified to throw an exception (with given/willThrow), `listAppsV2()` is not throwing an ApiException. I tried replacing the arguments with `any()` and it still failed to throw.  This is especially weird considering the WorkspaceManagerDaoTest didn't seem to have any problem throwing its exception.

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
```diff
@@ -55,36 +65,39 @@ public ResourceList enumerateDataRepoSnapshotReferences(UUID workspaceId, int of
...
+                    workspaceUUID, storageUUID, null, null, null, null);
+        CreatedAzureStorageContainerSasToken sasBundle =
+            restClientRetry.withRetryAndErrorHandling(sasBundleFunction, "WSM.sasBundle");
+        return sasBundle.getUrl();
+      } else {
+        throw new ApiException(
+            "WorkspaceManagerDao: Can't locate a storage resource matching workspace Id. ");
       }
+    } catch (ApiException e) {
+      throw new ResponseStatusException(HttpStatus.INTERNAL_SERVER_ERROR, null, e);
```
ashanhol, line 100: Do we want to keep WorkspaceManagerException? It seems like it was just a wrapper for this ^

service/src/main/java/org/databiosphere/workspacedataservice/sourcewds/WorkspaceDataServiceDao.java
```diff
@@ -25,16 +30,12 @@ public void setWorkspaceDataServiceUrl(String endpointUrl) {
...
-    } catch (ApiException e) {
-      LOGGER.warn("Error from remote WDS: {}", e.getMessage());
-      throw new WorkspaceDataServiceException(e);
-    }
+    var backupClient =
+        this.workspaceDataServiceClientFactory.getBackupClient(token, workspaceDataServiceUrl);
+    BackupRestoreRequest body = new BackupRestoreRequest();
+    body.setRequestingWorkspaceId(requesterWorkspaceId);
+    RestCall<BackupJob> createBackupFunction = () -> backupClient.createBackup(body, "v0.2");
+    return restClientRetry.withRetryAndErrorHandling(createBackupFunction, "WDS.createBackup");
```
ashanhol, line 38: This gives a second(+?) chance to create a backup in case the first fails. I've seen this flake so I took the liberty of adding a retry here.

service/src/main/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDao.java
```diff
@@ -5,37 +5,47 @@
...
-
-      throw new ApiException("Did not locate an app running WDS.");
-    } catch (ApiException e) {
-      throw new LeonardoServiceException(e);
+    RestCall<AppsApi> getAppsFunction = () -> this.leonardoClientFactory.getAppsV2Api(token);
+    var workspaceApps = restClientRetry.withRetryAndErrorHandling(getAppsFunction, "Leo.getApps");
+    RestCall<List<ListAppResponse>> listAppsFunction =
+        () -> workspaceApps.listAppsV2(workspaceId, null, false, null);
+    List<ListAppResponse> response =
+        restClientRetry.withRetryAndErrorHandling(listAppsFunction, "WorkspaceApps.listApps");
```
yuliadub, line 38: how many retries will this do? sorry if I am missing where this is set

yuliadub: or is this just one retry?

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
```diff
@@ -55,36 +65,39 @@ public ResourceList enumerateDataRepoSnapshotReferences(UUID workspaceId, int of
...
+      RestCall<ControlledAzureResourceApi> azureResourceApiFunction =
+          () -> this.workspaceManagerClientFactory.getAzureResourceApi(authToken);
+      ControlledAzureResourceApi azureResourceApi =
+          restClientRetry.withRetryAndErrorHandling(
+              azureResourceApiFunction, "WSM.azureResourceApi");
+      UUID workspaceUUID = UUID.fromString(storageWorkspaceId);
+      LOGGER.debug(
+          "Finding storage resource for workspace {} from Workspace Manager ...", workspaceUUID);
+      ResourceList resourceList =
+          enumerateResources(workspaceUUID, 0, 5, ResourceType.AZURE_STORAGE_CONTAINER, null);
```
yuliadub, line 78: this is the bug (there should be a token here, not null) that David fixed earlier - looks like you have some conflicts to fix in this file, just want to note to not miss this 😸

service/src/main/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDao.java
ashanhol: It's set as an environment variable used in[ RestClientRetry ](https://github.com/DataBiosphere/terra-workspace-data-service/pull/385/files#diff-be054dd87004e9dde5f95de0c49574e7d7c741087e9d7fb547e855ca0bf92b7eR35)

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
ashanhol: The test that's failing is in LeoDaoTest though... (will resolve those conflicts)

service/src/main/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDao.java
```diff
@@ -5,37 +5,47 @@
...
-      // unsure what the key would be if there is more than 1 wds present in the listed apps, but in
-      // this case our assumption is
-      // it is acceptable to fail if we cant find a single RUNNING wds in the proxy urls
-      var url = extractWdsUrl(response);
-      if (url != null) {
-        return url;
-      }
-
-      throw new ApiException("Did not locate an app running WDS.");
-    } catch (ApiException e) {
```
yuliadub, line 36: I have a theory that the reason the exception is no longer being thrown is because you took the catch exception throwing code out - if you bring this catch back, I bet the test will work.

yuliadub: the try/catch is there in case the WDS extraction parsing fails (like Leo changes what the response looks like) - so we should bring it back to handle that case

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
yuliadub: I think the reason for WorkspaceManagerException is that we can tell that this exception is coming from Workspace manager with a very specific error message vs having a more generic exception. I'd say in my opinion we keep that to have it be consistent with other clients.

service/src/test/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDaoTest.java
yuliadub: see this comment: https://github.com/DataBiosphere/terra-workspace-data-service/pull/396/files#r1380482256

service/src/test/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDaoTest.java
```diff
@@ -84,7 +86,7 @@ void testErrorThrown() throws ApiException {
         .willThrow(new ApiException(statusCode, "Intentional error thrown for unit test"));
     var exception =
         assertThrows(
-            WorkspaceManagerException.class,
+            AuthenticationException.class,
```
yuliadub, line 89: I am not sure I understand the changes to the type of exceptions (same applies to the Leo exception above) -- this is not even an exception that we throw anywhere inside WSM client?

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
yuliadub: yeah sorry not related to tests, just want to make sure we dont miss empty token vs not since we have beofre

service/src/test/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDaoTest.java
ashanhol: It's the exception thrown by RestClientRetry.withRetryAndErrorHandling (it throws 3 types of exception). I followed the same type of test changes made in David's PR

jladieu: I don't think this makes sense to change...if the mock is, in fact, throwing an `ApiException` from WSM, it should be bubbled up as a `WorkspaceManagerException` (not an Auth exception).  If it _is_, and this is the only way to get the test to pass II think it is pointing to a bug somewhere.

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerException.java
```diff
@@ -1,16 +0,0 @@
-package org.databiosphere.workspacedataservice.workspacemanager;
```
jladieu, line 1: Ohh, I see that you removed this class (which explains why you had to re-type the exception in the WorkspaceManagerDaoTest...

Without having looked at the full PR, I have a feeling we shouldn't need to remove this exception and am curious why you wanted/needed to?

service/src/test/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDaoTest.java
jladieu: Ok I understand this better after looking at the upstream PR...I think maybe instead of throwing an `HttpStatus.UNAUTHORIZED` out of WSM, it might make more sense to throw a different category of error like HttpStatus.CONFLICT, and then confirm it gets surfaced as a `RestException.class`.

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
```diff
@@ -105,11 +118,16 @@ private ResourceList enumerateResources(
...
       ResourceType resourceType,
-      StewardshipType stewardshipType)
-      throws ApiException {
-    ResourceApi resourceApi = this.workspaceManagerClientFactory.getResourceApi(null);
-    // TODO: retries
-    return resourceApi.enumerateResources(
-        workspaceId, offset, limit, resourceType, stewardshipType);
+      StewardshipType stewardshipType) {
+    RestCall<ResourceApi> resourceApiFunction =
+        () -> this.workspaceManagerClientFactory.getResourceApi(null);
```
jladieu, line 123: I'm not sure it makes sense to wrap the `getResourceApi` call with retry and error handling.  AFAIK `getResourceApi` is a synchronous internal call to just get an object reference, not something that's actually making a REST call.

ashanhol: I honestly wasn't sure. VSCode made them all seem like REST calls. I'll change them back

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
```diff
@@ -105,11 +118,16 @@ private ResourceList enumerateResources(
...
+    ResourceApi resourceApi =
+        restClientRetry.withRetryAndErrorHandling(resourceApiFunction, "WSM.resourceApi");
+    RestCall<ResourceList> enumerateResourcesFunction =
```
jladieu, line 126: How does this look when you inline `enumerateResourcesFunction`?  I have a hunch it might be more readable when paired with removing the unneeded retry handling for `getResourceApi`

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
```diff
@@ -55,36 +65,39 @@ public ResourceList enumerateDataRepoSnapshotReferences(UUID workspaceId, int of
...
-    int maxTries = 3;
-    while (true) {
-      try {
-        UUID workspaceUUID = UUID.fromString(storageWorkspaceId);
+    try {
+      RestCall<ControlledAzureResourceApi> azureResourceApiFunction =
+          () -> this.workspaceManagerClientFactory.getAzureResourceApi(authToken);
+      ControlledAzureResourceApi azureResourceApi =
+          restClientRetry.withRetryAndErrorHandling(
+              azureResourceApiFunction, "WSM.azureResourceApi");
```
jladieu, line 73: I don't think you need to wrap the `getAzureResourceApi` call in retry logic.

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
jladieu, line 123: should there be a token param here too instead of `null`, as @yuliadub noticed in https://github.com/DataBiosphere/terra-workspace-data-service/pull/396/files#r1380475932

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
```diff
@@ -55,36 +65,39 @@ public ResourceList enumerateDataRepoSnapshotReferences(UUID workspaceId, int of
...
+        RestCall<CreatedAzureStorageContainerSasToken> sasBundleFunction =
+            () ->
+                azureResourceApi.createAzureStorageContainerSasToken(
+                    workspaceUUID, storageUUID, null, null, null, null);
+        CreatedAzureStorageContainerSasToken sasBundle =
+            restClientRetry.withRetryAndErrorHandling(sasBundleFunction, "WSM.sasBundle");
+        return sasBundle.getUrl();
+      } else {
+        throw new ApiException(
+            "WorkspaceManagerDao: Can't locate a storage resource matching workspace Id. ");
```
jladieu, line 97: nit: remove trailing space from error message?
less nitty: add the workspaceUuid to the error message somehow?

service/src/main/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDao.java
```diff
@@ -5,37 +5,47 @@
...
-      throw new LeonardoServiceException(e);
+    RestCall<AppsApi> getAppsFunction = () -> this.leonardoClientFactory.getAppsV2Api(token);
+    var workspaceApps = restClientRetry.withRetryAndErrorHandling(getAppsFunction, "Leo.getApps");
```
jladieu, line 34: per similar other comments, you don't need to wrap the getAppsFunction call in retry logic since it's not going to result in an HTTP REST call.

That also means you don't need a variable called 'getAppsFunction' and can just put it back to the way it was before with:

`var workspaceApps = ...`

service/src/main/java/org/databiosphere/workspacedataservice/service/model/exception/RestServerException.java
```diff
@@ -2,7 +2,7 @@
 
 import org.springframework.http.HttpStatus;
 
-/** Indicates an error response from Sam of a 500, 502, 503, or 504 */
+/** Indicates an error response from our service client of a 500, 502, 503, or 504 */
```
jladieu, line 5: Nice catch :D

jladieu:
Just FYI, seeing how you used the Retry library made me have some thoughts about its usability, which I added as comments to the upstream https://github.com/DataBiosphere/terra-workspace-data-service/pull/385

FYI @davidangb 

Might be churny to try to incorporate my feedback there into _this_ PR, but I wanted to call attention to it as possible room to iterate on as a followup.

service/src/test/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDaoTest.java
davidangb, line 52: I haven't looked to see what changed elsewhere in the code, but the `given` expectation here seems wrong.

```java
given(mockAppsApi.listAppsV2(String.valueOf(UUID.randomUUID()), "things", null, null))
        .willThrow(
```
means: if and only if mockAppsApi.listAppsV2() is called with a workspaceId argument of this specific random UUID, a labels argument of "things", and two nulls, then throw an error. If mockAppsApi.listAppsV2() is called with any other arguments - as is the case inside `getWdsEndpointUrl` - it won't throw an error.

Try a less restrictive `given` for the mock, like:
```java
    given(
            mockAppsApi.listAppsV2(
                anyString(), nullable(String.class), anyBoolean(), nullable(String.class)))
        .willThrow(
```

service/src/main/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDao.java
```diff
@@ -5,37 +5,47 @@
...
+    // unsure what the key would be if there is more than 1 wds present in the listed apps, but in
+    // this case our assumption is
+    // it is acceptable to fail if we cant find a single RUNNING wds in the proxy urls
+    var url = extractWdsUrl(response);
+    if (url != null) {
+      return url;
     }
+
+    throw new ResponseStatusException(
+        HttpStatus.INTERNAL_SERVER_ERROR, "Did not locate an app running WDS.");
```
davidangb, line 48: If we reach this point, this PR now throws a `ResponseStatusException` instead of a `LeonardoServiceException`. What's the reasoning behind this change?

If we _do_ leave it as-is, I believe `LeonardoServiceException` is now unused and could be deleted.

see also Yulia's comments about try/catch above.

service/src/main/java/org/databiosphere/workspacedataservice/sourcewds/WorkspaceDataServiceDao.java
davidangb, line 38: Similar comment as above, I believe this will change the thrown exception from `WorkspaceDataServiceException` to `RestException`; is this intentional?

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
```diff
@@ -8,42 +8,52 @@
...
   }
 
   /** Creates a snapshot reference in workspaces manager and creates policy linkages. */
   public void createDataRepoSnapshotReference(SnapshotModel snapshotModel) {
-    final ReferencedGcpResourceApi resourceApi =
-        this.workspaceManagerClientFactory.getReferencedGcpResourceApi(null);
+    RestCall<ReferencedGcpResourceApi> gcpResourceApiFunction =
+        () -> this.workspaceManagerClientFactory.getReferencedGcpResourceApi(null);
+    ReferencedGcpResourceApi resourceApi =
+        restClientRetry.withRetryAndErrorHandling(gcpResourceApiFunction, "WSM.gcpResourceApi");
```
davidangb, line 39: `getReferencedGcpResourceApi` is not a REST API call, and should not be retried (see also https://github.com/DataBiosphere/terra-workspace-data-service/pull/396/files?diff=unified&w=1#r1380579896)

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
```diff
@@ -8,42 +8,52 @@
...
+                        new DataRepoSnapshotAttributes()
+                            .instanceName(INSTANCE_NAME)
+                            .snapshot(snapshotModel.getId().toString()))
+                    .metadata(
+                        new ReferenceResourceCommonFields()
+                            .cloningInstructions(CloningInstructionsEnum.REFERENCE)
+                            .name("%s_%s".formatted(snapshotModel.getName(), timeStamp))),
+                UUID.fromString(workspaceId));
+    restClientRetry.withRetryAndErrorHandling(
+        createSnapshotFunction, "WSM.createSnapshotReference");
```
davidangb, line 56: and, this will now throw `RestException` instead of `WorkspaceManagerException`; intentional?

davidangb (commented):
see comments inline!

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
ashanhol: conflicts should be fixed now

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerException.java
ashanhol: I removed it because I wasn't using it in code, but consensus was I should put it back. Restored now.

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
ashanhol: Fixed

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
ashanhol: Fixed

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
ashanhol: Should be fixed now with merge conficts resolved

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
ashanhol: Fixed

service/src/main/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDao.java
ashanhol: Fixed

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
ashanhol: Yeah, lots of comments so easy to miss Josh and mine talking about this. fixed now

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
ashanhol: Put it back, ty

service/src/test/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDaoTest.java
ashanhol: That was totally it, thank you. I had tried less restrictive arguments but didn't quite use the same ones.

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
ashanhol: reworked it per our DMs. Added constructor for RestException to WSMException and catch and route appropriately

service/src/main/java/org/databiosphere/workspacedataservice/sourcewds/WorkspaceDataServiceDao.java
ashanhol: reworked it per our DMs. Added constructor for RestException to WDSException and catch and route appropriately

service/src/main/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDao.java
ashanhol: Readded the leo exception and put everything in a try/catch

service/src/main/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDao.java
ashanhol: Fixed, added try/catches around all the rest calls catching the RestException

service/src/main/java/org/databiosphere/workspacedataservice/leonardo/LeonardoServiceException.java
```diff
@@ -13,4 +14,16 @@ public LeonardoServiceException(ApiException cause) {
         null,
         cause);
   }
+
+  public LeonardoServiceException(RestException cause) {
+    super(cause.getStatus(), cause.getMessage(), cause);
+  }
+
+  // public LeonardoServiceException(AuthenticationException cause) {
```
yuliadub, line 22: are these commented out lines needed?

yuliadub: applies to the other 2 exception classes as well

yuliadub (approved):
just one comment about some commented out code, but otherwise LGTM!

service/src/main/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDao.java
```diff
@@ -35,6 +46,8 @@ public String getWdsEndpointUrl(String token) {
       throw new ApiException("Did not locate an app running WDS.");
     } catch (ApiException e) {
```
davidangb, line 47: If you change `throw new ApiException("Did not locate an app running WDS.");` to throw a RestException instead of an ApiException, then the `catch (ApiException e)` is extraneous and can be removed. This would also mean you could remove the `LeonardoServiceException(ApiException cause)` constructor.

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
```diff
@@ -81,37 +91,40 @@ public ResourceList enumerateDataRepoSnapshotReferences(UUID workspaceId, int of
...
-          LOGGER.debug(
-              "Requesting SAS token-enabled storage url or workspace {} from Workspace Manager ...",
-              workspaceUUID);
-          CreatedAzureStorageContainerSasToken sasBundle =
-              azureResourceApi.createAzureStorageContainerSasToken(
-                  workspaceUUID, storageUUID, null, null, null, null);
-          return sasBundle.getUrl();
-        } else
-          throw new ApiException(
-              "WorkspaceManagerDao: Can't locate a storage resource matching workspace Id. ");
```
davidangb, line 111: same comment as in `LeonardoDao` - if you threw a `RestException` here instead of an `ApiException`, then the `catch (ApiException e)` is no longer necessary. Note: this also requires changes to `enumerateResources`; see comment below.

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
```diff
@@ -136,8 +149,15 @@ ResourceList enumerateResources(
       String authToken)
       throws ApiException {
```
davidangb, line 150: I believe the `throws ApiException` in the method signature is now incorrect and can be removed. This is necessary to remove the `catch (ApiException e)` in `getBlobStorageUrl()`, otherwise the compiler will complain.

davidangb (commented):
looking good, some more comments inline about further cleaning up exceptions, this is close!

service/src/main/java/org/databiosphere/workspacedataservice/leonardo/LeonardoServiceException.java
ashanhol: ty, fixed

service/src/main/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDao.java
ashanhol: Gotcha, wasn't sure if it was appropriate to switch that out. Fixed :)

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
ashanhol: Done

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
```diff
@@ -81,37 +90,39 @@ public ResourceList enumerateDataRepoSnapshotReferences(UUID workspaceId, int of
...
+    try {
+      final ControlledAzureResourceApi azureResourceApi =
+          this.workspaceManagerClientFactory.getAzureResourceApi(authToken);
+      UUID workspaceUUID = UUID.fromString(storageWorkspaceId);
+      LOGGER.debug(
+          "Finding storage resource for workspace {} from Workspace Manager ...", workspaceUUID);
+      ResourceList resourceList =
+          enumerateResources(
+              workspaceUUID, 0, 5, ResourceType.AZURE_STORAGE_CONTAINER, null, authToken);
+      // note: it is possible a workspace may have more than one storage container associated with
```
jladieu, line 102: supernit: comment needs re-wrapping

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
```diff
@@ -81,37 +90,39 @@ public ResourceList enumerateDataRepoSnapshotReferences(UUID workspaceId, int of
...
-          throw new ApiException(
-              "WorkspaceManagerDao: Can't locate a storage resource matching workspace Id. ");
-      } catch (ApiException e) {
-        if (++count == maxTries) throw new WorkspaceManagerException(e);
+            "Requesting SAS token-enabled storage url or workspace {} from Workspace Manager ...",
+            workspaceUUID);
+        RestCall<CreatedAzureStorageContainerSasToken> sasBundleFunction =
+            () ->
+                azureResourceApi.createAzureStorageContainerSasToken(
+                    workspaceUUID, storageUUID, null, null, null, null);
```
jladieu, line 114: I know you're not newly introducing this LOC here, but it's tough to discern the meaning of all the naked `null` parameters here.  A helpful trick I like to use to make these hard-to-read param lists is to add some inline comments ilke so:

```suggestion
                    workspaceUUID, 
                    storageUUID,
                    /* sasIpRange=*/ null, 
                    /* sasExpirationDuration=*/ null,
                    /* sasPermissions= */ null,
                    /* sasBlobName= */ null);
```

This applies for any large param list whose type and/or meaning are hard to infer without an IDE giving you hints.

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
```diff
@@ -133,11 +144,17 @@ ResourceList enumerateResources(
...
+      String authToken) {
     ResourceApi resourceApi = this.workspaceManagerClientFactory.getResourceApi(authToken);
-    // TODO: retries
-    return resourceApi.enumerateResources(
-        workspaceId, offset, limit, resourceType, stewardshipType);
+    try {
+      RestCall<ResourceList> enumerateResourcesFunction =
+          () ->
+              resourceApi.enumerateResources(
+                  workspaceId, offset, limit, resourceType, stewardshipType);
```
jladieu, line 153: Just crosslinking this as a counterpoint to my earlier comment about leaving helper comments on a long arglist, all these parameters are easy to understand based on their name, so no comments needed.

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java
jladieu: Also just [linking](https://github.com/DataBiosphere/terra-workspace-data-service/pull/396/files#r1383367587) a counter example where the parameter names are self-documenting and where this kind of comment isn't necessary.

service/src/test/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDaoTest.java
```diff
@@ -45,7 +49,9 @@ void setUp() {
   @Test
   void testWdsUrlNotReturned() throws ApiException {
     final int statusCode = HttpStatus.INTERNAL_SERVER_ERROR.value();
-    given(mockAppsApi.listAppsV2(String.valueOf(UUID.randomUUID()), "things", null, null))
+    given(
+            mockAppsApi.listAppsV2(
+                anyString(), nullable(String.class), anyBoolean(), nullable(String.class)))
```
jladieu, line 54: TIL there's a `nullable` matcher!

jladieu (approved):
Some nit / readability comments, but LGTM!

davidangb (approved):
looks good, thank you!

PR #401, opened 2023-11-03
AJ-1234: Relax regex matching conflict message.

jladieu:
The actual message differed slightly from expectation, so I've adjusted the regex to allow both forms of pluralization; what matters most is that the conflict was due to something relating to policies (and not one of the other myriad ways a 409 can be returned).


davidangb (approved):
I did not test myself, code changes look :+1:

nawatts (approved):

PR #405, opened 2023-11-09
AJ-1362 - Add retries in WDS end to end test 

yuliadub:
After reviewing all the failed runs from the past month, the following workflows cause the most failures:
- bee creation
- workspace creation
- polling for Leo Urls (occasional 503)

For retrying workspace creation and Leo url polling, I added a simple while loop to do the retries (all other options seemed more complicated than they needed to be). Everyone using the create workspace and polling for app status functions will get these retries for free. See those changes here: https://github.com/broadinstitute/dsp-reusable-workflows/pull/17

To retry bee creation, since that is a workflow steps that runs in another repo as a github action, it seemed like the easiest way was to just run the same step again if the first attempt failed. To do this, a recoded create-bee steps as its own action that I could call from within the larger e2e github action, without having to duplicate the steps. The create-bee action will only run a second time if the first time fails. 

Reminder:


.github/workflows/run-e2e-tests.yaml
```diff
@@ -76,7 +79,7 @@ jobs:
     permissions:
       contents: read
       id-token: write
-    uses: broadinstitute/dsp-reusable-workflows/.github/workflows/run-e2e-tests.yaml@main
+    uses: broadinstitute/dsp-reusable-workflows/.github/workflows/run-e2e-tests.yaml@aj-1362-retries
```
yuliadub, line 82: just here for testing

.github/actions/create-bee/action.yml
```diff
@@ -0,0 +1,29 @@
...
+      with:
+        workflow: bee-create
+        repo: broadinstitute/terra-github-workflows
+        ref: refs/heads/main
+        token: '${{ inputs.token }}'
+        inputs: >-
+          { 
+          "bee-name": "${{ inputs.bee_name }}", 
+          "bee-template-name":
+          "rawls-e2e-azure-tests",
```
jladieu, line 27: nit: i think this should be on the same line as "bee-template-name"

also just want to confirm..is it supposed to be rawls?

jladieu (approved):

.github/actions/create-bee/action.yml
yuliadub: ha good question, Bria set these up and I didnt change this but I assume yes :)

davidangb (commented):
I wonder if there's something prebuilt you can use for this, like https://github.com/marketplace/actions/retry-action

yuliadub:
> I wonder if there's something prebuilt you can use for this, like https://github.com/marketplace/actions/retry-action

I looked into that - from what I could tell there is no way to run another workflow with retries using that, only direct commands. I tried a few things but couldnt get it to work. We can try another round in office hours in case I missed I something :)

davidangb (approved):

jladieu (approved):
Great writeup!

PR #406, opened 2023-11-09
Enable can-i-deploy in WDS CI

jladieu:
[AJ-1454](https://broadworkbench.atlassian.net/browse/AJ-1454)

This is followup to initial pact verification work to actually enable deployment protection.

This is step 6 of: https://broadworkbench.atlassian.net/wiki/spaces/IRT/pages/2660368406/Getting+Started+with+Pact+Contract+Testing

[AJ-1454]: https://broadworkbench.atlassian.net/browse/AJ-1454?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Based heavily on: https://github.com/broadinstitute/rawls/blob/develop/.github/workflows/consumer_contract_tests.yml

### DIFF
--- a/.github/workflows/publish-pacts.yml
+++ b/.github/workflows/publish-pacts.yml
@@ -12,6 +12,7 @@ on:
 env:
   WDS_RUN_ID: ${{ github.event.repository.name }}-${{ github.run_id }}
   PUBLISH_CONTRACT_RUN_NAME: 'publish-contracts-${{ github.event.repository.name }}-${{ github.run_id }}-${{ github.run_attempt }}'
+  CAN_I_DEPLOY_RUN_NAME: 'can-i-deploy-${{ github.event.repository.name }}-${{ github.run_id }}-${{ github.run_attempt }}'
   WDS_PACTS_ARTIFACT: wds-pacts-${{ github.event.repository.name }}-${{ github.run_id }}
   WDS_PACTS_OUTPUT_DIR: service/build/pacts/
 
@@ -161,10 +162,10 @@ jobs:
             "repo-name": "${{ github.event.repository.name }}",
             "repo-branch": "${{ needs.init-github-context.outputs.repo-branch }}",
             "release-tag": "${{ needs.regulated-tag-job.outputs.new-tag }}"
-          }'
+            }'
+
   can-i-deploy:
     runs-on: ubuntu-latest
-    if: ${{ needs.init-github-context.outputs.fork == 'false' }}
     needs:
       - init-github-context
       - publish-pacts-job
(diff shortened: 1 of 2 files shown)


.github/workflows/consumer-contract-tests.yml
```diff
@@ -0,0 +1,198 @@
...
+jobs:
+  init-github-context:
+    runs-on: ubuntu-latest
+    outputs:
+      repo-branch: ${{ steps.extract-branch.outputs.repo-branch }}
+      repo-version: ${{ steps.extract-branch.outputs.repo-version }}
+      fork: ${{ steps.extract-branch.outputs.fork }}
+
+    steps:
+      - uses: actions/checkout@v3
```
yuliadub, line 61: here you use checkout@v3 but on line 105 you use checkout@v4, is that intentional?

yuliadub (approved):
aside from the one question, and my limited knowledge of pact tests, LGTM :D

.github/workflows/consumer-contract-tests.yml
jladieu: Oops, not intentional, thanks for spotting!  Fixed in 36c63668f3be1497584b76ac31c924257dc5850d

davidangb (approved):
giving a thumb. That said: what provider versions does this trigger tests against? From the list of all the providers that WDS cares about, which ones are properly setting their version upon deploy?

jladieu:
@davidangb I think we're still working out how to correctly set version upon deploy throughout the various pacticipants (see recent threads in #dsp-contract-testing, one in particular: https://broadinstitute.slack.com/archives/C043YJ40719/p1699888771297559)

AFAIK once everything is settled and integrated correctly, the pactbroker will conduct tests across an entire matrix of deployed versions of providers/consumers to prevent breaking deploys.

davidangb:
> I think we're still working out how to correctly set version upon deploy throughout the various pacticipants

@jladieu  as long as we don't end up in a place where we have false positives or false negatives because we're testing incorrect versions. I'd rather not run can-i-deploy at all than end up with misleading results. Thank you!

jladieu:
I'll hold off on merging this until I have evidence that all the collaborators are where they need to be.

Currently several pacticipants are reporting the github SHA as the version rather than the app semantic version, and @ichengchang is working on reference implementations for switching over to the correct app semantic version.

Once that work is done and all the correct webhooks have been configured appropriately, it's my understanding that `can-i-deploy` is supposed to check the contracts against all currently live deployed versions of its collaborator pacticipants.

https://docs.pact.io/pact_nirvana/step_6

jladieu:
Moving this back to draft state as it has aged quite a bit and I feel it'll be out of date by the time I get back to it.

jladieu:
I'm going to close this stale PR until we have need of a concerted push to enable this.  I'm hoping if I wait on this long enough there will be a better/templated solve.

PR #409, opened 2023-11-14
AJ-1451: Fix pact provider state

jladieu:
This includes a couple fixes that I discovered while working on the provider side of this contract.

These bugs are a reminder why it's helpful to complete the provider tests in lockstep with the consumer tests!

### DIFF
--- a/service/src/test/java/org/databiosphere/workspacedataservice/pact/WsmPactTest.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/pact/WsmPactTest.java
@@ -54,7 +54,7 @@ public class WsmPactTest {
   @Pact(consumer = "wds")
   RequestResponsePact linkSnapshotForPolicySuccess(PactDslWithProvider builder) {
     return builder
-        .given("a workspace with the given {id} exists", Map.of("id", WORKSPACE_UUID.toString()))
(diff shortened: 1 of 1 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/pact/WsmPactTest.java
```diff
@@ -337,14 +341,17 @@ RequestResponsePact createAzureStorageContainerSasTokenSuccess(PactDslWithProvid
     return builder
         .given("a workspace with the given id exists", Map.of("id", WORKSPACE_UUID.toString()))
         .given(
-            "a storage container with the given id exists for the given workspace_id",
-            Map.of("id", RESOURCE_UUID.toString(), "workspace_id", WORKSPACE_UUID.toString()))
+            "a storage container resource exists for the given workspace_id",
```
jladieu, line 344: By using the `FromProviderState` helpers, this lets the provider generate a resource ID for us, and draw our expectations around that, rather than having the client specify the IDs up front.

Since the resourceId is something WDS learns about by querying the `enumerateResources` endpoint, it doesn't make sense for WDS to specify it up front.

This is contrast to the workspace ID, which we have to know about to initiate the interaction.

davidangb (approved):

nawatts (approved):

PR #415, opened 2023-11-17
AJ-1452: pfb data types

davidangb:
I have two PRs for this ticket, which take different approaches to the problem. We will only merge one of them:

## #415 (this PR)
* Before processing any records, read the Avro schema from the PFB.
* Translate the Avro schema into a WDS schema, e.g. map Avro's `Schema.Type.DOUBLE` to WDS's `DataTypeMapping.NUMBER`.
* Pass this WDS schema down through the call stack.
* When converting an Avro record to a WDS record, look up the expected datatype from the schema. Based on the expected schema AND the actual Java object type returned by the Avro API, translate Avro fields into WDS attributes.

## #416
* No inspection of the Avro schema.
* When converting an Avro record to a WDS record, base the translation purely on the actual Java object type returned by the Avro API.

The former PR - #415 (this one) - has more safeguards, at the expense of being notably more complex. The latter PR - #416 - relies on the Avro API being correct and uses that trust to simplify the code.

_my recommendation is to merge #416 but I am looking for feedback!_

Note that both PRs leave outstanding work; I will file follow-on tickets for these:
* decoding enum values; see https://github.com/uc-cdis/pypfb/blob/master/docs/detailed_pfb_doc.md#enum
* #415 only handles scalar strings, enums, booleans, and numbers; other datatypes are TODOs
* #416 handles both scalar and array strings, enums, booleans and numbers, but does not handle Avro datatypes map, record, bytes, or fixed.

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -80,7 +80,7 @@ dependencies {
     implementation 'com.squareup.okhttp3:okhttp:4.11.0' // required by Sam client
     implementation "bio.terra:datarepo-client:1.537.0-SNAPSHOT"
     implementation "bio.terra:workspace-manager-client:0.254.717-SNAPSHOT"
-    implementation "bio.terra:java-pfb-library:0.13.0"
+    implementation "bio.terra:java-pfb-library:0.14.0"
     implementation project(path: ':client')
 
     // hk2 is required to use WSM client, but not correctly exposed by the client
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
@@ -6,8 +6,10 @@ import static org.databiosphere.workspacedataservice.shared.model.Schedulable.AR
 
 import bio.terra.datarepo.model.SnapshotModel;
 import bio.terra.pfb.PfbReader;
+import java.io.IOException;
 import java.net.URL;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
@@ -17,6 +19,7 @@ import java.util.UUID;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
+import org.apache.avro.Schema;
 import org.apache.avro.file.DataFileStream;
 import org.apache.avro.generic.GenericRecord;
 import org.databiosphere.workspacedataservice.activitylog.ActivityLogger;
@@ -25,6 +28,7 @@ import org.databiosphere.workspacedataservice.jobexec.QuartzJob;
 import org.databiosphere.workspacedataservice.retry.RestClientRetry;
 import org.databiosphere.workspacedataservice.service.BatchWriteService;
 import org.databiosphere.workspacedataservice.service.model.BatchWriteResult;
+import org.databiosphere.workspacedataservice.service.model.DataTypeMapping;
 import org.databiosphere.workspacedataservice.service.model.exception.PfbParsingException;
 import org.databiosphere.workspacedataservice.service.model.exception.RestException;
 import org.databiosphere.workspacedataservice.shared.model.RecordType;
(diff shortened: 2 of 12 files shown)


service/build.gradle
```diff
@@ -80,7 +80,7 @@ dependencies {
     implementation 'com.squareup.okhttp3:okhttp:4.11.0' // required by Sam client
     implementation "bio.terra:datarepo-client:1.537.0-SNAPSHOT"
     implementation "bio.terra:workspace-manager-client:0.254.717-SNAPSHOT"
-    implementation "bio.terra:java-pfb-library:0.13.0"
+    implementation "bio.terra:java-pfb-library:0.14.0"
```
davidangb, line 83: necessary to get access to the `PfbReader.getPfbSchema` method.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -78,25 +82,50 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
...
 
+    // Retrieve the schema(s) as reported by PfbReader for use later.
+    // N.B. It does not seem worthwhile to create/modify Postgres tables here, based on
+    // this retrieved schema. This schema does not handle file, date, datetime etc. columns yet -
+    // those are only determined after sampling values within the records. So, if we updated tables
+    // now we'd still have to do it again later.
+    //
+    // This is HTTP connection #1 to the PFB.
+    Map<String, Map<String, DataTypeMapping>> recordTypeSchemas = calculateSchema(url);
+
```
davidangb, line 93: `Map<String, Map<String, DataTypeMapping>>` is a really ugly datatype, and I would love to see a better encapsulation for this. However, `Map<String, DataTypeMapping>` is already used extensively in the codebase; refactoring it would touch a lot of code. I'd like to save any refactor for a separate PR to keep this one from blowing up.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
```diff
@@ -14,24 +17,36 @@ public class PfbRecordConverter {
...
     Record converted =
         new Record(
             genRec.get(ID_FIELD).toString(),
             RecordType.valueOf(genRec.get(TYPE_FIELD).toString()),
             RecordAttributes.empty());
 
-    // contains attributes
+    // retrieve the expected schema for this record type
+    Map<String, DataTypeMapping> wdsTypeSchema =
+        recordTypeSchemas.getOrDefault(genRec.get(TYPE_FIELD).toString(), Map.of());
```
davidangb, line 36: We retrieve the PFB schema and translate it to a WDS schema once up front, in `PfbQuartzJob`, then pass that translated schema all the way down to here … even though we have access to the PFB schema here. The alternative is to translate the PFB schema to a WDS schema for _every_ record we import; that feels wasteful.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
```diff
@@ -40,16 +59,50 @@ public Record genericRecordToRecord(GenericRecord genRec) {
...
+    if (returnValue == null) {
+      // If we reach here, it means that the actual object returned by Avro did not
+      // match the expected WDS data type, and the "instanceof" clauses in the cases above
+      // don't satisfy. This should only happen if our logic is faulty.
+      logger.warn(
+          "mismatched attribute datatype: expected {} but found {}",
+          targetDataType,
+          attribute.getClass().getSimpleName());
+      returnValue = attribute.toString();
+    }
```
yuliadub, line 104: this is the one possible thing we loose (ability to tell that this happened) with the other approach - I wonder if it makes sense to add this somewhere in the UI (like when you say okay, import my PFB) where we basically tell the user something like (in a much better PM way :D) "if the schema has issues, WDS wont correct them as it trusts that the avro schema is correct"

davidangb:
Closing in favor of #416 416

PR #421, opened 2023-11-28
AJ-1463: update client dependencies too

davidangb:
Update dependencies in the `:client` project to match versions in `:service` and/or to use latest versions.

This PR has a bunch of auto-formatter whitespace changes; you may want to ignore whitespace in the diff.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,6 +10,7 @@ plugins {
     id 'java'
     id 'com.gorylenko.gradle-git-properties' version '2.4.1'
     id 'jacoco'
+    id 'com.diffplug.spotless' version '6.23.0' apply false
 }
 
 repositories {
--- a/client/build.gradle
+++ b/client/build.gradle
@@ -5,7 +5,7 @@ plugins {
 
     id 'com.jfrog.artifactory' version '5.1.10'
     id 'org.openapi.generator'
-    id 'com.diffplug.spotless' version '6.21.0'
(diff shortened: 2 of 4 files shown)


build.gradle
```diff
@@ -10,6 +10,7 @@ plugins {
     id 'java'
     id 'com.gorylenko.gradle-git-properties' version '2.4.1'
     id 'jacoco'
+    id 'com.diffplug.spotless' version '6.23.0' apply false
```
davidangb, line 13: this specifies the version of the spotless plugin once, in the parent project; the `:client` and `:service` subprojects can just inherit this version instead of specifying it themselves.

client/build.gradle
```diff
@@ -42,7 +42,7 @@ spotless {
         targetExclude "${buildDir}/**"
         targetExclude "**/swagger-code/**"
         targetExclude "**/generated/**"
-        googleJavaFormat()
+        googleJavaFormat('1.18.1')
```
davidangb, line 45: this (and the corresponding change in the `:service` project) updates `google-java-format` from 1.17.0 to latest 1.18.1, which also carries an upgrade to guava. This is specific to the spotless gradle configuration - i.e. not to runtime or test configurations - but Dependabot finds those configurations and reports on the outdated guava.

yuliadub (approved):

jladieu (approved):

PR #422, opened 2023-11-28
AJ-1489: handle arrays of nulls in PFB import

davidangb:
Allow WDS to handle arrays of nulls. There were a couple places we didn't handle those - see code comments inline for explanations.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -712,7 +712,7 @@ public class RecordDao {
     return inferer.getArrayOfType(attVal.toString(), typeMapping.getJavaArrayTypeForDbWrites());
   }
 
-  private Object[] getListAsArray(List<?> attVal, DataTypeMapping typeMapping) {
+  Object[] getListAsArray(List<?> attVal, DataTypeMapping typeMapping) {
     return switch (typeMapping) {
       case ARRAY_OF_STRING,
           ARRAY_OF_FILE,
@@ -720,7 +720,10 @@ public class RecordDao {
           ARRAY_OF_DATE,
           ARRAY_OF_DATE_TIME,
           ARRAY_OF_NUMBER,
-          EMPTY_ARRAY -> attVal.stream().map(Object::toString).toList().toArray(new String[0]);
+          EMPTY_ARRAY -> attVal.stream()
+          .map(e -> Objects.toString(e, null))
+          .toList()
+          .toArray(new String[0]);
       case ARRAY_OF_BOOLEAN ->
       // accept all casings of True and False if they're strings
       attVal.stream()
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -9,6 +9,7 @@ import static org.databiosphere.workspacedataservice.service.model.exception.Inv
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Maps;
 import java.math.BigDecimal;
 import java.sql.Date;
@@ -712,6 +713,7 @@ public class RecordDao {
     return inferer.getArrayOfType(attVal.toString(), typeMapping.getJavaArrayTypeForDbWrites());
   }
 
+  @VisibleForTesting
   Object[] getListAsArray(List<?> attVal, DataTypeMapping typeMapping) {
     return switch (typeMapping) {
       case ARRAY_OF_STRING,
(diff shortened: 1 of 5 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -712,15 +712,18 @@ private Object getArrayValues(Object attVal, DataTypeMapping typeMapping) {
     return inferer.getArrayOfType(attVal.toString(), typeMapping.getJavaArrayTypeForDbWrites());
   }
 
-  private Object[] getListAsArray(List<?> attVal, DataTypeMapping typeMapping) {
+  Object[] getListAsArray(List<?> attVal, DataTypeMapping typeMapping) {
```
davidangb, line 715: visibility changed to allow for unit testing.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -712,15 +712,18 @@ private Object getArrayValues(Object attVal, DataTypeMapping typeMapping) {
...
           ARRAY_OF_FILE,
           ARRAY_OF_RELATION,
           ARRAY_OF_DATE,
           ARRAY_OF_DATE_TIME,
           ARRAY_OF_NUMBER,
-          EMPTY_ARRAY -> attVal.stream().map(Object::toString).toList().toArray(new String[0]);
+          EMPTY_ARRAY -> attVal.stream()
+          .map(e -> Objects.toString(e, null))
+          .toList()
+          .toArray(new String[0]);
```
davidangb, line 726: if an array contained nulls, we'd get a NullPointerException on the previous code `Object::toString`. I changed this to `Objects.toString(e, null)` to be null-safe.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -119,7 +119,7 @@ <T> T withPfbStream(URL url, PfbStreamConsumer<T> consumer) {
         PfbReader.getGenericRecordsStream(url.toString())) {
       return consumer.run(dataStream);
     } catch (Exception e) {
-      throw new PfbParsingException("Error processing PFB", e);
+      throw new PfbParsingException("Error processing PFB: " + e.getMessage(), e);
```
davidangb, line 122: not strictly necessary - if a PFB import job fails and the exception is caught here, we only delivered a generic message "Error processing PFB" with no indication of the cause. Now, we indicate the cause. In both cases you still have to dive into logs to get the stack trace.

service/src/main/java/org/databiosphere/workspacedataservice/service/model/DataTypeMapping.java
```diff
@@ -70,7 +70,7 @@ public static DataTypeMapping getArrayTypeForBase(DataTypeMapping baseType) {
       return EMPTY_ARRAY;
     }
     return switch (baseType) {
-      case STRING -> ARRAY_OF_STRING;
+      case STRING, NULL -> ARRAY_OF_STRING;
```
davidangb, line 73: this was the top-level problem: if WDS tries to determine the array type but encounters a base type of `NULL`, it previously failed (see the `default` case). Now, it treats it as an array of strings, whose array elements happen to be null.

service/src/test/java/org/databiosphere/workspacedataservice/dao/RecordDaoTest.java
```diff
@@ -957,4 +962,55 @@ void testRemoveFromJoin() {
...
+      })
+  void testOnlyNullsInList(DataTypeMapping typeMapping) {
+    List<?> input = new ArrayList<>();
+    input.add(null);
+    input.add(null);
+    input.add(null);
+    String[] expected = new String[] {null, null, null};
+    Object[] actual = recordDao.getListAsArray(input, typeMapping);
+
+    assertArrayEquals(expected, actual);
```
jladieu, line 986: Are you using `getListAsArray` to have access to `assertArrayEquals`?  Is there not a readily available `List`y equivalent that would obviate the need for the second step?

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
jladieu: How do you feel about marking such methods with `@VisibleForTesting` annotation?

I think the only immediate benefit from doing so is easy greppability, but I think with some tooling support you can also use it as a way to prevent unintended invocation from production codepaths with linters (eg: https://googlesamples.github.io/android-custom-lint-rules/checks/VisibleForTests.md.html#:~:text=With%20the%20%40VisibleForTesting%20annotation%20you,widely%20visible%20for%20the%20tests.)

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
jladieu: IMO this is worth being an inline comment, maybe just with an explanatory arg comment for the `null`:

```/* nullDefault= */```

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
davidangb: lol I just added that annotation and came here to find that you had already reviewed this PR!

service/src/test/java/org/databiosphere/workspacedataservice/dao/RecordDaoTest.java
davidangb: in this case, `getListAsArray` is the method I want to test; I changed its behavior in the presence of nulls.

jladieu (approved):

calypsomatic (approved):

PR #426, opened 2023-11-30
AJ-1393 parse relations in pfb

calypsomatic:
https://broadworkbench.atlassian.net/browse/AJ-1393
Adds a second pass to PFB parsing to add relations between rows.  The double-pass approach is necessary to deal with potential forward-references (a row relates to a row that only appears later in the PFB).  Adds a `PfbImportMode` enum to `PfbStreamWriteHandler` to indicate whether to parse for record attributes or relations only, and propagates/refers to that flag throughout the process.

Reminder:


davidangb (commented):
I think this is conceptually correct - make an additional pass over the PFB, during that pass identify any relations and update the underlying record with anything you found.

I'd love to eliminate some of the copy/paste in BatchWriteService and PfbStreamWriteHandler - I don't have great ideas on that right now. Also make sure that if a given record (or stretch of records in a row) have no relations, we don't issue any update statements for those.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -144,6 +149,27 @@ BatchWriteResult importTables(DataFileStream<GenericRecord> dataStream, UUID tar
...
+              entry -> {
+                RecordType recordType = entry.getKey();
+                int quantity = entry.getValue();
+                activityLogger.saveEventForCurrentUser(
+                    user ->
+                        user.upserted().record().withRecordType(recordType).ofQuantity(quantity));
+              });
+    }
+    return result;
+  }
```
davidangb, line 171: Can you consolidate `importTables` and `addRelationsToTables` into one method which accepts an argument for relations=true|false? Or perhaps extract a private method that has the implementation, and leave `importTables` and `addRelationsToTables` as shell/wrapper functions?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
```diff
@@ -46,6 +48,40 @@ public Record genericRecordToRecord(GenericRecord genRec) {
...
+        */
+        attributes.putAttribute(
+            relationType + "_id",
+            RelationUtils.createRelationString(RecordType.valueOf(relationType), relationId));
+      }
+      converted.setAttributes(attributes);
+    }
+
+    return converted;
+  }
```
davidangb, line 83: I also wonder if `genericRecordToRelations` and `genericRecordToRecord` can be consolidated, something like:
```java
  public Record convert(GenericRecord genRec, boolean relationsOnly) {
    // create the WDS record shell (id, record type, empty attributes)
    Record converted =
        new Record(
            RecordAttributes.empty());
    
    if (!relationsOnly) {
      addAttributes(genRec, converted);
    } else {
      addRelations(genRec, converted);
    }
    
    return converted;
  }
```

service/src/main/java/org/databiosphere/workspacedataservice/service/PfbStreamWriteHandler.java
```diff
@@ -26,6 +26,15 @@ public PfbStreamWriteHandler(DataFileStream<GenericRecord> inputStream) {
   }
 
   public WriteStreamInfo readRecords(int numRecords) throws IOException {
+    return convertStream(numRecords, false);
+  }
+
+  public WriteStreamInfo readRelations(int numRecords) throws IOException {
+    return convertStream(numRecords, true);
+  }
```
davidangb, line 34: suggestion: add a `boolean relationsOnly` argument to the constructor of `PfbStreamWriteHandler`. You can pass the true/false from `PfbQuartzJob.importTables()` and `PfbQuartzJob.addRelationsToTables()` through `batchWriteService.batchWritePfbStream()` to this constructor.

Then, since the  `PfbStreamWriteHandler` object will know if it's working with relations or not, you can do away with the separate `readRecords` and `readRelations` methods. Internally to `PfbStreamWriteHandler`, it can return different data from `readRecords` depending on how it was constructed.

I think this might help to clean up/consolidate code elsewhere, too!

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
calypsomatic: Done

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
```diff
@@ -42,7 +51,32 @@ public Record genericRecordToRecord(GenericRecord genRec) {
       }
       converted.setAttributes(attributes);
     }
+    return converted;
+  }
 
+  private Record addRelations(GenericRecord genRec, Record converted) {
+    // get the relations array from the record
+    if (genRec.get(RELATIONS_FIELD) instanceof Collection relationArray
```
davidangb, line 59: nit: use `Collection<?>` to avoid IntelliJ squigglies

davidangb (commented):
love the most recent changes!  A few more comments inline …

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -127,20 +131,27 @@ <T> T withPfbStream(URL url, PfbStreamConsumer<T> consumer) {
...
-            entry -> {
-              RecordType recordType = entry.getKey();
-              int quantity = entry.getValue();
-              activityLogger.saveEventForCurrentUser(
-                  user -> user.upserted().record().withRecordType(recordType).ofQuantity(quantity));
-            });
+        batchWriteService.batchWritePfbStream(
+            dataStream, targetInstance, Optional.of(ID_FIELD), relationsOnly);
+
+    if (result != null) {
```
yuliadub, line 143: when would a result be null? would this happen when there is no a relation for a given row of data or something like that?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
```diff
@@ -18,15 +19,23 @@ public class PfbRecordConverter {
...
         new Record(
             genRec.get(ID_FIELD).toString(),
             RecordType.valueOf(genRec.get(TYPE_FIELD).toString()),
             RecordAttributes.empty());
+    if (relationsOnly) {
+      return addRelations(genRec, converted);
+    } else {
+      return addAttributes(genRec, converted);
+    }
+  }
```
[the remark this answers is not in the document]
yuliadub: I second this suggestion!

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
calypsomatic: Right, if none of the records in the batch have relations, this will output null and scream.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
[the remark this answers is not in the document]
calypsomatic: Ok I did it, feel free to check if the enum is in the right place/done correctly/etc

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
```diff
@@ -94,7 +95,18 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
...
+        stream ->
+            importTables(
+                stream, targetInstance, PfbStreamWriteHandler.PfbImportMode.BASE_ATTRIBUTES));
+
+    // This is HTTP connection #3 to the PFB.
+    logger.info("Updating tables and rows from this PFB with relations...");
+    withPfbStream(
+        url,
+        stream ->
+            importTables(stream, targetInstance, PfbStreamWriteHandler.PfbImportMode.RELATIONS));
```
davidangb, line 109: optional; your call: if you
```
import static org.databiosphere.workspacedataservice.service.PfbStreamWriteHandler.PfbImportMode.BASE_ATTRIBUTES;
import static org.databiosphere.workspacedataservice.service.PfbStreamWriteHandler.PfbImportMode.RELATIONS;
```
then you can skip the `PfbStreamWriteHandler.PfbImportMode.` verbosity here.

IntelliJ: right-click on "RELATIONS" on this line -> Show Context Actions -> Add static import for …

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
```diff
@@ -42,7 +53,28 @@ public Record genericRecordToRecord(GenericRecord genRec) {
...
+    // get the relations array from the record
+    if (genRec.get(RELATIONS_FIELD) instanceof Collection<?> relationArray
+        && !relationArray.isEmpty()) {
+      RecordAttributes attributes = RecordAttributes.empty();
+      for (Object relationObject : relationArray) {
+        // Here we assume that the relations object is a GenericRecord with keys "dst_name" and
+        // "dst_id"
+        if (relationObject instanceof GenericRecord relation) {
+          String relationType = relation.get("dst_name").toString();
+          String relationId = relation.get("dst_id").toString();
```
davidangb, line 69: consider making "dst_name" and "dst_id" as constants, since they're copy/pasted in a few of the tests

davidangb (approved):
two minor comments inline as optional - nice job getting this through!

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbQuartzJob.java
calypsomatic: Oh thank you that was really bugging me!

yuliadub (approved):
LGTM!

PR #430, opened 2023-12-01
AJ-1337: example of custom Micrometer metrics

davidangb:
This PR adds a custom metric, via Micrometer, for tracking when we issue an "alter table … alter column" statement to change the datatype of a pre-existing column.

For instance, if a user uploads a TSV containing only numbers in a column "myColumn", WDS will create that column as `numeric` in Postgres. If the user then uploads another TSV that contains non-numeric string values in "myColumn", WDS will alter the Postgres table and set the column to be `text`. This metric tracks every time we do that.

This new "column_change_datatype" metric is visible in Application Insights:
Screenshot 12-1-2023 at 11 58 AM
Screenshot 12-1-2023 at 11 58 AM (1)


service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
```diff
@@ -208,6 +218,18 @@ public Map<String, DataTypeMapping> addOrUpdateColumnIfNeeded(
...
+      Counter counter =
+          Counter.builder(COUNTER_COLCHANGE)
+              .tag(TAG_RECORDTYPE, recordType.getName())
+              .tag(TAG_ATTRIBUTENAME, column)
+              .tag(TAG_INSTANCE, instanceId.toString())
+              .tag("OldDataType", valueDifference.leftValue().toString())
+              .tag("NewDataType", updatedColType.toString())
+              .description("Column schema changes")
+              .register(meterRegistry);
+      counter.increment();
```
davidangb, line 232: I tried to guess at which tags will be reused across WDS - like `TAG_INSTANCE` - and which are specific to this metric, like "OldDataType" and "NewDataType".

service/src/main/java/org/databiosphere/workspacedataservice/metrics/MetricsDefinitions.java
```diff
@@ -0,0 +1,16 @@
+package org.databiosphere.workspacedataservice.metrics;
+
+public class MetricsDefinitions {
+
+  /** counter for column schema changes, i.e. "alter column" sql statements */
+  public static final String COUNTER_COLCHANGE = "column_change_datatype";
```
jladieu, line 8: micronit: can you share how you're thinking about the _ delimeter in the `SCREAMING_SNAKE_CASE`?  I instinctively want to make `COLCHANGE` -> `COL_CHANGE` and `RECORDTYPE` -> `RECORD_TYPE` and `ATTRIBUTENAME` -> `ATTRIBUTE_NAME` but am flexible.

davidangb: I … did not really put any thought into it. I'm fine renaming things if that's helpful.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
jladieu: Interesting to think about the tags we'd want to slice on!

I'm used to thinking about metrics as `path.of.some.structure.to.metric`, do all these tags combine to form an ordered path/namespace for the metric that starts with the first tag and proceeds to the last, or does micrometer have a different model?

I ask because I've encountered issues with high cardinality metrics (where N scaled by some factor proportional to users), and I wonder what kind of cardinality this might expand out to with real world usage, customer-entered values, etc.

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordServiceSpec.java
```diff
@@ -0,0 +1,102 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.ActiveProfiles;
+
+@SpringBootTest
+@ActiveProfiles(profiles = {"mock-sam"})
+class RecordServiceTest {
```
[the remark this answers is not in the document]
davidangb: lol you're right I originally named it RecordServiceSpec and then fixed that - but obv failed to rename the FILE as well. Good catch!

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
davidangb: I don't think Micrometer cares too much. It does consider each unique combination of tags to be a different counter object on the registry, but creation of dynamic counters like this appears to be common, for example:
* https://stackoverflow.com/a/74098618
* https://medium.com/@aviadpines/dynamic-meters-and-metrics-with-micrometer-e7e16ce52371

I'd be more concerned about how Application Insights - or other repository - stores these. App Insights seems to just consider these tags as extra dimensions on the entry; see the first screenshot from the description.

We could totally go with some kind of dot-hierarchy for names if you think it makes it more portable!

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
```diff
@@ -208,6 +217,18 @@ public Map<String, DataTypeMapping> addOrUpdateColumnIfNeeded(
           inferer.selectBestType(valueDifference.leftValue(), valueDifference.rightValue());
       recordDao.changeColumn(instanceId, recordType, column, updatedColType);
       schema.put(column, updatedColType);
+
+      // update a metrics counter with this schema change
+      Counter counter =
+          Counter.builder(COUNTER_COLCHANGE)
```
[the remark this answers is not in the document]
davidangb: See my other response at https://github.com/DataBiosphere/terra-workspace-data-service/pull/430#discussion_r1412463070 … it's a different `Counter` object as far as Micrometer is concerned, but once it gets saved to Application Insights we can query to find all entries that have the name `column_change_datatype` and aggregate on them.

jladieu (approved):

service/src/main/java/org/databiosphere/workspacedataservice/metrics/MetricsDefinitions.java
davidangb: renamed!

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
jladieu: Meta note about this comment:  I am starting from the position that the custom metric introduced by this PR is really just here as a POC and not a critical must-have metric.  LMK if you disagree about this!

Any discussion about the merits of whether/how to keep it here are for the sake of the thought exercise of "how to deal with high cardinality metrics".

After more exploration around Micrometer & Prometheus, I want to circle back around to this metric and assert that my hunch about exploding cardinalities is still relevant.  Prometheus explicitly [warns](https://www.javadoc.io/doc/io.prometheus/simpleclient/0.0.5/io/prometheus/client/SimpleCollector.html) against using high cardinality values for dynamic metrics

> Warning #2: While labels are very powerful, avoid overly granular metric labels. The combinatorial explosion of breaking out a metric in many dimensions can produce huge numberts of timeseries, which will then take longer and more resources to process.
As a rule of thumb aim to keep the cardinality of metrics below ten, and limit where the cardinality exceeds that value. For example rather than breaking out latency by customer and endpoint in one metric, you might have two metrics with one breaking out by each. If the cardinality is in the hundreds, you may wish to consider removing the breakout by one of the dimensions altogether.

That said, I don't think _MicroMeter_ has opinions on this, as different metrics systems can have different levels of support for high cardinality tags.  For example, App Insights just happily includes this as extra metadata on a given sample, but doesn't expose it as an entirely new dimension that causes.

I think if we think there's sufficient value, we _can_ still support high cardinality Meters, but we need to actively manage them and prevent them from being scraped by Prometheus, otherwise the combinatorial explosion will make all the computers upstream sad.  Some options to proceed:

1) Avoid high cardinality metrics entirely and catch & prevent them in code review.
   1.a) That could mean undoing/deleting this one.  If this is the route we take, I think we can probably still use logging to record this info for ad hoc analysis.
   1.b) We could also just _reduce_ the cardinality of this one by removing tags that have O(>1) cardinality.  For that, I'd say we would only keep the old/new data types tags, as the data types are a fixed O(1) list of values.

2) Support high cardinality metrics by explicitly mark/tag high cardinality metrics with a tag about cardinality, and use some kind of mechanism to exclude these from prometheus scraping (while letting them out through whatever other surfaces we think can support them, like App Insights).

3) Explicitly mark/tag metrics with a tag asserting their _low_ cardinality, and only allow these to be scraped by prometheus.  Combine with code review to ensure this tag is used thoughtfully.

Regardless: We also can (& should) on the global config `maximumAllowable{Tags,Meters}` that cap metrics if we blow our budget anyway.

Of these options (or any I haven't suggested), how do you think we should proceed?

I'm inclined to go with (1.a) for the time being until we have an urgent need for a detailed high-cardinality metric.

jladieu: Looks like SpringBoot 3 Observability provides a formal construct for recognizing high vs. low cardinality observation info: https://docs.spring.io/spring-boot/docs/3.0.0/reference/html/actuator.html#actuator.observability

PR #434, opened 2023-12-07
AJ-1478: Add Avro `map` type.

jladieu:
[AJ-1478](https://broadworkbench.atlassian.net/browse/AJ-1478): Respect additional datatypes when parsing PFBs.

Main change:
* Add support for Avro `map` type.

Also:
* Rearranges the type conversion logic to match the order specified in the linked Avro documentation.  This is just to make it easier to eye both the code and the docs and make sure they are the same list of things.
* Print a warning on unexpected type.

[AJ-1478]: https://broadworkbench.atlassian.net/browse/AJ-1478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
@@ -13,7 +13,6 @@ import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericEnumSymbol;
 import org.apache.avro.generic.GenericFixed;
 import org.apache.avro.generic.GenericRecord;
-import org.apache.commons.lang3.NotImplementedException;
 import org.databiosphere.workspacedataservice.service.PfbStreamWriteHandler;
 import org.databiosphere.workspacedataservice.service.RelationUtils;
 import org.databiosphere.workspacedataservice.shared.model.Record;
@@ -100,8 +99,7 @@ public class PfbRecordConverter {
 
     // Avro records
     if (attribute instanceof GenericRecord recordAttr) {
-      throw new NotImplementedException("Avro `record` attributes not yet supported (AJ-1478)");
-      // TODO AJ-1478: Make these into WDS json?
+      return recordAttr.toString(); // TODO AJ-1478: Make these into WDS json?
     }
 
     // Avro enums


service/src/test/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverterTest.java
```diff
@@ -142,14 +152,24 @@ void valuesInObjectAttributes() {
         actual.getAttributeValue("arrayOfNumbers"));
     assertEquals(List.of("one", "two", "three"), actual.getAttributeValue("arrayOfStrings"));
     assertEquals(List.of("enumValue2", "enumValue1"), actual.getAttributeValue("arrayOfEnums"));
+    assertEquals(
```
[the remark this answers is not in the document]
jladieu: Oops, yes!  Fixed.

calypsomatic (approved):

nawatts (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
```diff
@@ -89,24 +95,13 @@ Object convertAttributeType(Object attribute) {
...
-      return BigDecimal.valueOf(doubleAttr);
-    }
 
-    // Avro booleans
-    if (attribute instanceof Boolean boolAttr) {
-      return boolAttr;
+    // Avro records
+    if (attribute instanceof GenericRecord recordAttr) {
+      throw new NotImplementedException("Avro `record` attributes not yet supported (AJ-1478)");
+      // TODO AJ-1478: Make these into WDS json?
```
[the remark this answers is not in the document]
jladieu: Fixed in ac6bc010f548e03f2f82ddb2827df25f01ae37d0

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
```diff
@@ -120,21 +113,59 @@ Object convertAttributeType(Object attribute) {
...
-    // Avro bytes
-    if (attribute instanceof ByteBuffer byteBufferAttr) {
-      return new String(byteBufferAttr.array());
+    // Avro maps
+    if (attribute instanceof Map<?, ?> mapAttr) {
+      // recurse
+      return BiStream.from(mapAttr)
+          .mapKeys(Object::toString)
+          .mapValues(this::convertAttributeType)
+          .toMap();
```
davidangb, line 122: Brainstorming: does `attribute` actually come back as some higher-level class, perhaps in the `org.apache.avro.generic` package? Something that extends Map? If so, does that higher-level class have a `.toString()` that knows how to turn it into json?

My other thought here is that it may be safer to just do a Jackson ObjectMapper.writeValueAsString(mapAttr) instead of recursing on it and potentially resulting in map values that WDS doesn't know how to handle.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
```diff
@@ -89,24 +94,12 @@ Object convertAttributeType(Object attribute) {
...
-    }
-    if (attribute instanceof Integer intAttr) {
-      return BigDecimal.valueOf(intAttr);
-    }
-    if (attribute instanceof Float floatAttr) {
-      return BigDecimal.valueOf(floatAttr);
-    }
-    if (attribute instanceof Double doubleAttr) {
-      return BigDecimal.valueOf(doubleAttr);
-    }
```
davidangb, line 105: fwiw I had ordered this with the most-common types first and least-common types last (except for the catch-all String at the end), to optimize runtime … I suspect that's a fairly low-impact prioritization though.

davidangb (commented):
some thinking-out-loud comments inline …

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
jladieu: I see this and will followup when I come back for the final type.  Thanks for taking a look!

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
jladieu: With another nudge, I'd be happy put this back the way it was, although definitely welcome your hunch of where the new types belong in the frequency ordering.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
jladieu: Fixed with https://github.com/DataBiosphere/terra-workspace-data-service/pull/435#discussion_r1419783800

PR #435, opened 2023-12-08
Add support for Avro `record` type.

jladieu:
[AJ-1478](https://broadworkbench.atlassian.net/browse/AJ-1478): Respect additional datatypes when parsing PFBs.

Main change:
* Add support for Avro `record` type.

Also:
* Use `objectMapper` to stringify `map` attributes, instead of recursively serializing them.
* Test enums using parameterized test.
* Switch to using `GenericRecordBuilder`.
* Make `PfbRecordConverter` a Spring-managed `@Bean`.

[AJ-1478]: https://broadworkbench.atlassian.net/browse/AJ-1478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
@@ -3,13 +3,15 @@ package org.databiosphere.workspacedataservice.dataimport;
 import static bio.terra.pfb.PfbReader.convertEnum;
 import static org.databiosphere.workspacedataservice.service.PfbStreamWriteHandler.PfbImportMode.RELATIONS;
 
-import com.google.mu.util.stream.BiStream;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import java.math.BigDecimal;
 import java.nio.ByteBuffer;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericEnumSymbol;
 import org.apache.avro.generic.GenericFixed;
 import org.apache.avro.generic.GenericRecord;
@@ -32,6 +34,12 @@ public class PfbRecordConverter {
   public static final String RELATIONS_ID = "dst_id";
   public static final String RELATIONS_NAME = "dst_name";
 
+  private final ObjectMapper objectMapper;
+
+  public PfbRecordConverter(ObjectMapper objectMapper) {
+    this.objectMapper = objectMapper;
+  }
+
   public Record genericRecordToRecord(
       GenericRecord genRec, PfbStreamWriteHandler.PfbImportMode pfbImportMode) {
     // create the WDS record shell (id, record type, empty attributes)
@@ -40,11 +48,10 @@ public class PfbRecordConverter {
             genRec.get(ID_FIELD).toString(),
             RecordType.valueOf(genRec.get(TYPE_FIELD).toString()),
             RecordAttributes.empty());
-    if (pfbImportMode == RELATIONS) {
-      return addRelations(genRec, converted);
-    } else {
-      return addAttributes(genRec, converted);
-    }
+    return switch (pfbImportMode) {
+      case RELATIONS -> addRelations(genRec, converted);
+      case BASE_ATTRIBUTES -> addAttributes(genRec, converted);
+    };
   }
 
   private Record addAttributes(GenericRecord genRec, Record converted) {
@@ -99,7 +106,8 @@ public class PfbRecordConverter {
 
     // Avro records
     if (attribute instanceof GenericRecord recordAttr) {
-      return recordAttr.toString(); // TODO AJ-1478: Make these into WDS json?
+      // According to its Javadoc, GenericData#toString() renders the given datum as JSON
+      return GenericData.get().toString(recordAttr);
     }
 
     // Avro enums
@@ -115,11 +123,7 @@ public class PfbRecordConverter {
 
     // Avro maps
     if (attribute instanceof Map<?, ?> mapAttr) {
-      // recurse
-      return BiStream.from(mapAttr)
-          .mapKeys(Object::toString)
-          .mapValues(this::convertAttributeType)
-          .toMap();
+      return convertToString(mapAttr);
     }
 
     // Avro fixed
@@ -167,5 +171,18 @@ public class PfbRecordConverter {
         attribute,
         attribute.getClass());
     return attribute.toString();
+  }
+
+  private String convertToString(Object attribute) {
+    try {
+      return objectMapper.writeValueAsString(attribute);
+    } catch (JsonProcessingException e) {
+      LOGGER.warn(
+          String.format(
+              "Unable to convert attribute \"%s\" to JSON string, falling back to toString()",
+              attribute),
+          e);
+      return attribute.toString();
+    }
   }
 }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
@@ -1,7 +1,6 @@
 package org.databiosphere.workspacedataservice.dataimport;
 
 import static bio.terra.pfb.PfbReader.convertEnum;
-import static org.databiosphere.workspacedataservice.service.PfbStreamWriteHandler.PfbImportMode.RELATIONS;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
@@ -107,6 +107,10 @@ public class PfbRecordConverter {
     // Avro records
     if (attribute instanceof GenericRecord recordAttr) {
       // According to its Javadoc, GenericData#toString() renders the given datum as JSON
+      // However, it may contribute to numeric precision loss (see:
+      // https://broadworkbench.atlassian.net/browse/AJ-1292)
+      // If that's the case, then it may be necessary to traverse the record recursively and do a
+      // less lossy conversion process.
       return GenericData.get().toString(recordAttr);
     }
 
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
@@ -136,8 +136,8 @@ public class PfbRecordConverter {
     }
 
     // Avro strings
-    if (attribute instanceof String stringAttr) {
-      return stringAttr;
+    if (attribute instanceof CharSequence charSequenceAttr) {
+      return charSequenceAttr.toString();
     }
 
     // Avro bytes
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
@@ -20,6 +20,7 @@ import java.util.UUID;
 import org.apache.avro.file.DataFileStream;
 import org.apache.avro.generic.GenericRecord;
 import org.databiosphere.workspacedataservice.dao.RecordDao;
+import org.databiosphere.workspacedataservice.dataimport.PfbRecordConverter;
 import org.databiosphere.workspacedataservice.service.model.BatchWriteResult;
 import org.databiosphere.workspacedataservice.service.model.DataTypeMapping;
 import org.databiosphere.workspacedataservice.service.model.exception.BadStreamingWriteRequestException;
@@ -44,6 +45,7 @@ public class BatchWriteService {
   private final ObjectReader tsvReader;
 
   private final RecordService recordService;
+  private final PfbRecordConverter pfbRecordConverter;
 
   public BatchWriteService(
       RecordDao recordDao,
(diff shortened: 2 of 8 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
```diff
@@ -40,11 +48,10 @@ public Record genericRecordToRecord(
             genRec.get(ID_FIELD).toString(),
             RecordType.valueOf(genRec.get(TYPE_FIELD).toString()),
             RecordAttributes.empty());
-    if (pfbImportMode == RELATIONS) {
-      return addRelations(genRec, converted);
-    } else {
-      return addAttributes(genRec, converted);
-    }
+    return switch (pfbImportMode) {
```
jladieu, line 51: Since this is an enum value, I started manually switching over to an oldschool switch statement...but the IDE recommended this fancy new [switch expression](https://blog.gypsyengineer.com/en/tech/new-switch-expressions-in-java-14.html) style and then also coached me to drop the `default` clause.  I thought this was cool because it is required to be exhaustive.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
```diff
@@ -115,11 +123,7 @@ Object convertAttributeType(Object attribute) {
 
     // Avro maps
     if (attribute instanceof Map<?, ?> mapAttr) {
-      // recurse
-      return BiStream.from(mapAttr)
-          .mapKeys(Object::toString)
-          .mapValues(this::convertAttributeType)
-          .toMap();
+      return convertToString(mapAttr);
```
jladieu, line 126: Per @davidangb 's [suggestion](https://github.com/DataBiosphere/terra-workspace-data-service/pull/434#discussion_r1419472036)

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -214,7 +214,7 @@ public BatchWriteResult batchWritePfbStream(
       Optional<String> primaryKey,
       PfbStreamWriteHandler.PfbImportMode pfbImportMode) {
     try (PfbStreamWriteHandler streamingWriteHandler =
-        new PfbStreamWriteHandler(is, pfbImportMode)) {
+        new PfbStreamWriteHandler(is, pfbImportMode, objectMapper)) {
```
jladieu, line 217: I happened to have `objectMapper` in hand so I unceremoniously used it.  I'd be open to doing some fancier config to have it AutoWired but it didn't seem like a big deal to just provide one in the few places where the `PfbStreamWriteHandler` is constructed.

service/src/main/java/org/databiosphere/workspacedataservice/service/PfbStreamWriteHandler.java
```diff
@@ -13,24 +14,27 @@
...
    * @param inputStream the PFB stream
    */
   public PfbStreamWriteHandler(
-      DataFileStream<GenericRecord> inputStream, PfbImportMode pfbImportMode) {
+      DataFileStream<GenericRecord> inputStream,
+      PfbImportMode pfbImportMode,
+      ObjectMapper objectMapper) {
     this.inputStream = inputStream;
     this.pfbImportMode = pfbImportMode;
+    this.pfbRecordConverter = new PfbRecordConverter(objectMapper);
```
jladieu, line 37: Also thought about having the `PfbRecordConverter` `@AutoWired` but wasn't sure if it made sense here so went with the oldschool construction.  Open to nudging here.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverterTest.java
```diff
@@ -94,31 +102,40 @@ void valuesInObjectAttributes() {
...
+                new Schema.Field("mapOfEnums", Schema.createMap(enumSchema)),
+                new Schema.Field("embeddedObject", embeddedObjectSchema)));
+
+    GenericData.Record embeddedObjectRecord =
+        new GenericRecordBuilder(embeddedObjectSchema)
+            .set("embeddedLong", 123L)
+            .set("embeddedString", "embeddedString")
+            .build();
+    GenericData.Record objectAttributes =
+        new GenericRecordBuilder(myObjSchema)
```
jladieu, line 114: Switching to the `Builder` provides some additional correctness guarantees during creation, which is what resulted in me removing the test for `null` object.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/PfbTestUtils.java
```diff
@@ -101,12 +102,12 @@ public static GenericRecord makeRecord(
...
-    GenericRecord rec = new GenericData.Record(RECORD_SCHEMA);
-    rec.put("id", id);
-    rec.put("name", name);
-    rec.put("object", objectAttributes);
-    rec.put("relations", relations);
-    return rec;
+    return new GenericRecordBuilder(RECORD_SCHEMA)
+        .set("id", id)
+        .set("name", name)
+        .set("object", objectAttributes)
```
jladieu, line 108: The `GenericRecordBuilder` throws if you give it a `null` value for `objectAttributes`, which is why I removed the test for this, since I could no longer actually set up the fixture.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
jladieu: Should I give this same treatment to the block for `array`?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
calypsomatic: Neat!

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverterTest.java
```diff
@@ -48,17 +56,7 @@ void recordIdentifiers() {
   @Test
   void emptyObjectAttributes() {
     GenericRecord input = PfbTestUtils.makeRecord("my-id", "my-name");
-    Record actual = new PfbRecordConverter().genericRecordToRecord(input, BASE_ATTRIBUTES);
-
-    assertThat(actual.attributeSet()).isEmpty();
-  }
-
-  // if the GenericRecord has null in the "object" column, the WDS record should have no attributes
```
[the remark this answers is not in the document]
davidangb: great question, if you think it isn't possible in the real world, it's great to omit this test

davidangb (commented):
A few things to think about inline.

I know we are struggling to construct arbitrary PFBs of our own. I want to make sure at some point (doesn't have to block this PR) that we test this end-to-end and ensure the actual schema in Postgres is as expected. This could be in PfbQuartzJobE2ETest, or a manual test.

service/src/main/java/org/databiosphere/workspacedataservice/service/PfbStreamWriteHandler.java
jladieu: I'd be happy to do the refactor to make it so in a followup if you think we've passed the inflection point where it'd make more sense to be a bean.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverterTest.java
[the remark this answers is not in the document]
jladieu: Looking into the validation logic which triggered here:

```
  /**
   * Validates that a particular value for a given field is valid according to the
   * following algorithm: 1. If the value is not null, or the field type is null,
   * or the field type is a union which accepts nulls, returns. 2. Else, if the
   * field has a default value, returns. 3. Otherwise throws AvroRuntimeException.
   * 
   * @param field the field to validate.
   * @param value the value to validate.
   * @throws AvroRuntimeException if value is null and the given field does not
   *                              accept null values.
   */
```

So perhaps it's possible for a `null` object, but the types we set up in our test case weren't configured to allow it.

I'll see if I can back up the removal of this test and figure out the correct incantation to maintain the test coverage.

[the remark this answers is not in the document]
jladieu: Ok after fighting more with Avro to set up this test scenario and losing, I'm back to thinking that this scenario won't ever practically happen; while the Javadoc I linked earlier suggests that a null value would potentially be considered valid in the presence of a field default, I don't think we need to worry about a default:

Looking at https://avro.apache.org/docs/1.11.1/specification/#schema-record, which explains the defaults:

default: A default value for this field, *only used when reading instances that lack the field for schema evolution purposes.* The presence of a default value does not make the field optional at encoding time. Permitted values depend on the field’s schema type, according to the table below. Default values for union fields correspond to the first schema in the union. Default values for bytes and fixed fields are JSON strings, where Unicode code points 0-255 are mapped to unsigned 8-bit byte values 0-255. Avro encodes a field even if its value is equal to its default.

So my takeaway is that unless we're going to support serialized PFBs in some version that predate the `object` field being in the schema (if such a version ever existed), we should be in the clear.  Avro wouldn't allow serialization of records without an object field.

Totally open to having a real world scenario prove my understanding wrong here, but unless we encounter it ourselves  in the wild I think we should be okay to assume it won't happen.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
```diff
@@ -99,7 +106,8 @@ Object convertAttributeType(Object attribute) {
 
     // Avro records
     if (attribute instanceof GenericRecord recordAttr) {
-      return recordAttr.toString(); // TODO AJ-1478: Make these into WDS json?
+      // According to its Javadoc, GenericData#toString() renders the given datum as JSON
+      return GenericData.get().toString(recordAttr);
```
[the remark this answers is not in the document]
bethsheets: @jladieu that plan is fine with me.

service/src/main/java/org/databiosphere/workspacedataservice/service/PfbStreamWriteHandler.java
jladieu: I ended up converting to a bean in 36826f25e0206a7885287bb137f81d1c698f890c

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverterTest.java
```diff
@@ -31,14 +36,17 @@
 
 class PfbRecordConverterTest {
 
+  private final PfbRecordConverter converter =
+      new PfbRecordConverter(new JsonConfig().objectMapper());
+
```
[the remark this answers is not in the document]
jladieu: This prompted me to switch `PfbRecordConverter` over to be a bean in 36826f25e0206a7885287bb137f81d1c698f890c, namely because I didn't want to introduce another testing idiom into the codebase.  So instead of doing your suggested fix, I ended up with an `@AutoWired` instance variable.  LMK if this seems reasonable or not.

jladieu:
This is ready for another round of review.  @davidangb I saw your comment about validating the resulting schema, and I'd like to figure out a way to automate that with `PfbQuartzJobE2ETest` but didn't want to hold up this PR on that task.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
```diff
@@ -128,8 +136,8 @@ Object convertAttributeType(Object attribute) {
     }
 
     // Avro strings
-    if (attribute instanceof String stringAttr) {
-      return stringAttr;
+    if (attribute instanceof CharSequence charSequenceAttr) {
+      return charSequenceAttr.toString();
```
davidangb, line 140: I'm curious about the switch from `String` to `CharSequence` here. Are you seeing inbound text as a class other than String? Specifically, are you seeing it as `org.apache.avro.util.Utf8`?

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverterTest.java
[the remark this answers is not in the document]
davidangb: sounds good to me, thanks for all the research on this.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
```diff
@@ -3,13 +3,15 @@
 import static bio.terra.pfb.PfbReader.convertEnum;
 import static org.databiosphere.workspacedataservice.service.PfbStreamWriteHandler.PfbImportMode.RELATIONS;
```
davidangb, line 4: nit: unused import (Sonar is flagging this)

davidangb (approved):
one nit and one question inline - proactively approving, thank you for the diligence!

re: validating the resulting schema via `PfbQuartzJobE2ETest` … I agree on pushing that off; let's get AJ-1512 done and then come back and figure out how to improve test coverage.

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverter.java
jladieu: Yes, that's exactly it.  Also, the linked doc specifies that Avro returns `CharSequence` for strings, _not_ that it returns Strings.  TIL that String extends CharSequence and not the other way around :)

`Utf8` also extends `CharSequence`, but I think is intended to be an implementation detail, so I didn't want to check against that type.

jladieu: From: https://avro.apache.org/docs/current/api/java/org/apache/avro/generic/package-summary.html

image

PR #437, opened 2023-12-12
AJ-1013: TDR snapshot import

davidangb:
Highlights:
* bring in the `parquet-avro` library and supporting libraries, which allows reading Parquet files into Avro Java classes
* create new `ParquetRecordConverter`
* factor out a bunch of shared logic from `PfbRecordConverter` and `ParquetRecordConverter` into a new abstract base class `AvroRecordConverter`
* create new `ParquetStreamWriteHandler`, which mirrors `PfbStreamWriteHandler`
* create new `TdrManifestQuartzJob`
* factor out shared logic from `PfbQuartzJob` and `TdrManifestQuartzJob` into `TdrSnapshotSupport`

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -85,7 +85,7 @@ dependencies {
     // Parquet, used for TDR snapshot import
     implementation "org.apache.parquet:parquet-common:$parquet_mr_version"
     implementation "org.apache.parquet:parquet-avro:$parquet_mr_version"
-    implementation("org.apache.parquet:parquet-hadoop:$parquet_mr_version")
+    implementation "org.apache.parquet:parquet-hadoop:$parquet_mr_version"
 
     // we need Hadoop libraries to read Parquet files for TDR import. However, Hadoop brings
     // a boatload of transitive dependencies we don't need, some of which (e.g. com.sun.xml.bind)
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -87,12 +87,10 @@ dependencies {
     implementation "org.apache.parquet:parquet-avro:$parquet_mr_version"
     implementation("org.apache.parquet:parquet-hadoop:$parquet_mr_version")
 
-    // Hadoop libraries, used for TDR snapshot import
     // we need Hadoop libraries to read Parquet files for TDR import. However, Hadoop brings
     // a boatload of transitive dependencies we don't need, some of which (e.g. com.sun.xml.bind)
     // cause conflicts. Exclude a lot of the transitive deps. Inspired by https://stackoverflow.com/a/60067027
-    // TODO AJ-1013 can the excludes be factored out into a common method?
-    implementation("org.apache.hadoop:hadoop-common:$hadoop_version") {
+    def withoutHadoopExcludes = {
         exclude(group: 'log4j')
         exclude(group: 'org.slf4j')
         exclude(group: 'org.mortbay.jetty')
@@ -104,18 +102,8 @@ dependencies {
         exclude(group: 'org.apache.kerby')
         exclude(group: 'com.google.protobuf')
     }
-    implementation("org.apache.hadoop:hadoop-mapreduce-client-core:$hadoop_version") {
-        exclude(group: 'log4j')
-        exclude(group: 'org.slf4j')
-        exclude(group: 'org.mortbay.jetty')
-        exclude(group: 'javax.servlet.jsp')
-        exclude(group: 'com.sun.jersey')
-        exclude(group: 'com.sun.xml.bind')
-        exclude(group: 'org.apache.curator')
-        exclude(group: 'org.apache.zookeeper')
-        exclude(group: 'org.apache.kerby')
-        exclude(group: 'com.google.protobuf')
-    }
+    implementation "org.apache.hadoop:hadoop-common:$hadoop_version", withoutHadoopExcludes
+    implementation "org.apache.hadoop:hadoop-mapreduce-client-core:$hadoop_version", withoutHadoopExcludes
 
     // hk2 is required to use WSM client, but not correctly exposed by the client
     implementation "org.glassfish.jersey.inject:jersey-hk2:$jersey_version"
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -37,6 +37,8 @@ jib {
 
 ext {
     jersey_version = "2.36"
+    parquet_mr_version = "1.13.1"
+    hadoop_version = "3.3.6"
 }
 
 dependencies {
@@ -79,6 +81,41 @@ dependencies {
     implementation "bio.terra:workspace-manager-client:0.254.717-SNAPSHOT"
     implementation "bio.terra:java-pfb-library:0.15.0"
     implementation project(path: ':client')
+
+    // Parquet, used for TDR snapshot import
+    implementation "org.apache.parquet:parquet-common:$parquet_mr_version"
+    implementation "org.apache.parquet:parquet-avro:$parquet_mr_version"
+    implementation("org.apache.parquet:parquet-hadoop:$parquet_mr_version")
+
+    // Hadoop libraries, used for TDR snapshot import
+    // we need Hadoop libraries to read Parquet files for TDR import. However, Hadoop brings
+    // a boatload of transitive dependencies we don't need, some of which (e.g. com.sun.xml.bind)
+    // cause conflicts. Exclude a lot of the transitive deps. Inspired by https://stackoverflow.com/a/60067027
+    // TODO AJ-1013 can the excludes be factored out into a common method?
+    implementation("org.apache.hadoop:hadoop-common:$hadoop_version") {
+        exclude(group: 'log4j')
+        exclude(group: 'org.slf4j')
+        exclude(group: 'org.mortbay.jetty')
+        exclude(group: 'javax.servlet.jsp')
+        exclude(group: 'com.sun.jersey')
+        exclude(group: 'com.sun.xml.bind')
+        exclude(group: 'org.apache.curator')
+        exclude(group: 'org.apache.zookeeper')
+        exclude(group: 'org.apache.kerby')
+        exclude(group: 'com.google.protobuf')
+    }
+    implementation("org.apache.hadoop:hadoop-mapreduce-client-core:$hadoop_version") {
+        exclude(group: 'log4j')
+        exclude(group: 'org.slf4j')
+        exclude(group: 'org.mortbay.jetty')
+        exclude(group: 'javax.servlet.jsp')
+        exclude(group: 'com.sun.jersey')
+        exclude(group: 'com.sun.xml.bind')
+        exclude(group: 'org.apache.curator')
+        exclude(group: 'org.apache.zookeeper')
+        exclude(group: 'org.apache.kerby')
+        exclude(group: 'com.google.protobuf')
+    }
 
     // hk2 is required to use WSM client, but not correctly exposed by the client
     implementation "org.glassfish.jersey.inject:jersey-hk2:$jersey_version"
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/AvroRecordConverter.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/AvroRecordConverter.java
@@ -22,10 +22,7 @@ import org.databiosphere.workspacedataservice.shared.model.RecordType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-/**
- * Logic to convert Avro GenericRecord to WDS's Record Avro GenericRecord is used by PFB import and
- * TDR import.
- */
+/** Logic to convert Avro GenericRecord to WDS's Record. Used by PFB import and TDR import. */
 public abstract class AvroRecordConverter {
   private static final Logger LOGGER = LoggerFactory.getLogger(AvroRecordConverter.class);
 
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/AvroRecordConverter.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/AvroRecordConverter.java
@@ -3,15 +3,16 @@ package org.databiosphere.workspacedataservice.dataimport;
 import static bio.terra.pfb.PfbReader.convertEnum;
 import static org.databiosphere.workspacedataservice.service.PfbStreamWriteHandler.PfbImportMode.RELATIONS;
 
-import com.google.mu.util.stream.BiStream;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import java.math.BigDecimal;
 import java.nio.ByteBuffer;
-import java.nio.charset.StandardCharsets;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericEnumSymbol;
 import org.apache.avro.generic.GenericFixed;
 import org.apache.avro.generic.GenericRecord;
@@ -28,7 +29,11 @@ import org.slf4j.LoggerFactory;
 public abstract class AvroRecordConverter {
   private static final Logger LOGGER = LoggerFactory.getLogger(AvroRecordConverter.class);
 
-  public AvroRecordConverter() {}
+  private final ObjectMapper objectMapper;
+
+  public AvroRecordConverter(ObjectMapper objectMapper) {
+    this.objectMapper = objectMapper;
+  }
 
   /**
    * Create the "shell" Record, with a valid RecordId and a valid RecordType, but no attributes.
@@ -77,7 +82,7 @@ public abstract class AvroRecordConverter {
     return converted;
   }
 
-  protected Object convertAttributeType(Object attribute) {
+  Object convertAttributeType(Object attribute) {
 
     if (attribute == null) {
       return null;
@@ -88,7 +93,12 @@ public abstract class AvroRecordConverter {
 
     // Avro records
     if (attribute instanceof GenericRecord recordAttr) {
-      return recordAttr.toString(); // TODO AJ-1478: Make these into WDS json?
+      // According to its Javadoc, GenericData#toString() renders the given datum as JSON
+      // However, it may contribute to numeric precision loss (see:
+      // https://broadworkbench.atlassian.net/browse/AJ-1292)
+      // If that's the case, then it may be necessary to traverse the record recursively and do a
+      // less lossy conversion process.
+      return GenericData.get().toString(recordAttr);
     }
 
     // Avro enums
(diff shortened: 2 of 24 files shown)


service/build.gradle
```diff
@@ -80,6 +82,29 @@ dependencies {
     implementation "bio.terra:java-pfb-library:0.15.0"
     implementation project(path: ':client')
 
+    // Parquet, used for TDR snapshot import
+    implementation "org.apache.parquet:parquet-common:$parquet_mr_version"
+    implementation "org.apache.parquet:parquet-avro:$parquet_mr_version"
+    implementation("org.apache.parquet:parquet-hadoop:$parquet_mr_version")
```
jladieu, line 88: nit: drop the parens for consistency

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/AvroRecordConverter.java
```diff
@@ -0,0 +1,205 @@
...
+import org.apache.avro.generic.GenericRecord;
+import org.databiosphere.workspacedataservice.service.PfbStreamWriteHandler;
+import org.databiosphere.workspacedataservice.shared.model.Record;
+import org.databiosphere.workspacedataservice.shared.model.RecordAttributes;
+import org.databiosphere.workspacedataservice.shared.model.RecordType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Logic to convert Avro GenericRecord to WDS's Record Avro GenericRecord is used by PFB import and
```
jladieu, line 26: Formatter overzealously wrapped this line on you

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ParquetRecordConverter.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.List;
+import java.util.Set;
+import org.apache.avro.generic.GenericRecord;
+import org.databiosphere.workspacedataservice.service.RelationUtils;
+import org.databiosphere.workspacedataservice.service.model.TdrManifestImportTable;
+import org.databiosphere.workspacedataservice.shared.model.RecordAttributes;
+import org.databiosphere.workspacedataservice.shared.model.RecordType;
+
+/** Logic to convert a TDR Parquet's GenericRecord to WDS's Record */
+// TODO AJ-1522 should ParquetRecordConverter be a bean, like PfbRecordConverter? Right now,
```
jladieu, line 16: For the TODO: Maybe make a `RecordConverterFactory` `@Bean` with a method like `createParquetConverter(TdrManifestImportTable table)`, and also a method for creating a `PfbRecordConverter`?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/AvroRecordConverter.java
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  /**
+   * What is the id for this record? Subclasses must implement this method, which is called for
+   * every GenericRecord in the inbound import data.
+   *
+   * @param genericRecord the inbound Avro GenericRecord to be converted
+   * @return the id to use for the WDS record
+   */
+  abstract String extractRecordId(GenericRecord genericRecord);
```
jladieu, line 45: Sorry in advance if this is too much churn, but I can't help myself.

Whenever I see the [Template Pattern](https://www.tutorialspoint.com/design_pattern/template_pattern.htm) in play I always try to figure out how to accomplish the same thing with composition instead of inheritance, so I couldn't help but squint at this class.

I notice all the methods that remain `abstract` on this class are of the form `extract<Something>(GenericRecord genericRecord)`.  What do you think of moving these methods to their own interface whose job is to wrap a single `GenericRecord` and return info about it that's format specific?  Maybe its usage could look like:

```
RecordReader reader = recordReaderFactory.createReader(genRec);

String id = reader.recordId();
RecordType = reader.recordType();

... and so on
```

This allows `AvroRecordConverter` to maintain its single `converter` responsibility, but shifts the extraction/reading functionality over to another class whose only job is to read fancy formats and return standardized stuff.

As my suggested code snippet depicts, I'm guessing you'll also need a factory for the `RecordReader` instances, and this would be the new actor that takes on the extra state & responsibility of paying attention to, for example, the the `TdrManifestImportTable` when generating records from parquet.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -3,66 +3,329 @@
...
+   * @param targetInstance instance into which to import
+   * @param pfbImportMode mode for this invocation
+   */
+  private void importTables(
+      List<TdrManifestImportTable> importTables,
+      UUID targetInstance,
+      PfbStreamWriteHandler.PfbImportMode pfbImportMode) {
+    // loop through the tables that have data files. In the TDR manifest, for Azure snapshots only,
+    // the first file in the list will always be a directory. Attempting to import that directory
+    // will fail; it has no content. Ensure we are resilient to those failures.
```
jladieu, line 205: This comment is helpful extra context but I can't see the connection from the comment to where you ensure resilience to the failures -- can you give a more specific trailmarker to where this happens?  Or maybe move the comment closer to it?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -3,66 +3,329 @@
...
+   * @param snapshotExportResponseModel the inbound manifest
+   * @return information necessary for the import
+   */
+  List<TdrManifestImportTable> extractTableInfo(
+      SnapshotExportResponseModel snapshotExportResponseModel) {
+
+    TdrSnapshotSupport tdrSnapshotSupport =
+        new TdrSnapshotSupport(workspaceId, wsmDao, restClientRetry);
+
+    // find all the exported tables in the manifest.
```
jladieu, line 253: These three methods, with their corresponding helpful comments, might make sense driven down into the `TdrSnapshotSupport` behind one-liner methods with intention revealing names (which perhaps obviate the need for comment explanation?)

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -3,66 +3,329 @@
...
+
+              // determine data files for this table
+              List<URL> dataFiles =
+                  table.getPaths().stream()
+                      .map(
+                          x -> {
+                            try {
+                              return new URL(x);
+                            } catch (MalformedURLException e) {
+                              throw new TdrManifestImportException(e.getMessage());
```
jladieu, line 310: can you pass the `e` in as a cause?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -3,66 +3,329 @@
...
+                            return requestedPrimaryKey != null
+                                && requestedPrimaryKey.equals(actualPrimaryKey);
+                          })
+                      .toList();
+
+              // determine data files for this table
+              List<URL> dataFiles =
+                  table.getPaths().stream()
+                      .map(
+                          x -> {
```
jladieu, line 306: nit: rename `x` to `path`? (I try to avoid mystery letter variable names if possible)

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -3,66 +3,329 @@
...
+              String primaryKey = primaryKeys.get(recordType);
+
+              // this table may or may not have relations
+              Collection<RelationshipModel> possibleRelations = relationsByTable.get(recordType);
+
+              // filter relations to those that point at a valid primary key
+              List<RelationshipModel> relations =
+                  possibleRelations.stream()
+                      .filter(
+                          relationshipModel -> {
```
jladieu, line 290: Can you extract the contents of the express into a private method to make the filter a one-liner?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -3,66 +3,329 @@
...
+                                && requestedPrimaryKey.equals(actualPrimaryKey);
+                          })
+                      .toList();
+
+              // determine data files for this table
+              List<URL> dataFiles =
+                  table.getPaths().stream()
+                      .map(
+                          x -> {
+                            try {
```
jladieu, line 307: Could this try / catch block be extracted into a method to make the map a one-liner?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrSnapshotSupport.java
```diff
@@ -125,4 +137,70 @@ protected UUID safeGetSnapshotId(ResourceDescription resourceDescription) {
...
+   * that does not already have a reference.
+   *
+   * @param snapshotIds the list of snapshot ids to create or verify references.
+   */
+  protected void linkSnapshots(Set<UUID> snapshotIds) {
+    // list existing snapshots linked to this workspace
+    List<UUID> existingSnapshotIds = existingPolicySnapshotIds(/* pageSize= */ 50);
+    // find the snapshots that are not already linked to this workspace
+    List<UUID> newSnapshotIds =
+        snapshotIds.stream().filter(id -> !existingSnapshotIds.contains(id)).toList();
```
jladieu, line 152: You've already got a set in hand with `snapshotIds` -- would this be worth converting existing to a Set and using `Sets.difference`?

service/src/main/java/org/databiosphere/workspacedataservice/service/TwoPassStreamingWriteHandler.java
jladieu, line 1: Nice :)

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJobTest.java
```diff
@@ -0,0 +1,104 @@
...
+    // schema,
+    // but only contains export data files for project, edges, and test_result.
+    List<TdrManifestImportTable> expected =
+        List.of(
+            // single primary key
+            new TdrManifestImportTable(
+                RecordType.valueOf("project"),
+                "project_id",
+                List.of(
+                    new URL(
```
jladieu, line 72: Consider extracting these URLs to constants with intention revealing names?

jladieu (approved):
Left a few comments and questions!  Nothing critical so leaving a pre-emptive approval.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/AvroRecordConverter.java
```diff
@@ -0,0 +1,205 @@
...
+   * for every GenericRecord in the inbound import data.
+   *
+   * @param genericRecord the inbound Avro GenericRecord to be converted
+   * @return the RecordType to use for the WDS record
+   */
+  abstract RecordType extractRecordType(GenericRecord genericRecord);
+
+  /**
+   * When operating in {@see PfbStreamWriteHandler.PfbImportMode.BASE_ATTRIBUTES} mode, what
+   * attributes should be upserted for this GenericRecord?
```
calypsomatic, line 58: Not terribly important, but maybe a bit confusing for the `BASE_ATTRIBUTES` and `RELATIONS` enums to be `PfbImportMode` in `PfbStreamWriterHandler` when they're used/referred to by non-pfb classes

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/AvroRecordConverter.java
davidangb: I've refactored and renamed these - in a different way than you suggested. Does this work better for you?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/AvroRecordConverter.java
davidangb: moved!

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/PfbRecordConverterTest.java
```diff
@@ -49,8 +49,7 @@ void recordIdentifiers() {
     GenericRecord input = PfbTestUtils.makeRecord(inputId, inputName);
 
     Record actual =
-        converter.genericRecordToRecord(
-            input, TwoPassStreamingWriteHandler.ImportMode.BASE_ATTRIBUTES);
+        converter.convert(input, TwoPassStreamingWriteHandler.ImportMode.BASE_ATTRIBUTES);
```
jladieu, line 52: supernit: can you change the import so that this reads like `converter.convert(input, ImportMode.BASE_ATTRIBUTES);`

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
davidangb: I did not do this; lmk if you feel strongly. These are already one-liners, albeit dense ones, and 2 of the 3 already delegate most of their work to `TdrSnapshotSupport`, so I don't know if we can do a lot better. I considered having `tdrSnapshotSupport.identifyPrimaryKeys()` and `tdrSnapshotSupport.identifyRelations()` accept the entire `SnapshotExportResponseModel` as their arguments, which would make these lines more readable; but that would also make those methods more difficult to unit-test because they need larger inputs.

service/build.gradle
```diff
@@ -85,7 +85,7 @@ dependencies {
     // Parquet, used for TDR snapshot import
     implementation "org.apache.parquet:parquet-common:$parquet_mr_version"
     implementation "org.apache.parquet:parquet-avro:$parquet_mr_version"
-    implementation("org.apache.parquet:parquet-hadoop:$parquet_mr_version")
+    implementation "org.apache.parquet:parquet-hadoop:$parquet_mr_version"
```
jladieu, line 88: I can't not read this as "Parquet Mister Version".

calypsomatic: Same

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -3,66 +3,331 @@
...
   }
 
+  // TODO AJ-1523 unit tests
   @Override
   protected void executeInternal(UUID jobId, JobExecutionContext context) {
-    // retrieve the Quartz JobDataMap, which contains arguments for this execution
+    // Grab the manifest url from the job's data map
     JobDataMap jobDataMap = context.getMergedJobDataMap();
+    URL url = getJobDataUrl(jobDataMap, ARG_URL);
+    UUID targetInstance = getJobDataUUID(jobDataMap, ARG_INSTANCE);
```
calypsomatic, line 88: What's the difference between this instanceId and the `workspaceId` on the class that's pulled from application.yml?

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJobTest.java
```diff
@@ -59,28 +59,34 @@ void extractSnapshotInfo() throws IOException {
     SnapshotExportResponseModel snapshotExportResponseModel =
         tdrManifestQuartzJob.parseManifest(manifestAzure.getURL());
 
-    // this manifest describes tables for project, edges, test_result, genome in the snapshot
-    // schema,
+    // URLs for use in the manifest below
+    List<URL> projectParquetUrls =
```
jladieu, line 63: Hmm, I'm still not loving how much of the long winded URL strings I have to read to get through the test fixture, so I'm trying to come up with a way to punt that stuff out.

One possible option is I noticed there's a bunch of reused text in between each URL, is it possible to to extract some kind of String template out to a method like:

```
private getFileParquetUrl(String importantPartThatChanges) {
 return new URL(String.format("https://mysnapshotsa.blob.core.windows.net/metadata/parquet/9516afec-583f-11ec-bf63-0242ac130002/%s/F0EE365B-314D-4E19-A177-E8F63D883716_9274_0-1.parquet?sp=r&st=2022-08-04T15:31:55Z&se=2022-08-06T23:31:55Z&spr=https&sv=2021-06-08&sr=b&sig=bogus", importantPartThatChanges));
}
```

Then you might end up with:

```
List<URL> projectParquetUrls = List.of(getFileParquetUrl("project.parquet");
List<URL> edgesParquetUrls = List.of(getFileParquetUrl("edges.parquet");
...
```

jladieu: Also, once you've got that shorter form...maybe re-inlining them again would be readable enough as to not necessitate the variable name?

jladieu (approved):
Approved with tiny optional suggestions.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrSnapshotSupport.java
```diff
@@ -125,4 +138,69 @@ protected UUID safeGetSnapshotId(ResourceDescription resourceDescription) {
...
+
+  String identifyPrimaryKey(List<String> snapshotKeys) {
+    if (snapshotKeys != null && snapshotKeys.size() == 1) {
+      return snapshotKeys.get(0);
+    }
+    return DEFAULT_PRIMARY_KEY;
+  }
+
+  String getDefaultPrimaryKey() {
+    return DEFAULT_PRIMARY_KEY;
```
calypsomatic, line 190: Why is `DEFAULT_PRIMARY_KEY` private with a getter instead of just public?

calypsomatic (approved):
It works!

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
davidangb: I'm trying to future-proof for a time when workspaces could have multiple instances attached - so I want to use the instance here, not the workspace … though in practice with end users, they'll be the same.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrSnapshotSupport.java
davidangb: eh, just good encapsulation practice - a getter method is easier to mock, easier to override, etc. if we ever needed to do so

calypsomatic: Roger

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJobTest.java
davidangb: @jladieu okay, what about the latest commit? I am hesitant to create something like `getFileParquetUrl(String importantPartThatChanges)` … the fact that the URLs in this test file have so much similarity is an artifact of them being test data; in practice they will all be different, and I don't want to rely too much on their similarity.

jladieu: Looks great!

jladieu (approved):

PR #438, opened 2023-12-15
AJ-1300: capabilities API

davidangb:
This adds a "capabilities" API to WDS:
Screenshot 12-15-2023 at 04 30 PM

The response is a set of KVPs, where the value is always a boolean. Terra UI - or other clients - can call this API to determine if the WDS app they're talking to supports a given feature. If the feature they are asking about is not present in the response, callers should _assume the value is false_. This helps with backwards compatibility: WDS will only add a feature key with a value of true if the feature is enabled.

Until we get more real-world use cases, it seems a flat String->Boolean structure for KVPs is fine. Obviously if we discover that Terra UI or other callers need something more, we can expand.

Inspired by:
* https://developer.mozilla.org/en-US/docs/Web/API/Media_Capabilities_API
* https://stripe.com/docs/api/capabilities/object
* https://developers.google.com/android/reference/com/google/android/gms/wearable/CapabilityClient


service/src/main/resources/static/swagger/swagger-ui.html
```diff
@@ -4,7 +4,7 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
-  <title>Swagger UI</title>
+  <title>Workspace Data Service API</title>
```
davidangb, line 7: just cuz. I kept losing my tab because it had a generic name.

service/src/main/java/org/databiosphere/workspacedataservice/controller/CapabilitiesController.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+    capabilitiesServerModel = tempModel;
+  }
+
+  @Override
+  @JsonPropertyOrder(alphabetic = true)
+  public ResponseEntity<CapabilitiesServerModel> capabilities() {
+    return new ResponseEntity<>(capabilitiesServerModel, HttpStatus.OK);
+  }
+}
```
davidangb, line 41: The `@JsonPropertyOrder(alphabetic = true)` doesn't work here. The only way I could get the response to be sorted alphabetically by key was to change a setting globally on the ObjectMapper in JsonConfig. I would love to have the output be nicely sorted.

service/src/main/resources/static/swagger/apis-v1.yaml
```diff
@@ -116,6 +135,13 @@ components:
         format: uuid
 
   schemas:
+    Capabilities:
+      type: object
+      properties:
+        capabilities:
+          type: boolean
+      additionalProperties:
+        type: boolean
```
davidangb, line 144: this is a quirk of openapi-generator. By specifying at least one named/expected property - in this case, "capabilities" - it generates a more usable model class than if we didn't define anything and just relied on the `additionalProperties` setting.

service/src/main/java/org/databiosphere/workspacedataservice/controller/CapabilitiesController.java
jladieu: I agree sorting these would be nice! But also not super critical, this list will be small for awhile.  Doing so in a targeted way for this endpoint also seems reasonable, but also wouldn't be opposed to doing it globally.  Does something making you hesitate to do so?

service/src/main/resources/capabilities.json
```diff
@@ -0,0 +1,6 @@
+{
+  "capabilities": true,
+  "cloning": true,
```
jladieu, line 3: I'm going to put on my over-engineering hat for a minute...but this is a top level namespace, so you're claiming a _very_ broad scope by grabbing `cloning`.  I want to suggest potentially adding a concept of a namespace to give as more specificity here so you don't have to carry forward an ambiguous name once we add some _other_ kind of cloning, eg: `workspace-cloning`, so that in the future maybe we'll want to add a 'column-cloning' feature.

Also, some other thoughts:
- decide on and document a naming convention you like, eg: dot notation for namespacing, or dot-hyphen if you want to use hyphens for word separators but still have a namespace construct
- is there a distinguishing namespace that would be worth adding right now to definitely avoid future collisions?  Would it be too leaky to include a java package in a capability name?
- just hypothetically, if you _were_ to use a java package, maybe it'd be cool to use some kind of interface/annotation construct so you could get some compiler help with generating this file, or skip the file entirely and just generate a capabilities list at runtime; totally not necessary now but might be a nice way to get this list to be managed relatively for 'free' when we have enough capabilities to want it

jladieu (changes requested):
I think you should add some kind of namespace disambiguation to cloning at the least, but maybe for all keys too.  Otherwise looks good!   Some optional suggestions for future improvement if you're interested.

service/src/test/java/org/databiosphere/workspacedataservice/controller/CapabilitiesControllerTest.java
```diff
@@ -0,0 +1,53 @@
...
+  @Value("classpath:capabilities.json")
+  Resource capabilitiesResource;
+
+  @Test
+  void resourceFileIsValid() {
+    assertDoesNotThrow(
+        () ->
+            objectMapper.readValue(
+                capabilitiesResource.getInputStream(), CapabilitiesServerModel.class),
+        "Have you modified capabilities.json? Is it still valid JSON?");
```
calypsomatic, line 36: This is a good catch for updates, but I would advocate also adding a section about this in the README

davidangb: I've added a page to the wiki, does that work? https://github.com/DataBiosphere/terra-workspace-data-service/wiki/Capabilities-API (I will update that page with links once this PR merges)

calypsomatic: Looks good!  Still not sure we have a solid rubric for when something goes in the README and when in the wiki, but this at least documents it somewhere.

service/src/main/resources/capabilities.json
davidangb: okay, I've updated in the latest commit to use namespaces, and documented the structure in https://github.com/DataBiosphere/terra-workspace-data-service/wiki/Capabilities-API. I've also removed the "cloning" key altogether to avoid problems there.

We know we want an API like this to support rollout of search/edit features in data tables. However, without more detail and concrete use cases I don't want to prematurely optimize by building out auto-detection or annotations … keeping it simple until we realize it isn't simple.

jladieu (approved):
Great writeup!

service/src/main/java/org/databiosphere/workspacedataservice/controller/CapabilitiesController.java
davidangb: oh I didn't respond to this - no, I don't have hesitations to turning on the global sorting - but I'd like to do that in a separate PR in case there are any repercussions to unit tests or anything like that

calypsomatic (approved):

PR #439, opened 2023-12-19
AJ-1500: Enable Prometheus metrics

jladieu:
[AJ-1500](https://broadworkbench.atlassian.net/browse/AJ-1500)

For prometheus output format and usage, see:
https://docs.spring.io/spring-boot/docs/current/actuator-api/htmlsingle/#prometheus

Also:
* Allow any metrics starting with `MetricsConfig.ALLOWED_PREFIXES`;
this list currently includes everything enabled out of the box, plus a
`wds` prefix for our custom metrics.
* Deny everything not explicitly allowlisted.  This is to help manage egress cost.
* Add a tag to every gathered metric which includes WDS version.

Note: we're not actually tagging metrics with a workspace ID, as this
would be a high cardinality metric if aggregated globally.

Sample requests while running locally:

```
curl http://localhost:8080/prometheus

curl http://localhost:8080/prometheus?includedNames=jvm_memory_used_bytes%2Cjvm_memory_committed_bytes
```

[AJ-1500]: https://broadworkbench.atlassian.net/browse/AJ-1500?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/metrics/MetricsConfig.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/metrics/MetricsConfig.java
@@ -1,0 +1,63 @@
+package org.databiosphere.workspacedataservice.metrics;
+
+import static io.micrometer.core.instrument.config.MeterFilter.acceptNameStartsWith;
+import static io.micrometer.core.instrument.config.MeterFilter.deny;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import java.util.Set;
+import org.springframework.boot.actuate.autoconfigure.metrics.MeterRegistryCustomizer;
+import org.springframework.boot.info.BuildProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class MetricsConfig {
+  private final BuildProperties buildProperties;
+
+  public MetricsConfig(BuildProperties buildProperties) {
+    this.buildProperties = buildProperties;
+  }
+
+  @VisibleForTesting
+  static final Set<String> ALLOWED_PREFIXES =
+      // please keep this list alphabetized
+      ImmutableSet.of(
+          "application",
+          "cache",
+          "disk",
+          "executor",
+          "hikaricp",
+          "http",
+          "jdbc",
+          "jvm",
+          "logback",
+          "process",
+          "spring",
+          "system",
+          "tomcat",
+          "wds");
+
+  @Bean
+  MeterRegistryCustomizer<MeterRegistry> allowlistMetrics() {
+    return registry -> {
+      ALLOWED_PREFIXES.forEach(
+          prefix -> registry.config().meterFilter(acceptNameStartsWith(prefix)));
+      registry.config().meterFilter(deny()); // deny everything else
+    };
+  }
+
+  @Bean
+  MeterRegistryCustomizer<MeterRegistry> addWdsVersionTag() {
+    return registry ->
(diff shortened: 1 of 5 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/metrics/MetricsConfig.java
```diff
@@ -0,0 +1,63 @@
...
+  private final BuildProperties buildProperties;
+
+  public MetricsConfig(BuildProperties buildProperties) {
+    this.buildProperties = buildProperties;
+  }
+
+  @VisibleForTesting
+  static final Set<String> ALLOWED_PREFIXES =
+      // please keep this list alphabetized
+      ImmutableSet.of(
```
jladieu, line 28: This is pretty much everything; I'd be good with significantly slimming this list down to reduce overhead, but don't have a good sense of what is considered operationally valuable, so would welcome nudging.

service/src/main/resources/application.yml
```diff
@@ -23,14 +23,18 @@ management:
...
         liveness:
           include: livenessState,db
+    prometheus:
+      enabled: true
   endpoints:
     enabled-by-default: false
     web:
-      exposure.include: info,health
+      exposure:
+        include: "info,health,prometheus"
```
yuliadub, line 32: just making sure the following need quotes around them?

yuliadub (approved):

service/src/main/java/org/databiosphere/workspacedataservice/metrics/MetricsConfig.java
yuliadub: would having all these prefixes mean more data may end being sent that we will never look at?

jladieu: Yes, correct!  There is a whole bunch of stuff that comes for free.  The whole payload is in the low tens of kilobytes.  I think we can probably significantly pare down this list, but I wanted to get input first from anyone with opinions over what might be worthwhile to keep.

service/src/main/resources/application.yml
jladieu: They _may_ not need them, but the reference [documentation](https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html#actuator.endpoints.exposing) I was looking at had them quoted, so I decided to go with that.

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -507,6 +507,15 @@ paths:
...
           $ref: '#/components/responses/VersionResponseBody'
+  /prometheus:
+    get:
+      summary: App health metrics for Prometheus -- generated via Spring Boot Actuator (see https://docs.spring.io/spring-boot/docs/current/actuator-api/htmlsingle/#prometheus for details)
+      operationId: prometheusGet
+      tags:
+        - General WDS Information
+      responses:
+        200:
+          $ref: '#/components/responses/PrometheusResponseBody'
```
davidangb, line 518: adding the `/prometheus` endpoint to our OpenAPI docs means it will also be visible in the Java/Python clients. Do we want that? Especially since we haven't figured out auth, it might be best to leave this out - a "hidden" API (barely hidden).

davidangb (commented):
one question inline, otherwise looks good!

service/src/main/resources/static/swagger/openapi-docs.yaml
jladieu: Oh!  Good call.  I totally just copypasta'd this under the assumption that it was required boilerplate to make a new endpoint.  Didn't realize the OpenAPI stuff was orthogonal to what SpringBoot was doing.  Removed.

davidangb (approved):

PR #440, opened 2023-12-21
AJ-1157: Spring Boot 3.2.1

davidangb:
Updates WDS to Spring Boot 3.2.1. Supersedes #210.

Highlights of this PR:
* Update Spring Boot dependency to 3.2.1
* Update Data Repo and WSM clients to Jakarta-compatible versions
* Update a bunch of other deps to compatible versions
* Remove dependency overrides and version specifications where possible, since Spring Boot 3 [updates a lot of those for us](https://docs.spring.io/spring-boot/docs/current/reference/html/dependency-versions.html)
* bunch of small syntax compatibility changes due to the javax -> jakarta change and to Spring Boot 3 changes
* Refactor our logging config to include the default Spring Boot defaults; this fixes syntax errors with our previous config

Recommended Reading:
* https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Migration-Guide
* https://github.com/spring-projects/spring-framework/wiki/Upgrading-to-Spring-Framework-6.x
* https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Release-Notes
* https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.1-Release-Notes
* https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.2-Release-Notes

Still TODO:
- [x] a few unit tests are breaking, at least one because some error messages have changed. Address these.
- [x] fix Prometheus metrics

For future PRs:
- [ ] Re-enable [Problem Detail](https://github.com/spring-projects/spring-framework/issues/27052) support, which may have compatibility issues for clients
- [ ] stop publishing a javax client, once CBAS updates to Spring Boot 3 as well

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -56,17 +56,17 @@ dependencies {
     implementation 'org.springframework.boot:spring-boot-starter-actuator'
     implementation 'io.micrometer:micrometer-registry-prometheus'
     implementation 'org.springframework.integration:spring-integration-jdbc'
-    implementation 'org.aspectj:aspectjweaver:1.8.9' // required by spring-retry, not used directly
+    implementation 'org.aspectj:aspectjweaver' // required by spring-retry, not used directly
     implementation 'com.google.guava:guava:32.1.3-jre'
     implementation 'org.postgresql:postgresql'
     implementation 'org.webjars:webjars-locator-core' // versioned by spring dependency management
-    implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-csv:2.13.5'
+    implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-csv'
     implementation 'io.sentry:sentry-logback:6.34.0'
-    implementation 'org.liquibase:liquibase-core:4.25.1'
+    implementation 'org.liquibase:liquibase-core'
     implementation 'javax.cache:cache-api'
     implementation 'org.ehcache:ehcache:3.10.8:jakarta'
     implementation 'org.hashids:hashids:1.0.3'
-    implementation 'jakarta.ws.rs:jakarta.ws.rs-api:3.1.0'
+    implementation 'jakarta.ws.rs:jakarta.ws.rs-api'
     implementation 'com.google.mug:mug:6.6'
 
     // required by openapi-generated models and api interfaces
@@ -76,7 +76,7 @@ dependencies {
     // Terra libraries
     implementation group: 'org.broadinstitute.dsde.workbench', name: 'sam-client_2.13', version: '0.1-752b4f3'
     implementation group: 'org.broadinstitute.dsde.workbench', name: 'leonardo-client_2.13', version: '1.3.6-22ee00b'
-    implementation 'com.squareup.okhttp3:okhttp:4.12.0' // required by Sam client
+    implementation 'com.squareup.okhttp3:okhttp' // required by Sam client
     implementation "bio.terra:datarepo-jakarta-client:1.557.0-SNAPSHOT"
     implementation "bio.terra:workspace-manager-client:0.254.983-SNAPSHOT"
     implementation "bio.terra:java-pfb-library:0.15.0"
@@ -113,7 +113,7 @@ dependencies {
     annotationProcessor 'org.springframework.boot:spring-boot-configuration-processor'
     runtimeOnly 'org.webjars.npm:swagger-ui-dist:5.9.0'
     testImplementation 'org.springframework.boot:spring-boot-starter-test'
-    testImplementation 'org.junit.jupiter:junit-jupiter:5.8.2'
+    testImplementation 'org.junit.jupiter:junit-jupiter'
     testImplementation project(':client')
     testImplementation 'au.com.dius.pact.consumer:junit5:4.6.1'
     testImplementation 'org.junit-pioneer:junit-pioneer:2.1.0'
@@ -241,7 +241,7 @@ testing {
                 implementation project(':client')
                 implementation 'org.springframework.boot:spring-boot-starter-test'
                 implementation 'org.springframework.boot:spring-boot-starter-jdbc'
-                implementation 'org.liquibase:liquibase-core:4.25.1'
+                implementation 'org.liquibase:liquibase-core'
                 implementation "org.glassfish.jersey.core:jersey-client"
                 implementation "org.glassfish.jersey.core:jersey-common"
                 implementation "org.glassfish.jersey.inject:jersey-hk2"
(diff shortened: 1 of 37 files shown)


service/src/main/resources/application.yml
```diff
@@ -39,6 +39,10 @@ management:
   info:
     env:
       enabled: true
+  prometheus:
+    metrics:
+      export:
+        enabled: true
```
davidangb, line 45: https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Migration-Guide#actuator-metrics-export-properties

service/build.gradle
```diff
@@ -57,29 +56,29 @@ dependencies {
...
     implementation 'org.postgresql:postgresql'
     implementation 'org.webjars:webjars-locator-core' // versioned by spring dependency management
-    implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-csv:2.13.5'
+    implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-csv'
     implementation 'io.sentry:sentry-logback:6.34.0'
-    implementation 'org.liquibase:liquibase-core:4.21.1'
+    implementation 'org.liquibase:liquibase-core'
     implementation 'javax.cache:cache-api'
-    implementation 'org.ehcache:ehcache:3.10.8'
+    implementation 'org.ehcache:ehcache:3.10.8:jakarta'
```
davidangb, line 67: the ehcache version is managed by Spring Dependency Management, so ideally we'd leave the version number off here. However, it requires the `jakarta` classifier, and we can't specify a classifier without also specifying a version using this syntax, so I left the version in.

```
implementation group: 'org.ehcache', name: 'ehcache', classifier: 'jakarta'
```
would work, but results in IntelliJ complaining about the missing version.

service/src/main/java/org/databiosphere/workspacedataservice/controller/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,112 @@
...
+    // the real message we want to display. Gather all the nested messages and display the last one.
+    List<String> errorMessages = gatherNestedErrorMessages(ex, new ArrayList<>());
+    if (!errorMessages.isEmpty()) {
+      errorBody.put(MESSAGE, errorMessages.get(errorMessages.size() - 1));
+    } else {
+      errorBody.put(MESSAGE, "Unexpected error: " + ex.getClass().getName());
+    }
+
+    return ResponseEntity.badRequest().contentType(MediaType.APPLICATION_JSON).body(errorBody);
+  }
```
davidangb, line 100: In these methods, I wanted to use the `ErrorResponse` class instead of a raw `Map<String, Object>`. However, `ErrorResponse` does not serialize to json consistent with other classes. We could use its generated `.toJson()` method, but that uses gson instead of Jackson. Or, we could serialize it with Jackson …. but then, the `timestamp` field is serialized as a string instead of a Date, and we'd have to do our own custom Date formatting inside the exception handler. So, I'm sticking with the raw Map. Hopefully we upgrade to Problem Details soon and this code can mostly go away.

service/src/main/java/org/databiosphere/workspacedataservice/controller/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,112 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.ControllerAdvice;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.context.request.ServletWebRequest;
+import org.springframework.web.context.request.WebRequest;
+import org.springframework.web.method.annotation.MethodArgumentTypeMismatchException;
+import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExceptionHandler;
+
+@ControllerAdvice
+public class GlobalExceptionHandler extends ResponseEntityExceptionHandler {
```
jladieu, line 26: This looks like a lot of new/novel code -- was this logic handled somehow for us automatically by something that got removed in the upgrade to SpringBoot 3 or did I miss a move in the red parts of the diff?

jladieu (approved):
All looks great to me!  Nice work.

I don't think I understand what precipitated the introduciton of `GlobalExceptionHandler` but it looks fine to me either way.

jladieu:
LMK if I can help with Prometheus being busted!

service/src/main/java/org/databiosphere/workspacedataservice/controller/GlobalExceptionHandler.java
calypsomatic: I'm also confused what this is all about.  Did you hand write this code?

davidangb: apologies, I meant to document this in the PR!

Spring Boot 3 / Spring Framework 6 enables, by default, the [RFC 7807](https://datatracker.ietf.org/doc/html/rfc7807) "Problem Details for HTTP APIs" standard for structuring its error responses. In WDS today on Spring Boot 2.7, Spring uses the [`ErrorResponse`](https://github.com/DataBiosphere/terra-workspace-data-service/blob/b61b420df35e93c7a3df65cfde1aea0f6dd5e3a7/service/src/main/resources/static/swagger/openapi-docs.yaml#L864-L888) structure instead.

If we blindly accepted the change to use Problem Details responses, we'd break anyone using the Java or Python clients to handle error messages, and potentially (probably?) break Terra UI when it tries to parse error messages from WDS.

So, in this PR, I have set `mvc.problemdetails.enabled: false` in application.yml to maintain the previous behavior. But that's not sufficient, because in multiple places WDS's exception classes extend  `ResponseStatusException` (example [here](https://github.com/DataBiosphere/terra-workspace-data-service/blob/b61b420df35e93c7a3df65cfde1aea0f6dd5e3a7/service/src/main/java/org/databiosphere/workspacedataservice/service/model/exception/RestException.java#L6)). When those exceptions are serialized to the response, they serialize using the Problem Details structure. Therefore, I've also added this global exception handler to intercept any Problem Details being sent to the response and transform them to ErrorResponse instead.

You can see this in practice if you run this branch locally by turning off the global exception handler (comment out the `@ControllerAdvice` annotation) and doing something like sending an invalid UUID or an invalid version string in various APIs. You will also see a unit test fail because it asserts on the old error messages.

@calypsomatic yes, this is hand-written code.

calypsomatic (approved):

jladieu:
@davidangb I tried locally and prometheus LGTM too...I just noticed that you had left it in your TODOs and thought there might still be something up

jladieu (approved):

davidangb:
> I just noticed that you had left it in your TODOs and thought there might still be something up

@jladieu ah yeah - I have that still listed under TODO, but I ticked the box next to it to show it's complete. This a note-to-self technique I use when iterating on a PR, I realize in hindsight it is not obvious to others!

PR #444, opened 2024-01-05
AJ-1518 deal with empty file in parquet

calypsomatic:
https://broadworkbench.atlassian.net/browse/AJ-1518
Azure snapshot parquets have an "extra" empty file that throws an error when parsing.  To avoid this failed parsing attempt, this PR checks files to see if they have empty before attempting to parse.

Reminder:


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -151,45 +149,52 @@ private BatchWriteResult importTable(
...
-                            user.upserted()
-                                .record()
-                                .withRecordType(entry.getKey())
-                                .ofQuantity(entry.getValue()));
-                  });
+      // In the TDR manifest, for Azure snapshots only,
+      // the first file in the list will always be a directory. Attempting to import that directory
+      // will fail; it has no content. To avoid those failures,
+      // check files for length and ignore any that are empty
+      FileSystem fileSystem = FileSystem.get(new Configuration());
```
calypsomatic, line 156: Is this the best way to check file length?  This is just a suggestion the internet gave me.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -151,45 +149,52 @@ private BatchWriteResult importTable(
...
+      FileStatus fileStatus = fileSystem.getFileStatus(hadoopFilePath);
+      if (fileStatus.getLen() == 0) {
+        logger.info("Empty file in parquet, skipping");
+        return BatchWriteResult.empty();
```
davidangb, line 160: nit: move this 0-length check above the `// generate the HadoopInputFile` block - there's no need to create the `InputFile` if we're not going to use it.

Another nit: no need for the `else` here - it can just be:
```java
if (fileStatus.getLen() == 0) {
    return …;
}

… other code …
```

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
davidangb: This looks good to me, especially if it works :)

Might want to create a `Configuration()` separately, then reuse it for both the `FileSystem.get` and the ` HadoopInputFile.fromPath` - that way, if we find we need to make changes to the config, it'll take effect in both places.

davidangb (commented):
nice! minor comments inline.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJobTest.java
```diff
@@ -92,4 +101,67 @@ void extractSnapshotInfo() throws IOException {
...
+
+    // Make sure real errors on parsing parquets are not swallowed
+    assertThrows(
+        Exception.class,
+        () ->
+            tdrManifestQuartzJob.importTable(
+                manifestAzure.getURL(),
+                table,
+                workspaceId,
+                TwoPassStreamingWriteHandler.ImportMode.BASE_ATTRIBUTES));
```
davidangb, line 165: nit: assert this throws `TdrManifestImportException` instead of just `Exception` to make sure it's the exception we expect

davidangb (approved):
thank you! A couple nits inline, proactively approving.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -179,17 +188,11 @@ private BatchWriteResult importTable(
         return result;
       } catch (IOException e) {
         logger.error("Hit an error on file: {}", e.getMessage(), e);
-        return new BatchWriteResult(Map.of());
-        // throw new TdrManifestImportException(e.getMessage());
-        // TODO AJ-1518 more specific handling for 0-length directories; other errors should
-        //     be true failures
+        throw new TdrManifestImportException(e.getMessage());
```
calypsomatic, line 191: Just noticed that this error is caught and rethrown only to be caught and rethrown by the next catch statement.  Is there a better way to do this?

yuliadub: I think the first catch can now be removed since the empty file is handled -- and then the Throwable in the second catch can become IOException and that should be enough error handling? but I am not quite sure why both catches were there to be honest

yuliadub (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
calypsomatic: Hmm @davidangb did you have these separate just as an artifact of the empty file handling or do we want to have different try-catch loops (including the one after `InputFile inputFile` above)?  If we have different try-catches seems like we should distinguish that in the error messages as well.

PR #445, opened 2024-01-05
AJ-1520 - Add activity logging for imports

yuliadub:
Update: adjusting per David's [suggestion ](https://github.com/DataBiosphere/terra-workspace-data-service/pull/445#pullrequestreview-1809472497)to consolidate activity logging into a single place. 

_____
previous approach: 

Since an import can have many tables, as many relations, the approach I took with logging around if the import was successful or not is the following: 
- Have a logging event that import was started with specific snapshotId
- Log count of records of each of the tables that were actually imported (applies for both records and relations)
- Have a logging event that import was completed mentioning the same snapshotId. 

This means that to know if the import truly worked or not, you need to look for the "complete" activity logging statement. If you cant find it, you can assume that the import failed. My hope is that having the logging statements for what records were imported before the failure occurred could be beneficial when debugging the failure. 

Reminder:

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
@@ -102,14 +102,31 @@ public class TdrManifestQuartzJob extends QuartzJob {
         extractTableInfo(snapshotExportResponseModel);
 
     // loop through the tables to be imported and upsert base attributes
-    importTables(
-        tdrManifestImportTables,
-        targetInstance,
-        TwoPassStreamingWriteHandler.ImportMode.BASE_ATTRIBUTES);
+    var result =
+        importTables(
+            tdrManifestImportTables,
+            targetInstance,
+            TwoPassStreamingWriteHandler.ImportMode.BASE_ATTRIBUTES);
 
     // add relations to the existing base attributes
     importTables(
         tdrManifestImportTables, targetInstance, TwoPassStreamingWriteHandler.ImportMode.RELATIONS);
+
+    // activity logging for import status
+    // no specific activity logging for relations since main import is a superset
+    if (result != null) {
+      result
+          .entrySet()
+          .forEach(
+              entry -> {
+                activityLogger.saveEventForCurrentUser(
+                    user ->
+                        user.upserted()
+                            .record()
+                            .withRecordType(entry.getKey())
+                            .ofQuantity(entry.getValue()));
+              });
+    }
   }
 
   /**
@@ -160,22 +177,6 @@ public class TdrManifestQuartzJob extends QuartzJob {
             batchWriteService.batchWriteParquetStream(
                 avroParquetReader, targetInstance, table, importMode);
 
-        // activity logging
-        // TODO AJ-1520 this activity logging can be a "false positive" - we will log here,
-        //     but if the overall transaction fails and is rolled back these logs will be false
-        if (result != null) {
-          result
-              .entrySet()
-              .forEach(
-                  entry -> {
-                    activityLogger.saveEventForCurrentUser(
-                        user ->
-                            user.upserted()
-                                .record()
-                                .withRecordType(entry.getKey())
-                                .ofQuantity(entry.getValue()));
-                  });
-        }
         return result;
       } catch (IOException e) {
         logger.error("Hit an error on file: {}", e.getMessage(), e);
@@ -200,10 +201,12 @@ public class TdrManifestQuartzJob extends QuartzJob {
    * @param targetInstance instance into which to import
    * @param importMode mode for this invocation
    */
-  private void importTables(
+  private BatchWriteResult importTables(
       List<TdrManifestImportTable> importTables,
       UUID targetInstance,
       TwoPassStreamingWriteHandler.ImportMode importMode) {
+
+    var combinedResult = BatchWriteResult.empty();
     // loop through the tables that have data files.
     importTables.forEach(
         importTable -> {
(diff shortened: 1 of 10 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/pact/WsmPactTest.java
```diff
@@ -59,6 +61,7 @@ public class WsmPactTest {
   private static final int NUM_SNAPSHOTS_THAT_EXIST = 3;
   private static final int NUM_SNAPSHOTS_REQUESTED = NUM_SNAPSHOTS_THAT_EXIST + 2;
   private static final String BEARER_TOKEN = "fakebearertoken";
+  @MockBean ActivityLogger activityLogger;
```
yuliadub, line 64: Not sure if this is potentially screwing up pact tests? they didnt complain but want to make sure this is not completely wrong :D

davidangb (commented):
Would you consider an alternate approach to doing this?

Currently:
* `TdrManifestQuartzJob.executeInternal() calls `importTables()`
* `importTables` returns void, and calls `importTable` for each table in the TDR manifest
* `importTable` returns a `BatchWriteResult`, which is ignored.

My suggestion:
* `importTables`, before doing anything else, creates an "accumulator" `BatchWriteResult.empty()`
* every time `importTable` succeeds, append its result to the accumulator
* modify `importTables` to return the accumulator when it succeeds
* add activity logging into `executeInternal()`, and only log the result of the accumulator

I know that `executeInternal` calls `importTables` twice - once for base attributes, once for relations. I don't think that's too relevant - we can probably just log the result of the base attributes, since that would always be the superset.

LMK if you want to discuss!

yuliadub:
> Would you consider an alternate approach to doing this?
> 
> Currently:
> 
> * `TdrManifestQuartzJob.executeInternal() calls `importTables()`
> * `importTables` returns void, and calls `importTable` for each table in the TDR manifest
> * `importTable` returns a `BatchWriteResult`, which is ignored.
> 
> My suggestion:
> 
> * `importTables`, before doing anything else, creates an "accumulator" `BatchWriteResult.empty()`
> * every time `importTable` succeeds, append its result to the accumulator
> * modify `importTables` to return the accumulator when it succeeds
> * add activity logging into `executeInternal()`, and only log the result of the accumulator
> 
> I know that `executeInternal` calls `importTables` twice - once for base attributes, once for relations. I don't think that's too relevant - we can probably just log the result of the base attributes, since that would always be the superset.
> 
> LMK if you want to discuss!

This was the original approach I was considering, but then decided it wasn't worth to change all the function definitions just for activity logging. Thinking more it, it is much better to consolidate all events into one!

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -102,14 +102,31 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
         extractTableInfo(snapshotExportResponseModel);
 
     // loop through the tables to be imported and upsert base attributes
-    importTables(
-        tdrManifestImportTables,
-        targetInstance,
-        TwoPassStreamingWriteHandler.ImportMode.BASE_ATTRIBUTES);
+    var result =
```
yuliadub, line 105: likely a non issue but came to mind as I was making changes: if this result is quite large (for a large import), it will have to be held in memory when we do the relations, could potentially slow things down? but maybe I am overthinking this!

ashanhol: is there an opportunity to use a Stream?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -102,14 +102,31 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
...
     importTables(
         tdrManifestImportTables, targetInstance, TwoPassStreamingWriteHandler.ImportMode.RELATIONS);
+
+    // activity logging for import status
+    // no specific activity logging for relations since main import is a superset
+    if (result != null) {
+      result
+          .entrySet()
+          .forEach(
+              entry -> {
```
ashanhol, line 121: Is there a chance there could be a large amount of entries? Wondering if this has the ability to generate too many events and could we condense the info into one if so

ashanhol: i see that was code that was just moved from a different part so nvm if out of scope

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
yuliadub: previously the implementation kind of acted like a stream, but the problem was that we would get events even if import failed which is not the behavior that we want.

ashanhol (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
davidangb: This `BatchWriteResult` just holds a `Map<RecordType, Integer>` representing the tables being written and the count of rows written for each table. I don't anticipate this becoming a memory problem unless we were writing hundreds of thousands or millions of tables … in which case we likely have more pressing problems.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -219,9 +222,20 @@ private void importTables(
...
+                var result = importTable(path, importTable, targetInstance, importMode);
+
+                if (result != null) {
+                  result
+                      .entrySet()
+                      .forEach(
+                          entry -> {
+                            combinedResult.increaseCount(entry.getKey(), entry.getValue());
+                          });
+                }
```
davidangb, line 234: for nice OO encapsulation, it would be great to move this logic inside `BatchWriteResult`. Just like `BatchWriteResult` has `increaseCount(RecordType recordType, int count)`, it could also have a method that accepts an entire `BatchWriteResult` and combines the two. Something like this (totally untested!!!):

```java
  public void merge(BatchWriteResult other) {
    other.resultCounts.forEach((key, value) -> this.resultCounts.merge(key, value, Integer::sum));
  }
```

bonus: the `merge` method is nicely unit-testable!

davidangb (approved):
one nit and one comment about OO design inline. I'm giving this a thumb because the OO design stuff is just a nice-to-have; I am ok if this PR merges without that.

PR #447, opened 2024-01-08
AJ-1529: json parsing should be resilient to unknown properties

davidangb:
turns out that AJ-1529 was not a problem with the TDR client; it was a problem with our own JSON parsing. Our `ObjectMapper` was not configured to ignore unknown properties, so when TDR added new properties to the manifest and we tried to parse them into an older version of the model which didn't have those properties, we hit an error.

This PR fixes that by setting `DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES` to false, and adding a unit test.

The unit test is in `TdrManifestQuartzJobTest` because that's our driving use case, but the setting applies to all JSON parsing.


calypsomatic (approved):
Nice find!

jladieu (approved):
LGTM

PR #448, opened 2024-01-09
AJ-1517 download files only once

calypsomatic:
https://broadworkbench.atlassian.net/browse/AJ-1507
Although I couldn't find it explicitly stated in any documentation, it's strongly implied and empirically verified that avro's `ParquetReader` simply cannot handle URLs with query parameters.  So this PR still downloads parquets to a temporary file, but now does so ahead of time so that each file is downloaded only once. 

Todo:

- [x] Return a custom object instead of a directory path?
- [x] Validate with AppSec
- [x] Will there be problems downloading all parquet files at once?
- [x] Add retry for file downloads?
- [ ] Verify/allowlist URLs?

Reminder:


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -205,25 +190,75 @@ private BatchWriteResult importTables(
...
-                var result = importTable(path, importTable, targetInstance, importMode);
-
-                if (result != null) {
-                  combinedResult.merge(result);
+                try {
+                  // download the file from the URL to a temp file on the local filesystem
+                  // Azure urls, with SAS tokens, don't need any particular auth.
+                  File tempFile = File.createTempFile("tdr-", "download");
+                  logger.info("downloading to temp file {} ...", tempFile.getPath());
+                  FileUtils.copyURLToFile(path, tempFile);
```
jladieu, line 236: Coincidentally, I just completed server-side request forgery training, so I have a heightened awareness of downloading arbitrary URLs...one of the attack vectors that was suggested was to not properly lock down what scheme was used in URLs.  Do we know anything about the source location of the files being downloaded here that we can validate against?

Possible ideas:
- allowlist of domains where expect these files to live
- allowlist of schemes (eg: http / https) that we expect these URLs to be composed from

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -205,25 +190,75 @@ private BatchWriteResult importTables(
...
+    return combinedResult;
+  }
+
+  /**
+   * Given the list of tables/data files to be imported, loop through and import each one
+   *
+   * @param importTables tables to be imported
+   */
+  @VisibleForTesting
+  Map<String, List<InputFile>> prepareTableImport(List<TdrManifestImportTable> importTables) {
```
jladieu, line 213: Consider renaming this method from `prepare` to clarify that it's going to fetch a bunch of files.  Not sure what the better name might look like, but it'd be good to make it obvious that it's going to do a bunch of I/O.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -205,25 +190,75 @@ private BatchWriteResult importTables(
...
           paths.forEach(
               path -> {
-                var result = importTable(path, importTable, targetInstance, importMode);
-
-                if (result != null) {
-                  combinedResult.merge(result);
+                try {
+                  // download the file from the URL to a temp file on the local filesystem
+                  // Azure urls, with SAS tokens, don't need any particular auth.
+                  File tempFile = File.createTempFile("tdr-", "download");
```
jladieu, line 234:
```suggestion
                  File tempFile = File.createTempFile(/* prefix= */ "tdr-", /* suffix= */ "download");
```

jladieu: Multiple string parameters in a row can be hard to reason about on their own...when I read this I assumed it was just components of a filename, and also assumed that the file would be named `tdr-download`.  Some helper comments might demystify the params.

jladieu, line 234: How do the temp files get cleaned up?  Might be worth adding a code comment explaining their lifecycle.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -205,25 +190,75 @@ private BatchWriteResult importTables(
...
+
+                  // In the TDR manifest, for Azure snapshots only,
+                  // the first file in the list will always be a directory. Attempting to import
+                  // that directory
+                  // will fail; it has no content. To avoid those failures,
+                  // check files for length and ignore any that are empty
+                  FileSystem fileSystem = FileSystem.get(configuration);
+                  FileStatus fileStatus = fileSystem.getFileStatus(hadoopFilePath);
+                  if (fileStatus.getLen() == 0) {
+                    logger.info("Empty file in parquet, skipping");
```
jladieu, line 248: Would it be worth logging out the URL path here in the log statement so we know _which_ file is considered empty?

(Assuming the path includes enough information to tell us the original filename)

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -205,25 +190,75 @@ private BatchWriteResult importTables(
...
+                  } else {
+                    // generate the HadoopInputFile
+                    InputFile inputFile = HadoopInputFile.fromPath(hadoopFilePath, configuration);
+                    files.add(inputFile);
+                  }
+
+                } catch (IOException e) {
+                  throw new TdrManifestImportException(e.getMessage());
```
jladieu, line 256: can you pass `e` in as the second arg to `TdrManifestImportException` to rewrap it and preserve the stacktrace?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
calypsomatic: Ha I actually did already rename it just haven't pushed that yet.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -196,6 +180,7 @@ BatchWriteResult importTable(
    */
   private BatchWriteResult importTables(
       List<TdrManifestImportTable> importTables,
+      Map<String, List<InputFile>> fileMap,
```
jladieu, line 183: Purely for fun...`Map<K, List<V>>` is a good candidate to be replaced with a `Multimap<K, V>` ([ref](https://guava.dev/releases/19.0/api/docs/com/google/common/collect/Multimap.html)) which can make interacting with the data structure a little bit easier if you end up doing more complicated stuff with it.

That said, I don't think it would change/unlock/simplify anything in this method other than to move you to a less clunky/non-nested generic type in the method arg, so take with a grain of salt.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -205,25 +190,75 @@ private BatchWriteResult importTables(
...
+  }
+
+  /**
+   * Given the list of tables/data files to be imported, loop through and import each one
+   *
+   * @param importTables tables to be imported
+   */
+  @VisibleForTesting
+  Map<String, List<InputFile>> prepareTableImport(List<TdrManifestImportTable> importTables) {
+    Map<String, List<InputFile>> fileMap = new HashMap<>();
```
jladieu, line 214: Paired with my other comment...Map<K, List<V>> is a good candidate to be replaced with a Multimap<K, V> ([ref](https://guava.dev/releases/19.0/api/docs/com/google/common/collect/Multimap.html)).

In this case (unlike in the other spot where I had this suggestion), I think it actually might be helpful for this method impl; I'll add a couple other comments of where it might let you tweak things for the better.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -205,25 +190,75 @@ private BatchWriteResult importTables(
...
+          logger.info("Fetching files for table '{}' ...", importTable.recordType().getName());
+
           // find all Parquet files for this table
           List<URL> paths = importTable.dataFiles();
           logger.debug(
               "Table '{}' has {} export file(s) ...",
               importTable.recordType().getName(),
               paths.size());
 
+          List<InputFile> files = new ArrayList<>();
```
jladieu, line 227: You can drop this variable entirely with multimap

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -205,25 +190,75 @@ private BatchWriteResult importTables(
...
+                  // will fail; it has no content. To avoid those failures,
+                  // check files for length and ignore any that are empty
+                  FileSystem fileSystem = FileSystem.get(configuration);
+                  FileStatus fileStatus = fileSystem.getFileStatus(hadoopFilePath);
+                  if (fileStatus.getLen() == 0) {
+                    logger.info("Empty file in parquet, skipping");
+                  } else {
+                    // generate the HadoopInputFile
+                    InputFile inputFile = HadoopInputFile.fromPath(hadoopFilePath, configuration);
+                    files.add(inputFile);
```
jladieu, line 252: this would become `fileMap.add(importTable.recordType().getName(), inputFile);` if you switched it to a multimap

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
```diff
@@ -205,25 +190,75 @@ private BatchWriteResult importTables(
...
+                  }
+
+                } catch (IOException e) {
+                  throw new TdrManifestImportException(e.getMessage());
                 }
               });
+          fileMap.put(importTable.recordType().getName(), files);
```
jladieu, line 259: this line could go away if you switch `fileMap` to a `Multimap`

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJobTest.java
```diff
@@ -130,15 +133,9 @@ void parseEmptyParquet() throws IOException {
...
-            () ->
-                tdrManifestQuartzJob.importTable(
-                    emptyParquet.getURL(),
-                    table,
-                    workspaceId,
-                    TwoPassStreamingWriteHandler.ImportMode.BASE_ATTRIBUTES));
-    assertThat(actual.entrySet()).isEmpty();
+    Map<String, List<InputFile>> actual =
+        assertDoesNotThrow(() -> tdrManifestQuartzJob.prepareTableImport(List.of(table)));
+    assert (actual.get("data").isEmpty());
```
jladieu, line 138: Not sure, but using a Multimap might let you still do the syntactic-sugary assertThat(actual.get("data")).isEmpty()

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
calypsomatic: Good question!  Seems like they get cleaned up on JVM exit - which is not useful for us.  I guess I assume they were garbage collected, but since not, we should delete them manually.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
davidangb: note that this class already uses Multimap, as the return value from `identifyRelations`

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
davidangb, line 236: We probably want a retry (with backoff) for these downloads, for resilience against temporary network conditions. This could be a separate/follow-on PR.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
davidangb: There may be security considerations with logging the full original URL, since it's a signed URL - we could log a sanitized version of it.

davidangb (commented):
I'm disappointed (in avro-parquet/Hadoop, not you!) that we can't read these files directly over HTTP.

A few comments inline about temp dirs. I agree with Josh's comment about making sure these temp files get cleaned up.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -129,59 +138,39 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
                           .withRecordType(entry.getKey())
                           .ofQuantity(entry.getValue()));
             });
+    // delete temp files after everything else is completed
+    try {
+      FileUtils.deleteDirectory(fileDir.toFile());
```
yuliadub, line 143: is it possible that two separate people could import a file into the same workspace at the same time (ish) and the first import would delete some of the not completed second import files? (looks like this may be possible since we are storing all of it in the same temp directory - unless I am misreading, would it make sense to make the temp directory import file specific?

calypsomatic: I was worried about that, but the directory is actually not the same - it uses the parameter I pass to it to give it a prefix but also uses a random string to name the directory.

yuliadub: am I missing where you are doing that, sorry (my brain is failing me 😆 ) -- seems like the main directory is a hardcoded name:  https://github.com/DataBiosphere/terra-workspace-data-service/pull/448/files#diff-922f9c8ce63f97c9b9c2b988bf7f11cd174de5326a807c58fee214c553aae11dR253

calypsomatic: I don't do it explicitly, but the `createTempDirectory` method generates a random string on top of the name I pass it.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -219,15 +260,34 @@ private BatchWriteResult importTables(
...
+                  logger.info("downloading to temp file {} ...", tempFile.getPath());
+                  FileUtils.copyURLToFile(path, tempFile);
+                  // In the TDR manifest, for Azure snapshots only,
+                  // the first file in the list will always be a directory.
+                  // Attempting to import that directory
+                  // will fail; it has no content. To avoid those failures,
+                  // check files for length and ignore any that are empty
+                  if (tempFile.length() == 0) {
+                    logger.info("Empty file in parquet, skipping");
+                    Files.delete(tempFile.toPath());
```
calypsomatic, line 277: If this empty temp file fails to delete, then as currently written it'll fail the import.  Unsure how likely this is to happen, but if we want to avoid it I could:
- Put the deletion in its own try/catch
- Not delete the file since it's empty anyway and assume the accumulation of empty temp files will be small and not an issue
- Check for emptiness before creating the temp file, which I believe means making an extra http call to ask for the content length of the file located at the remote URL

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -106,16 +107,24 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
     List<TdrManifestImportTable> tdrManifestImportTables =
         extractTableInfo(snapshotExportResponseModel);
 
+    // get all the parquet files from the manifests
+    // TODO do we really want to download them all at once ahead of time?
+    Multimap<String, File> fileMap = getFilesForImport(tdrManifestImportTables);
```
calypsomatic, line 112: Not sure if there's a good way to avoid downloading all parquet files ahead of time, given that we're downloading them, but I don't know how large we expect them to be and if there might be any performance or memory issues with the pod.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/TdrManifestQuartzJob.java
calypsomatic: Without a custom object for holding the files and directory information, it seems easier to leave out the dedicated subdirectory.  Is there a big difference between setting permissions on each file and setting one on an entire directory?

calypsomatic:
Added custom download class in https://github.com/DataBiosphere/terra-workspace-data-service/pull/448/commits/71222902d3e1bf305d47a99e1bf4f3d65c89a18f .  Please to give opinion on whether this is better than without it.  I moved the actual downloading of the files in there as well, could then be a place to add in retries.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/FileDownload.java
```diff
@@ -0,0 +1,73 @@
...
+  private final Path tempFileDir;
+  private final Multimap<String, File> fileMap;
+  private final Set<PosixFilePermission> permissions =
+      EnumSet.of(
+          PosixFilePermission.OWNER_READ,
+          PosixFilePermission.GROUP_READ,
+          PosixFilePermission.OTHERS_READ);
+
+  public FileDownload(String dirName) throws IOException {
+    this.tempFileDir = Files.createTempDirectory(dirName);
```
jladieu, line 31: It doesn't look like `tempFileDir` is used yet...do you mean to put the tempFiles created on line 37 underneath this dir?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -106,16 +102,24 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
     List<TdrManifestImportTable> tdrManifestImportTables =
         extractTableInfo(snapshotExportResponseModel);
 
+    // get all the parquet files from the manifests
+
+    FileDownload fileMap = getFilesForImport(tdrManifestImportTables);
```
jladieu, line 107: nit: rename variable to better describe the type...in this case `FileDownload`, but I also wonder if that should be renamed to something less singular like `FileDownload{Helper,Support}` since it gets used for multiple files.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -106,16 +102,24 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
...
+
     // loop through the tables to be imported and upsert base attributes
     var result =
         importTables(
             tdrManifestImportTables,
+            fileMap,
             targetInstance,
             TwoPassStreamingWriteHandler.ImportMode.BASE_ATTRIBUTES);
```
jladieu, line 115: Can you fiddle with the imports to make this `ImportMode.BASE_ATTRIBUTES`?  I think the fully qualified class name isn't really necessary, although the `ImportMode` remains important.  Making this terser might let the code wrap onto one line (or be a little closer to being able to do so).

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/FileDownload.java
```diff
@@ -0,0 +1,73 @@
...
+import java.nio.file.attribute.PosixFilePermission;
+import java.util.Collection;
+import java.util.EnumSet;
+import java.util.Set;
+import org.apache.commons.io.FileUtils;
+import org.databiosphere.workspacedataservice.service.model.exception.TdrManifestImportException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class FileDownload {
```
jladieu, line 19: Is this class eligible to implement `AutoCloseable` such that the final teardown can happen automatically in a try with resources block?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -130,61 +134,38 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
                           .withRecordType(entry.getKey())
                           .ofQuantity(entry.getValue()));
             });
+    // delete temp files after everything else is completed
+    // Any failed deletions will be removed if/when pod restarts
+    fileMap.deleteFileDirectory();
```
jladieu, line 139: could this be eligible to be called `close()` and benefit from being an `AutoCloseable` try with resource block?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -200,6 +181,7 @@ BatchWriteResult importTable(
    */
   private BatchWriteResult importTables(
       List<TdrManifestImportTable> importTables,
+      FileDownload fileDir,
```
jladieu, line 184: I think this could just be the `Multimap` instead of the full `FileDownload` object.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/FileDownload.java
```diff
@@ -0,0 +1,73 @@
...
+          PosixFilePermission.OWNER_READ,
+          PosixFilePermission.GROUP_READ,
+          PosixFilePermission.OTHERS_READ);
+
+  public FileDownload(String dirName) throws IOException {
+    this.tempFileDir = Files.createTempDirectory(dirName);
+    this.fileMap = HashMultimap.create();
+  }
+
+  public void downloadFileFromURL(String tableName, URL pathToRemoteFile) {
```
jladieu, line 35: This class is juggling three different responsibilities (being a multimap data structure, dealing with temp directory shenanigans, and doing file downloads and permissions), so I'm thinking about ways you could break those responsibilities apart.

I have an idea to kick out the `Multimap` responsibility:

1) Change `downloadFileFromUrl` to return `Optional<File>` from this class.  The file would only be absent in the empty file scenario.

2) Move all the `Multimap` creation and management out of this class and into the caller.  This would let you eliminate the `isEmpty()` and `get()` methods, and leave this class with just two responsibilities (which maybe can be broken apart further after the dust settles).

jladieu: And here's my followup idea to eliminate the temp directory shenanigans.

Make the caller supply `Path fileDir` as a constructor arg.  Then this class just has the responsibility of downloading its stuff into that directory.  This class doesn't even have to know its a temporary, or that everything will be deleted some day.  Then the caller can decide when it's time to blow everything up.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -106,16 +102,24 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
     List<TdrManifestImportTable> tdrManifestImportTables =
         extractTableInfo(snapshotExportResponseModel);
 
+    // get all the parquet files from the manifests
+
```
jladieu, line 106: I think this class should be responsible for creating the temp directory, and passing the Path to that directory down through `getFilesForImport`

Then, later on, when this class wants to delete all the downloaded files, it can just delete the directory it already has a reference to.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/FileDownload.java
calypsomatic: Ah yeah I went back and forth on using the directory or not, will make sure that is all cleared up

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
calypsomatic: Thank you, I had no idea what to call this class

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/FileDownload.java
calypsomatic: That almost makes me think it's not necessary to have the class at all... originally the idea of the class was to hold both the multimap and the path to the directory so that I could know where to find the files, but also know which files were there so as to easily iterate over them.  Once the class existed, it seemed to make sense to give it some methods to deal with the file download.  I agree it's kind of a mess as written, but if all it does is download files it's not a big win, plus I still have to deal with keeping track of a multimap and a path to a directory separately.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
yuliadub: aha, that was the peice I was missing! thank you!

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/FileDownloadHelper.java
```diff
@@ -0,0 +1,94 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.retry.annotation.Backoff;
+import org.springframework.retry.annotation.Retryable;
+
+public class FileDownloadHelper {
+
+  private final DownloadHelper downloadHelper;
+
+  public interface DownloadHelper {
+    default void copyURLToFile(URL sourceUrl, File destinationFile) throws IOException {
```
calypsomatic, line 25: Needed this method to not be abstract in order to mock it

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/FileDownloadHelperTest.java
```diff
@@ -0,0 +1,67 @@
...
+  @Test
+  void testRetry() throws Exception {
+    FileDownloadHelper.DownloadHelper mockDownloadHelper =
+        Mockito.mock(FileDownloadHelper.DownloadHelper.class);
+    doThrow(new IOException("Simulated connection error"))
+        .doCallRealMethod() // Succeed on the second attempt
+        .when(mockDownloadHelper)
+        .copyURLToFile(any(URL.class), any(File.class));
+
+    // Create a RetryTemplate to set off Spring's retryable
```
calypsomatic, line 45: Spring's `@Retryable` wasn't working and among the several hideous options I could find to try to rig it to trigger this one seemed the least convoluted

yuliadub (approved):
looks like sonar is still unhappy about us writing to a "public" dir, but otherwise the changes make sense to me! 

still sad there isnt a straightforward way to stream this, not sure if we want to create a ticket to revisit in a bit (6 monts?) to see if anything changes

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/FileDownloadHelper.java
jladieu: Very weird, I wasn't expecting that

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/FileDownloadHelper.java
```diff
@@ -41,7 +45,14 @@ public FileDownloadHelper(String dirName, DownloadHelper downloadHelper) throws
   }
 
   public FileDownloadHelper(String dirName) throws IOException {
-    this(dirName, FileUtils::copyURLToFile);
+    this(
+        dirName,
+        new DownloadHelper() {
```
jladieu, line 50: I think this doesn't need the override anymore since the default impl does the right thing.  Just pass in `new DownloadHelper()`

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/FileDownloadHelper.java
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  private final Logger logger = LoggerFactory.getLogger(this.getClass());
+  private final Path tempFileDir;
+  private final Multimap<String, File> fileMap;
+  private final Set<PosixFilePermission> permissions =
+      EnumSet.of(
+          PosixFilePermission.OWNER_READ,
+          PosixFilePermission.GROUP_READ,
+          PosixFilePermission.OTHERS_READ);
```
davidangb, line 39: could this be locked down even further, to just `OWNER_READ`? Is there a need for group/others to also read?

jladieu (approved):
Bummer about `@Retryable` not doing the trick; I think that's worthy of a followup ticket to sort out as it'd be really nice to have access to those annotation based features.  I think @ashanhol encountered problems with the`@Observable` annotation too.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/FileDownloadHelperTest.java
davidangb: The fact that you need to include a `RetryTemplate` here is indication that `@Retryable` is not working correctly in `FileDownloadHelper`. From this test, you *should* be able to just call `helper.downloadFileFromURL()` outside of a `RetryTemplate` and it will work.

I am pretty certain the reason that `@Retryable` is not working is that `FileDownloadHelper` is not a Spring bean, and therefore Spring can't set up any proxying for it and therefore can't implement retries. To use `@Retryable` you'll have to lean into Spring-ness. I can see a few options:

1. `FileDownloadHelper` is a singleton bean (which is what we use everywhere else), and you move all statefulness about the temp dir into `DownloadHelper`
2. `FileDownloadHelper` is a [prototype bean](https://docs.spring.io/spring-framework/reference/core/beans/factory-scopes.html) which gets created on demand, including its temp-dir statefulness. You'll probably need a separate singleton factory bean to do the creation.

there are probably other ways to do it … 

( I could pair on this/contribute code if my explanations don't make sense)

calypsomatic:
Removed retry in https://github.com/DataBiosphere/terra-workspace-data-service/pull/448/commits/51e0682546b220ea4195ce11f436271a155b6972 and created https://broadworkbench.atlassian.net/browse/AJ-1555 to put it back and also take a look at possible memory issues.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/FileDownloadHelperTest.java
```diff
@@ -0,0 +1,23 @@
...
+import org.springframework.core.io.Resource;
+
+@SpringBootTest
+public class FileDownloadHelperTest {
+
+  @Value("classpath:parquet/empty.parquet")
+  Resource emptyParquet;
+
+  @Test
+  void downloadEmptyFile() throws IOException {
```
calypsomatic, line 18: This test is a duplicate of the test in `TdrManifestQuartzJobTest`.  This seems like the place for it, but should I alter the other test to verify the behavior of the actual `TdrManifestQuartzJob` with an empty file, or is this test sufficient?

davidangb: I think it's worth testing both `TdrManifestQuartzJob.getFilesForImport()` in `TdrManifestQuartzJobTest` and `FileDownloadHelper.downloadFileFromURL()` here, even though they're very similar.

davidangb (approved):
thaaaank you

PR #451, opened 2024-01-10
[AJ-1280] Add API for deleting an attribute from a record type

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1280

This adds an API that accepts delete requests at `{instanceId}/types/{v}/{type}/{attribute}` and deletes the given attribute from the given record type by dropping the column in the database. It rejects requests to delete the primary key column.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -1024,4 +1024,14 @@ public class RecordDao {
       throw e;
     }
   }
+
+  public void deleteAttribute(UUID instanceId, RecordType recordType, String attribute) {
+    namedTemplate
+        .getJdbcTemplate()
+        .update(
+            "alter table "
+                + getQualifiedTableName(recordType, instanceId)
+                + " drop column "
+                + quote(SqlUtils.validateSqlString(attribute, ATTRIBUTE)));
+  }
 }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
@@ -239,7 +239,7 @@ public class RecordOrchestratorService { // TODO give me a better name
     validateAttributeExistsAndIsEditable(instanceId, recordType, attribute);
     recordService.deleteAttribute(instanceId, recordType, attribute);
     activityLogger.saveEventForCurrentUser(
-        user -> user.updated().table().ofQuantity(1).withRecordType(recordType));
+        user -> user.deleted().attribute().withRecordType(recordType).withId(attribute));
   }
 
   private void validateAttributeExistsAndIsEditable(
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
@@ -238,6 +238,8 @@ public class RecordOrchestratorService { // TODO give me a better name
     checkRecordTypeExists(instanceId, recordType);
     validateAttributeExistsAndIsEditable(instanceId, recordType, attribute);
     recordService.deleteAttribute(instanceId, recordType, attribute);
+    activityLogger.saveEventForCurrentUser(
+        user -> user.updated().table().ofQuantity(1).withRecordType(recordType));
   }
 
   private void validateAttributeExistsAndIsEditable(
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
@@ -20,6 +20,7 @@ import org.databiosphere.workspacedataservice.service.model.DataTypeMapping;
 import org.databiosphere.workspacedataservice.service.model.RecordTypeSchema;
 import org.databiosphere.workspacedataservice.service.model.Relation;
 import org.databiosphere.workspacedataservice.service.model.exception.AuthorizationException;
+import org.databiosphere.workspacedataservice.service.model.exception.DeleteAttributeRequestException;
 import org.databiosphere.workspacedataservice.service.model.exception.MissingObjectException;
 import org.databiosphere.workspacedataservice.shared.model.Record;
 import org.databiosphere.workspacedataservice.shared.model.RecordQueryResponse;
@@ -229,6 +230,27 @@ public class RecordOrchestratorService { // TODO give me a better name
     recordService.deleteRecordType(instanceId, recordType);
     activityLogger.saveEventForCurrentUser(
         user -> user.deleted().table().ofQuantity(1).withRecordType(recordType));
+  }
+
+  public void deleteAttribute(
+      UUID instanceId, String version, RecordType recordType, String attribute) {
+    validateAndPermissions(instanceId, version);
+    checkRecordTypeExists(instanceId, recordType);
+    validateAttributeExistsAndIsEditable(instanceId, recordType, attribute);
+    recordService.deleteAttribute(instanceId, recordType, attribute);
+  }
+
+  private void validateAttributeExistsAndIsEditable(
+      UUID instanceId, RecordType recordType, String attribute) {
+    RecordTypeSchema schema = getSchemaDescription(instanceId, recordType);
+
+    if (schema.attributes().stream()
+        .noneMatch(attributeSchema -> attributeSchema.name().equals(attribute))) {
+      throw new MissingObjectException("Attribute");
+    }
+    if (attribute.equals(schema.primaryKey())) {
+      throw new DeleteAttributeRequestException("Unable to delete ID attribute");
+    }
   }
 
   @ReadTransaction
--- a/service/src/main/resources/capabilities.json
+++ b/service/src/main/resources/capabilities.json
@@ -1,5 +1,6 @@
 {
   "capabilities": true,
   "dataimport.pfb": true,
-  "dataimport.tdrmanifest": false
+  "dataimport.tdrmanifest": false,
+  "edit.deleteAttribute": true
 }
--- a/service/src/main/resources/static/swagger/openapi-docs.yaml
+++ b/service/src/main/resources/static/swagger/openapi-docs.yaml
@@ -483,6 +483,35 @@ paths:
           description: Success
         409:
           description: at least one of the records to be deleted is a relation target
+  /{instanceid}/types/{v}/{type}/{attribute}:
+    delete:
+      summary: Delete attribute from record type
+      description: Delete attribute from record type. This attribute will be removed from all records of this type.
+      operationId: deleteAttribute
+      security:
+        - bearerAuth: [ ]
+      tags:
+        - Schema
+      parameters:
+        - $ref: '#/components/parameters/instanceIdPathParam'
+        - $ref: '#/components/parameters/versionPathParam'
+        - $ref: '#/components/parameters/recordTypePathParam'
+        - $ref: '#/components/parameters/attributePathParam'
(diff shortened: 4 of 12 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -1971,6 +1971,54 @@ void deleteReferencedRecordTypeWithNoRecords() throws Exception {
...
+    createSomeRecords(recordType, 3);
+    mockMvc
+        .perform(
+            delete(
+                "/{instanceId}/types/{v}/{type}/{attribute}",
+                instanceId,
+                versionId,
+                recordType,
+                "attr1"))
+        .andExpect(status().isNoContent());
```
calypsomatic, line 1987: Do we want to fetch the schema and verify the attribute is gone?

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -483,6 +483,35 @@ paths:
           description: Success
         409:
           description: at least one of the records to be deleted is a relation target
+  /{instanceid}/types/{v}/{type}/{attribute}:
```
calypsomatic, line 486: Just double checking - should we be putting new APIs into the v1 swagger?

calypsomatic (approved):
Looks great!  Just a few non-blocking questions.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/451/commits/ea621b4235f6a8174ddefe98db30a903502d11a4.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -231,6 +232,29 @@ public void deleteRecordType(UUID instanceId, String version, RecordType recordT
...
   }
 
+  public void deleteAttribute(
+      UUID instanceId, String version, RecordType recordType, String attribute) {
+    validateAndPermissions(instanceId, version);
+    checkRecordTypeExists(instanceId, recordType);
+    validateAttributeExistsAndIsEditable(instanceId, recordType, attribute);
+    recordService.deleteAttribute(instanceId, recordType, attribute);
+    activityLogger.saveEventForCurrentUser(
+        user -> user.updated().table().ofQuantity(1).withRecordType(recordType));
```
davidangb, line 242: hm, I think we need to update the activity logger here to be more specific, and record that the user deleted an entire column, instead of logging that the user updated the table.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -1024,4 +1024,14 @@ public void deleteRecordType(UUID instanceId, RecordType recordType) {
       throw e;
     }
   }
+
+  public void deleteAttribute(UUID instanceId, RecordType recordType, String attribute) {
+    namedTemplate
+        .getJdbcTemplate()
+        .update(
```
jladieu, line 1031: Can you use the varargs update here?  I am guessing it's safer than doing string interpolation:
```suggestion
        .update("alter table ? drop column ?", getQualifiedTableName(recordType, instanceId), quote(SqlUtils.validateSqlString(attribute, ATTRIBUTE)))
```

jladieu, line 1031: Can you use the varargs update here?  I am guessing it's safer than doing string interpolation:
```suggestion
        .update("alter table ? drop column ?", getQualifiedTableName(recordType, instanceId), quote(SqlUtils.validateSqlString(attribute, ATTRIBUTE)))
```

jladieu, line 1031: Can you use the varargs update here?  I am guessing it's safer than doing string interpolation:
```suggestion
        .update("alter table ? drop column ?", getQualifiedTableName(recordType, instanceId), quote(SqlUtils.validateSqlString(attribute, ATTRIBUTE)))
```

nawatts: I can try that. I was following the example of other methods in this class.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -231,6 +232,29 @@ public void deleteRecordType(UUID instanceId, String version, RecordType recordT
...
+  }
+
+  private void validateAttributeExistsAndIsEditable(
+      UUID instanceId, RecordType recordType, String attribute) {
+    RecordTypeSchema schema = getSchemaDescription(instanceId, recordType);
+
+    if (schema.attributes().stream()
+        .noneMatch(attributeSchema -> attributeSchema.name().equals(attribute))) {
+      throw new MissingObjectException("Attribute");
+    }
```
davidangb, line 252: minor nit: `schema.attributes().stream().findAny(…)` will short-circuit as soon as it matches, and thus is more performant than `schema.attributes().stream().noneMatch(…)`, which by definition must check all elements in the stream.

In this use case, it's almost completely irrelevant, since record types won't have a large enough number of attributes to make a real difference. But, I always want to call these out so we don't adopt too many bad habits in our code - using noneMatch when streaming through a large PFB, for instance, could be an issue. So, your call on changing this PR or not; it's not critical in this case.

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
```diff
@@ -127,6 +130,63 @@ void deleteRecordType() {
...
+    assertAttributes(Set.of("id", "attr1"));
+  }
+
+  @Test
+  void deletePrimaryKeyAttribute() {
+    // Arrange
+    setUpDeleteAttributeTest();
+
+    // Act/Assert
+    assertThrows(
```
jladieu, line 151: When it's possible for an exception of the caught type (`DeleteAttributeRequestException`) to be thrown for multiple reasons, I think it's a good idea to also assert that the message (and therefore the reason) matches what you're expecting.

I think you can use this approach.
```suggestion
    var thrown = assertThrows(...)
    assertThat(var.getMessage()).contains("Unable to delete ID");
```

service/src/main/resources/capabilities.json
```diff
@@ -1,5 +1,6 @@
 {
   "capabilities": true,
   "dataimport.pfb": true,
-  "dataimport.tdrmanifest": false
+  "dataimport.tdrmanifest": false,
+  "edit.deleteAttribute": true
```
davidangb, line 5: nice!

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
```diff
@@ -127,6 +130,63 @@ void deleteRecordType() {
...
+  }
+
+  private void setUpDeleteAttributeTest() {
+    RecordRequest recordRequest =
+        new RecordRequest(
+            RecordAttributes.empty().putAttribute("attr1", "foo").putAttribute("attr2", "bar"));
+
+    ResponseEntity<RecordResponse> response =
+        recordOrchestratorService.upsertSingleRecord(
+            INSTANCE, VERSION, TEST_TYPE, "1", Optional.of("id"), recordRequest);
```
davidangb, line 179: for good measure, toss in `assertAttributes(Set.of("id", "attr1", "attr2"));` here … ensure the tests are set up correctly before performing the deletions?

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
davidangb: SQL doesn't accept bind parameters in certain positions - I believe table names and column names do not support bind params. I agree with Josh that we should _always_ prefer params instead of string interpolation - but I don't think it'll work here. That said, give it a try, and if you can get it to work, great!

davidangb (commented):
one important comment about the activity logger, a few other minor comments. Looking good!

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
nawatts: `getQualifiedTableName` includes quotes, which I'm guessing would interfere with using it as a parameter.

I'm wondering how to pass both the schema and table name. I tried:
```
.update(
            "alter table ?.? drop column ?",
            instanceId.toString(),
            recordType.getName(),
            SqlUtils.validateSqlString(attribute, ATTRIBUTE));
```

And got an error:
```
org.springframework.jdbc.BadSqlGrammarException: PreparedStatementCallback; bad SQL grammar
```

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
nawatts: `findAny` doesn't accept a predicate. I originally has `!schema.attributes().stream().anyMatch(...)`, but Intellij suggested replacing it with `noneMatch`. The docs say that `noneMatch` also short circuits.
https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#noneMatch-java.util.function.Predicate-

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
jladieu: I'd say don't hold off on blocking this PR based on this.  We have precedent to do it the way you did it...so if it just doesn't work for _reasons_ I'm ok with accepting it and moving on for the time being.

jladieu (approved):

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
nawatts: Oops. Misunderstood the "message" parameter of `assertThrows` here. Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/451/commits/64ce7dd0a5b1e6ab8b42b2925e53348d2c8aa173.

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/451/commits/0458f9baefa0c46ab9cbfa4ad7b161e15f4e5d17.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
nawatts: Do you want to add another type of action here (as opposed to "updated") or extend activity events to have sub-types of actions?

nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/451/commits/0c534b1fcd9c1ec4bf05f8264f46c1038fa076e0.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
davidangb: you're right! I stand corrected - of course `noneMatch` short-circuits, that makes sense.

PR #452, opened 2024-01-10
Restructure src/test/resources

jladieu:
This is in prep for being able to add/use test resources as part of enhancing test coverage with [AJ-1523](https://broadworkbench.atlassian.net/browse/AJ-1523).

* Nest like files in subdirectories that logically group them.
* Rename some files to eliminate parts of the filename which would have been redundant with the directory name.
* All changes were made using the IDE refactor/move or refactor/rename functions.

[AJ-1523]: https://broadworkbench.atlassian.net/browse/AJ-1523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service/src/test/java/org/databiosphere/workspacedataservice/controller/TsvDownloadTest.java
```diff
@@ -121,7 +121,7 @@ void tsvUploadWithChosenPrimaryKeyFollowedByDownload(String primaryKeyName) thro
   void batchWriteFollowedByTsvDownload() throws IOException {
     RecordType recordType = RecordType.valueOf("bar");
 
-    InputStream is = TsvDownloadTest.class.getResourceAsStream("/batch_write_tsv_data.json");
+    InputStream is = TsvDownloadTest.class.getResourceAsStream("/batch-write/write-tsv-data.json");
```
nawatts, line 124: It looks like most files are organized by format. But this one looks like it's organized by API/function. Could that lead to cases where it's unclear where a file should go? Like what if batch write can accept a TSV?

nawatts (approved):

davidangb (approved):
:+1:, assuming all tests pass

service/src/test/java/org/databiosphere/workspacedataservice/controller/TsvDownloadTest.java
jladieu: Yes, totally.  I had the same thought.  That said...I think JSON is a bit of an outlier on formats, since it's used as both our API format as well as data formats, so I didn't think it'd make sense to have a JSON folder which might eventually fill up with both requests, responses, and data files.

I'm hoping if the next person who comes along to add some stuff encounters the problem you're describing that they can deal with figuring that out :D

PR #454, opened 2024-01-11
[AJ-1532] Sanitize null characters from Postgres error message

ashanhol:
https://broadworkbench.atlassian.net/browse/AJ-1532

When feeding a malformed parquet file to dataImport, the job would fail to terminate correctly. This is due to the fact the error message we fed the jobDao included the malformed string, which in our bug example included null bytes. That would error our Postgres upsert, so the job wouldn't get terminated to failed correctly and remain "running". 

This PR does a string replace in the error message replacing null bytes with "". 

I originally attempted to solve this by converting the error string to a UTF8 byte array and back, but that didn't seem to actually get rid of the null characters. Theoretically we'd want to make sure all of our strings were sanitized, but it seems an exception message is the most likely place we'd run into that. 

------------
Reminder:


davidangb:
Where is the original exception being thrown? If we can catch it closer to the source, I think it's more likely we can handle the problem gracefully.

calypsomatic (approved):
Are there any other places in the code we might run into something like this where we'd want to put the same fix?

nawatts (approved):

davidangb (commented):
Couple questions: 

1. if this error message contains the null bytes from the Parquet file itself, it has the possibility of echoing sensitive user data - maybe we want to mask the entire error message. Can you provide an example of the error message in question? Is it useful, and is there anything else we can display instead?
2. Do we need a fallback/resilient code path in PostgresJobDao when we update the job's status, in cases where updating the job's status failed?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -239,8 +239,9 @@ SnapshotExportResponseModel parseManifest(URL manifestUrl) {
     try {
       return mapper.readValue(manifestUrl, SnapshotExportResponseModel.class);
     } catch (IOException e) {
+      // Sanitize potential null characters that caused the mapper to fail
       throw new JobExecutionException(
-          "Error reading TDR snapshot manifest: %s".formatted(e.getMessage()), e);
+          "Error reading TDR snapshot manifest: %s".formatted(e.getMessage().replace("\0", "")), e);
```
davidangb, line 244: what about using `StringEscapeUtils.escapeJson` here instead? That'll need a `import org.apache.commons.text.StringEscapeUtils;`. There may be other methods in `StringEscapeUtils` that work as well.

ashanhol:
@davidangb replied in slack but can post here for documentation. 
1. The error message is `Error reading TDR snapshot manifest: Unrecognized token 'PAR1PPL'` . That could have the possibility of being something sensitive given on how the objectmapper errors. The exception is a DataIntegrityViolationException, we could specifically check for that and just print `Error reading TDR snapshot manifest from unrecognized token`, but that wouldn't help anyone debug what failed in the file. 
2. We could do a try/catch around  namedTemplate.update(sb.toString(), params); that calls update again with a new error message and throws an exception?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
davidangb: Alternatively, we could look for the "Unrecognized token" you mentioned, and do something like:
```java
  SnapshotExportResponseModel parseManifest(URL manifestUrl) {
    // read manifest
    try {
      return mapper.readValue(manifestUrl, SnapshotExportResponseModel.class);
    } catch (IOException e) {
      String errMsg = e.getMessage();
      if (e.getMessage().startsWith("Unrecognized token")) {
        // no need to output the token itself
        errMsg = "Unrecognized token";
      }
      throw new JobExecutionException(
          "Error reading TDR snapshot manifest: %s".formatted(errMsg), e);
    }
  }
```

davidangb (commented):
> The error message is Error reading TDR snapshot manifest: Unrecognized token 'PAR1���P�PL�����' . That could have the possibility of being something sensitive given on how the objectmapper errors. The exception is a DataIntegrityViolationException, we could specifically check for that and just print Error reading TDR snapshot manifest from unrecognized token, but that wouldn't help anyone debug what failed in the file.

I'm waffling a bit here, it feels awkward to potentially be outputting a whole bunch of gibberish to the end user (we could log it separately for debugging purposes). I'm ok leaving this in the output, until/unless we see it become a problem in practice.

> We could do a try/catch around namedTemplate.update(sb.toString(), params); that calls update again with a new error message and throws an exception?

yeah, I think we might want that - I am fine saving that for a follow-on PR with its own unit tests and all that.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
ashanhol: Tagging @jladieu since he discovered a way we could end up in this state

Not sure if we could face multiple scenarios where we could end up with file corruption gibberish. An incorrectly formatted parquet file would also throw an exception and it wouldn't necessarily be gibberish. If my import was failing saying there was an "unrecognized token" but it didn't tell me what it was, I personally would be a bit annoyed since I couldn't fix it. 

If Josh has a fix that solidly prevents this scenario though, we can just close this PR and keep the error message as is.

PR #455, opened 2024-01-11
[AJ-1282] Add API for renaming an attribute

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1282

This adds an API that accepts patch requests at `{instanceId}/types/{v}/{type}/{attribute}` and renames the given attribute from the given record type by altering the database table. It rejects requests to rename the primary key column and requests where an attribute with the new name already exists.


service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
```diff
@@ -145,22 +211,22 @@ void deleteAttribute() {
...
+    setUpRenameOrDeleteAttributeTest();
 
     // Act/Assert
     DeleteAttributeRequestException e =
         assertThrows(
             DeleteAttributeRequestException.class,
             () -> recordOrchestratorService.deleteAttribute(INSTANCE, VERSION, TEST_TYPE, "id"),
             "deleteAttribute should have thrown an error");
-    assertEquals(e.getMessage(), "Unable to delete ID attribute");
+    assertEquals("Unable to delete ID attribute", e.getMessage());
```
nawatts, line 222: After I merged the last PR, I noticed SonarCloud was telling me the order of these arguments were backwards. The expected value should be passed first.

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -687,6 +730,14 @@ components:
         relatesTo:
           type: string
           description: Name of type to which this attribute relates. Only present if this is a relation attribute.
+    AttributeUpdateRequest:
```
nawatts, line 733: My thinking here is that the same API endpoint will be used for editing an attribute data type. When that's available, this would let the user either specify a new name _or_ a new data type (or maybe both?).

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -167,6 +169,28 @@ public ResponseEntity<Void> deleteRecordType(
...
+      @PathVariable("type") RecordType recordType,
+      @PathVariable("attribute") String attribute,
+      @RequestBody AttributeUpdateRequest attributeUpdateRequest) {
+    String newAttributeName = attributeUpdateRequest.name();
+    recordOrchestratorService.renameAttribute(
+        instanceId, version, recordType, attribute, newAttributeName);
+
+    RecordTypeSchema recordTypeSchema =
+        recordOrchestratorService.describeRecordType(instanceId, version, recordType);
+    AttributeSchema attributeSchema =
```
jladieu, line 186: Curious about why this check is needed...In what circumstance would everything before this line succeed, but this line fails?  I'm worried about things being left in a weird / partial / incomplete state if that ever actually happens to necessitate this check.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -232,18 +234,46 @@ public void deleteRecordType(UUID instanceId, String version, RecordType recordT
...
+  private void validateRenameAttribute(
+      UUID instanceId, RecordType recordType, String attribute, String newAttributeName) {
+    RecordTypeSchema schema = getSchemaDescription(instanceId, recordType);
+
+    if (schema.attributes().stream()
+        .noneMatch(attributeSchema -> attributeSchema.name().equals(attribute))) {
+      throw new MissingObjectException("Attribute");
+    }
+    if (attribute.equals(schema.primaryKey())) {
+      throw new UpdateAttributeRequestException("Unable to rename ID attribute");
```
jladieu, line 258: Will "ID" always be the name of the attribute?  Otherwise it might be better to include the actual `schema.primaryKey` in the message.

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
nawatts: What's the "check" you're referring to? `orElseThrow`? If so, `orElseThrow` shouldn't throw since the new attribute should be present in the list. But `orElseThrow` seemed preferable to `get`, which produces Intellij warnings.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
nawatts: No. I was using "ID" as a generic term for "primary key". But yeah, it would be better to either say "primary key" or the actual attribute name.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -232,18 +234,46 @@ public void deleteRecordType(UUID instanceId, String version, RecordType recordT
...
+
+  private void validateRenameAttribute(
+      UUID instanceId, RecordType recordType, String attribute, String newAttributeName) {
+    RecordTypeSchema schema = getSchemaDescription(instanceId, recordType);
+
+    if (schema.attributes().stream()
+        .noneMatch(attributeSchema -> attributeSchema.name().equals(attribute))) {
+      throw new MissingObjectException("Attribute");
+    }
+    if (attribute.equals(schema.primaryKey())) {
```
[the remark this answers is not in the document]
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/455/commits/ae1d5a0953dce6db3e43fa19f86c2d138cab9a76.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
nawatts: Replaced "ID" with "primary key" in https://github.com/DataBiosphere/terra-workspace-data-service/pull/455/commits/684acdf28113863b677a226c8e5bda20c05d17a6.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -232,27 +234,55 @@ public void deleteRecordType(UUID instanceId, String version, RecordType recordT
...
+  public void renameAttribute(
+      UUID instanceId,
+      String version,
+      RecordType recordType,
+      String attribute,
+      String newAttributeName) {
+    validateAndPermissions(instanceId, version);
+    checkRecordTypeExists(instanceId, recordType);
+    validateRenameAttribute(instanceId, recordType, attribute, newAttributeName);
+    recordService.renameAttribute(instanceId, recordType, attribute, newAttributeName);
```
yuliadub, line 246: do we need to log an attribute rename to the activity logger? its not technically changing the data itself but it is changing data properties

nawatts: Oops, forgot to add that.

I wonder what the best way to log it would be. Two events: delete the old attribute and add the new attribute? Or one rename event with both attribute names (that seems like kind of a misuses of the IDs field of activity events).

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -484,6 +484,42 @@ paths:
...
+        - $ref: '#/components/parameters/versionPathParam'
+        - $ref: '#/components/parameters/recordTypePathParam'
+        - $ref: '#/components/parameters/attributePathParam'
+      requestBody:
+        $ref: '#/components/requestBodies/AttributeUpdateRequestBody'
+      responses:
+        204:
+          description: Success
+        400:
+          description: Attribute is the ID for record type
```
yuliadub, line 506: does ID imply primary key?

nawatts: Yes, but it should really match the terms used elsewhere. Changed "ID" to "primary key" in some error messages in https://github.com/DataBiosphere/terra-workspace-data-service/pull/455/commits/684acdf28113863b677a226c8e5bda20c05d17a6. Will change it here as well.

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -484,6 +484,42 @@ paths:
...
+        - bearerAuth: [ ]
+      tags:
+        - Schema
+      parameters:
+        - $ref: '#/components/parameters/instanceIdPathParam'
+        - $ref: '#/components/parameters/versionPathParam'
+        - $ref: '#/components/parameters/recordTypePathParam'
+        - $ref: '#/components/parameters/attributePathParam'
+      requestBody:
+        $ref: '#/components/requestBodies/AttributeUpdateRequestBody'
```
davidangb, line 501: Could this accept the existing `AttributeSchema` model as its payload? This would keep our APIs consistent - you could retrieve an `AttributeSchema` from a GET request, modify it, and reupload it to this API. It would mean you'd have to add validation now that disallows changes to the `datatype` or `relatesTo` fields, but it would make this API forward-compatible to when we do allow those.

nawatts: That seems like it would make the API more difficult to use, since a client would have to send all the parts of the existing attribute schema regardless of whether they wanted to change them.

service/src/main/resources/static/swagger/openapi-docs.yaml
nawatts: Changed in https://github.com/DataBiosphere/terra-workspace-data-service/pull/455/commits/97b1817b3964a7ee869954dc997f8a833c0e5f63.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
nawatts: Added in https://github.com/DataBiosphere/terra-workspace-data-service/pull/455/commits/83847ffc36ec2299b5df78936be429009fbc3a3e.

service/src/main/resources/static/swagger/openapi-docs.yaml
nawatts: More generally, do we favor put or patch APIs? It looks like the only update API we currently have is for records, where we offer both put and patch.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
jladieu: Ahh 💡   I get it.  Carry on....

service/src/main/resources/static/swagger/openapi-docs.yaml
davidangb: For an API like this, where we're allowing partial modifications, I agree that PATCH is the right thing to do. And I agree that we shouldn't force users to specify all of the existing attribute schema , if they're only changing one of those. I am agreeing with you on that!

What I'm pushing for here is avoiding proliferation of model classes and API payload definitions. Is there a way to reuse the models and payloads we already have? For instance, could we just change `AttributeSchema` so its `datatype` property is not required, use it here, and then in the Java code get rid of the `AttributeUpdateRequest` model and use `AttributeSchema`? I think this would simplify the codebase overall.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -1974,6 +1975,116 @@ void deleteReferencedRecordTypeWithNoRecords() throws Exception {
...
+                .contentType(MediaType.APPLICATION_JSON)
+                .content(mapper.writeValueAsString(new AttributeUpdateRequest(newAttributeName))))
+        .andExpect(status().isOk());
+
+    mockMvc
+        .perform(
+            get("/{instanceId}/types/{version}/{recordType}", instanceId, versionId, recordType))
+        .andExpect(content().string(not(containsString(attributeToRename))))
+        .andExpect(content().string(containsString(newAttributeName)));
+  }
```
davidangb, line 2023: It's more precise to return the mockMvc result as a Java object - otherwise this test could hit false positives/negatives if the underlying data changed to have attributes that were substrings of each other, like "foo" and "foo1". Use `.andReturn` to do this, a la:
```java
    // use .andReturn() to get the result
    MvcResult mvcResult =
        mockMvc
            .perform(
                get(
                    "/{instanceId}/types/{version}/{recordType}",
                    instanceId,
                    versionId,
                    recordType))
            .andReturn();

    // parse the result into the type we expect
    RecordTypeSchema actual =
        mapper.readValue(mvcResult.getResponse().getContentAsString(), RecordTypeSchema.class);

    // extract the set of attribute names for easier processing
    Set<String> attributeNames =
        actual.attributes().stream().map(AttributeSchema::name).collect(Collectors.toSet());

    assertThat(attributeNames).doesNotContain(attributeToRename);
    assertThat(attributeNames).contains(newAttributeName);
```

davidangb (commented):
couple tweaks inline!

service/src/main/java/org/databiosphere/workspacedataservice/service/model/exception/AttributeExistsException.java
```diff
@@ -0,0 +1,11 @@
+package org.databiosphere.workspacedataservice.service.model.exception;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(code = HttpStatus.CONFLICT)
+public class AttributeExistsException extends UpdateAttributeRequestException {
```
jladieu, line 7: I'm not entirely opposed to adding explicit `AttributeExistsException` and `UpdateAttributeRequestException`, but I also think they don't add a lot of extra value and may potentially create clutter in the exception package.

I think it would be equally appropriate to just throw a more generic base class or framework-provided exception (which is already mapped to the desired HTTP code) with a custom error message.  Then in your tests, just assert that the error message is what you expect.

I don't know if I have the mappings exactly correct here, but I feel like `IllegalStateException` might map well to `CONFLICT`, and `IllegalArgumentException` might map well to `BAD_REQUEST`.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -1974,6 +1975,116 @@ void deleteReferencedRecordTypeWithNoRecords() throws Exception {
         .andExpect(status().isConflict());
   }
 
+  @Test
+  @Transactional
+  void updateAttributeNoUpdate() throws Exception {
+    String recordType = "recordType";
+    createSomeRecords(recordType, 3);
+    String attributeToUpdate = "attr1";
```
jladieu, line 1983: Since this variable is only used once, consider inlining it.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -1974,6 +1975,116 @@ void deleteReferencedRecordTypeWithNoRecords() throws Exception {
...
+        .andExpect(content().string(not(containsString(attributeToRename))))
+        .andExpect(content().string(containsString(newAttributeName)));
+  }
+
+  @Test
+  @Transactional
+  void renameNonExistentAttribute() throws Exception {
+    String recordType = "recordType";
+    createSomeRecords(recordType, 3);
+    String attributeToRename = "doesNotExist";
```
jladieu, line 2030: Since these variables are only used once, consider inlining.  Ok if you want to leave them as is for consistency with the other tests though.

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/AttributeUpdateRequest.java
```diff
@@ -0,0 +1,5 @@
+package org.databiosphere.workspacedataservice.shared.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public record AttributeUpdateRequest(@JsonProperty("name") String name) {}
```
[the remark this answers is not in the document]
davidangb: Ignoring the question of using `AttributeUpdateRequest` vs `AttributeSchema` - if these were v1 APIs in`apis-v1.yaml`, all the models would be auto-generated. Since these are v0.2 APIs in `openapi-docs.yaml`, we do need to hand-write them.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -1974,6 +1975,116 @@ void deleteReferencedRecordTypeWithNoRecords() throws Exception {
...
+                    attributeToRename)
+                .contentType(MediaType.APPLICATION_JSON)
+                .content(mapper.writeValueAsString(new AttributeUpdateRequest(newAttributeName))))
+        .andExpect(status().isNotFound());
+  }
+
+  @Test
+  @Transactional
+  void renamePrimaryKeyAttribute() throws Exception {
+    String recordType = "recordType";
```
jladieu, line 2049: I can't tell from reading this test which of the attributes is considered the primary key.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -1974,6 +1975,116 @@ void deleteReferencedRecordTypeWithNoRecords() throws Exception {
...
+                    attributeToRename)
+                .contentType(MediaType.APPLICATION_JSON)
+                .content(mapper.writeValueAsString(new AttributeUpdateRequest(newAttributeName))))
+        .andExpect(status().isBadRequest());
+  }
+
+  @Test
+  @Transactional
+  void renameAttributeConflict() throws Exception {
+    String recordType = "recordType";
```
jladieu, line 2070: I can't tell from reading the test _why_ there is a conflict.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -1974,6 +1975,116 @@ void deleteReferencedRecordTypeWithNoRecords() throws Exception {
         .andExpect(status().isConflict());
   }
 
+  @Test
+  @Transactional
+  void updateAttributeNoUpdate() throws Exception {
```
jladieu, line 1980: I'm not sure I understand what 'no update' means.    Is it that the JSON body is empty?

Devil's advocate, maybe this should be okay, but print a warning on the backend since it might indicate something is fishy?

I would compare this to a case where you update an attribute but don't actually end up change its value.  I'd rather that be an idempotent operation that doesn't have side effects.

jladieu (approved):
I left some thoughts/comments/suggestions that can all be treated as optional.  Provided you can address @davidangb's feedback, I'm happy to provide a pre-emptive 👍

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
nawatts: Yes, that's the default name for the primary key column used by the existing test setup method `createSomeRecords`.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
nawatts: Yeah, the attribute names come from `generateRandomAttributes` via `createSomeRecords`.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
nawatts: Yes, this is testing a request body containing an empty JSON object. This was kind of thinking ahead to when edit data type is added and we'll probably have to validate the presence of keys in the request body in the controller instead of relying on Spring.

I think this should not be logged on the server, since it's a client/user error.

service/src/main/resources/static/swagger/openapi-docs.yaml
nawatts: Is there a way in the OpenAPI configuration to define an object that extends another?

If we mark the fields optional in AttributeSchema, that makes our documentation for response formats less accurate.

davidangb: > Is there a way in the OpenAPI configuration to define an object that extends another?

Yes, and we're using it in the newer v1 APIs; check out GenericJob for instance.  https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/ has the details (I find it to be ugly code, but it all makes sense)

nawatts: It doesn't look like there's a way to override properties of a schema.

service/src/main/java/org/databiosphere/workspacedataservice/service/model/exception/AttributeExistsException.java
nawatts: Changed these to more generic ValidationException and ConflictException in https://github.com/DataBiosphere/terra-workspace-data-service/pull/455/commits/3efae2da0438cdcfb1bae19bc8933deff51d763d.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/455/commits/e5c2682bc96b96a8291d7c3aaec291dc792afaa4.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/455/commits/e5c2682bc96b96a8291d7c3aaec291dc792afaa4.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
nawatts: Added a comment about this in https://github.com/DataBiosphere/terra-workspace-data-service/pull/455/commits/2c9910722575b757e6ede9b68323084de9a83873.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
nawatts: Added a comment about this in https://github.com/DataBiosphere/terra-workspace-data-service/pull/455/commits/2c9910722575b757e6ede9b68323084de9a83873.

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
```diff
@@ -170,18 +236,18 @@ void deleteNonexistentAttribute() {
                 recordOrchestratorService.deleteAttribute(
                     INSTANCE, VERSION, TEST_TYPE, "doesnotexist"),
             "deleteAttribute should have thrown an error");
-    assertEquals(e.getMessage(), "Attribute does not exist");
+    assertEquals("Attribute does not exist", e.getMessage());
     assertAttributes(Set.of("id", "attr1", "attr2"));
   }
 
-  private void setUpDeleteAttributeTest() {
+  private void setUpRenameOrDeleteAttributeTest() {
```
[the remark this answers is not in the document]
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/455/commits/d55521e8dcca7421b302cf41de0fb24123cd552d.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/455/commits/ef32a31a4dbc71cdba992e2ce4feed908ba3fe72.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -232,27 +234,61 @@ public void deleteRecordType(UUID instanceId, String version, RecordType recordT
...
+    }
+    if (schema.attributes().stream()
+        .noneMatch(attributeSchema -> attributeSchema.name().equals(attribute))) {
+      throw new MissingObjectException("Attribute");
+    }
+    if (schema.attributes().stream()
+        .anyMatch(attributeSchema -> attributeSchema.name().equals(newAttributeName))) {
+      throw new AttributeExistsException("Attribute already exists");
+    }
+  }
```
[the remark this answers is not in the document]
nawatts: 👍 I will opt to do this in a follow up.

service/src/main/resources/static/swagger/openapi-docs.yaml
nawatts: Thinking about adding edit data type, I think this gets more complex if we allow renaming or changing data type in a request, but not both in the same request. I'm not convinced there's a good way to reuse AttributeSchema for this.

nawatts: Tangentially related... I think I'll rename this AttributeSchemaUpdate to more clearly link it to AttributeSchema.

davidangb: > It doesn't look like there's a way to override properties of a schema.

In OpenAPI, I don't think so. But you could create a base model that is lacking that property, then every inherited model specifies its own implementation of that property.

> I'm skeptical there's a good way to reuse AttributeSchema for this

I won't die on this hill, I am ok with different models in the OpenAPI. At the Java level, however, can you reuse `AttributeSchema` instead of creating new classes?

nawatts: Yes. Though I'm still not thrilled about that, since we'll be passing around an object with a `relatesTo` field that's never used.

nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/455/commits/80a0d679ddffa88e771575ed3516c110e5cc8416.

davidangb (approved):
thank you for sticking through all the comments!

PR #456, opened 2024-01-11
AJ-1523: Add `TdrManifestQuartzJobE2ETest` and fix discovered issues.

jladieu:
[AJ-1523](https://broadworkbench.atlassian.net/browse/AJ-1523): Improve unit test coverage for the snapshot import process.

This follows the model of `PfbQuartzJobE2ETest` using a snapshot of the v2f dataset.

The test is also paired with a couple fixes to the code under test for issues that caused the test to initially fail.

The process to get the V2F snapshot parquet files is documented [here](https://docs.google.com/document/d/1MiFdUZsWm5PBnrKSlqZoarU7ngwO-tO0ohlwPOJMLlY/edit)

[AJ-1523]: https://broadworkbench.atlassian.net/browse/AJ-1523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/AvroRecordConverter.java
```diff
@@ -138,8 +138,8 @@ public Object convertAttributeType(Object attribute) {
     }
```
jladieu, line 138: The changes in this file address a bug surfaced by this test and discussed in [slack thread](https://broadinstitute.slack.com/archives/C02SJQ0FKHQ/p1705007984663199).

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -173,7 +174,9 @@ BatchWriteResult importTable(
 
       // upsert this parquet file's contents
       try (ParquetReader<GenericRecord> avroParquetReader =
-          AvroParquetReader.<GenericRecord>builder(inputFile).build()) {
+          AvroParquetReader.<GenericRecord>builder(inputFile)
```
jladieu, line 177: The V2F file contained deprecated values of `int96` type.  This flag directs the `AvroParquetReader` not to crash and instead return these as a `fixed` type.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -313,7 +316,8 @@ private boolean isValidRelation(
     return requestedPrimaryKey != null && requestedPrimaryKey.equals(actualPrimaryKey);
   }
 
-  private URL parseUrl(String path) {
+  @VisibleForTesting
```
jladieu, line 319: This change is to provide an override seam used by `TdrTestUtils`

nawatts (approved):

calypsomatic (approved):

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbRecordConverterTest.java
```diff
@@ -165,8 +165,14 @@ void valuesInObjectAttributes() {
     assertEquals(BigDecimal.valueOf(3.14159), actual.getAttributeValue("pi"));
     assertEquals(Boolean.TRUE, actual.getAttributeValue("booly"));
     assertEquals("enumValue2", actual.getAttributeValue("enum"));
-    assertEquals("some bytes", actual.getAttributeValue("bytesOfStuff"));
-    assertEquals("fixedBytes", actual.getAttributeValue("tenFixedBytesOfStuff"));
+    assertEquals(
```
jladieu, line 168: Note: this is what the UI will start to look like for these attributes...so the compatibility with weird characters comes at the expense of deserializing these fields in a human-friendly fashion.

An alternative might be to attempt to Stringify, and only if that fails, return the array of bytes, but I thought having the behavior change depending on the content would be more confusing than beneficial.

Would welcome thoughts!

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/AvroRecordConverter.java
```diff
@@ -138,8 +138,8 @@ public Object convertAttributeType(Object attribute) {
...
 
     // Avro fixed
-    if (attribute instanceof GenericFixed fixedAttr) {
-      return new String(fixedAttr.bytes());
+    if (attribute instanceof GenericData.Fixed fixedAttr) {
```
jladieu, line 141: I forgot to mention when I first pushed this up that I explicitly changed the type here from the interface `GenericFixed` to the implementation `GenericData.Fixed`

Nothing forced me to do this, but the `toString()` method that I depend on is implemented as part of `GenericData.Fixed` and doesn't actually appear at all in `GenericFixed`.  So if you use the IDE to click through to the definition of `toString()` it would get lost and end up on the implementation from `Object.toString()` which I thought was confusing.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJobE2ETest.java
```diff
@@ -0,0 +1,240 @@
...
+    List<RecordTypeSchema> allTypes =
+        recordOrchestratorService.describeAllRecordTypes(instanceId, "v0.2");
+
+    // spot-check some column data types to see if they are good
+    assertDataType(allTypes, "all_data_types", "bool_column", DataTypeMapping.BOOLEAN);
+    assertDataType(
+        allTypes,
+        "all_data_types",
+        "date_column",
+        DataTypeMapping.NUMBER); // actual data: 19174 (days since epoch)
```
jladieu, line 120: Anywhere I left a comment with the actual data is a spot where I thought we might want to deserialize into a different type.  However I believe the Avro file's schema is what's driving this type decision, and they appear to be incorrect (mostly strings where a more specific type should be given).

nawatts (approved):

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbRecordConverterTest.java
calypsomatic: So the original data is a byte array of the string "some bytes" but it comes out as an array of numbers? 😬 Would we do some extra work to display it in the UI then?  Seems rough

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJobE2ETest.java
```diff
@@ -0,0 +1,240 @@
...
+        allTypes,
+        "all_data_types",
+        "date_array_column",
+        DataTypeMapping
+            .STRING); // actual data is string: "[\"2022-7-02\",\"2022-7-03\",\"2022-7-04\"]"
+    assertDataType(
+        allTypes,
+        "all_data_types",
+        "date_time_column",
+        DataTypeMapping.STRING); // actual data is string: "2022-07-05T12:00:01Z[UTC]"
```
calypsomatic, line 132: These are strings because they don't match our Date format exactly?

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJobE2ETest.java
```diff
@@ -0,0 +1,240 @@
...
+        allTypes,
+        "all_data_types",
+        "file_ref_column",
+        DataTypeMapping.STRING); // actual data is null
+    assertDataType(allTypes, "all_data_types", "float_column", DataTypeMapping.NUMBER);
+    assertDataType(
+        allTypes,
+        "all_data_types",
+        "float_array_column",
+        DataTypeMapping.STRING); // actual data is string: "[1.232,123.0]"
```
calypsomatic, line 153: Shouldn't this be an array of number?

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJobE2ETest.java
jladieu: Nope...unfortunately, it's because the type information stored in parquet is lossy due to conversion through synapse.  We need to look at the schema in the manifest and trust _that_ type for deserialization.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbRecordConverterTest.java
jladieu: I'm not exactly sure what the right thing to do here is.  The byte array _could_ have non-displayable bytes in it like the null byte that's been giving us a hard time.  In that case, we have no way to meaningfully display that to the user.

jladieu: Bringing [AJ-1537](https://broadworkbench.atlassian.net/browse/AJ-1537) to refinement for this.

PR #460, opened 2024-01-16
Aj 1531 python import test

calypsomatic:
https://broadworkbench.atlassian.net/browse/AJ-1531
Adds a test of TDR import for the python client.  This requires serving the manifest and parquet files from the nginx server and waiting for the import job to complete.

Reminder:


service/src/test/resources/tdrmanifest/v2f_for_python.json
```diff
@@ -0,0 +1,1164 @@
...
+      "shared" : null
+    }
+  },
+  "validatedPrimaryKeys" : null,
+  "format" : {
+    "parquet" : {
+      "location" : {
+        "tables" : [ {
+          "name" : "ancestry_specific_meta_analysis",
+          "paths" : [ "https://github.com/DataBiosphere/terra-workspace-data-service/raw/main/service/src/test/resources/parquet/v2f/ancestry_specific_meta_analysis.parquet" ]
```
calypsomatic, line 1136: Undoubtedly skipped something real obvious, but struggled to find a good way to define the paths to these files ahead of time.  One alternative is to live-update/generate this manifest from within the python test.  Is there another way?  Which is best?

jladieu: Might the process that needs to read the paths be able to read URLs of the form: `file://./relative/path/location.parquet` ?

Or alternately, you could do a hack like I did for the Java side, and mock something out to add support for something like a `local://./relative/path/location.parquet`

docs/WDS Running Python tests locally.md
```diff
@@ -36,6 +36,13 @@ To run tests, WDS uses pytest. Ensure that is installed by running:
 pip install pytest
 ```
 
+Some tests also require a server from which to fetch test files.  For these to work, you'll have to set up the nginx docker cotainer to serve them:
+```bash
+# start the server as a docker container in detached mode
+docker run -v `pwd`/service/src/test/resources/nginx.conf:/etc/nginx/nginx.conf -v `pwd`/service/src/test/resources:/usr/share/nginx/html -p 9889:80 -d nginx:1.23.3
```
calypsomatic, line 42: I left this change off of the readme, since presumably in everyday use we don't need to have these files.

service/src/test/resources/tdrmanifest/v2f.json
```diff
@@ -1157,7 +1157,7 @@
           "paths" : [ "classpath:parquet/v2f/feature_consequence.parquet" ]
         } ]
       },
-      "manifest" : "classpath:tdrmanifest/v2f/manifest.json"
+      "manifest" : "classpath:tdrmanifest/v2f.json"
```
calypsomatic, line 1160: Unrelated to this PR, just noticed it was incorrect.

service/src/test/python/test.py
```diff
@@ -236,3 +239,33 @@ def test_upload_download_tsv(self):
         response = self.schema_client.delete_record_type(self.current_workspaceId, self.version, self.cvsUpload_test);
         workspace_ent_type = self.schema_client.describe_all_record_types(self.current_workspaceId, self.version)
         self.assertTrue(len(workspace_ent_type) == 0)
+
+    # SCENARIO 6
+    # import snapshot from TDR with appropriate permissions
+    def test_import_snapshot(self):
+        import_request = { "type": "TDRMANIFEST", "url": "http://localhost:9889/tdrmanifest/v2f_for_python.json"}
```
calypsomatic, line 246: The file we're importing is relatively large for a test, but it's what we already had with a matching manifest.  Locally a single 10-second wait was enough for it to complete.  If we're concerned about it being too big/taking too long I'm open to crafting a smaller test parquet/manifest, it's just more work than it sounds like.

service/src/test/resources/tdrmanifest/v2f_for_python.json
```diff
@@ -0,0 +1,1164 @@
+{
+  "snapshot" : {
```
calypsomatic, line 2: Because the python test requires different paths for the parquet files, i just copied this manifest and changed only the paths.  Seems a bit silly and may be an argument in favor of making a different manifest/parquet combo, but maybe it doesn't matter.

service/src/test/resources/tdrmanifest/v2f_for_python.json
calypsomatic: I've tried a different hack (serving files from nginx) but I'm not married to it if we don't like the extra infrastructure required for that.

docs/WDS Running Python tests locally.md
```diff
@@ -36,6 +36,13 @@ To run tests, WDS uses pytest. Ensure that is installed by running:
 pip install pytest
 ```
 
+Some tests also require a server from which to fetch test files.  For these to work, you'll have to set up the nginx docker cotainer to serve them:
```
davidangb, line 39: typo: "cotainer"

davidangb (approved):
typo noted inline, otherwise lgtm and thank you!

service/src/test/python/test.py
nawatts: Nit: it would be nice if the test server URL was defined in a constant/configuration somewhere, so it doesn't have to be repeated/hard coded in each test that uses it.

service/src/test/python/test.py
```diff
@@ -236,3 +239,33 @@ def test_upload_download_tsv(self):
...
+        job_response = self.import_client.import_v1(self.current_workspaceId, import_request)
+        job_status_response = self.job_client.job_status_v1(job_response.job_id)
+        job_status = job_status_response.status
+        while job_status in ['RUNNING', 'QUEUED']:
+            time.sleep(10) #sleep ten seconds then try again
+            job_status_response = self.job_client.job_status_v1(job_response.job_id)
+            job_status = job_status_response.status
+        self.assertEqual(job_status, 'SUCCEEDED')
```
nawatts, line 254: FYI, since these tests are run using Pytest, they can use plain asserts instead of unittest methods.

```python
assert job_status == 'SUCCEEDED'
```

https://docs.pytest.org/en/7.1.x/how-to/assert.html

nawatts (approved):

PR #461, opened 2024-01-18
Add async observability to QuartzJob

ashanhol:
Added Observability to QuartzJob.execute denoting running, failure, and succeed states. The [`DefaultMeterObservationHandler`](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/observation/DefaultMeterObservationHandler.java#L23) gives us a built in Timer, Counter, and LongTaskTimer.

Here's how it reports in Prometheus. 

```
# HELP wds_job_execute_job_failed_total
# TYPE wds_job_execute_job_failed_total counter
wds_job_execute_job_failed_total{jobType="TdrManifestQuartzJob",wds_version="0.2.113-SNAPSHOT",} 1.0
# HELP wds_job_execute_job_running_total
# TYPE wds_job_execute_job_running_total counter
wds_job_execute_job_running_total{jobType="TdrManifestQuartzJob",wds_version="0.2.113-SNAPSHOT",} 1.0
# HELP wds_job_execute_seconds
# TYPE wds_job_execute_seconds summary
wds_job_execute_seconds_count{error="JobExecutionException",jobType="TdrManifestQuartzJob",wds_version="0.2.113-SNAPSHOT",} 1.0
wds_job_execute_seconds_sum{error="JobExecutionException",jobType="TdrManifestQuartzJob",wds_version="0.2.113-SNAPSHOT",} 0.522719647
# HELP wds_job_execute_seconds_max
# TYPE wds_job_execute_seconds_max gauge
wds_job_execute_seconds_max{error="JobExecutionException",jobType="TdrManifestQuartzJob",wds_version="0.2.113-SNAPSHOT",} 0.522719647
# HELP wds_job_execute_active_seconds
# TYPE wds_job_execute_active_seconds summary
wds_job_execute_active_seconds_active_count{wds_version="0.2.113-SNAPSHOT",} 1.0
wds_job_execute_active_seconds_duration_sum{wds_version="0.2.113-SNAPSHOT",} 3.752484771
# HELP wds_job_execute_active_seconds_max
# TYPE wds_job_execute_active_seconds_max gauge
wds_job_execute_active_seconds_max{wds_version="0.2.113-SNAPSHOT",} 3.752493404
```

----------------
Reminder:


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJob.java
```diff
@@ -58,7 +59,9 @@ public PfbQuartzJob(
       RestClientRetry restClientRetry,
       BatchWriteService batchWriteService,
       ActivityLogger activityLogger,
-      @Value("${twds.instance.workspace-id}") UUID workspaceId) {
+      @Value("${twds.instance.workspace-id}") UUID workspaceId,
```
jladieu, line 62: nit: I think leaving the most dynamic/variable constructor arg as the last one makes sense, so I'd recommend moving the @Value param to the end to separate it out from the others

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
```diff
@@ -29,16 +38,29 @@
...
 
   @Override
   public void execute(JobExecutionContext context) throws org.quartz.JobExecutionException {
     // retrieve jobId
     UUID jobId = UUID.fromString(context.getJobDetail().getKey().getName());
-    try {
+    Observation observation =
+        Observation.start(OBSERVE_JOB_EXECUTE, observationRegistry)
+            .contextualName(NAME_JOB_EXECUTION)
+            .lowCardinalityKeyValue(TAG_JOB_TYPE, getClass().getName())
```
jladieu, line 57: Nice use of the new API! :)

service/build.gradle
```diff
@@ -54,6 +54,7 @@ dependencies {
     implementation 'org.springframework.boot:spring-boot-starter-batch'
     implementation 'org.springframework.boot:spring-boot-starter-quartz'
     implementation 'org.springframework.boot:spring-boot-starter-actuator'
+    implementation 'org.springframework.boot:spring-boot-starter-aop'
```
jladieu, line 57: Did you follow some kind of instructions to know you had to add `spring-boot-starter-aop` and the `observedAspect` stuff?  If so, do you mind linking out to that in the PR description?  I'd love to have a read.

service/src/main/java/org/databiosphere/workspacedataservice/metrics/MetricsDefinitions.java
```diff
@@ -15,4 +16,26 @@ private MetricsDefinitions() {}
 
   /** tag for a record attribute name */
   public static final String TAG_ATTRIBUTE_NAME = "AttributeName";
+
+  // Quartz Job Metrics
+  /** observable name for running job */
```
jladieu, line 21: I bet this class will get _very_ cluttered, so I'm actually considering recommending getting rid of it entirely and just inlining all the strings until we have a real compelling need to factor them out.

If you don't want to touch what's already here, how would you feel about just inlining these newly added variables?

cc: @davidangb who introduced the class

service/src/main/java/org/databiosphere/workspacedataservice/metrics/MetricsDefinitions.java
```diff
@@ -15,4 +16,26 @@ private MetricsDefinitions() {}
...
+  public static final String EVENT_JOB_RUNNING = "job.running";
+
+  /** event name for job failed */
+  public static final String EVENT_JOB_FAILED = "job.failed";
+
+  /** name for job execution (used in span) */
+  public static final String NAME_JOB_EXECUTION = "job-execution";
+
+  /** tag for the type of job */
+  public static final String TAG_JOB_TYPE = "jobType";
```
jladieu, line 37: Tags should also use the lowercase dot notation adopted for the metric name: https://github.com/micrometer-metrics/micrometer-docs/blob/main/src/docs/concepts/naming.adoc#tag-naming

So I think that suggests `job.type` and `job.id`

service/src/main/java/org/databiosphere/workspacedataservice/metrics/MetricsDefinitions.java
```diff
@@ -15,4 +16,26 @@ private MetricsDefinitions() {}
...
+  public static final String EVENT_JOB_SUCCEEDED = "job.succeeded";
+
+  /** event name for job running */
+  public static final String EVENT_JOB_RUNNING = "job.running";
+
+  /** event name for job failed */
+  public static final String EVENT_JOB_FAILED = "job.failed";
+
+  /** name for job execution (used in span) */
+  public static final String NAME_JOB_EXECUTION = "job-execution";
```
jladieu, line 34: I'm not sure what contextualName is used for, but I wonder if `job-execution` should follow the same lowercase dot naming convention as tags: `job.execution`

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobE2ETest.java
```diff
@@ -127,7 +129,8 @@ void importTestResource() throws IOException, JobExecutionException {
     when(wsmDao.enumerateDataRepoSnapshotReferences(any(), anyInt(), anyInt()))
         .thenReturn(new ResourceList());
 
-    buildPfbQuartzJob(jobDao, wsmDao, restClientRetry, batchWriteService, activityLogger)
+    buildPfbQuartzJob(
```
jladieu, line 132: I'm thinking out loud on this one...you don't have to do anything with this comment, but if you're interested, I have some thoughts about what I think could be a perfect _prefactoring_ for this change.

Based on the noisy diff in this file, it looks like you had to change every invocation of `buildPfbQuartzJob`, which, if IDE supported, was probably straightforward...but if you had to do it manually, or read the change manually, it'd have been a bummer.

I think this is a good place where you could factor the logic out into a private helper method for this class that encapsulated all the dependency injection, such that you end up with a method like:

`buildPfbQuartzJob(<minimalArgs>)` where `minimalArgs` only contains the stuff that actually changes from test to test.  This might even be a completely empty param list if nothing actually changes.

I gave this treatment to the `TdrManifestImportE2ETest` and found that helped slim down the per-test changes required when changing the constructor args.

Tdr also had a non-E2E test which would benefit from it, so I factored it out entirely into:

https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrTestSupport.java#L24

I bet we could do the same treatment for Pfb.

Now, if this had been done for you ahead of time (as in, a prefactor), you'd only have to change one line of code in the file, rather than each test case.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobTest.java
```diff
@@ -72,7 +74,13 @@ void linkAllNewSnapshots() {
 
     // call linkSnapshots
     PfbQuartzJob pfbQuartzJob =
-        buildPfbQuartzJob(jobDao, wsmDao, restClientRetry, batchWriteService, activityLogger);
+        buildPfbQuartzJob(
```
jladieu, line 77: See other comment about how I ended up with https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrTestSupport.java#L24 which factors out the QuartzJob instantiation into one place so adding another arg to the constructor doesn't result in this form of [shotgun surgery](https://refactoring.guru/smells/shotgun-surgery)

jladieu (approved):
I have a few comments/suggestions but nothing showstopper.  LGTM with pre-emptive approval!

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
jladieu, line 57: One late thought after my prior LGTM...is there a straightforward way to get the classname _without_ the fully qualified package name?  I'm thinking that we're much more likely to move class files around (and change the package name), which would create metric churn.  It'd be great to avoid it by just keying off the classname `TdrSnapshotImportJob` or whatever, which is less likely to change than the package.

service/src/main/java/org/databiosphere/workspacedataservice/metrics/MetricsDefinitions.java
ashanhol: I knew you were going to comment, I followed the naming conventions from the blog post https://spring.io/blog/2022/10/12/observability-with-spring-boot-3/#webmvc-server-code

service/src/main/java/org/databiosphere/workspacedataservice/metrics/MetricsDefinitions.java
ashanhol: a tag is not the metric name. the name the (observability) metric is "job.execution". I followed the naming convention in the blog post https://spring.io/blog/2022/10/12/observability-with-spring-boot-3/#webmvc-server-code

service/build.gradle
ashanhol: So I thin it's only strictly necessary to include in order to use the @Observable annotation, which is how I originally attempted to wrap QuartzJob.execute. I left them in after I refactored it though thinking it'd be nice to have the option to use the @Observable annotation for other methods in the future. Happy to remove if you think it's too preemptive

service/src/main/java/org/databiosphere/workspacedataservice/metrics/MetricsDefinitions.java
davidangb: I'll follow your recommendation here. My original thinking is that it would be good to have a "table of contents" that shows what all our metric names/tags/etc are … but that is likely available elsewhere.

service/src/main/java/org/databiosphere/workspacedataservice/metrics/MetricsDefinitions.java
jladieu: Huh, I think the blogpost and the tag naming recommendations are in conflict.   That said I don't have a great path to resolve that conflict so I think it's fine to stick with what you've got until I have more signal of one being a better source of truth over the other.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobE2ETest.java
jladieu: Couldn't help myself so I went ahead and pushed up https://github.com/DataBiosphere/terra-workspace-data-service/pull/462 for this if you'd like to benefit from the prefactor.

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
ashanhol: Fixed with getSimpleName, thank you

service/build.gradle
ashanhol: removed 👍

service/src/main/java/org/databiosphere/workspacedataservice/metrics/MetricsDefinitions.java
ashanhol: Ok removed this file and inlined all the strings. Please review to confirm things went where we expect 🙏

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
```diff
@@ -29,16 +33,32 @@
...
+        .observationHandler(new DefaultMeterObservationHandler(new SimpleMeterRegistry()));
+    Observation observation =
+        Observation.start("wds.job.execute", observationRegistry)
+            .contextualName("job-execution")
+            .lowCardinalityKeyValue("jobType", getClass().getSimpleName())
+            .highCardinalityKeyValue("jobId", jobId.toString());
+    try (Observation.Scope scope = observation.openScope()) {
       // mark this job as running
       getJobDao().running(jobId);
+      observation.event(Observation.Event.of("job.running"));
```
ashanhol, line 61: @jladieu So now thinking about it more, I wonder if we might want to extend the DefaultMeterObservationHandler since the only thing it does onEvent is increment a counter. We might want to Timer.record a timestamp instead. What do you think?

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
```diff
@@ -220,10 +216,10 @@ public Map<String, DataTypeMapping> addOrUpdateColumnIfNeeded(
...
       // update a metrics counter with this schema change
       Counter counter =
-          Counter.builder(COUNTER_COL_CHANGE)
-              .tag(TAG_RECORD_TYPE, recordType.getName())
-              .tag(TAG_ATTRIBUTE_NAME, column)
-              .tag(TAG_INSTANCE, instanceId.toString())
+          Counter.builder("column.change.datatype")
+              .tag("RecordType", recordType.getName())
+              .tag("AttributeName", column)
+              .tag("Instance", instanceId.toString())
```
davidangb, line 222: should this change to using the `Observation` approach?

davidangb (commented):
one question inline, otherwise looking good!

I see this is failing unit tests; it needs some updates for that.

Can you write a test to ensure that `QuartzJob` is properly recording metrics? If you pass a `@MockBean ObservationRegistry observationRegistry` (or a `@SpyBean`), can you verify in the mock/spy that it's doing the right thing?

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
jladieu: Thanks for thinking more about this.  Your comment prompting me to take a look at `DefaultMeterObservationHandler`'s implementation, and I think it supports what we need to do, but we need to use its interface differently.  If you take a look at its `onStart` and `onStop` methods, they are interacting with a `Timer.Sample` and `LongTaskTimer.Sample`, both of which I think have something valuable to offer.

This also brings to mind something that I recall reading when I was going through the micrometer concepts [doc](https://docs.micrometer.io/micrometer/reference/concepts/counters.html)

>Be sure to read through the timer section before jumping into using counters, as timers record a count of timed events as part of the suite of metrics that go into timing. For those pieces of code you intend to time, you do NOT need to add a separate counter.

So I *think* we are encouraged to use either a `Timer` or a `Counter`, but not both, as it's redundant.  Since what we want here is timing, I think we just need to adjust how we interact with the `Observation` API.

I am guessing what you might want to do is:
- get rid of the `Observation.Scope scope` line; afaik this doesn't do anything yet; we might be able to use it later on but it looks like it's all no-ops in the implementation; (hang onto the `try` though)
- get rid of the `observation.event(...running...)` line, I think that is implied by starting the observation a couple lines earlier
- _maybe, but I'm not sure this will be necessary_, replace `observation.event(...job.running...)` with `observation.getContext().lowCardinalityKeyValue("outcome", "success")`
- _maybe, but I'm not sure this will be necessary_, replace `observation.event(...job.failed...)` with `observation.getContext().lowCardinalityKeyValue("outcome", "failure")`

If the last two swaps aren't necessary, I think you can remove the `observation.event` lines entirely, as your later `observation.stop()` should record the appropriate timings, and the presence or absence of an `error` tag on the observation is what we can key off to determine success/error rate.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
jladieu: I think this would be good from the perspective of calling out the high vs. low cardinality keys, but ok to leave it for a followup ticket.

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
```diff
@@ -29,16 +33,32 @@
...
   /** implementing classes are expected to be beans that inject a JobDao */
   protected abstract JobDao getJobDao();
 
   @Override
   public void execute(JobExecutionContext context) throws org.quartz.JobExecutionException {
     // retrieve jobId
     UUID jobId = UUID.fromString(context.getJobDetail().getKey().getName());
-    try {
+    // configure observation to collect counter, timer, and longtasktimer metrics
+    observationRegistry
```
jladieu, line 50: After thinking about testability, I think this `observationRegistry` configuration stuff should be factored out into a `ObservationRegistryCustomizer` `@Bean`.  You can probably copy how I did this from `MetricsConfig` with the `MeterRegistryCustomizer`, and stand up your own `ObservationsConfig`.

Also, I suspect you shouldn't have to explicitly provide your own `new SimpleMeterRegistry()` and in fact this probably inhibits this correctly working with springboot, so instead you want to use an autowired `MeterRegistry` that is provided by Spring and already pre-configured.  I think this could be the first argument of your `ObservationsConfig` constructor and hopefully can just be autowired.

jladieu: I put up this [PR](https://github.com/DataBiosphere/terra-workspace-data-service/pull/466) which you can just merge into this one if you think this makes sense.

jladieu:
@ashanhol I saw your question about whether we might need to customize `DefaultMeterObservationHandler` and I left some [suggestions](https://github.com/DataBiosphere/terra-workspace-data-service/pull/461#discussion_r1461959026).

Regarding your question in Slack about whether/how to test this, I think if you merge the PR mentioned [here](https://github.com/DataBiosphere/terra-workspace-data-service/pull/461#discussion_r1462008771) you should be able to adjust `QuartzJobTest` to use an `@Autowired MeterRegistry` and then check meters in a test as described in this StackOverflow: https://stackoverflow.com/a/74182551

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
ashanhol: Thank you for the PR, it does make sense for this context. However, if we ever want to override that handler, we could register a new observationHandler, but I'm not sure what consequences that has on a bean?

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
ashanhol: Scope deleted- no issues. Not sure why it was included in every example, made it seem needed. 
Gotcha, yeah that's a better way to denote the state. 
I had that thought- that start/stop time would automatically record state and we'd be getting duplicate info. Technically there's a low cardinality key that's included when a task errors by default (error="JobExecutionException") so outcome would be redundant too (no error tag= success). BUT, I think the extra tag might still help with easy metric creation and it is low cardinality 🤷‍♀️

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
```diff
@@ -32,6 +43,9 @@
 class QuartzJobTest {
 
   @MockBean JobDao jobDao;
+  @Autowired ObservationRegistry observationRegistry;
+  @Autowired MeterRegistry meterRegistry;
+  private TestObservationRegistry testObservationRegistry;
```
ashanhol, line 48: Need both regular observationRegistry+meterRegistry to test timer/counter/longtesttimer, and testObsrevationRegistry to test the observation itself. Metrics are null in testObservationRegistry and afaik there's not an easy way to test an observation without it

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
```diff
@@ -80,21 +105,62 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
...
+    new TestableQuartzJob(randomToken, observationRegistry).execute(mockContext);
+
+    // metrics provisioned should be longTaskTimer and a Timer, confirm both ran
+    LongTaskTimer longTaskTimer = meterRegistry.find("wds.job.execute.active").longTaskTimer();
+    assertNotEquals(longTaskTimer.duration(TimeUnit.SECONDS), 0);
+    Timer timer = meterRegistry.find("wds.job.execute").timer();
+    assertNotEquals(timer.count(), 0);
+  }
+
+  // sets up a job and returns the job context
```
ashanhol, line 155: @davidangb pulled this code into a function. From what I can tell you need all this to set up running any job (at least it failed when I didn't)

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
jladieu: I think it's fine to leave in the low card tags for now, and we can strip them out if they end up being superfluous.

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
jladieu: IMO, from a testing perspective, I care a lot more about what ends up landing in the MeterRegistry and a lot less about _how_ it gets there (whether by `Observation` or direct metering), so I would say there's no need to test the `Observation` stuff if you can make assertions about the `Meter`s gathered.

ashanhol: per our irl discussion- keeping it so we have an example of how to test the observability stuff plus it's important to test at least one

jladieu (approved):

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
```diff
@@ -80,21 +105,62 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
...
+        .hasLowCardinalityKeyValue("outcome", "success")
+        .hasBeenStarted()
+        .hasBeenStopped();
+  }
+
+  @Test
+  void correctMetrics() throws JobExecutionException {
+    String randomToken = RandomStringUtils.randomAlphanumeric(10);
+    JobExecutionContext mockContext = setUpTestJob(randomToken, UUID.randomUUID().toString());
+
```
davidangb, line 144: suggestion: add assertions here that verify the timer and long task timer are empty to start:
```java
    // metrics should be null before we run the job
    assertNull(meterRegistry.find("wds.job.execute.active").longTaskTimer());
    assertNull(meterRegistry.find("wds.job.execute").timer());
```

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
```diff
@@ -80,21 +105,62 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
...
+    TestObservationRegistryAssert.assertThat(testObservationRegistry)
+        .doesNotHaveAnyRemainingCurrentObservation()
+        .hasObservationWithNameEqualTo("wds.job.execute")
+        .that()
+        .hasHighCardinalityKeyValue("jobId", jobUuid)
+        .hasLowCardinalityKeyValue("jobType", "TestableQuartzJob")
+        .hasLowCardinalityKeyValue("outcome", "success")
+        .hasBeenStarted()
+        .hasBeenStopped();
```
davidangb, line 137: add `.hasNumberOfObservationsWithNameEqualTo("wds.job.execute", 1)` for good measure?

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
```diff
@@ -80,21 +105,62 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
...
+  }
+
```
davidangb, line 139: could you add a test for a failing job that looks for `.hasLowCardinalityKeyValue("outcome", "failure")`? You probably need to modify `TestableQuartzJob` or create something new that `extends QuartzJob` but throws an error in `executeInternal`.

davidangb (commented):
comments inline, mostly about tests. Good stuff in this PR!

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
```diff
@@ -29,16 +31,29 @@
...
   @Override
   public void execute(JobExecutionContext context) throws org.quartz.JobExecutionException {
     // retrieve jobId
     UUID jobId = UUID.fromString(context.getJobDetail().getKey().getName());
+
+    Observation observation =
+        Observation.start("wds.job.execute", observationRegistry)
+            .contextualName("job-execution")
+            .lowCardinalityKeyValue("jobType", getClass().getSimpleName())
+            .highCardinalityKeyValue("jobId", jobId.toString());
```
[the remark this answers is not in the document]
jladieu: Good call @davidangb this is actually the spirit behind the measurement, which is to know how long the customer has to wait, so I'm in favor of this.  Ok to defer to another PR though if you just want to land this one @ashanhol

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
```diff
@@ -50,11 +65,15 @@ public void execute(JobExecutionContext context) throws org.quartz.JobExecutionE
       executeInternal(jobId, context);
       // if we reached here, mark this job as successful
       getJobDao().succeeded(jobId);
+      observation.lowCardinalityKeyValue("outcome", "success");
     } catch (Exception e) {
       // on any otherwise-unhandled exception, mark the job as failed
       getJobDao().fail(jobId, e);
+      observation.error(e);
+      observation.lowCardinalityKeyValue("outcome", "failure");
```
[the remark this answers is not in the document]
jladieu: I like this idea!

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
jladieu: Sorry, I'm not sure I entirely understand the hypothetical here...but I think you should be able to override/replace the common config on an as-needed basis if you have a specific scenario that needs to do something different.

Regardless, I think it makes more sense to start with a common config, and avoid doing any local overrides until we have a really compelling need to.

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
ashanhol: Thank you for making this change in your PR

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
```diff
@@ -80,21 +105,62 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
...
+  @Test
+  void correctMetrics() throws JobExecutionException {
+    String randomToken = RandomStringUtils.randomAlphanumeric(10);
+    JobExecutionContext mockContext = setUpTestJob(randomToken, UUID.randomUUID().toString());
+
+    // execute the TestableQuartzJob, then confirm metrics provisioned correctly
+    new TestableQuartzJob(randomToken, observationRegistry).execute(mockContext);
+
+    // metrics provisioned should be longTaskTimer and a Timer, confirm both ran
+    LongTaskTimer longTaskTimer = meterRegistry.find("wds.job.execute.active").longTaskTimer();
```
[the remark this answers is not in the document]
ashanhol: most of these comments are now resolved with your PR, will close

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
ashanhol: done

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
ashanhol: OK I'm losing my mind here. In [this current test implementation](https://github.com/DataBiosphere/terra-workspace-data-service/pull/461/commits/2d5f6f06de765245f0a1f28c616ad82d867af2ea#diff-ff441b6759d47109f28f9c84574847a094a371d887c244d0b18449a8650a691fR216) I have the execute function wrapped in a `assertThrows`, and it fails because it says it doesn't throw anything. But if I get rid of `assertThrows` it totally does. I feel like I'm missing something obvious

ashanhol: OK @jladieu and I figured this out, and it's confusing. We have two JobExecutionExceptions, one from Quartz and one we define. The one we define extends RuntimeException and therefore doesn't need to be caught/handled, but the quartz one does. QuartzJob.execute throws a org.quartz.JobExecutionException (and therefore needs the `throws org.quartz.JobExecutionException` in the method signature), but all our custom jobs throw various RuntimeException ones. So I had the right idea having the TestQuartzJob throw our custom one but it made writing this test VERY confusing

davidangb: Maybe this explanation helps to explain …

The typical data flow is:
* Quartz looks for an `execute()` method and runs that method when it kicks off a job
* in the WDS codebase, we define the `execute()` method in `QuartzJob`.
* `QuartzJob.execute()` wraps and calls the `executeInternal()` method defined in subclasses such as `TdrManifestQuartzJob` or `PfbQuartzJob`
* As part of its wrapping, `QuartzJob.execute()` !!_*catches any exceptions thrown by `executeInternal`*_!!. On catch, it marks the job as failed and swallows the exception so Quartz can mark its job as complete.

In this test:
* the test defines a `TestableQuartzJob` subclass of `QuartzJob`
* the test throws an exception !!_*from `executeInternal()`*_!!
* the `QuartzJob.execute()` wrapper properly catches the error thrown from `executeInternal()`, marks the job as failed, registers metrics, and swallows the error.

So, it's expected that when this test calls `TestableQuartzJob(...).execute`, it will not throw an error.

ashanhol: Yeah, I figured it out eventually. I personally didn't know that RuntimeException had different behavior than regular Exception and that required some compiler fighting and extra confusion with the namespaces.

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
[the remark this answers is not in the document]
ashanhol: I actually had this exact thought when doing this, but couldn't quite figure out how to "pass" an observation from one class to another. I'm happy to make a ticket to think about it further.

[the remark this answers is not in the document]
ashanhol: https://broadworkbench.atlassian.net/browse/AJ-1568

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
```diff
@@ -73,28 +107,138 @@ protected JobDao getJobDao() {
     @Override
     protected void executeInternal(UUID jobId, JobExecutionContext context) {
       assertEquals(expectedToken, TokenContextUtil.getToken().getValue());
+      try {
+        if (shouldThrowError) {
+          throw new JobExecutionException("Forced job to fail");
+        }
+      } catch (Exception e) {
+        throw new JobExecutionException("Forced job to fail", e);
+      }
```
davidangb, line 116: simplify to:
```suggestion
      if (shouldThrowError) {
        throw new JobExecutionException("Forced job to fail");
      }
```

davidangb (approved):
one nit/suggestion inline, giving this a proactive :+1:

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
ashanhol: ah thanks, forgot that when I cleaned up the other stuff

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
jladieu: Fixed in https://github.com/DataBiosphere/terra-workspace-data-service/pull/484

PR #463, opened 2024-01-19
[AJ-1279] Add RecordTypeSchema.getAttributeSchema

nawatts:
Small refactor to add a more convenient API for getting a specific attribute's schema.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/model/RecordTypeSchema.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/model/RecordTypeSchema.java
@@ -14,4 +14,11 @@ public record RecordTypeSchema(
     return attributes.stream()
         .anyMatch(attributeSchema -> attributeSchema.name().equals(attribute));
   }
+
+  public AttributeSchema getAttributeSchema(String attribute) {
+    return attributes.stream()
+        .filter(attr -> attr.name().equals(attribute))
+        .findFirst()
+        .orElseThrow();
+  }
 }
--- a/service/src/test/java/org/databiosphere/workspacedataservice/service/model/RecordTypeSchemaTest.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/service/model/RecordTypeSchemaTest.java
@@ -43,7 +43,9 @@ class RecordTypeSchemaTest {
 
   @Test
   void testGetAttributeSchema() {
-    assertEquals(schema.getAttributeSchema("attr1"), new AttributeSchema(/* name */ "attr1", /* datatype */ "STRING", /* relatesTo */ null));
+    assertEquals(
+        schema.getAttributeSchema("attr1"),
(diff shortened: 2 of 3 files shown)


calypsomatic (approved):

service/src/test/java/org/databiosphere/workspacedataservice/service/model/RecordTypeSchemaTest.java
```diff
@@ -37,4 +40,17 @@ void testContainsAttribute() {
     assertTrue(schema.containsAttribute("attr1"));
     assertFalse(schema.containsAttribute("doesNotExist"));
   }
+
+  @Test
+  void testGetAttributeSchema() {
+    assertEquals(schema.getAttributeSchema("attr1"), new AttributeSchema("attr1", "STRING", null));
```
jladieu, line 46:
```suggestion
    assertEquals(schema.getAttributeSchema("attr1"), new AttributeSchema(/* name= */ "attr1", /* datatype= "STRING", /* relatesTo= */ null));
```

I know the convention throughout the codebase is to just instantiate `AttributeSchema` as you have here, but where possible, and in the absence of a more expressive construction process like a builder, I recommend putting in this kind of arg comment to make it more scannable when the arg value itself isn't intention revealing enough to distinguish what the arg is.

jladieu (approved):

service/src/test/java/org/databiosphere/workspacedataservice/service/model/RecordTypeSchemaTest.java
nawatts: For what it's worth, Intellij has been helpfully showing me the argument names.

Screenshot 2024-01-19 at 12 45 18 PM

nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/463/commits/495b3ebc1c5fb6c7c32205d27be491c41a586296.

service/src/main/java/org/databiosphere/workspacedataservice/service/model/RecordTypeSchema.java
```diff
@@ -14,4 +14,11 @@ public boolean containsAttribute(String attribute) {
     return attributes.stream()
         .anyMatch(attributeSchema -> attributeSchema.name().equals(attribute));
   }
+
+  public AttributeSchema getAttributeSchema(String attribute) {
+    return attributes.stream()
+        .filter(attr -> attr.name().equals(attribute))
+        .findFirst()
+        .orElseThrow();
+  }
```
davidangb, line 23: I know this PR is already merged - I wonder if it makes sense by now to change this from a `record` to a `class` and store a member variable which is a map of attribute name -> attribute schema, instead of storing the list of attribute schemas and repeatedly streaming through them

PR #465, opened 2024-01-22
[AJ-1279] Add alternate constructors for AttributeSchema

nawatts:
Addressing @jladieu's comment here: https://github.com/DataBiosphere/terra-workspace-data-service/pull/463#discussion_r1459406188.

In tests, AttributeSchema is frequently instantiated with a null value for `relatesTo` and occasionally with a null value for `datatype`. This adds alternate constructors that allow omitting null values and also allow passing a DataTypeMapping instead of a string for the `datatype` argument. This clarifies the meaning of the second argument and reduces the chance of a typo causing a test failure due to an invalid data type.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/model/AttributeSchema.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/model/AttributeSchema.java
@@ -4,4 +4,21 @@ import com.fasterxml.jackson.annotation.JsonInclude;
 import org.databiosphere.workspacedataservice.shared.model.RecordType;
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
-public record AttributeSchema(String name, String datatype, RecordType relatesTo) {}
+public record AttributeSchema(String name, String datatype, RecordType relatesTo) {
+
+  public AttributeSchema(String name, DataTypeMapping datatype, RecordType relatesTo) {
+    this(name, datatype.name(), relatesTo);
+  }
+
+  public AttributeSchema(String name, String datatype) {
+    this(name, datatype, null);
+  }
+
+  public AttributeSchema(String name, DataTypeMapping datatype) {
+    this(name, datatype, null);
+  }
+
+  public AttributeSchema(String name) {
+    this(name, (String) null, null);
+  }
+}
--- a/service/src/test/java/org/databiosphere/workspacedataservice/service/model/AttributeSchemaTest.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/service/model/AttributeSchemaTest.java
@@ -13,6 +13,9 @@ class AttributeSchemaTest {
         new AttributeSchema("attr", DataTypeMapping.STRING, RecordType.valueOf("otherAttr")));
 
     assertEquals(
+        new AttributeSchema("attr", "STRING", null), new AttributeSchema("attr", "STRING"));
+
+    assertEquals(
         new AttributeSchema("attr", "STRING", null),
         new AttributeSchema("attr", DataTypeMapping.STRING));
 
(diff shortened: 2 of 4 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/service/model/RecordTypeSchemaTest.java
```diff
@@ -44,8 +44,7 @@ void testContainsAttribute() {
   @Test
   void testGetAttributeSchema() {
     assertEquals(
-        schema.getAttributeSchema("attr1"),
-        new AttributeSchema(/* name */ "attr1", /* datatype */ "STRING", /* relatesTo */ null));
+        new AttributeSchema("attr1", DataTypeMapping.STRING), schema.getAttributeSchema("attr1"));
```
nawatts, line 47: Correcting the order of expected/actual values while we're here.

calypsomatic (approved):

ashanhol (approved):
Looks like this addresses those comments :)

service/src/main/java/org/databiosphere/workspacedataservice/service/model/AttributeSchema.java
```diff
@@ -4,4 +4,21 @@
...
+  public AttributeSchema(String name, String datatype) {
+    this(name, datatype, null);
+  }
+
+  public AttributeSchema(String name, DataTypeMapping datatype) {
+    this(name, datatype, null);
+  }
+
+  public AttributeSchema(String name) {
+    this(name, (String) null, null);
```
jladieu, line 22: What does it actually mean to be an `AttributeSchema` with no datatype?  Are there any circumstances where we expect this to happen in the wild, or is this something we only encounter in test cases?  If the latter, I'd suggest dropping this constructor entirely and just putting in a reasonable value in the tests.

Also, if it really doesn't make sense to have a null datatype, this might be worth adding in a validation to protect against null args.  I found a quick read about how to do more interesting thing with Java record objects, including validation: https://xebia.com/blog/how-to-use-java-records/

service/src/test/java/org/databiosphere/workspacedataservice/service/model/AttributeSchemaTest.java
```diff
@@ -0,0 +1,24 @@
...
+    assertEquals(
+        new AttributeSchema("attr", "STRING", RecordType.valueOf("otherAttr")),
+        new AttributeSchema("attr", DataTypeMapping.STRING, RecordType.valueOf("otherAttr")));
+
+    assertEquals(
+        new AttributeSchema("attr", "STRING", null), new AttributeSchema("attr", "STRING"));
+
+    assertEquals(
+        new AttributeSchema("attr", "STRING", null),
+        new AttributeSchema("attr", DataTypeMapping.STRING));
```
jladieu, line 20: A small nit but which can be an annoying detour/refactor...but as I read this I really think `DataTypeMapping` just needs to be called `DataType`.

jladieu (approved):
Thanks for working on making `AttributeSchema` more usable :)

service/src/main/java/org/databiosphere/workspacedataservice/service/model/AttributeSchema.java
nawatts: I think the only place _we_ construct an AttributeSchema with no datatype is tests. But it's related to reusing the AttributeSchema class for the attribute update request body, where an update can specify only one field to update.

service/src/test/java/org/databiosphere/workspacedataservice/service/model/AttributeSchemaTest.java
nawatts: I have though the same thing. @davidangb do you know of any history/reasoning for the `DataTypeMapping` name?

nawatts: I guess maybe it was named "mapping" because it maps a WDS type to a Postgres type?

PR #471, opened 2024-01-24
[AJ-1279] Add basic attribute data type conversions

nawatts:
This adds a RecordOrchestrationService method that allows converting attribute data types between string, number, and boolean (including arrays of those). To limit the PR size, it's not yet wired up to the API through RecordController.

Data type conversions are handled by Postgres. WDS runs ` ALTER COLUMN <column> TYPE <new type> USING <expression>`, where `<expression>` is generated from the relevant `DataTypeMapping`s Postgres types. The expression  uses the double colon operator, which is equivalent to `CAST`. For example, `expression::text` is equivalent to `CAST(expression AS text)`.

https://www.postgresql.org/docs/current/sql-altertable.html
https://learnsql.com/blog/double-colon-operator-postgresql/

Other data types will be added in future PRs. For now, attempting to convert either from or to one of those types throw an error.


service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
```diff
@@ -198,6 +204,288 @@ void renameAttributeConflictingName() {
...
+    // Act
+    recordOrchestratorService.updateAttributeDataType(
+        INSTANCE, VERSION, TEST_TYPE, attributeName, newDataType.name());
+
+    // Assert
+    assertAttributeDataType(attributeName, newDataType);
+    assertAttributeValue(RECORD_ID, attributeName, expectedFinalAttributeValue);
+  }
+
+  private static Stream<Arguments> updateAttributeDataTypeArgsProvider() {
```
jladieu, line 233: I think `@MethodSource` can take an array of method name strings, using `@MethodSource({"1","2",...})`

Would you consider breaking apart this method into many, one for each origin type?  I think it might help improve scannability and replace all the comments `// xyz to...` with an intention revealing method name.

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
```diff
@@ -346,4 +634,23 @@ private void testContainsRecord(
...
+        recordOrchestratorService.describeRecordType(INSTANCE, VERSION, TEST_TYPE);
+    AttributeSchema attributeSchema = recordTypeSchema.getAttributeSchema(attributeName);
+    assertEquals(dataType.name(), attributeSchema.datatype());
+  }
+
+  private void assertAttributeValue(String recordId, String attributeName, Object expectedValue) {
+    RecordResponse record =
+        recordOrchestratorService.getSingleRecord(INSTANCE, VERSION, TEST_TYPE, recordId);
+    Object attributeValue = record.recordAttributes().getAttributeValue(attributeName);
+
```
jladieu, line 649: Most of the assert methods should take an optional first arg with a message, and in this case where you've got a custom method wrapping exceptions, it can be helpful to include the extra context from the args.

Maybe something like this:

```
String explanation = String.format("assertAttributeValue expected attribute %s of record %s to equal %s", attributeName, recordId, expectedValue);

if (expectedValue instanceof Object[]) {
   assertArrayEquals(explanation, (Object[]) expectedValue, ...);
} else {
   assertEquals(explanation, expectedValue...);
}
```

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
jladieu, line 1: Nice test coverage!

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -1038,6 +1039,80 @@ public void renameAttribute(
...
+            DataTypeMapping.JSON,
+            DataTypeMapping.RELATION);
+    if (unsupportedDataTypes.contains(dataType.getBaseType())
+        || unsupportedDataTypes.contains((newDataType.getBaseType()))) {
+      throw new IllegalArgumentException(
+          "Unable to convert attribute from %s to %s"
+              .formatted(dataType.name(), newDataType.name()));
+    }
+
+    String expression = quote(SqlUtils.validateSqlString(attribute, ATTRIBUTE));
```
jladieu, line 1088: When doing a lot of String concatenation, I believe it's recommended to use a `StringBuilder` which is a lot smarter about memory allocation with Strings.

https://www.baeldung.com/java-strings-concatenation

But it amounts to:

StringBuilder expressionBuilder = new StringBuilder();
expressionBuilder.append("... stuff ...");
...
...
return expressionBuilder.build();

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -1038,6 +1039,80 @@ public void renameAttribute(
...
+                .formatted(attribute, newDataType.name()));
+      }
+      throw e;
+    }
+  }
+
+  private String getPostgresTypeConversionExpression(
+      String attribute, DataTypeMapping dataType, DataTypeMapping newDataType) {
+    // Some data types are not yet supported.
+    Set<DataTypeMapping> unsupportedDataTypes =
```
jladieu, line 1074: I find it difficult to read logic any time negatives are used.  Is there any way to express this as `supportedDataTypes` instead?

Maybe you can use Set math:

```
Set<DataTypeMapping> typeConversionSupportedDataTypes = Sets.difference(<all data types>, Set.of(... unsupported));
```

Also, this might be worth factoring out into a `private static final` variable so it's only allocated once for the class.

jladieu (approved):
Nice work!  I have a few suggestions but pre-emptively approving.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -1038,6 +1039,80 @@ public void renameAttribute(
...
+    if (!dataType.isArrayType() && newDataType.isArrayType()) {
+      expression = "array_append('{}', " + expression + ")";
+    }
+    // When converting from an array type to a scalar type, take the first value in the array.
+    else if (dataType.isArrayType() && !newDataType.isArrayType()) {
+      expression = expression + "[1]";
+    }
+
+    // No direct conversion exists between numeric and boolean types, so go through int.
+    if ((dataType.getBaseType() == DataTypeMapping.BOOLEAN
```
jladieu, line 1100: Oops missed this in my first pass but saw it on the followup PR...

You should almost always use `equals(...)` instead of `==` for equality comparison in Java.

Although for enums, it looks like `==` is "ok" (pending some caveats about null safety): https://stackoverflow.com/questions/1750435/comparing-java-enum-members-or-equals

Despite that, I still generally try to use `equals` anyway to propagate that assumption through the codebase.

If you're worried about nullsafety, you can reverse order and take the thing you know is not-null as the subject, eg. here:
`DataTypeMapping.BOOLEAN.equals(dataType.getBaseType())`

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
nawatts: For what it's worth, this block should be fairly short-lived as subsequent PRs add support for more data types.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
nawatts: Good catch. I think this is from JavaScript habits.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
nawatts: Then again... there are NULL and EMPTY_ARRAY types which likely won't be supported.

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
nawatts: Split these up and moved them to a different file in https://github.com/DataBiosphere/terra-workspace-data-service/pull/471/commits/a1c9253192acacc91306dece242396e9f65833e5.

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
nawatts: Added explanations to assertions in these helpers in https://github.com/DataBiosphere/terra-workspace-data-service/pull/471/commits/e8bfcfcfe650c9112564fc5fe737e031a87ef500.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
nawatts: Switched in https://github.com/DataBiosphere/terra-workspace-data-service/pull/471/commits/c0b531b8fa8f7ccf390a18a6daf4d80681631bc0.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
nawatts: Refactored in https://github.com/DataBiosphere/terra-workspace-data-service/pull/471/commits/75893217f64f38e2fc53d189b935512067c86360.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
nawatts: Since there are some cases here where it's not just concatenation (for example, `expression = "array_append('{}', " + expression + ")";`), I think I'll hold off on switching to StringBuilder until after [part 2](#472) and then we can have a clearer sense of how it fits/reads.

service/src/test/java/org/databiosphere/workspacedataservice/DataTypeConversionTestCases.java
nawatts, line 1: Wasn't really sure where to put this, so I just dumped it at the top level.

jladieu: I think they make sense in the original location of the method that inspired them...as private methods of the `RecordOrchestratorServiceTest`, right alongside the test case that uses them.

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
```diff
@@ -198,6 +202,108 @@ void renameAttributeConflictingName() {
     assertAttributes(Set.of("id", "attr1", "attr2"));
   }
 
+  @ParameterizedTest(name = "update attribute data type from {1} to {2}")
+  @MethodSource({
+    "org.databiosphere.workspacedataservice.DataTypeConversionTestCases#stringConversions",
```
jladieu, line 207: I think you can just make all these methods sibling methods to this one, and then *I think* you won't have to specify the fully qualified classpath.

jladieu (approved):

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
nawatts: That does avoid the need for the fully qualified class path, but now I kind of like having them in a separate file...
1. to avoid the huge wall of text for this one test function
2. they're related to another huge list of test parameters that will be added for https://github.com/DataBiosphere/terra-workspace-data-service/pull/472#discussion_r1465451372

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -1038,6 +1039,80 @@ public void renameAttribute(
...
+                  + " using "
+                  + getPostgresTypeConversionExpression(attribute, currentDataType, newDataType));
+    } catch (DataIntegrityViolationException e) {
+      if (e.getRootCause() instanceof SQLException sqlEx
+          && sqlEx.getSQLState() != null
+          && sqlEx.getSQLState().equals("22P02")) {
+        throw new ConflictException(
+            "Unable to convert values for attribute %s to %s"
+                .formatted(attribute, newDataType.name()));
+      }
```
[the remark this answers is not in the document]
nawatts: I'm not sure. I can try some other conversions and see if anything comes up. Even if there are other codes, I think the API will still behave reasonably. A DataIntegrityViolationException would still be thrown, but instead of a 409 response with a descriptive error message, it would return a 500 response.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -1038,6 +1039,80 @@ public void renameAttribute(
...
+    String expression = quote(SqlUtils.validateSqlString(attribute, ATTRIBUTE));
+
+    // When converting from a scalar type to an array type, append value to an empty array.
+    if (!dataType.isArrayType() && newDataType.isArrayType()) {
+      expression = "array_append('{}', " + expression + ")";
+    }
+    // When converting from an array type to a scalar type, take the first value in the array.
+    else if (dataType.isArrayType() && !newDataType.isArrayType()) {
+      expression = expression + "[1]";
+    }
```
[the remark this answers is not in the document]
nawatts: This matches the behavior for GCP workspaces.

This also isn't the only type conversion that is lossy. Number to boolean, date/time to date are as well.

We can definitely warn in the UI that it's destructive, but I think, where possible, should give users sharp tools and let them decide if they want to use them.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -1038,6 +1039,80 @@ public void renameAttribute(
...
+
+    // No direct conversion exists between numeric and boolean types, so go through int.
+    if ((dataType.getBaseType() == DataTypeMapping.BOOLEAN
+            && newDataType.getBaseType() == DataTypeMapping.NUMBER)
+        || (dataType.getBaseType() == DataTypeMapping.NUMBER
+            && newDataType.getBaseType() == DataTypeMapping.BOOLEAN)) {
```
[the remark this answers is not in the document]
nawatts: 👍 I waffled between the clarity and concision of this vs the extra Set allocations.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -1038,6 +1048,73 @@ public void renameAttribute(
...
+      if (newDataType.isArrayType()) {
+        expression += "[]";
+      }
+    }
+
+    // Convert to desired type.
+    expression += "::" + newDataType.getPostgresType();
+
+    return expression;
+  }
```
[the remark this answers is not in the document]
nawatts: After enumerating all the generated expressions for tests, I'm wondering if I should have just done that for the implementation to begin with.

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
```diff
@@ -346,4 +452,29 @@ private void testContainsRecord(
...
+  }
+
+  private void assertAttributeValue(String recordId, String attributeName, Object expectedValue) {
+    RecordResponse record =
+        recordOrchestratorService.getSingleRecord(INSTANCE, VERSION, TEST_TYPE, recordId);
+    Object attributeValue = record.recordAttributes().getAttributeValue(attributeName);
+
+    String explanation =
+        "assertAttributeValue expected attribute %s of record %s to equal %s"
+            .formatted(attributeName, recordId, expectedValue);
```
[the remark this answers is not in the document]
nawatts: As is though, this helper is generic and _could_ be used by any other test, so I'm reluctant to tie the message to the data type conversion test. But alternatively, the test could pass a message to the helper.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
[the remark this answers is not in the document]
jladieu: Will `Sets.union(x, y)` work?  Keeping it to set math might be a bit more intention revealing.  Either way, I think this is worth an inline explanatory comment.

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
jladieu: I agree with having them in their only file if (2) means they need to be accessible from another Test.java file, but if they're only going to be used by _this_ test file, I think it's okay (and still preferable) for the wall to be embedded in this test case (principle: keep test fixture definition as close to where it's used as possible).

No biggie either way if you really really like them in their own file.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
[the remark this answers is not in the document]
nawatts: It looks like `Sets.union` takes `Set` arguments, so I think it would end up `Sets.union(Set.of(dataType.getBaseType()), Set.of(newDataType.getBaseType()))`.

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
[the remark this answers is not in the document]
nawatts: Changed in https://github.com/DataBiosphere/terra-workspace-data-service/pull/471/commits/b6d9ed61e515fe80710883095857bf5d1769c4c8.

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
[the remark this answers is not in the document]
nawatts: Changed in https://github.com/DataBiosphere/terra-workspace-data-service/pull/471/commits/9583c97a0aceecda76276cca35d2c770fd621e7a.

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
nawatts: Moved in https://github.com/DataBiosphere/terra-workspace-data-service/pull/471/commits/19a0155056241ce2e89aa5e5e0e2215dca5bf2cf.

service/src/test/java/org/databiosphere/workspacedataservice/DataTypeConversionTestCases.java
nawatts: Moved these back to RecordOrchestratorServiceTest in https://github.com/DataBiosphere/terra-workspace-data-service/pull/471/commits/19a0155056241ce2e89aa5e5e0e2215dca5bf2cf.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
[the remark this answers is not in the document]
nawatts: 22P01 (floating point exception) looks like it's thrown from arithmetic operations, so I don't think we'd encounter it here.

I did find at least one more code we might encounter: 22003 (numeric value out of range). That can happen when converting a string containing a number with tens or hundreds of thousands of digits to a number.

Refactored to handle multiple error codes in https://github.com/DataBiosphere/terra-workspace-data-service/pull/471/commits/7e6dd693cc1d0d10c8a0b90f2666c1c2321b1b75.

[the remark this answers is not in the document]
davidangb: Agreed the API will still behave well, it'll just give a less helpful error message. So, I think it's worth a reasonable best-effort to find those error codes, but not let that be a rathole. Looks like you've done that in https://github.com/DataBiosphere/terra-workspace-data-service/commit/7e6dd693cc1d0d10c8a0b90f2666c1c2321b1b75, thank you!

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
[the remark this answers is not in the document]
davidangb: I want to be careful about replicating what we do in GCP - I'd rather make the best decision even if it means diverging from GCP behavior. That said, I really think it's a Product call!

davidangb (commented):
All the code changes look good and I am :+1: at the syntax level. I do think we need to pull in Product to validate that the functionality is what we want to implement.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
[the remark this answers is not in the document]
nawatts: @bethsheets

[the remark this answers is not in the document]
nawatts: Per Slack conversation with @bethsheets, removed support for array to scalar type conversions in https://github.com/DataBiosphere/terra-workspace-data-service/pull/471/commits/2adfad11aaad6a64cb267c4a7de1c9e566f14660.

davidangb (approved):

PR #472, opened 2024-01-24
[AJ-1279] Add support for date and date time data type conversions 

nawatts:
Stacked on #471. This adds support for date and date time types.

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -229,6 +229,7 @@ jacocoTestReport {
 test {
     systemProperties['pact.rootDir'] = "$buildDir/pacts"
     systemProperties['pact.provider.version'] = "$project.version"
+    systemProperties['user.timezone'] = "UTC";
 }
 
 tasks.register("pactTests", Test) {
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -104,6 +104,11 @@ public class RecordDao {
           DataTypeMapping.DATE_TIME,
           DataTypeMapping.ARRAY_OF_DATE_TIME);
 
+  private static final Set<Set<DataTypeMapping>> invalidDataTypeConversions =
+      Set.of(
+          Set.of(DataTypeMapping.BOOLEAN, DataTypeMapping.DATE),
+          Set.of(DataTypeMapping.BOOLEAN, DataTypeMapping.DATE_TIME));
+
   /**
    * These error codes are expected when a valid update attribute data type request fails because
    * attribute values for some records cannot be converted to the new data type.
@@ -1101,9 +1106,12 @@ public class RecordDao {
 
   private String getPostgresTypeConversionExpression(
       String attribute, DataTypeMapping dataType, DataTypeMapping newDataType) {
-    // Some data types are not supported.
-    if (!(dataTypesSupportedForTypeConversion.contains(dataType)
-        && dataTypesSupportedForTypeConversion.contains(newDataType))) {
+    // Some data types are not yet supported.
+    // Some conversions don't make sense / are invalid.
+    if (!dataTypesSupportedForTypeConversion.contains(dataType)
+        || !dataTypesSupportedForTypeConversion.contains(newDataType)
+        || invalidDataTypeConversions.contains(
+            Set.copyOf(List.of(dataType.getBaseType(), newDataType.getBaseType())))) {
       throw new IllegalArgumentException(
           "Unable to convert attribute from %s to %s"
               .formatted(dataType.name(), newDataType.name()));
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -1113,9 +1113,9 @@ public class RecordDao {
 
     // Unable to cast numbers to dates or timestamps.
     // Convert number to timestamp using to_timestamp.
-    if (dataType.getBaseType() == DataTypeMapping.NUMBER
-        && (newDataType.getBaseType() == DataTypeMapping.DATE
-            || newDataType.getBaseType() == DataTypeMapping.DATE_TIME)) {
+    if (dataType.getBaseType().equals(DataTypeMapping.NUMBER)
+        && Set.of(DataTypeMapping.DATE, DataTypeMapping.DATE_TIME)
+            .contains(newDataType.getBaseType())) {
       if (dataType.isArrayType()) {
         expression = "(convert_array_of_numbers_to_timestamps(" + expression + "))";
       } else {
@@ -1125,9 +1125,8 @@ public class RecordDao {
 
     // Unable to cast dates or timestamps to numbers.
     // Extract epoch from timestamp to get a number.
-    if ((dataType.getBaseType() == DataTypeMapping.DATE
-            || dataType.getBaseType() == DataTypeMapping.DATE_TIME)
-        && newDataType.getBaseType() == DataTypeMapping.NUMBER) {
+    if (Set.of(DataTypeMapping.DATE, DataTypeMapping.DATE_TIME).contains(dataType.getBaseType())
+        && newDataType.getBaseType().equals(DataTypeMapping.NUMBER)) {
       if (dataType.isArrayType()) {
         expression = "convert_array_of_timestamps_to_numbers(" + expression + ")";
         // Truncate to an integer for dates.
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -98,7 +98,11 @@ public class RecordDao {
           DataTypeMapping.NUMBER,
           DataTypeMapping.ARRAY_OF_NUMBER,
           DataTypeMapping.BOOLEAN,
-          DataTypeMapping.ARRAY_OF_BOOLEAN);
+          DataTypeMapping.ARRAY_OF_BOOLEAN,
+          DataTypeMapping.DATE,
+          DataTypeMapping.ARRAY_OF_DATE,
+          DataTypeMapping.DATE_TIME,
+          DataTypeMapping.ARRAY_OF_DATE_TIME);
 
   /**
    * These error codes are expected when a valid update attribute data type request fails because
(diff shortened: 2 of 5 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -1087,6 +1082,41 @@ private String getPostgresTypeConversionExpression(
...
+      }
+    }
+
+    // Unable to cast dates or timestamps to numbers.
+    // Extract epoch from timestamp to get a number.
+    if ((dataType.getBaseType() == DataTypeMapping.DATE
+            || dataType.getBaseType() == DataTypeMapping.DATE_TIME)
+        && newDataType.getBaseType() == DataTypeMapping.NUMBER) {
+      if (dataType.isArrayType()) {
+        // TODO: confirm that this preserves order of array elements.
```
calypsomatic, line 1104: Is there a ticket for this TODO or is that as part of this PR?

nawatts: I was just in the process of removing that comment. It's from a previous version of these functions.

ashanhol (approved):
Nice job with thorough test cases. Glad to see number (array) to.. as well as all the Date (array) to.. Datetime (array) to...  👍

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -1072,12 +1072,7 @@ private String getPostgresTypeConversionExpression(
       String attribute, DataTypeMapping dataType, DataTypeMapping newDataType) {
     // Some data types are not yet supported.
```
jladieu, line 1073: I think you might have ample coverage of this in the test that checks the DB outcomes...but I wonder if there is enough going on in this method with enough potential edge cases that it might be worth making the method non-private and annotate it as @VisibleForTesting.

```
@VisibleForTesting
String getPostgresType...
```

And then add some unit tests that verify some of the edgiest-case String construction.

service/src/main/resources/liquibase/changesets/20240118_number_timestamp_array_conversions.yaml
```diff
@@ -0,0 +1,37 @@
+databaseChangeLog:
+  - changeSet:
+      id:  20240118_number_timestamp_array_conversions
+      author:  nwatts
+      changes:
+        -  sql:
+             dbms:  postgresql
+             splitStatements: false
+             sql:  |
+               create or replace function convert_array_of_numbers_to_timestamps(numeric[])
```
jladieu, line 10: This stuff is intense!  I'd love to see if you have a source recipe you followed for PG procs.

jladieu (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
nawatts: Ooh, I like that idea.

service/src/main/resources/liquibase/changesets/20240118_number_timestamp_array_conversions.yaml
nawatts: This was all adapted from StackOverflow answers.

nawatts:
Rebased after addressing feedback on #471.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -1089,6 +1093,39 @@ private String getPostgresTypeConversionExpression(
 
     String expression = quote(SqlUtils.validateSqlString(attribute, ATTRIBUTE));
 
+    // Unable to cast numbers to dates or timestamps.
+    // Convert number to timestamp using to_timestamp.
+    if (dataType.getBaseType() == DataTypeMapping.NUMBER
```
jladieu, line 1098: These newly introduced conditionals should use `.equals(` too :)

service/build.gradle
```diff
@@ -229,6 +229,7 @@ jacocoTestReport {
 test {
     systemProperties['pact.rootDir'] = "$buildDir/pacts"
     systemProperties['pact.provider.version'] = "$project.version"
+    systemProperties['user.timezone'] = "UTC";
```
davidangb, line 232: :+1: for this code change - could you explain it to the team in a sprint review or team topics or some such?

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -1089,6 +1093,39 @@ private String getPostgresTypeConversionExpression(
...
+        }
+        expression = "(" + expression + ")";
+      } else {
+        expression = "extract(epoch from " + expression + ")";
+        // Truncate to an integer for dates.
+        if (dataType.getBaseType() == DataTypeMapping.DATE) {
+          expression += "::bigint";
+        }
+      }
+    }
```
davidangb, line 1127: I think this needs Product input too - is this an appropriate conversion? Or should we disallow conversion between these datatypes? Any change in functionality here would of course extend to the `convert_array_of_timestamps_to_numbers` function

service/src/main/resources/liquibase/changesets/20240118_number_timestamp_array_conversions.yaml
```diff
@@ -0,0 +1,37 @@
...
+                 declare
+                    timestamps alias for $1;
+                    numbers numeric[];
+                  begin
+                    for i in array_lower(timestamps, 1)..array_upper(timestamps, 1) loop
+                      numbers[i] := extract(epoch from timestamps[i]);
+                    end loop;
+                    return numbers;
+                  end;
+               $$ language plpgsql stable;
```
davidangb, line 37: Could you declare these functions as [IMMUTABLE](https://www.postgresql.org/docs/14/xfunc-volatility.html), as a best practice? I believe these are immutable, correct? 

Also, for discussion: this creates these two functions inside the `public` schema. This is consistent with where we created the file, relation, array_of_file, and array_of_relation domains. But, should we create these functions in the `sys_wds` schema instead, and at some future point move our domains into `sys_wds` as well?

davidangb (commented):
questions/thoughts inline, nice work on this!

service/src/main/resources/liquibase/changesets/20240118_number_timestamp_array_conversions.yaml
nawatts: Ooh, I learned about STABLE from StackOverflow, but did not know about IMMUTABLE.

The `sys_wds` schema question is beyond my Postgres knowledge.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
nawatts: I did take a decision here on making numbers represent seconds since epoch when converting between numbers and dates or date/times. From a programmer's perspective, I think that's a natural conversion, though one _could_ argue for milliseconds vs seconds, etc.

service/build.gradle
yuliadub: yay UTC!

nawatts:
Rebased after addressing feedback on #471

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/472/commits/2d11823bbcb838694d446c5077c959cc54c59f5f.

service/src/main/resources/liquibase/changesets/20240118_number_timestamp_array_conversions.yaml
nawatts: Marked these functions immutable in https://github.com/DataBiosphere/terra-workspace-data-service/pull/472/commits/59d6c41ca72180399ef4a0776a22add7cc941012. Moved them to the `sys_wds` schema in https://github.com/DataBiosphere/terra-workspace-data-service/pull/472/commits/ce1a49d4328ca16757b4e0fd4e0d76d8ecf9d2ae.

service/build.gradle
nawatts: When converting between datetimes and strings or between dates/datetimes and numbers, the timezone for the database session affects the result. We want the tests to pass with the same fixture data regardless of what timezone they're run in (UTC for CI, US Eastern or Pacific for developers' machines). To accomplish that, this  overrides the system timezone and to ensure tests are always run with the same timezone setting.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
nawatts: Saving this for another PR since this one's already hundreds of lines.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -1079,12 +1084,16 @@ public void updateAttributeDataType(
...
-                .formatted(attribute, newDataType.name()));
+      if (e.getRootCause() instanceof SQLException sqlEx && sqlEx.getSQLState() != null) {
+        if (expectedDataTypeConversionErrorCodes.contains(sqlEx.getSQLState())) {
+          throw new ConflictException(
+              "Unable to convert values for attribute %s to %s"
+                  .formatted(attribute, newDataType.name()));
+        } else {
+          LOGGER.warn(
+              "updateAttributeDataType: DataIntegrityViolationException with unexpected error code {}",
+              sqlEx.getSQLState());
```
nawatts, line 1095: Thought this might help us discover other data type conversion related error codes that we should possibly handle.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
nawatts: @bethsheets

nawatts: Keeping this conversion per [Slack conversation](https://broadinstitute.slack.com/archives/C02SJQ0FKHQ/p1706544435880189?thread_ts=1706538890.918049&cid=C02SJQ0FKHQ) with @bethsheets.

yuliadub (approved):

davidangb (approved):

PR #475, opened 2024-01-26
suggested tweaks to observability setup

davidangb:
Note this is a PR to #461, not to `main`

See my comments inline for explanation.


service/src/main/java/org/databiosphere/workspacedataservice/metrics/ObservationConfig.java
davidangb, line 1: This is why we saw a count of 2 when running 1 job. Spring already connects its default `ObservationRegistry` to its default `MeterRegistry`. But we were adding an additional observation handler which connected it a second time … and thus were recording every metric twice.

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
```diff
@@ -36,16 +39,27 @@
...
+  @Autowired private ObservationRegistry observationRegistry;
+
+  // override Spring's standard ObservationRegistry with a TestObservationRegistry
+  @TestConfiguration
+  static class TestObservationConfig {
+    @Bean
+    ObservationRegistry testObservationRegistry() {
+      return TestObservationRegistry.create();
+    }
+  }
```
davidangb, line 62: This replaces the default `ObservationRegistry` bean with a `TestObservationRegistry` inside this test, instead of keeping the default `ObservationRegistry` bean and adding an additional (non-bean) `TestObservationRegistry`.

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
```diff
@@ -64,16 +78,14 @@ void beforeAll() {
         .thenThrow(new RuntimeException("test failed via jobDao.fail()"));
     when(jobDao.fail(any(), anyString()))
         .thenThrow(new RuntimeException("test failed via jobDao.fail()"));
-    // set up metrics registries
-    testObservationRegistry = TestObservationRegistry.create();
-    Metrics.globalRegistry.add(meterRegistry);
```
davidangb, line 69: this setup is not necessary; Spring handles it for us once we make `TestObservationRegistry` a Spring-managed bean.

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
```diff
@@ -64,16 +78,14 @@ void beforeAll() {
...
 
-  @AfterAll
-  void afterAll() {
-    testObservationRegistry.clear();
+  /** clear all observations and metrics prior to each test */
+  @BeforeEach
+  void beforeEach() {
     meterRegistry.clear();
     Metrics.globalRegistry.clear();
+    ((TestObservationRegistry) observationRegistry).clear();
```
davidangb, line 88: previously, this cleared everything after all tests had run. There's no point to that, as Spring will destroy the context after the whole test finishes anyway.

Instead, I am now clearing everything before each test; this prevents metrics recorded in one test bleeding into the metrics being read by another test.

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
```diff
@@ -142,14 +158,35 @@ void correctMetrics() throws JobExecutionException {
...
+    assertNotNull(longTaskTimer);
     Timer timer = meterRegistry.find("wds.job.execute").timer();
-    assertNotEquals(0, timer.count());
+    assertNotNull(timer);
+
+    // the LongTaskTimer for wds.job.execute.active tracks "in-flight long-running tasks".
+    // since the job we just ran is complete, we expect it to show zero active tasks
+    // and zero duration.
+    assertEquals(0, longTaskTimer.activeTasks());
+    assertEquals(0, longTaskTimer.duration(TimeUnit.SECONDS));
```
davidangb, line 178: previously, this test asserted that `longTaskTimer.duration` was nonzero. The fact it was nonzero seems to be an artifact of recording the same observation twice to the same meterRegistry. Since this is the "*active" metric, it should only be recording currently-in-flight observations, and we should expect it to be zero. Once I cleared up the double-recording, it is now zero.

service/src/main/java/org/databiosphere/workspacedataservice/metrics/ObservationConfig.java
davidangb, line 1: This is why we saw a count of 2 when running 1 job. Spring already connects its default `ObservationRegistry` to its default `MeterRegistry`. But we were adding an additional observation handler which connected it a second time … and thus were recording every metric twice.

davidangb, line 1: This is why we had a count of 2 after executing a job once. Spring already connects the `ObservationRegistry` to the `MeterRegistry`.  But here, we were connecting it a second time … and thus, we were recording each observation twice.

jladieu: Dang, introducing this config was my suggestion...I didn't realize Spring was already providing this exact thing out of the box.

I _think_ this perhaps remains a valuable reference snippet for how we might want to customize the observation config when we have a reason to do so, so I wanted to ask to check my own comprehension:

If we just leave the Config class in place and only delete line 23 (`.observationHandler(...)`), would it have the same/desired effect as dropping the whole file?  I imagine wanting to re-introduce this class at some point for the purpose of overriding config, but don't want to cause a regression.

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
```diff
@@ -36,16 +39,27 @@
...
   @MockBean JobDao jobDao;
-  @Autowired ObservationRegistry observationRegistry;
   @Autowired MeterRegistry meterRegistry;
-  private TestObservationRegistry testObservationRegistry;
+
+  @Autowired private ObservationRegistry observationRegistry;
+
+  // override Spring's standard ObservationRegistry with a TestObservationRegistry
+  @TestConfiguration
+  static class TestObservationConfig {
```
jladieu, line 57: Would you consider making this config universal for all tests?  If that's possible, it might be nice to provide an accessor method for the global `TestObservationRegistry` that tests can use instead of having to `@Autowire ObservationRegistry` and typecast anywhere that's needed.  The typecasting feels clumsy and it'd be nice if it wasn't necessary.

jladieu (approved):
Nice sleuthing!

service/src/main/java/org/databiosphere/workspacedataservice/metrics/ObservationConfig.java
yuliadub: tricky!

yuliadub (approved):
thanks for digging into this David - and the easy explanation of what you found through all the comments.

service/src/main/java/org/databiosphere/workspacedataservice/metrics/ObservationConfig.java
davidangb: @jladieu my $.02 would be to look back at this PR, or the PR that introduced this code, if we need a reference in the future instead of keeping it in the runtime codebase but doing nothing. That said, yeah, I think just removing the `.observationHandler(...)` line would have the same effect.

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
davidangb: Unfortunately making this config universal for all tests isn't super easy. I've moved the static inner class to a reusable standalone class … but it still requires any test that needs it to include it via `@Import` and to perform the appropriate casting.

fwiw, I kinda like the inclusion of
```java
    // this test requires a TestObservationRegistry
    TestObservationRegistry testObservationRegistry =
        assertInstanceOf(TestObservationRegistry.class, observationRegistry);
```
in the unit test, as it ensures the unit test is set up correctly.

davidangb:
@ashanhol feel free to merge this at will; I didn't want to merge directly into your branch myself.

service/src/main/java/org/databiosphere/workspacedataservice/metrics/ObservationConfig.java
```diff
@@ -1,25 +0,0 @@
...
-  public ObservationConfig(MeterRegistry meterRegistry) {
-    this.meterRegistry = meterRegistry;
-  }
-
-  @Bean
-  ObservationRegistryCustomizer<ObservationRegistry> useAutowiredMeterRegistry() {
-    return observationRegistry ->
-        observationRegistry
-            .observationConfig()
-            .observationHandler(new DefaultMeterObservationHandler(meterRegistry));
```
ashanhol, line 23: So I'm a little confused. Without explicitly defining the DefaultMeterObservationHandler in the normal ObservationConfig, i noticed the timers were coming out 0 in my manual testing pulling the prometheus dump, even though theoretically a default one is supposed to be created. Was that not a problem for you?

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
ashanhol: I'm glad that creating a TestObservationRegistry bean does get Metrics to work (it didn't in my implementation, hence why I had to create a test and a regular observation registry)

service/src/main/java/org/databiosphere/workspacedataservice/metrics/ObservationConfig.java
davidangb: > the timers were coming out 0

which timer, specifically? My understanding is that the `wds.job.execute` timer tracks historical data and persists timing info after the job is done. But, `wds.job.execute.active` - note the "active" suffix - is only supposed to track jobs that are currently in flight. So, the `wds.job.execute.active` timer _should_ be zero unless you catch it at exactly the right time, when a job has started but not yet completed.

When I ran manually via `bootRun`, I saw the expected results in the prometheus scrape API.

ashanhol: ah shoot it was active seconds so yeah I went out of my way to give wrong results. I had a really hard time finding definitions for all the metrics (timers, max, etc), how did you find clarification?

davidangb: I noticed the definition when reading the javadoc/comments for `LongTaskTimer`. I was clicking around in IntelliJ but you can see the comments here, too: https://www.javadoc.io/doc/io.micrometer/micrometer-core/latest/io/micrometer/core/instrument/LongTaskTimer.html

> A long task timer is used to track the total duration of all in-flight long-running tasks and the number of such tasks.

"in-flight" was the key description.

PR #480, opened 2024-01-31
Make `InstanceInitializer` load only for Azure.

jladieu:
For [AJ-1558](https://broadworkbench.atlassian.net/browse/AJ-1558): prove out how we might use annotation-based logic branches for deploying GCP or Azure-specific components.

The first commit is a prefactor to simplify `InstanceInitializer`, and is easier to review independently.

[AJ-1558]: https://broadworkbench.atlassian.net/browse/AJ-1558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service/src/test/java/org/databiosphere/workspacedataservice/InstanceInitializerTest.java
```diff
@@ -166,42 +166,47 @@ void cloneWithCloneTableAndNoInstance() {
...
   }
 
   @Test
   void sourceAndCurrentWorkspaceIdsMatch() {
-    boolean cloneMode = instanceInitializerBean.isInCloneMode(workspaceId);
+    boolean cloneMode = instanceInitializer.isInCloneMode(workspaceId);
     assertFalse(cloneMode);
   }
+
+  private void cleanupInstanceSchemas() {
```
jladieu, line 208: I added this to support reusing it in both the `@BeforeEach` and `@AfterEach` methods.

service/src/test/java/org/databiosphere/workspacedataservice/InstanceInitializerTest.java
```diff
@@ -85,21 +85,21 @@ class InstanceInitializerBeanTest {
   void setUp() throws InterruptedException {
     when(mockLock.tryLock(anyLong(), any())).thenReturn(true);
     when(registry.obtain(anyString())).thenReturn(mockLock);
+
+    cleanupInstanceSchemas();
```
jladieu, line 89: This is one of the only manual changes I had to make...the test started failing due to the instance schema lingering between tests, so I added this extra `cleanupInstanceSchemas()` call to happen before each test.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializer.java
jladieu, line 1: Note to reviewer: this looks like an enormous diff, but all I did here was copy the old (small) contents of `InstanceInitializer.java` into `InstanceInitializerBean.java`, and then rename `InstanceInitializerBean.java` to `InstanceInitializer.java`, effectively merging the two classes.

I'll try to call out the interesting bits to reduce diff burden.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializerConfig.java
jladieu, line 1: This class went away and instead is taken care of by `InstanceInitializerBean` declaring itself as a `@Component`.

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializer.java
```diff
@@ -1,23 +1,313 @@
...
+    this.leoDao = leoDao;
+    this.wdsDao = wdsDao;
+    this.cloneDao = cloneDao;
+    this.restoreService = restoreService;
+    this.lockRegistry = lockRegistry;
   }
 
   @Override
   public void onApplicationEvent(@NotNull ContextRefreshedEvent event) {
-    instanceInitializerBean.initializeInstance();
```
jladieu, line 21: this got changed simply to `this.initializeInstance()`;

Everything that follows is code moved in from `InstanceInitializerBean.java`

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializer.java
```diff
@@ -1,23 +1,313 @@
...
+import org.springframework.dao.DataAccessException;
+import org.springframework.http.HttpStatus;
+import org.springframework.integration.support.locks.LockRegistry;
 import org.springframework.stereotype.Component;
 
 @Component
 @Profile({"!local"})
 public class InstanceInitializer implements ApplicationListener<ContextRefreshedEvent> {
 
-  private final InstanceInitializerBean instanceInitializerBean;
```
jladieu, line 13: This lone field and single-arg constructor got replaced with all of the instance state previously in the `InstanceInitializerBean`.  It's now all inline here.

jladieu:
I don't have a straightforward path to resolving the integrationTest failures, so I tried to do this without the refactor in this alternate PR: https://github.com/DataBiosphere/terra-workspace-data-service/pull/483

service/src/main/java/org/databiosphere/workspacedataservice/annotations/CloudPlatform.java
```diff
@@ -0,0 +1,23 @@
...
+  @Retention(RetentionPolicy.RUNTIME)
+  @ConditionalOnProperty(value = "env.wds.cloudPlatform", havingValue = "gcp")
+  @interface GCP {}
+
+  /** The annotated type or method is conditionally enabled when the cloud platform is azure. */
+  @Target({ElementType.TYPE, ElementType.METHOD})
+  @Retention(RetentionPolicy.RUNTIME)
+  @ConditionalOnProperty(value = "env.wds.cloudPlatform", havingValue = "azure")
+  @interface Azure {}
+}
```
davidangb, line 23: I know this is just a POC - eventually, we might wanna make these a little more generic, like using `@DataPlane` vs `@ControlPlane`, or something like that, so we don't build ourselves into a corner where we're trying to run WDS in the data plane in GCP. I suppose we could also just rename the annotations at that point.

jladieu: A code-based rename would be super easy and I'd be happy to do that now or later.

But do you think the environment variable should be something different?  Since that spans the terra helmfile I think that'd be more annoying to rename down the road...would it still be `env.wds.cloudPlatform` or can you think of a better name.

jladieu:
The refactor seems to have messed up the integration tests and I haven't been able to unravel it, so I redid this PR over here without the refactor: https://github.com/DataBiosphere/terra-workspace-data-service/pull/480

PR #481, opened 2024-02-01
AJ-1548: Connect to prometheus rsocket proxy

jladieu:
Ticket [AJ-1548](https://broadworkbench.atlassian.net/browse/AJ-1548)

* Enabled in bees
* Disabled everywhere else
* Can be enabled locally with spring-profile `local-promprox`

[AJ-1548]: https://broadworkbench.atlassian.net/browse/AJ-1548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/resources/application-bee.yml
+++ b/service/src/main/resources/application-bee.yml
@@ -5,3 +5,9 @@ datarepourl: https://data.${beename}.bee.envs-terra.bio
 workspacemanagerurl: https://workspace.${beename}.bee.envs-terra.bio
 leoUrl: https://leonardo.${beename}.bee.envs-terra.bio
 
+# Note: this will probably get re-nested under `prometheus` if prometheus-rsocket-proxy is upgraded to support
+# spring-boot 3+
+management.metrics.export.prometheus.rsocket:
+  enabled: true
+  host: promproxy.bee.envs-terra.bio
+
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -38,6 +38,7 @@ jib {
 ext {
     parquet_mr_version = "1.13.1"
     hadoop_version = "3.3.6"
+    prometheus_rsocket_version = "1.5.3"
 }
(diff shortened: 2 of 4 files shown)


nawatts (approved):

service/src/main/resources/application-bee.yml
```diff
@@ -5,3 +5,9 @@ datarepourl: https://data.${beename}.bee.envs-terra.bio
 workspacemanagerurl: https://workspace.${beename}.bee.envs-terra.bio
 leoUrl: https://leonardo.${beename}.bee.envs-terra.bio
 
+# Note: this will probably get re-nested under `prometheus` if prometheus-rsocket-proxy is upgraded to support
+# spring-boot 3+
+management.metrics.export.prometheus.rsocket:
+  enabled: true
+  host: promproxy.bee.envs-terra.bio
```
nawatts, line 12: Is there one proxy for all BEEs? Is there a case where you'd want to / be able to run a proxy in your own BEE?

jladieu: One proxy to rule them all!

I think if we ever need to fork/experiment with changes to prometheus-rsocket-proxy itself we would probably need to figure out a way to override this too, but the intent of having all the bees funnel to the same proxy is to get (slightly) more lifelike and interesting metrics that we can work on dashboards for, and this more closely matches how production will work anyway, which is that all the LZs will phone home to the same proxy.

mflinn-broad:
there's a hacky demo dashboard I threw together this morning just to show a basic example of this working this morning [here](https://grafana.dsp-devops.broadinstitute.org/d/gVx5KLtSz/mikes-wds-data-plane-metrics-demo?orgId=1)

PR #483, opened 2024-02-01
AJ-1558: Make `InstanceInitializerBean` data-plane-only

jladieu:
For [AJ-1558](https://broadworkbench.atlassian.net/browse/AJ-1558): prove out how we might use annotation-based logic branches for deploying control-plane or data-plane-specific components.

In this case, we want `InstanceInitializerBean` to only be configured in DataPlane.

See `DeploymentMode.java` for the actual annotations.

[AJ-1558]: https://broadworkbench.atlassian.net/browse/AJ-1558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ashanhol:
In the description, could you link to the PR where you implemented annotations for reference?

service/src/main/java/org/databiosphere/workspacedataservice/annotations/CloudPlatform.java
```diff
@@ -0,0 +1,23 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+
+/** Conditional annotations for cloud platform specific beans. */
+public interface CloudPlatform {
+
+  /** The annotated type or method is conditionally enabled when the cloud platform is gcp. */
+  @Target({ElementType.TYPE, ElementType.METHOD})
```
ashanhol, line 13: I was going through all the types you could include. I wonder if we might also want PACKAGE since we know there's specific pub/sub behavior thats gcp only?

service/src/test/java/org/databiosphere/workspacedataservice/annotations/CloudPlatformTest.java
```diff
@@ -0,0 +1,53 @@
...
+      includeFilters = {
+        @Filter(type = ANNOTATION, classes = GCP.class),
+        @Filter(type = ANNOTATION, classes = Azure.class)
+      })
+  static class MinimalTestConfiguration {}
+
+  private ConfigurableApplicationContext loadApplicationContext(String... properties) {
+    ConfigurableEnvironment environment = new StandardEnvironment();
+
+    TestPropertyValues.of(properties).applyTo(environment);
```
ashanhol, line 47: This is really cool, I had no idea you could granularly provision a specific environment like this!

service/src/test/java/org/databiosphere/workspacedataservice/annotations/CloudPlatformTest.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  @Test
+  void instanceInitializerDisabledForGcp() {
+    var context = loadApplicationContext("env.wds.cloudPlatform=gcp");
+    assertThat(context.containsBean("instanceInitializerBean")).isFalse();
+  }
+
+  /* This is a minimal configuration for testing just the beans annotated with GCP & Azure. */
+  @TestConfiguration
```
ashanhol, line 35: [nit] I'm personally a fan of putting setup and configuration things before the tests. It makes it look less magic if you can see what you're defining before using

service/src/main/java/org/databiosphere/workspacedataservice/annotations/CloudPlatform.java
jladieu: The idea of annotating at package scope is a really good one, but I think we have to do that at build time as part of the gradle lifecycle.

This annotation essentially subclasses `@ConditionalOnProperty` annotation ([ref](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/condition/ConditionalOnProperty.html)) which is limited to being used on `TYPE` and `METHOD` only, so I don't think I can expand the scope any further without having to implement my own version of the aliased annotation.

service/src/test/java/org/databiosphere/workspacedataservice/annotations/CloudPlatformTest.java
jladieu: Neither did I!  I figured out how to do this by plagiarizing spring-boot's own [tests](https://github.com/spring-projects/spring-boot/blob/f9363569abbbce80c2c4f318d936365d9a7678ea/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/condition/ConditionalOnPropertyTests.java#L49) on the @ConditionalOnProperty annotation that I based this off of.

service/src/test/java/org/databiosphere/workspacedataservice/annotations/CloudPlatformTest.java
jladieu: I pulled this up to the top of the file in 157dc502f5a67f16858bdf1cdcb0fc6961d88ac1

I still left the `private` method that relies on it tucked at the bottom of the class, like a footnote, per my vague philosophy on code organization, elaborated in more detail [here](https://broadinstitute.slack.com/archives/C02SJQ0FKHQ/p1704819206137009?thread_ts=1704818907.823259&cid=C02SJQ0FKHQ).

jladieu:
> In the description, could you link to the PR where you implemented annotations for reference?

Added a comment in the PR description, but the annotations are introduced _in this PR_  within the new `CloudPlatform.java` file.

ashanhol (approved):
LGTM

service/src/main/java/org/databiosphere/workspacedataservice/InstanceInitializer.java
```diff
@@ -10,14 +12,16 @@
 @Profile({"!local"})
 public class InstanceInitializer implements ApplicationListener<ContextRefreshedEvent> {
```
davidangb, line 13: how would you feel about annotating this - `InstanceInitializer`, which is a `@Component` - with `@Azure` instead of `InstanceInitializerConfig`? There would then be no need to use an `Optional`.

davidangb: potentially - I haven't tested it - you could still annotate both. But because `InstanceInitializer` is annotated and won't get constructed, it won't complain if `InstanceInitializerBean` is null.

jladieu: That did the trick.

service/src/main/java/org/databiosphere/workspacedataservice/annotations/DeploymentMode.java
```diff
@@ -0,0 +1,28 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+
+/** Conditional annotations for deployment mode-specific beans. */
+public interface DeploymentMode {
+
+  /**
+   * The annotated type or method is conditionally enabled when the deployment mode is
+   * control-plane.
```
ashanhol, line 14: I know we're trying to keep cloud specifics out of this, but I still think it'd be helpful to have a mention in the comment "this is primarily used for GCP control plane services" and "this is primarily used for Azure Landing Zone deployment". We can change the comments when we add another (AWS, WDS @ LZ, etc)

davidangb (approved):
nice!

service/src/main/java/org/databiosphere/workspacedataservice/annotations/DeploymentMode.java
jladieu: Added 0e74ffcf2bac4e4a03718afdcdd0d67d73119502 to clarify

service/src/main/java/org/databiosphere/workspacedataservice/annotations/DeploymentMode.java
```diff
@@ -0,0 +1,29 @@
...
+   * control-plane. As of this writing, this equates to cWDS in GCP.
+   */
+  @Target({ElementType.TYPE, ElementType.METHOD})
+  @Retention(RetentionPolicy.RUNTIME)
+  @ConditionalOnProperty(value = "env.wds.deploymentMode", havingValue = "control-plane")
+  @interface ControlPlane {}
+
+  /**
+   * The annotated type or method is conditionally enabled when the deployment mode is data-plane.
+   * As of this writing, this equates to WDS in Azure Landing Zone.
```
ashanhol, line 23: a;lksjdf;lsakdjf nit: this implies we're running WDS at the LZ level (which we're not). "This equates to running WDS in an Azure workspace" might make more sense. No need to change it unless you're making other code changes tho

jladieu: Pushed an amendment, PTAL.

ashanhol (approved):

PR #485, opened 2024-02-01
AJ-1561: Associate import jobs with their containing instance

davidangb:
* Add an `instance_id` column to the `sys_wds.job` table
* When saving an async job to the `sys_wds.job` table, persist the instance id to the new `instance_id` column
* Add `instanceId` to the `JobV1` model in swagger and our clients. Since this is additive, and not required, it should be a non-breaking change. Please check my assumption here!
* Add a new `InstanceId` model class which wraps UUID. How do folks feel about this? I think it could be good to start using this in more places - as well as add `WorkspaceId` and `JobId` classes

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/shared/model/InstanceId.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/shared/model/InstanceId.java
@@ -10,6 +10,13 @@ import java.util.UUID;
  */
 public record InstanceId(UUID id) {
 
+  // disallow nulls
+  public InstanceId {
+    if (id == null) {
+      throw new IllegalArgumentException("Id cannot be null");
+    }
+  }
+
   /**
    * Create a new InstanceId using the given id
    *
@@ -22,6 +29,6 @@ public record InstanceId(UUID id) {
 
   @Override
   public String toString() {
-    return this.id() == null ? "null" : this.id().toString();
+    return this.id().toString();
   }
 }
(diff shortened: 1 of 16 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/shared/model/InstanceId.java
```diff
@@ -0,0 +1,34 @@
+package org.databiosphere.workspacedataservice.shared.model;
+
+import java.util.UUID;
+
+/**
+ * Model to represent an instance id, which currently is a UUID. Since we use UUIDs throughout our
+ * code for multiple use cases, this wrapper class exists to help disambiguate between those UUIDs.
+ *
```
calypsomatic, line 8: I think this is great for reducing confusion!

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/InstanceId.java
```diff
@@ -0,0 +1,34 @@
+package org.databiosphere.workspacedataservice.shared.model;
+
+import java.util.UUID;
+
+/**
```
jladieu, line 5: Love this :)

jladieu (approved):
Looks great!

service/src/test/java/org/databiosphere/workspacedataservice/dao/PostgresJobDaoTest.java
```diff
@@ -61,6 +65,7 @@ private GenericJobServerModel assertJobCreation(JobType jobType) {
         () ->
             namedTemplate.queryForObject(
                 "select id from sys_wds.job where id = :jobId and type = :type and status = :status "
+                    + "and instance_id = :instanceId "
```
davidangb, line 68: because this helper method `assertJobCreation` is used in multiple unit tests, tweaking it to check the instanceid gives good test coverage of the changes in this PR

PR #486, opened 2024-02-02
AJ-1594 convert logging to json

calypsomatic:
https://broadworkbench.atlassian.net/browse/AJ-1594

Configures logging to output in JSON format, which is required for control plane services, only in the case that the property `deploymentMode` is `control-plane`.  Otherwise keeps the non-JSON format.

Without JSON:
(local)
Screenshot 2024-02-02 at 10 22 52 AM
(in Azure portal)
Screenshot 2024-02-02 at 2 21 10 PM

With JSON:
(local)
Screenshot 2024-02-02 at 10 22 26 AM
(in Azure portal)
Screenshot 2024-02-02 at 2 20 57 PM

Reminder:

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -65,6 +65,7 @@ dependencies {
     implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-csv'
     implementation 'io.sentry:sentry-logback:6.34.0'
     implementation 'net.logstash.logback:logstash-logback-encoder:7.3'
+    implementation 'org.codehaus.janino:janino:3.1.11'
     implementation 'org.liquibase:liquibase-core'
     implementation 'javax.cache:cache-api'
     implementation 'org.ehcache:ehcache:3.10.8:jakarta'
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -65,7 +65,7 @@ dependencies {
     implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-csv'
     implementation 'io.sentry:sentry-logback:6.34.0'
     implementation 'net.logstash.logback:logstash-logback-encoder:7.3'
-    implementation 'org.codehaus.janino:janino:3.1.11'
+    implementation 'org.codehaus.janino:janino'
     implementation 'org.liquibase:liquibase-core'
     implementation 'javax.cache:cache-api'
     implementation 'org.ehcache:ehcache:3.10.8:jakarta'
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -64,6 +64,7 @@ dependencies {
     implementation 'org.webjars:webjars-locator-core' // versioned by spring dependency management
     implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-csv'
     implementation 'io.sentry:sentry-logback:6.34.0'
+    implementation 'net.logstash.logback:logstash-logback-encoder:7.3'
     implementation 'org.liquibase:liquibase-core'
     implementation 'javax.cache:cache-api'
     implementation 'org.ehcache:ehcache:3.10.8:jakarta'
--- a/service/src/main/resources/logback-spring.xml
+++ b/service/src/main/resources/logback-spring.xml
@@ -10,6 +10,10 @@
   <include resource="org/springframework/boot/logging/logback/defaults.xml"/>
   <include resource="org/springframework/boot/logging/logback/console-appender.xml"/>
 
+  <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
+    <encoder class="net.logstash.logback.encoder.LogstashEncoder" />
+  </appender>
+
   <appender name="Sentry" class="io.sentry.logback.SentryAppender">
     <!-- Optionally change minimum Event level. Default for Events is ERROR -->
     <minimumEventLevel>ERROR</minimumEventLevel>
--- a/service/src/main/resources/logback-spring.xml
+++ b/service/src/main/resources/logback-spring.xml
@@ -6,7 +6,7 @@
   <!-- inspired by Spring default ${FILE_LOG_PATTERN} from defaults.xml, modified to include requestId -->
   <property name="WDS_LOG_PATTERN"
             value="%d{${LOG_DATEFORMAT_PATTERN:-yyyy-MM-dd HH:mm:ss.SSS}} ${LOG_LEVEL_PATTERN:-%5p} [%X{requestId}] [%t] %-40.40logger{39} : %m%n${LOG_EXCEPTION_CONVERSION_WORD:-%wEx}"/>
-
+  
   <include resource="org/springframework/boot/logging/logback/defaults.xml"/>
   <include resource="org/springframework/boot/logging/logback/console-appender.xml"/>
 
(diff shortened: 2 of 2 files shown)


yuliadub:
@jladieu mentioned that it was possible for JSON logging to have an impact on system that consume WDS logs today but I cant seem to remember all the details of that

service/src/main/resources/logback-spring.xml
```diff
@@ -10,6 +10,8 @@
   <include resource="org/springframework/boot/logging/logback/defaults.xml"/>
   <include resource="org/springframework/boot/logging/logback/console-appender.xml"/>
 
+
+
```
ashanhol, line 14: optional nit: if you happen to make any other code changes maybe delete this whitespace

ashanhol:
+1 to yulia, I thought david actually said that library breaks things if turned on. I'm curious to know if this works in a bee+kibana?

calypsomatic:
> +1 to yulia, I thought david actually said that library breaks things if turned on. I'm curious to know if this works in a bee+kibana?

So WDS logs don't normally show up in kibana, right?  I think it's just control plane apps that send their logs there (unless I've been dumb this whole time).  But the logs are still showing up in azure portal, although it's now strictly worse to look at.  Is there some more formatting that could fix this?  Can/should we make a control-vs-data profile that determines the logging format??

nawatts:
> I thought david actually said that library breaks things if turned on

I believe that was referring to [terra-common-lib](https://github.com/DataBiosphere/terra-common-lib).

davidangb:
Another thought: what would it take to teach Azure how to understand JSON? https://learn.microsoft.com/en-us/azure/azure-monitor/agents/data-collection-text-log?tabs=portal

service/src/main/resources/logback-spring.xml
```diff
@@ -18,6 +18,18 @@
   </appender>
 
   <springProfile name="!local">
+    <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
+      <if condition='isDefined("deploymentMode") &amp;&amp; property("deploymentMode").equals("control-plane")'>
```
calypsomatic, line 22: This seems to work even though in `application.yaml` `deploymentMode` is nested under `env.wds`, but someone may want to verify to ensure I'm not crazy

service/build.gradle
```diff
@@ -64,6 +64,8 @@ dependencies {
     implementation 'org.webjars:webjars-locator-core' // versioned by spring dependency management
     implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-csv'
     implementation 'io.sentry:sentry-logback:6.34.0'
+    implementation 'net.logstash.logback:logstash-logback-encoder:7.3'
+    implementation 'org.codehaus.janino:janino:3.1.11'
```
ashanhol, line 68: what is janino? i know the logback encoder is for json but not sure what this other package is

calypsomatic: Necessary to use the conditional clause in logback

ashanhol (approved):
If this has been tested and only works for cWDS then this looks to be a great step forward :)

service/build.gradle
davidangb, line 68: nit: janino versions are [managed by Spring Boot](https://docs.spring.io/spring-boot/docs/3.2.2/reference/html/dependency-versions.html#appendix.dependency-versions), so leave the `:3.1.11` off.

calypsomatic:
> I'm still seeing warnings being logged about http://logback.qos.ch/codes.html#nested_if_element, could you address those?

@davidangb I think I have those fixed now, although it does look like there is a lot of extra chatter that I think wasn't there before, so let me know if there's anything else in there that seems like it should also be addressed.

service/src/main/resources/logback-spring.xml
```diff
@@ -18,6 +20,18 @@
...
+    <if condition='isDefined("deploymentMode") &amp;&amp; property("deploymentMode").equals("control-plane")'>
+      <then>
+        <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
+          <encoder class="net.logstash.logback.encoder.LogstashEncoder"/>
+        </appender>
+        </then>
+        <else>
+          <encoder>
+            <pattern>${WDS_LOG_PATTERN}</pattern>
+          </encoder>
```
davidangb, line 32: I think this needs to declare a full `<appender>` as well, not just a bare `<encoder>`. We have been using the console appender from org/springframework/boot/logging/logback/console-appender.xml, so if we copy/paste its definition into here and use `WDS_LOG_PATTERN`, it would be:
```suggestion
	<appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
		<filter class="ch.qos.logback.classic.filter.ThresholdFilter">
			<level>${CONSOLE_LOG_THRESHOLD}</level>
		</filter>
		<encoder>
			<pattern>${WDS_LOG_PATTERN}</pattern>
			<charset>${CONSOLE_LOG_CHARSET}</charset>
		</encoder>
	</appender>
```

Interesting: I broke usage of `WDS_LOG_PATTERN` in #440, we haven't been using at all since then.

service/src/main/resources/logback-spring.xml
```diff
@@ -18,6 +20,18 @@
   </appender>
 
   <springProfile name="!local">
+    <if condition='isDefined("deploymentMode") &amp;&amp; property("deploymentMode").equals("control-plane")'>
+      <then>
+        <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
+          <encoder class="net.logstash.logback.encoder.LogstashEncoder"/>
+        </appender>
```
davidangb, line 27: This overwrites the `CONSOLE` definition from org/springframework/boot/logging/logback/console-appender.xml, if we wanted to replicate it as closely as possible but just use a different encoder, it would be:
```xml
        <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
          <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
            <level>${CONSOLE_LOG_THRESHOLD}</level>
          </filter>
          <encoder class="net.logstash.logback.encoder.LogstashEncoder"/>
          <charset>${CONSOLE_LOG_CHARSET}</charset>
        </appender>
```

service/src/main/resources/logback-spring.xml
```diff
@@ -7,6 +7,8 @@
   <property name="WDS_LOG_PATTERN"
             value="%d{${LOG_DATEFORMAT_PATTERN:-yyyy-MM-dd HH:mm:ss.SSS}} ${LOG_LEVEL_PATTERN:-%5p} [%X{requestId}] [%t] %-40.40logger{39} : %m%n${LOG_EXCEPTION_CONVERSION_WORD:-%wEx}"/>
 
+  <springProperty scope="context" name="deploymentMode" source="env.wds.deploymentMode" defaultValue="data-plane" />
+
   <include resource="org/springframework/boot/logging/logback/defaults.xml"/>
```
davidangb, line 12: you didn't change this in your PR, but this line duplicates line 4

davidangb (commented):
Two comments inline with the goal of changing only what's necessary.

I also notice that when this PR goes through and we return to using the `WDS_LOG_PATTERN` pattern, logging for the local profile will be different than for the !local profile. We probably want to align these, could certainly do that as a follow-on PR.

service/src/main/resources/logback-spring.xml
calypsomatic: Ok is [this](https://github.com/DataBiosphere/terra-workspace-data-service/pull/486/commits/01159c256328a585f9e439df395bd5d9843dec7b) what you were envisioning?  I think we can't use the spring console-appender directly because we're adding an encoder.  Should I just remove that `include resource` line then?

davidangb (approved):
thanks for seeing this one through!

PR #489, opened 2024-02-02
AJ-1591: instance to workspace

davidangb:
### Changes in this PR
* adds `workspace_id`, `name`, and `description` columns to the `sys_wds.instance` table
* via a Liquibase migration, backfill those columns so they can be marked non-null
* add a unique index on `workspace_id` + `name` so we can use that as a future identifier for instances instead of instanceid
* when creating an instance programmatically, smartly populate the new columns

### NOT in this PR; for future work:
* any new APIs that display or allow editing of the new `sys_wds.instance` columns
* any permission changes that would consume the `workspace_id` column value

in other words, this just sets up invisible infrastructure which we'll take advantage of later.

### Open questions:
1. the Liquibase migration is hairy. I like doing it in Liquibase because it keeps all our db-migration code in one place. But, would it be better to write this hairy logic in Java and execute it on startup?
2. The non-nullness of the `workspace_id` column requires that the `$WORKSPACE_ID` env var is populated. This is expected in the Azure data plane today, but was not true in many of our tests and will not be true in GCP. However, we also expect that in GCP control plane, there will be no rows in the `sys_wds.instance` table, so it's a non-issue. Does this feel safe?
3. Should we proactively add the `description` column to `sys_wds.instance` now, while we're also adding the `name` column, or save that for later? See discussion thread at https://github.com/DataBiosphere/terra-workspace-data-service/pull/489#discussion_r1480610380


service/src/test/java/org/databiosphere/workspacedataservice/service/InstanceServiceTest.java
```diff
@@ -92,11 +87,4 @@ void deleteInstance() {
         () -> instanceService.validateInstance(INSTANCE),
         "validateInstance should have thrown an error");
   }
-
-  @Test
-  void getWorkspaceId() {
-    assertEquals(
-        workspaceIdProperty,
-        instanceService.getWorkspaceId(InstanceId.of(UUID.randomUUID())).toString());
-  }
```
davidangb, line 101: moved to `InstanceServiceGetWorkspaceIdTest`

jladieu:
Without actually looking at the PR yet, I have some existing hunches about your open questions:

1) Trust Liquibase with the job of doing the migration.  Adding a Java layer on top of it could just add bugs.  I think if it is still feeling risky, maybe the way to improve confidence would be to add some Java (Gradle?) tasks that just check the db integrity and assumptions before & after the migration, then exercise them by doing:

```
gradlew db:prechecks
gradlew db:migrate
gradlew db:postcheck
gradlew db:rollback
gradlew db:prechecks
```

BTW, this is based on how I used to test db migrations in Rails...I'm not sure if this maps over to how liquibase migrations work.

2) If your code crashes fast and hard when the nullness constraint is violated, then I think that's all the safety we need. We want to detect issues related to the nullness ASAP rather than letting them propagate through the codebase.  Is it possible to enlist some extra help at compilation time by starting to pull in `@Nullable` and `@NonNull` annotations?

build.gradle
```diff
@@ -37,7 +37,12 @@ subprojects {
             mavenBom(SpringBootPlugin.BOM_COORDINATES)
         }
         dependencies {
-            dependency 'org.liquibase:liquibase-core:4.25.1'
+            // latest is 4.25.1
+            // https://github.com/liquibase/liquibase/pull/5391
+            // "This bug existed from 4.22.0 to 4.25.1"
+            // so, we roll back to before 4.22.0
+            dependency 'org.liquibase:liquibase-core:4.21.1'
+
```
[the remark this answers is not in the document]
jladieu: Ok I happened to be looking back in history for something unrelated, and it looks like we only just introduced the version bump to liquibase as part of upgrading to SpringBoot 3: https://github.com/DataBiosphere/terra-workspace-data-service/pull/440/files

So we haven't been running in a version greater than what you'd need to pin to for very long.  I agree with the hesitation about backsliding, but I think in this case it's not a dramatic reversion.

build.gradle
```diff
@@ -37,7 +37,12 @@ subprojects {
             mavenBom(SpringBootPlugin.BOM_COORDINATES)
         }
         dependencies {
-            dependency 'org.liquibase:liquibase-core:4.25.1'
+            // latest is 4.25.1
+            // https://github.com/liquibase/liquibase/pull/5391
+            // "This bug existed from 4.22.0 to 4.25.1"
+            // so, we roll back to before 4.22.0
+            dependency 'org.liquibase:liquibase-core:4.21.1'
```
jladieu, line 44: How do you feel about also documenting this in the service's constraints block, belt & suspenders style, as I've done in similar circumstances: https://github.com/DataBiosphere/terra-workspace-data-service/pull/496/files

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresInstanceDao.java
```diff
@@ -43,9 +50,9 @@ public List<UUID> listInstanceSchemas() {
   @WriteTransaction
   @SuppressWarnings("squid:S2077") // since instanceId must be a UUID, it is safe to use inline
   public void createSchema(UUID instanceId) {
-    namedTemplate
-        .getJdbcTemplate()
-        .update("insert into sys_wds.instance(id) values (?)", instanceId);
+    // insert to instance table
+    insertInstanceRow(instanceId, false);
```
jladieu, line 54:
```suggestion
    insertInstanceRow(instanceId, /* ignoreConflict= */ false);
```

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresInstanceDao.java
```diff
@@ -74,13 +81,52 @@ public void alterSchema(UUID oldSchemaId, UUID newSchemaId) {
...
+            workspaceId,
+            oldSchemaId);
     // ensure new exists in sys_wds.instance. When this alterSchema() method is called after
     // restoring from a pg_dump,
     // the oldSchema doesn't exist, so is not renamed in the previous statement.
-    namedTemplate
-        .getJdbcTemplate()
-        .update(
-            "insert into sys_wds.instance(id) values (?) on conflict(id) do nothing", newSchemaId);
+    insertInstanceRow(newSchemaId, true);
```
jladieu, line 92:
```suggestion
    insertInstanceRow(newSchemaId, /* ignoreConflict= */ true);
```

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresInstanceDao.java
```diff
@@ -14,6 +18,9 @@ public class PostgresInstanceDao implements InstanceDao {
 
   private final NamedParameterJdbcTemplate namedTemplate;
 
+  @Value("${twds.instance.workspace-id}")
+  private UUID workspaceId;
```
jladieu, line 22: I think to support the notion that WDS should fail fast if there's anything fishy with `WORKSPACE_ID` the creation of the `PostgresInstanceDao` bean should be `DeploymentMode` aware and strongly enforce the non-null `workspaceId` assumption, perhaps in both directions.

Maybe sibling bean methods somewhere that roughly do the following:

```
@Value("${twds.instance.workspace-id}")
private UUID workspaceId;

@ControlPlane
InstanceDao instanceDao instanceDao() {
    assert(workspaceId == null, "workspaceId must NOT be set in control plane");
    return // new instance of PosgresInstanceDao using the existing single-arg constructor
}

@DataPlane
InstanceDao instanceDao instanceDao() {
    assertLegitWorkspaceId(workspaceId, "workspaceId must legit in the data plane");
    return // new instance of PosgresInstanceDao using an overloaded two-arg constructor with a `@NonNull WorkspaceId` param
}
```

jladieu: (or, thinking further on this)...maybe the `InstanceDaoProvider` itself should be the bean, and you get one instance or the other based on the DeploymentMode, and the provider either has (or doesn't have) a `WorkspaceId`?

jladieu: Regardless of _how_, I just want to stress the main point which is that it'd be good to fail as early as possible if our assumptions around `WORKSPACE_ID` don't match up with the deploymentMode.

service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
```diff
@@ -129,6 +131,30 @@ public void validateInstance(UUID instanceId) {
    * @return the workspace containing the given instance.
    */
   public WorkspaceId getWorkspaceId(InstanceId instanceId) {
-    return Objects.requireNonNullElseGet(workspaceId, () -> WorkspaceId.of(instanceId.id()));
+    // look up the workspaceId for this instance in the instance table
+    WorkspaceId rowWorkspaceId = null;
+    try {
+      rowWorkspaceId = instanceDao.getWorkspaceId(instanceId);
+    } catch (EmptyResultDataAccessException e) {
+      // swallow not-found errors; it's valid for rows to not exist in the case of
```
jladieu, line 139: This might be a good spot to tick a counter if `WORKSPACE_ID` is present, as it would mean a data-plane WDS instance whose instances table hasn't yet been backfilled.  Having this signal drive to (and remain at) zero might help us build confidence that the data migration is complete for all WDS instances past a given version.

jladieu (changes requested):
After review I think I share your concern about (2) and I think the best way to protect against that is to introduce some kind of mechanism with the bean configuration / construction to crash fast when the `WORKSPACE_ID` assumption != `deploymentMode` assumption.  I offered an idea in the comment but not too hung up on how.

service/src/main/resources/liquibase/changesets/20240131_add_workspace_to_instance.sql
```diff
@@ -0,0 +1,40 @@
...
+set name = 'default'
+where name is null
+  and id = workspace_id;
+
+-- and if id != workspace_id, copy the id into the name
+update sys_wds.instance
+set name = id
+where name is null;
+
+-- finally, set the description, equal to the name
```
jladieu, line 37: Would you consider leaving description blank?

service/src/main/resources/liquibase/changesets/20240131_add_workspace_to_instance.yaml
```diff
@@ -0,0 +1,57 @@
...
+        # Now that the columns are all populated, set them to non-null
+        - addNotNullConstraint:
+            schemaName: sys_wds
+            tableName: instance
+            columnName: workspace_id
+        - addNotNullConstraint:
+            schemaName: sys_wds
+            tableName: instance
+            columnName: name
+        - addNotNullConstraint:
```
jladieu, line 48: Does description need to be non-null?

build.gradle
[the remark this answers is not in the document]
yuliadub: I dont love the downgrade, but I hate implementing in java way more :D

service/src/integrationTest/java/org/databiosphere/workspacedataservice/InstanceInitializerCloneTest.java
```diff
@@ -159,6 +161,18 @@ void cloneSuccess() throws ApiException {
     assertEquals(List.of(workspaceUuid), actualInstances);
     List<RecordType> actualTypes = recordDao.getAllRecordTypes(workspaceUuid);
     assertEquals(List.of(RecordType.valueOf("thing")), actualTypes);
+
+    // the restored instance should be associated with the current workspace, and should
+    // have its name and description populated correctly
+    Map<String, Object> rowMap =
+        namedTemplate.queryForMap(
+            "select id, workspace_id, name, description from sys_wds.instance where id = :id",
```
yuliadub, line 169: is there a reason why we need to store name and description of workspace in wds? from my understanding this is something WSM manages but it is likely I am missing some context here

service/src/main/resources/liquibase/changesets/20240131_add_workspace_to_instance.yaml
yuliadub: why do we need a description? (it seems like it would always be =name?)

yuliadub:
> The non-nullness of the workspace_id column requires that the $WORKSPACE_ID env var is populated. This is expected in the Azure data plane today, but was not true in many of our tests and will not be true in GCP. However, we also expect that in GCP control plane, there will be no rows in the sys_wds.instance table, so it's a non-issue. Does this feel safe?

I am struggling with this one - it almost seems like we may want to have a layer above this that will help guide when it is okay to not have a Workspace ID be populated vs not. I dont think I like the concept of overloading workspace id variable in this context. What if we included a new env param that said which environment WDS is being run in (or something equivalent, like control pane vs application pane, or e**ven just renaming the workspace id env variable and containing the value it can be to default or uuid and then basing logic on what to do next based on how we parse it (maybe I like this one best? )** ), that way we can have much more control over what we do at start up (and hopefully prevent cases where things go really wrong since we strongly type what the value can be) 

going to think more about this...will add more notes as more thoughts come

build.gradle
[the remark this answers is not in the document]
davidangb: now using the just-released 4.26.0, which I feel better about.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresInstanceDao.java
davidangb: As of #509 , all this validation has moved to `StartupConfig`, which should handle validation for InstanceDao as well as any other classes that rely on the `WORKSPACE_ID` value

service/src/main/resources/liquibase/changesets/20240131_add_workspace_to_instance.yaml
davidangb: This is all pretty future-looking and speculative. I am picturing being able to refer to a collection _either_ by its UUID, or by the combination of workspaceId + collection name. In such a case, I expect that collection name will be populated in various API signatures as part of an API's path … which means we will likely want to impose some syntax restrictions on the name, e.g. `a-zA-Z0-9\\-_`. And if we impose syntax restrictions on name … we may want a human-readable description. 

Since the code in this PR is the only place we set the name - we don't yet allow end users to set the name - I haven't implemented any syntax restrictions for it yet.

I'm fine leaving the description column out of this PR entirely and only adding it when we need it … would that be preferable?

build.gradle
davidangb: I don't really like having active code in both places, but I have added a comment in service/build.gradle referencing this change

davidangb:
@jladieu @yuliadub major changes since you last reviewed:
* Liquibase 4.26.0 is out, and this PR uses it
* #509 added validation of the $WORKSPACE_ID env var at startup, and this PR relies on that
* I believe I have replied to or adopted all feedback

Given all this, I am resetting your reviews, thank you!

service/src/main/resources/liquibase/changesets/20240131_add_workspace_to_instance.yaml
jladieu: I have no problem with pre-fetching the description because I think it's a pretty obvious field -- I just don't necessarily think we need to fill it with anything up front as then it becomes something that customers have to clear out if they don't like what it is :)

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresInstanceDao.java
```diff
@@ -74,13 +86,52 @@ public void alterSchema(UUID oldSchemaId, UUID newSchemaId) {
     // rename any rows in sys_wds.instance from old to new
     namedTemplate
         .getJdbcTemplate()
-        .update("update sys_wds.instance set id = ? where id = ?", newSchemaId, oldSchemaId);
+        .update(
```
jladieu, line 89: nit: would you consider extracting an `update` method too?  I like `insertInstanceRow` and think this method's readability would benefit from the update getting the same treatment

service/src/main/java/org/databiosphere/workspacedataservice/service/InstanceService.java
jladieu: I think the case I was thinking of wouldn't be practical/possible...I was thinking it'd be nice to measure how many unmigrated workspaces exist and watch that inventory go to zero.   But for that to happen, the measurement would have to precede this PR and be installed everywhere already.

So I think my comment is a big no-op.

jladieu (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresInstanceDao.java
davidangb: I was actively staying away from this, and thought about renaming this existing method to `adoptSchema` or something like that. The current method is only called from the clone/restore path and is not a generic "make changes to a collection" method. We'll need one of those eventually and it should handle name and description too - I'd look to extract reusability at that point.

service/src/main/resources/liquibase/changesets/20240131_add_workspace_to_instance.yaml
yuliadub: part of me thinks that we should just add when we have a use case for it, but at the same time, it's just a small column and you already did the work to have it in there, so I'm fine if it stays 😄 (especially if it ends up being null for now)

yuliadub (approved):
LGTM - thanks for working through this and all the helpful comments!

PR #491, opened 2024-02-02
[AJ-1279] Refactor check for supported data type conversion

nawatts:
After moving past the basic data types, it's become clear that it's simpler and clearer to explicitly list supported data type conversions vs assuming that any supported data type can be converted to any other supported data type and listing exceptions to that.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -91,6 +91,13 @@ public class RecordDao {
   @Value("${twds.streaming.fetch.size:5000}")
   int fetchSize;
 
+  /**
+   * Each member of this set is expected to be a set of two data types. The presence of a set of
+   * types in this set implies that conversions between the types are supported (in either
+   * direction). Note that arrays do not need to be listed separately here. For example, a set of
+   * STRING and NUMBER implies that all the following conversions are supported: STRING to NUMBER,
+   * NUMBER to STRING, ARRAY_OF_STRING to ARRAY_OF_NUMBER, and ARRAY_OF_NUMBER to ARRAY_OF_STRING.
+   */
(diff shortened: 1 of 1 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -91,23 +91,16 @@ public class RecordDao {
...
-          DataTypeMapping.DATE,
-          DataTypeMapping.ARRAY_OF_DATE,
-          DataTypeMapping.DATE_TIME,
-          DataTypeMapping.ARRAY_OF_DATE_TIME);
-
-  private static final Set<Set<DataTypeMapping>> invalidDataTypeConversions =
-      Set.of(
-          Set.of(DataTypeMapping.BOOLEAN, DataTypeMapping.DATE),
-          Set.of(DataTypeMapping.BOOLEAN, DataTypeMapping.DATE_TIME));
+          Set.of(DataTypeMapping.STRING, DataTypeMapping.NUMBER),
```
calypsomatic, line 96: This means that these can always be converted in either direction?  (i.e. string -> date AND date -> string).  I think it makes sense, just double-checking that that's correct

nawatts: Yes, that's the intent.

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
nawatts: Added a comment explaining this in https://github.com/DataBiosphere/terra-workspace-data-service/pull/491/commits/bfd07447195439e46fea581951bb5c4b6eaa7c49.

davidangb (approved):

PR #492, opened 2024-02-02
AJ-1591: no-db solution for instance-to-workspace association

davidangb:
Create a new method - `InstanceService.getWorkspaceId()` - which returns the proper workspace to use when authenticating to an instance. As compared to https://github.com/DataBiosphere/terra-workspace-data-service/pull/489, this PR makes no changes to the db. Instead, it relies heavily on assumptions which are _currently_ safe:
* in Azure, we deploy WDS per-workspace and set a `$WORKSPACE_ID` value for each deployment
* in the upcoming cWDS, since data tables will continue to be powered by Rawls, we can assume `instanceId=workspaceId`.

I really want to break that last assumption, but I don't want to break it at the expense of convoluted code which could incur bugs. So, here's a simple solution.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/shared/model/InstanceId.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/shared/model/InstanceId.java
@@ -27,6 +27,16 @@ public record InstanceId(UUID id) {
     return new InstanceId(id);
   }
 
+  /**
+   * Create a new InstanceId using the given id
+   *
+   * @param id the instance's id
+   * @return new InstanceId
+   */
+  public static InstanceId fromString(String id) {
+    return InstanceId.of(UUID.fromString(id));
+  }
+
   @Override
   public String toString() {
     return this.id().toString();
--- a/service/src/main/java/org/databiosphere/workspacedataservice/shared/model/WorkspaceId.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/shared/model/WorkspaceId.java
@@ -1,0 +1,41 @@
+package org.databiosphere.workspacedataservice.shared.model;
+
+import java.util.UUID;
+
+/**
+ * Model to represent a workspace id, which currently is a UUID. Since we use UUIDs throughout our
+ * code for multiple use cases, this wrapper class exists to help disambiguate between those UUIDs.
+ *
+ * @param id the workspace's id
+ */
+public record WorkspaceId(UUID id) {
+
+  // disallow nulls
+  public WorkspaceId {
+    if (id == null) {
+      throw new IllegalArgumentException("Id cannot be null");
+    }
+  }
+
+  /**
+   * Create a new WorkspaceId using the given id
+   *
+   * @param id the workspace's id
+   * @return new WorkspaceId
+   */
+  public static WorkspaceId of(UUID id) {
+    return new WorkspaceId(id);
+  }
+
+  /**
+   * Create a new WorkspaceId using the given id
+   *
+   * @par public static WorkspaceId fromString(String id) { return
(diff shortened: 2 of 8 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/shared/model/InstanceId.java
```diff
@@ -27,6 +27,16 @@ public static InstanceId of(UUID id) {
...
+  /**
+   * Create a new InstanceId using the given id
+   *
+   * @param id the instance's id
+   * @return new InstanceId
+   */
+  public static InstanceId fromString(String id) {
+    return InstanceId.of(UUID.fromString(id));
+  }
+
```
davidangb, line 39: not strictly necessary in this PR, just an additional helper method.

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/WorkspaceId.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.UUID;
+
+/**
+ * Model to represent a workspace id, which currently is a UUID. Since we use UUIDs throughout our
+ * code for multiple use cases, this wrapper class exists to help disambiguate between those UUIDs.
+ *
+ * @param id the workspace's id
+ */
+public record WorkspaceId(UUID id) {
```
davidangb, line 11: companion class to `InstanceId`, this one represents a workspace.

service/src/test/java/org/databiosphere/workspacedataservice/service/InstanceServiceTest.java
```diff
@@ -35,17 +35,13 @@
     })
 class InstanceServiceTest {
 
-  private InstanceService instanceService;
+  @Autowired private InstanceService instanceService;
```
davidangb, line 38: This test - and the other `InstanceService*Test` classes in this PR - were set up (probably by me!) in a very un-Spring-like way. They didn't rely on `@Autowired` to create the `InstanceService` objects, and thus didn't inherit any of the test context. Specifically, even though these tests set a `twds.instance.workspace-id` value, `InstanceService` didn't inherit that value. This broke the new test I added. I've cleaned these up to use `@Autowired`.

service/src/test/java/org/databiosphere/workspacedataservice/service/InstanceServiceNoWorkspaceTest.java
```diff
@@ -0,0 +1,39 @@
...
+@TestPropertySource(properties = {"twds.instance.workspace-id="})
+class InstanceServiceNoWorkspaceTest {
+
+  @Autowired private InstanceService instanceService;
+
+  @Value("${twds.instance.workspace-id:}")
+  private String workspaceIdProperty;
+
+  @Test
+  void assumptions() {
```
jladieu, line 27: I like this pattern of calling out assumptions in their own self-documenting test!

service/src/test/java/org/databiosphere/workspacedataservice/service/InstanceServiceSamTest.java
```diff
@@ -31,21 +27,15 @@
 
 @ActiveProfiles(profiles = "mock-instance-dao")
 @DirtiesContext
-@SpringBootTest(
-    classes = {
```
jladieu, line 35: Thanks for all this cleanup here and elsewhere.

jladieu (approved):
I like this PR because it promotes an assumption to a documented and tested implementation.  When it's time to break our assumption (which I'm in agreement we should do), we'll have to express that change in a PR which will include logic and test changes, and serve as a further source of explanation.

PR #497, opened 2024-02-07
[AJ-1279] Support conversions between strings and files (without validation)

nawatts:
Alternative to #493...

This treats the File type as simply an annotation/hint for the user. It allows creating File columns that contain values that are not valid URLs.


nawatts:
Closing until [AJ-1617](https://broadworkbench.atlassian.net/browse/AJ-1617) is resolved.

PR #498, opened 2024-02-07
add read check in getjob

calypsomatic:
Reminder:

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
@@ -90,6 +90,21 @@ public class HttpSamDao implements SamDao {
     return hasPermission(ACTION_WRITE, "Sam.hasWriteInstancePermission", token);
   }
 
+  /**
+   * Check if the current user has permission to read a workspace resource from Sam.
+   *
+   * @return true if the user has permission
+   */
+  @Override
+  public boolean hasReadInstancePermission() {
+    return hasReadInstancePermission(null);
+  }
+
+  @Override
+  public boolean hasReadInstancePermission(String token) {
+    return hasPermission(ACTION_READ, "Sam.hasReadInstancePermission", token);
+  }
+
   // this cache uses token.hashCode as its key. This prevents any logging such as
   // in CacheLogger from logging the raw token.
   @Cacheable(cacheNames = "tokenResolution", key = "#token.hashCode()")
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
@@ -3,6 +3,8 @@ package org.databiosphere.workspacedataservice.service;
 import java.util.UUID;
 import org.databiosphere.workspacedataservice.dao.JobDao;
 import org.databiosphere.workspacedataservice.generated.GenericJobServerModel;
+import org.databiosphere.workspacedataservice.sam.SamDao;
+import org.databiosphere.workspacedataservice.service.model.exception.AuthorizationException;
 import org.databiosphere.workspacedataservice.service.model.exception.MissingObjectException;
 import org.springframework.dao.EmptyResultDataAccessException;
 import org.springframework.stereotype.Service;
@@ -12,15 +14,21 @@ public class JobService {
 
   JobDao jobDao;
   InstanceService instanceService;
+  SamDao samDao;
 
-  public JobService(JobDao jobDao, InstanceService instanceService) {
+  public JobService(JobDao jobDao, InstanceService instanceService, SamDao samDao) {
     this.jobDao = jobDao;
     this.instanceService = instanceService;
+    this.samDao = samDao;
   }
 
   public GenericJobServerModel getJob(UUID jobId) {
     try {
-      return jobDao.getJob(jobId);
+      GenericJobServerModel result = jobDao.getJob(jobId);
+      if (!samDao.hasReadInstancePermission()) {
+        throw new AuthorizationException("Caller does not have permission to view this job.");
+      }
+      return result;
     } catch (EmptyResultDataAccessException e) {
       throw new MissingObjectException("Job");
     }
--- a/service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceTest.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceTest.java
@@ -1,0 +1,86 @@
+package org.databiosphere.workspacedataservice.service;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.BDDMockito.given;
+
+import java.time.OffsetDateTime;
+import java.time.ZoneId;
+import java.util.UUID;
+import org.broadinstitute.dsde.workbench.client.sam.ApiException;
+import org.broadinstitute.dsde.workbench.client.sam.api.ResourcesApi;
+import org.databiosphere.workspacedataservice.dao.JobDao;
+import org.databiosphere.workspacedataservice.generated.GenericJobServerModel;
+import org.databiosphere.workspacedataservice.sam.SamClientFactory;
+import org.databiosphere.workspacedataservice.sam.SamDao;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.test.context.ActiveProfiles;
+
+@ActiveProfiles(profiles = {"mock-sam"})
+@SpringBootTest
+public class JobServiceTest {
+  @MockBean JobDao jobDao;
+  @Autowired SamDao samDao;
+  @Autowired JobService jobService;
+  @MockBean SamClientFactory mockSamClientFactory;
+
+  ResourcesApi mockSamResourcesApi = Mockito.mock(ResourcesApi.class);
+
+  @BeforeEach
+  void setUp() {
+    given(mockSamClientFactory.getResourcesApi(null)).willReturn(mockSamResourcesApi);
+  }
+
+  @Test
+  void returnJobStatusWithPermission() throws ApiException {
+    // Make sure user has permission
+    given(mockSamResourcesApi.resourcePermissionV2(anyString(), anyString(), anyString()))
+        .willReturn(true);
+    // Tell jobDao to return
+    UUID instanceID = UUID.randomUUID();
+    UUID jobId = UUID.randomUUID();
+    GenericJobServerModel expectedJob =
+        new GenericJobServerModel(
+            jobId,
+            GenericJobServerModel.JobTypeEnum.DATA_IMPORT,
+            GenericJobServerModel.StatusEnum.RUNNING,
+            // set created and updated to now, but in UTC because that's how Postgres stores it
+            OffsetDateTime.now(ZoneId.of("Z")),
+            OffsetDateTime.now(ZoneId.of("Z")));
+    expectedJob.instanceId(instanceID);
+    given(jobDao.getJob(jobId)).willReturn(expectedJob);
+    // Ask jobService to get it
+    GenericJobServerModel fetchedJob = jobService.getJob(jobId);
+    // Success
+    assert (expectedJob.equals(fetchedJob));
+  }
+
+  @Test
+  void doNotReturnWithoutPermission() throws ApiException {
+    // Make sure user does not have permission
+    given(mockSamResourcesApi.resourcePermissionV2(anyString(), anyString(), anyString()))
+        .willReturn(false);
+    // Tell jobDao to return
+    UUID instanceID = UUID.randomUUID();
+    UUID jobId = UUID.randomUUID();
+    GenericJobServerModel expectedJob =
+        new GenericJobServerModel(
+            jobId,
+            GenericJobServerModel.JobTypeEnum.DATA_IMPORT,
+            GenericJobServerModel.StatusEnum.RUNNING,
+            // set created and updated to now, but in UTC because that's how Postgres stores it
+            OffsetDateTime.now(ZoneId.of("Z")),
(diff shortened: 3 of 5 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
```diff
@@ -90,6 +90,31 @@ public boolean hasWriteInstancePermission(String token) {
...
+   *
+   * @return true if the user has permission
+   */
+  @Override
+  public boolean hasReadInstancePermission(String instanceId) {
+    return hasReadInstancePermission(instanceId, null);
+  }
+
+  @Override
+  public boolean hasReadInstancePermission(String instanceId, String token) {
```
calypsomatic, line 104: Since the idea behind this ticket is that we won't be able to rely on the in-built WDS `instanceId`, I emulated the code from `HttpSamDao.hasPermission` but with the `instanceId` included as a parameter.  I could also modify `hasPermission` to always take in `instanceId`, and just have the other methods (e.g. `hasWriteInstancePermission`) pass the builtin `instanceId`.

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
```diff
@@ -12,15 +14,21 @@ public class JobService {
...
+  public JobService(JobDao jobDao, InstanceService instanceService, SamDao samDao) {
     this.jobDao = jobDao;
     this.instanceService = instanceService;
+    this.samDao = samDao;
   }
 
   public GenericJobServerModel getJob(UUID jobId) {
     try {
-      return jobDao.getJob(jobId);
+      GenericJobServerModel result = jobDao.getJob(jobId);
```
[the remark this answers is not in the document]
jladieu: Based on my read of @davidangb's [PR](https://github.com/DataBiosphere/terra-workspace-data-service/pull/482/files) (where he modifies `JobDao` to take an `InstanceId` as the first param) I think the caller of the API should specify the `instanceId` (or eventually, instance name, if we choose to go that route) in addition to the `jobId` -- then drill that down through `jobDao` rather than fetching the job first and then checking it.

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
jladieu: What you've got here seems reasonable to me, but I am not sure I have enough knowledge about Sam/permissions to nudge you away or toward this solution.  

A question I'd ask though...if we adopted this approach for the read path, should we require an `instanceId` for the write permission check too?  (I realize you can't provide an instanceId if checking for _create_ permission, but how about modification?)

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceTest.java
```diff
@@ -0,0 +1,86 @@
...
+            GenericJobServerModel.JobTypeEnum.DATA_IMPORT,
+            GenericJobServerModel.StatusEnum.RUNNING,
+            // set created and updated to now, but in UTC because that's how Postgres stores it
+            OffsetDateTime.now(ZoneId.of("Z")),
+            OffsetDateTime.now(ZoneId.of("Z")));
+    expectedJob.instanceId(instanceID);
+
+    given(jobDao.getJob(any())).willReturn(expectedJob);
+    // Ask jobService to get it
+    assertThrows(Exception.class, () -> jobService.getJob(jobId));
```
jladieu, line 84: Nit: check for a more specific exception here

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceTest.java
```diff
@@ -0,0 +1,86 @@
...
+            GenericJobServerModel.StatusEnum.RUNNING,
+            // set created and updated to now, but in UTC because that's how Postgres stores it
+            OffsetDateTime.now(ZoneId.of("Z")),
+            OffsetDateTime.now(ZoneId.of("Z")));
+    expectedJob.instanceId(instanceID);
+    given(jobDao.getJob(jobId)).willReturn(expectedJob);
+    // Ask jobService to get it
+    GenericJobServerModel fetchedJob = jobService.getJob(jobId);
+    // Success
+    assert (expectedJob.equals(fetchedJob));
```
jladieu, line 61: Nit: use fluent assertion methods over the native assertions, at least in tests.

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
calypsomatic: I think the write permissions should be good as is just for doing import jobs, since we're not actually writing to the WDS, but sending the data over to rawls/entity service.  But this would definitely need to be re-visited as we develop cWDS.

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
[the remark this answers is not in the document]
calypsomatic: I think that makes sense.  I guess the only concern would be if you sent an `instanceId` that didn't match the job - presumably wouldn't happen through the UI but you would just by calling the API.  So you'd want to check that the given `instanceId` matches that in the job anyway.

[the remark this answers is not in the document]
jladieu: I think in that case, where the `instanceId` mismatches the Job, I think a common way that gets handled is for the Dao to throw a RecordNotFound (or similarly named) error that gets turned into a 404, because the `instanceId` would become part of the SQL predicate (`WHERE instanceId = "<instanceId>" AND id="<jobId>"`) looking for a matching job record.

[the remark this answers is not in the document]
davidangb: > I think the caller of the API should specify the instanceId (or eventually, instance name, if we choose to go that route) in addition to the jobId

This was my original thought, but when we refined this ticket, team discussion landed on NOT asking the caller to specify the instanceid, and instead fetching it right from the job. Now that we're looking at the code, is this no longer the case?

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
davidangb: I'm ok with this as-is; I do think before long this class will need a makeover and that eventually the read and write methods should have parallel signatures.

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
```diff
@@ -12,15 +14,21 @@ public class JobService {
...
+      if (!samDao.hasReadInstancePermission(result.getInstanceId().toString())) {
+        throw new AuthorizationException("Caller does not have permission to view this job.");
+      }
+      return result;
```
davidangb, line 31: What @ashanhol found is that the `.getJob()` method does not yet return the instanceId value, and it needs a SQL update under the covers. so I expect the code you have here in this class will eventually work, but only after `PostgresJobDao` has some updates.

As of right now, though, I expect that if you run this code, `result.getInstanceId().toString()` will throw a null pointer exception. Are you seeing that?

calypsomatic: Yep looks like I'll need to update `PostgresJobDao`.  I almost did, thought maybe there was some reason not to, and then forgot about it.

davidangb: Adina might be working on this already, definitely check in!

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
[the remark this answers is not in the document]
calypsomatic: Hmm I'm starting to lean toward including `instanceId` in the API.  Would that be a v2 or something then since it's changing the signature?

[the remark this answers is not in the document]
calypsomatic: Although on attempting to implement it, it involves a *lot* of code change - mostly busywork but it makes me nervous that there will be some small detail that'll be missed

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
ashanhol: I am, PRs out

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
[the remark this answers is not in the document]
yuliadub: given that the JobId is a primary key in the Job table - I think it makes more sense not to include instanceId in the call (another thing for caller to figure out and supply, while JobId already infers it). Is our main motivation to keep all api signatures the same or is there something else behind wanting to include instanceId?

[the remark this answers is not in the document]
calypsomatic: Ok, per [discussion in slack](https://broadinstitute.slack.com/archives/C02SJQ0FKHQ/p1707415112147329), am keeping `instanceId` out of the API signature.  We can add it later if we realize it's needed.

service/src/test/java/org/databiosphere/workspacedataservice/controller/JobControllerMockMvcTest.java
```diff
@@ -33,6 +33,7 @@ void smokeTestGetJob() throws Exception {
             // set created and updated to now, but in UTC because that's how Postgres stores it
             OffsetDateTime.now(ZoneId.of("Z")),
             OffsetDateTime.now(ZoneId.of("Z")));
+    expected.instanceId(UUID.randomUUID());
```
ashanhol, line 36: I think I set this in my PR too

ashanhol (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
```diff
@@ -239,6 +239,7 @@ public GenericJobServerModel mapRow(ResultSet rs, int rowNum) throws SQLExceptio
           new GenericJobServerModel(jobId, jobType, status, created, updated);
 
       job.errorMessage(rs.getString("error"));
+      job.instanceId(rs.getObject("instance_id", UUID.class));
```
davidangb, line 242: is this safe when `instance_id` is null in the db? I think it is; just double-checking.

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
```diff
@@ -12,15 +14,21 @@ public class JobService {
...
+    this.samDao = samDao;
   }
 
   public GenericJobServerModel getJob(UUID jobId) {
     try {
-      return jobDao.getJob(jobId);
+      GenericJobServerModel result = jobDao.getJob(jobId);
+      if (!samDao.hasReadInstancePermission(result.getInstanceId().toString())) {
+        throw new AuthorizationException("Caller does not have permission to view this job.");
+      }
```
davidangb, line 30: until/unless we make instanceId required and enforce it is non-null, this will need a null check. Pre-existing jobs before #485 will have nulls here.

davidangb (changes requested):
Some concerns about null handling for legacy jobs, before jobs were associated with an instance.

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
calypsomatic: In that case, would we always allow the job to be viewed?

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
calypsomatic: Yes this part is safe.

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
davidangb: What do you think of this logic?
* if instanceId is not null, get the workspace associated with that instance and check permission on that workspace
* if instanceId is null (i.e. legacy job), check permission on the $WORKSPACE_ID env var

I don't think we should have an "always allow" clause, there should be *some* kind of check

service/src/main/resources/liquibase/changesets/20240212_backfill_instanceid_to_job.yaml
```diff
@@ -0,0 +1,43 @@
...
+            sql:  |
+              -- if WORKSPACE_ID did not exist, or any other problems occurred, simply remove jobs
+              -- with null instance_id
+              delete from sys_wds.job where instance_id is null;
+        - sql:
+            dbms:  postgresql
+            splitStatements: false
+            sql:  |
+              -- now set instance_id to be non-null
+              delete from sys_wds.job where instance_id is null;
```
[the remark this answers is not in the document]
calypsomatic: ah copy-pasta error

service/src/main/resources/liquibase/changesets/20240212_backfill_instanceid_to_job.yaml
```diff
@@ -0,0 +1,43 @@
...
+              $$ language plpgsql immutable;
+        - sql:
+            dbms:  postgresql
+            splitStatements: false
+            sql:  |
+              -- set instance_id to the env var WORKSPACE_ID for pre-existing jobs where instance_id is null
+              -- or leave null if WORKSPACE_ID does not exist
+              update sys_wds.job
+              set instance_id = sys_wds.uuid_or_null('${WORKSPACE_ID}')
+              where instance_id is null;
```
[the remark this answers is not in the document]
calypsomatic: Is that going to happen?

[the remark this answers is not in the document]
davidangb: Because the symptom is so drastic - WDS fails to start at all - I think we need to handle it. If we ever allowed "promoting" a WDS deployment from per-workspace to per-LZ, we could run into this. It also feels like using a `sqlFile` instead of `sql` is safe enough. Do you think it's not worth handling?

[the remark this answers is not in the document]
calypsomatic: Ok I'll steal the rest of your PR then ;)

davidangb (approved):
this was a tricky one; thank you!

ashanhol (approved):

PR #499, opened 2024-02-07
AJ-1591: profile-driven properties

davidangb:
Up for discussion: using Spring profiles, annotations, and injection to control individual properties of data plane vs. control plane.

In this PR:
* introduce a Spring standard in which properties are represented as Java classes annotated by `@ConfigurationProperties` and managed by Spring injection. I think this is the approach we should take going forward. In this PR I've only illustrated it for two properties, which I intend to make use of in #489.
* two new .yaml properties files corresponding to the "control-plane" and "data-plane" profiles, and which will populate the new `WdsTenancyProperties` class
* Unit tests to show behavior

The expectation is that our Helm charts will set either "control-plane" or "data-plane" into the `SPRING_PROFILES_ACTIVE` env var, which we already have in place for Azure: https://github.com/broadinstitute/terra-helmfile/blob/bffc0f84c80811c927cd837b52ce05a0dcb12585/charts/wds/templates/deployment.yaml#L109-L117. This will activate one of the two profiles.

Then, spring will load the corresponding `application-${profile}.yaml` file, which will populate the `WdsTenancyProperties` class. Finally, Java code throughout WDS can look at `WdsTenancyProperties` values.

Note: if we have the "data-plane" and "control-plane" profiles, we could rework the implementation of the `ControlPlane` and `DataPlane` annotations to be profile-driven instead of annotation-driven; I have not tried to tackle that in this PR.


.gitignore
```diff
@@ -39,8 +39,9 @@ out/
 
 ### VS Code ###
 .vscode/
-config/
 
+# output of scripts/render-config.sh
+/config/
```
davidangb, line 44: This was preventing me from committing the service/src/.…/config directories!

service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
```diff
@@ -17,6 +19,7 @@
 @EnableRetry
 @EnableTransactionManagement
 @EnableCaching
+@EnableConfigurationProperties(WdsTenancyProperties.class)
```
davidangb, line 22: we need to list each _top-level_ `@ConfigurationProperties`-annotated class here. By top-level, I mean that if we have a parent config class for the `twds` hierarchy, we'd only need to list that parent. To illustrate what I mean, see https://github.com/DataBiosphere/terra-azure-relay-listeners/blob/8c96e08fd2f7751b30c1d9d4fd6d15ae7b429f33/service/src/main/java/org/broadinstitute/listener/config/AppConfiguration.java#L32 and check out the `ListenerProperties` hierarchy.

service/src/main/java/org/databiosphere/workspacedataservice/config/WdsTenancyProperties.java
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  public Boolean getRequireEnvWorkspace() {
+    return requireEnvWorkspace;
+  }
+
+  public void setRequireEnvWorkspace(Boolean requireEnvWorkspace) {
+    this.requireEnvWorkspace = requireEnvWorkspace;
+  }
+}
```
davidangb, line 25: TODO: add javadoc

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -717,23 +718,24 @@ private Object getArrayValues(Object attVal, DataTypeMapping typeMapping) {
...
-          .map(Boolean::parseBoolean)
-          .toList()
-          .toArray(new Boolean[0]);
+          // accept all casings of True and False if they're strings
+          attVal.stream()
+              .map(Object::toString)
+              .map(String::toLowerCase)
+              .map(Boolean::parseBoolean)
+              .toList()
+              .toArray(new Boolean[0]);
```
davidangb, line 738: bah, well, this has formatting changes.

service/src/main/resources/application-data-plane.yml
```diff
@@ -0,0 +1,7 @@
+twds:
+  tenancy:
+    # Should this WDS deployment allow virtual collections, i.e. collections that have no
```
calypsomatic, line 3: I'm already confused by what "collections" means

service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
jladieu: Seems reasonable to me...for my understanding, what's `twds` stand for, as opposed to `wds` or `cWds`?

calypsomatic: terra-wds

service/src/test/java/org/databiosphere/workspacedataservice/config/WdsTenancyPropertiesNoProfileTest.java
```diff
@@ -0,0 +1,27 @@
+package org.databiosphere.workspacedataservice.config;
```
jladieu, line 1: Not a big deal, but I'd be lying if I didn't think "Aww, I wish we could have these three tests all in the same test file." when I saw that there was one test file per profile.

I think my reasoning is:
- Something idealistic about having the `Code.java` 1:1 correspondence with `CodeTest.java`
- Fear of exploding combinations of profiles (eg: DataPlane + Bee) resulting in even more files with ugly names.

jladieu (approved):
LGTM.  My one wish was that we could do a more targeted form of profile overriding on a per-test basis, but I'm fine with shipping as is.

service/src/test/java/org/databiosphere/workspacedataservice/config/WdsTenancyPropertiesNoProfileTest.java
calypsomatic: I feel the same way, but my memory of trying to do such a thing is that you just CAN'T

jladieu: Possibly a lead...I figured out how to get an instance of `ConfigurableApplicationContext` in https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/service/src/test/java/org/databiosphere/workspacedataservice/annotations/DeploymentModeTest.java#L52-L59 -- and I think that can be used to `getEnvironment().setActiveProfiles(....)` which might be doable on a per-test basis with some well factored helper constructs.

service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
jladieu: (more squinting....when do we choose to add that extra namespacing...is there a notion of a non-Terra WDS?)

service/src/test/java/org/databiosphere/workspacedataservice/config/WdsTenancyPropertiesNoProfileTest.java
jladieu: Ehh I dug in more and it looks like SpringBoot really doesn't want you to do mess around with ActiveProfiles more than once per test suite.

service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
davidangb: I would be totally up for renaming this property, because we just never use the "twds" acronym … but it does match the name of this repo

service/src/test/java/org/databiosphere/workspacedataservice/config/WdsTenancyPropertiesNoProfileTest.java
davidangb: I'd also accept that these tests are not super useful. They're 90% testing Spring Boot behavior, and maybe 10% testing that we've written our properties files/classes correctly. I think they're valuable to illustrate the behavior I wanted to showcase in this PR, but I'd also be ok not merging them into main if you think they add too much clutter.

jladieu: Whether it's 90/10 or not, IMO that 10% of ensuring we wrote our config correctly is an important 10% because those config files are so structurally important.  I think I'm okay with this kind of test as long as we can make them as lightweight and easy to adjust/modify as possible.  Right now I think what you've got satisfies both lightweight ( (at least from the runtime perspective, if not the code / class / # files perspective), and easy to change.

service/src/main/resources/application-data-plane.yml
jladieu: I think you can probably eliminate these inline comments since they are so well documented in now in the Java files.

service/src/test/java/org/databiosphere/workspacedataservice/config/WdsTenancyPropertiesNoProfileTest.java
```diff
@@ -0,0 +1,27 @@
...
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+
+/**
+ * when neither the "data-plane" nor "control-plane" profile is active, WdsTenancyProperties are
+ * null
+ */
```
jladieu, line 12: Is there a reasonable way to express `@ActiveProfiles({})` to make it explicit that there shouldn't be any active?  I could foresee some other more global profile configuration (eg: the test profile) incidentally causing this to fail in the absence of explicitly preventing it.

jladieu:
Approved with minor / optional comments.

service/src/main/java/org/databiosphere/workspacedataservice/config/WdsTenancyProperties.java
davidangb: javadoc added.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
davidangb: I removed the formatting changes; they might be valid but I was annoyed they were in this PR

service/src/test/java/org/databiosphere/workspacedataservice/config/WdsTenancyPropertiesNoProfileTest.java
davidangb: is this better?

calypsomatic (approved):

PR #501, opened 2024-02-08
List import jobs API

ashanhol:
https://broadworkbench.atlassian.net/browse/AJ-1560

Add API `/job/v1/instance/{instanceUuid}?statuses=` that pulls all jobs of type `status` that share the same `instanceId`. The UI will only need a status of type `Running` but having more options allows for future proofing & modular options. 

---------------
Reminder:


service/src/main/java/org/databiosphere/workspacedataservice/controller/JobController.java
```diff
@@ -29,4 +33,13 @@ public ResponseEntity<GenericJobServerModel> jobStatusV1(UUID jobId) {
 
     return new ResponseEntity<>(job, responseCode);
   }
+
+  @Override
+  public ResponseEntity<List<GenericJobServerModel>> jobsInInstanceV1(
+      UUID instanceUuid, String status) {
```
yuliadub, line 39: could this take the status enum instead of string right away?

ashanhol: I tried, I don't think so since it's autogenerated. Also wanted to see if it could take the InstanceId class David defined instead of UUID :/

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
```diff
@@ -25,4 +29,12 @@ public GenericJobServerModel getJob(UUID jobId) {
       throw new MissingObjectException("Job");
     }
   }
+
+  public List<GenericJobServerModel> getJobsForInstance(InstanceId instanceId, StatusEnum status) {
+    try {
+      return jobDao.getJobsForInstance(instanceId, status);
+    } catch (EmptyResultDataAccessException e) {
+      throw new MissingObjectException("Jobs for InstanceId: " + instanceId);
```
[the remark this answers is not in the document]
calypsomatic: I think we could definitely return zero jobs; if the UI just automatically checks to see if any are running presumably most of the time there wouldn't be any.

service/src/test/java/org/databiosphere/workspacedataservice/controller/JobControllerMockMvcTest.java
```diff
@@ -48,4 +53,55 @@ void smokeTestGetJob() throws Exception {
...
+    List<GenericJobServerModel> expected = new ArrayList<GenericJobServerModel>(2);
+    expected.add(
+        new GenericJobServerModel(
+                jobId1,
+                GenericJobServerModel.JobTypeEnum.DATA_IMPORT,
+                GenericJobServerModel.StatusEnum.RUNNING,
+                time,
+                time)
+            .instanceId(instanceId.id()));
+    expected.add(
```
yuliadub, line 75: just to expand the test case a bit more, maybe add another value that is not in state Running?

yuliadub: and maybe one for in a different instance too

yuliadub: ah wait, I am reading the test case wrong - this is the expected output, not the set you are querying against - ignore the comment

ashanhol: You mean add a job of a different status than Running to the `expected` list? or do a second call to the API with a different status than Running?

yuliadub (approved):

service/src/main/java/org/databiosphere/workspacedataservice/controller/JobController.java
yuliadub: gotcha, I couldnt remember if it could do that or not. Figures!

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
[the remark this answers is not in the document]
ashanhol: huh. as a sanity check I did just that and it returned fine.
image
Just stepped through the debugger, it's not even getting hit. I'll take it out since yeah it is valid to get no jobs of that type back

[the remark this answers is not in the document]
ashanhol: Removed

service/src/main/resources/static/swagger/apis-v1.yaml
```diff
@@ -96,7 +96,30 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/GenericJob'
-
+  /job/v1/{instanceUuid}/{status}:
```
nawatts, line 99: Would it be preferable to pass status as a query parameter and potentially allow filtering for multiple statuses in one request?

ashanhol: hmm. Is there a benefit to allowing multiple statuses vs just returning all jobs for an instanceid? Right now the only use case is getting all running jobs. I'm fine with making the status optional so we can return all jobs for an ID

nawatts: It's a REST nitpick, but it seems like the list of jobs is the resource here and the status is a way of filtering it. So, to me, a query parameter seems more appropriate for the status. I think my vote would be return running jobs by default and (now or later if needed) add a query parameter that allowed returning other statuses.

davidangb: +1 on allowing queries on multiple statuses. What the UI really cares about is the list of jobs that are in queued+created+running ... I forget the actual status names but I think you get the drift. 

It would also be nice eventually to allow listing all jobs, not just running ones; this would include jobs in error or succeeded. We don't need that functionality in this PR, but I'd like the method signature to eventually allow it.

(We might eventually need pagination, too, which would be query params)

davidangb: Another option is to copy the existing API in Import Service and have a "active=(true|false)" query params (feel free to rename it). This would absolve callers from knowing the various status names. WDS would encapsulate knowledge of which statuses are active vs. which are not.

ashanhol: Ok I modified it to be an optional array query param. From manual testing it seems to behave as expected (ie status still works and if you don't specify any status it returns all jobs). Still trying (tomorrow) to debug why test is failing.

service/src/main/resources/static/swagger/apis-v1.yaml
```diff
@@ -96,7 +96,31 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/GenericJob'
-
+  /job/v1/all/{instanceUuid}:
```
ashanhol, line 99: In moving status to be an optional query param, it got dropped from the url. buuut GET /job/v1/{SOME UUID} is already used above for getjob, so I included /all/ to disambiguate it

service/src/main/java/org/databiosphere/workspacedataservice/controller/JobController.java
```diff
@@ -29,4 +33,24 @@ public ResponseEntity<GenericJobServerModel> jobStatusV1(UUID jobId) {
 
     return new ResponseEntity<>(job, responseCode);
   }
+
+  @Override
+  public ResponseEntity<List<GenericJobServerModel>> jobsInInstanceV1(
+      UUID instanceUuid, List<String> status) {
```
calypsomatic, line 39: Nit: if there are multiple statuses, then shouldn't the variable name be `statuses`?

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
```diff
@@ -25,4 +27,9 @@ public GenericJobServerModel getJob(UUID jobId) {
       throw new MissingObjectException("Job");
     }
   }
+
+  public List<GenericJobServerModel> getJobsForInstance(
+      InstanceId instanceId, List<String> statuses) {
+    return jobDao.getJobsForInstance(instanceId, statuses);
```
calypsomatic, line 33: Do we need to do a read permission check like we're doing in `getJobStatus`?

service/src/test/java/org/databiosphere/workspacedataservice/controller/JobControllerMockMvcTest.java
```diff
@@ -48,4 +53,50 @@ void smokeTestGetJob() throws Exception {
     // which is equal to the expected job
     assertEquals(expected, actual);
   }
+
+  @Test
+  void smokeTestJobsInInstanceV1() throws Exception {
+    // return a test job from the mocked JobDao
```
calypsomatic, line 59: Is it worth it to also test for either no statuses or multiple statuses?

service/src/main/resources/static/swagger/apis-v1.yaml
nawatts: Alternatively, something like `/jobs/v1/instance/{instanceId}`?

service/src/test/java/org/databiosphere/workspacedataservice/controller/JobControllerMockMvcTest.java
ashanhol: I wasn't sure whether or not to do that in a smoke test since what we're returning is extremely mocked. I thought that behavior would be better for a unittest but i didn't see any for getstatus so I wasn't sure

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
ashanhol: Ah yeah, is your PR in? I can add the behavior you added

service/src/test/java/org/databiosphere/workspacedataservice/controller/JobControllerMockMvcTest.java
ashanhol: k unittests created

service/src/main/resources/static/swagger/apis-v1.yaml
ashanhol: i liked that better, went with yours

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
ashanhol: added, thank you for your hard work on that PR :)

service/src/test/java/org/databiosphere/workspacedataservice/controller/JobControllerTest.java
```diff
@@ -0,0 +1,115 @@
...
+    // cleanup: delete everything from the job table
+    namedTemplate.getJdbcTemplate().update("delete from sys_wds.job;");
+  }
+
+  @Test
+  void instanceJobsReturnAllNoStatus() {
+    HttpHeaders headers = new HttpHeaders();
+    // ParameterizedTypeReference<List<GenericJobServerModel>> returnType = new
+    // ParameterizedTypeReference<List<GenericJobServerModel>>() {};
+    ResponseEntity<List<GenericJobServerModel>> result =
```
yuliadub, line 77: seems like the unit test is unhappy with being able to convert response to this object?

ashanhol: no- bria and I just figured it out. I'm missing the mock sam profile. it's mad b/c it got a 500 error (JSON) in response

ashanhol: I spent a bunch of time yesterday trying to figure out why it was passing for me locally and not on github. apparently it'll pass if you have the workspaceid environment variable set

service/src/main/java/org/databiosphere/workspacedataservice/controller/JobController.java
```diff
@@ -29,4 +33,24 @@ public ResponseEntity<GenericJobServerModel> jobStatusV1(UUID jobId) {
...
   }
+
+  @Override
+  public ResponseEntity<List<GenericJobServerModel>> jobsInInstanceV1(
+      UUID instanceUuid, List<String> statuses) {
+    // status is an optional parameter
+    if (statuses != null) {
+      try {
+        // validate the strings in status are all valid statuses before proceeding
+        for (var statusValue : statuses) {
```
jladieu, line 44: Since you're doing the work to check these against the Enum right here, would you consider mapping these to Enums and passing them down to `getJobsForInstance`?  I think using stronger types when they're available is always a good move.

EG:
```
var statusEnums = statuses.stream().map(status -> StatusEnum.fromValue(statusValue))
```

service/src/main/java/org/databiosphere/workspacedataservice/dao/JobDao.java
```diff
@@ -27,4 +29,6 @@ public interface JobDao {
   GenericJobServerModel fail(UUID jobId, Exception e);
 
   GenericJobServerModel getJob(UUID jobId);
+
+  List<GenericJobServerModel> getJobsForInstance(InstanceId instanceId, List<String> statuses);
```
jladieu, line 33: Does the order of the statuses matter?  If not, would you consider taking in a `Collection` instead?  Then you could pass in a `Set`, a `List`, whatever.

service/src/main/java/org/databiosphere/workspacedataservice/controller/JobController.java
ashanhol: I was doing that, but I converted it back to strings when we decided to make the params a list since it'd require serializing them to a new data structure, then deserializing them back into strings as soon as we use them in PostgresJobDao.getJobsForInstance. It's actually just simpler to keep them as strings to save ourselves the extra step at the end and be able to pass them as is.

service/src/test/java/org/databiosphere/workspacedataservice/controller/JobControllerTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+  private InstanceId instanceId = InstanceId.of(UUID.randomUUID());
+
+  private UUID jobId;
+
+  @BeforeAll
+  void beforeAll() {
+    // push jobs into the db, one of different instanceId
+    Job<JobInput, JobResult> testJob1 =
+        Job.newJob(instanceId, JobType.DATA_IMPORT, JobInput.empty());
```
jladieu, line 53: nit: there's a lot of `Job` showing up in the test setup...most of them are fine, but I think the `Job.newJob` would benefit from using a static import to just become:

```suggestion
       newJob(instanceId, JobType.DATA_IMPORT, JobInput.empty());
```

Would require:
```
import static org.databiosphere.workspacedataservice.shared.model.job.Job.newJob;
```

service/src/main/java/org/databiosphere/workspacedataservice/dao/JobDao.java
ashanhol: I'm specifying an array in apis-v1.yaml

jladieu (approved):
All looks good to me!  Some minor nitty comments/suggestions that you can do with what ye will.

service/src/test/java/org/databiosphere/workspacedataservice/controller/JobControllerTest.java
yuliadub: ahhh got it, thanks for following up!

PR #503, opened 2024-02-08
[AJ-1605] Make RecordRowMapper sensitive to attribute case

nawatts:
Currently, WDS supports inserting records with attributes that differ only in case. For example, a record with two attributes: "Value" and "value". Such records are stored correctly in the database, but when queried through WDS, the same value will be returned for both attributes.

For example, a record inserted as `{"value": "foo", "Value": "bar"}` would be returned as `{"value": "foo", "Value": "foo"}`.

This is because of an issue in RecordRowMapper. It gets column values from the ResultSet by column name.

From [java.sql.ResultSet's documentation](https://docs.oracle.com/en/java/javase/17/docs/api/java.sql/java/sql/ResultSet.html):
> Column names used as input to getter methods are case insensitive. When a getter method is called with a column name and several columns have the same name, the value of the first matching column will be returned. The column name option is designed to be used when column names are used in the SQL query that generated the result set. For columns that are NOT explicitly named in the query, it is best to use column numbers. If column names are used, the programmer should take care to guarantee that they uniquely refer to the intended columns, which can be assured with the SQL AS clause.

The solution is to get attribute values by column index instead of name. For the `getAttributes` method, this is a trivial change. Getting the primary key column in the `mapRow` method was a little more cumbersome.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -922,7 +922,8 @@ public class RecordDao {
         }
       }
       if (primaryKeyColumnIndex == -1) {
-        throw new RuntimeException("Primary key column not found");
+        throw new RuntimeException(
+            "Primary key column \"%s\" not found".formatted(primaryKeyColumn));
       }
 
       try {
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -909,10 +909,10 @@ public class RecordDao {
       ResultSetMetaData metaData = rs.getMetaData();
 
       int primaryKeyColumnIndex = -1;
-      for (int i = 1; i <= metaData.getColumnCount(); i++) {
-        String columnName = metaData.getColumnName(i);
+      for (int columnIndex = 1; columnIndex <= metaData.getColumnCount(); columnIndex++) {
+        String columnName = metaData.getColumnName(columnIndex);
         if (columnName.equals(primaryKeyColumn)) {
-          primaryKeyColumnIndex = i;
+          primaryKeyColumnIndex = columnIndex;
           break;
         }
       }
@@ -932,10 +932,10 @@ public class RecordDao {
         ResultSetMetaData metaData = rs.getMetaData();
         RecordAttributes attributes = RecordAttributes.empty(primaryKeyColumn);
 
-        for (int j = 1; j <= metaData.getColumnCount(); j++) {
-          String columnName = metaData.getColumnName(j);
+        for (int columnIndex = 1; columnIndex <= metaData.getColumnCount(); columnIndex++) {
+          String columnName = metaData.getColumnName(columnIndex);
           if (columnName.equals(primaryKeyColumn)) {
-            attributes.putAttribute(primaryKeyColumn, rs.getString(j));
+            attributes.putAttribute(primaryKeyColumn, rs.getString(columnIndex));
             continue;
           }
           if (referenceColToTable.size() > 0
@@ -944,10 +944,11 @@ public class RecordDao {
             attributes.putAttribute(
                 columnName,
                 RelationUtils.createRelationString(
-                    referenceColToTable.get(columnName), rs.getString(j)));
+                    referenceColToTable.get(columnName), rs.getString(columnIndex)));
           } else {
             attributes.putAttribute(
-                columnName, getAttributeValueForType(rs.getObject(j), schema.get(columnName)));
+                columnName,
+                getAttributeValueForType(rs.getObject(columnIndex), schema.get(columnName)));
           }
         }
         return attributes;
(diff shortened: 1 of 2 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -906,8 +906,22 @@ public RecordRowMapper(RecordType recordType, ObjectMapper objectMapper, UUID in
...
+      for (int i = 1; i <= metaData.getColumnCount(); i++) {
+        String columnName = metaData.getColumnName(i);
+        if (columnName.equals(primaryKeyColumn)) {
+          primaryKeyColumnIndex = i;
+          break;
+        }
+      }
+      if (primaryKeyColumnIndex == -1) {
+        throw new RuntimeException("Primary key column not found");
+      }
```
nawatts, line 921: I tried this with `IntStream.range(..).filter(...).findFirst()` but it was just as verbose and swallowed SQLExceptions from `getColumnName`.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -921,7 +935,7 @@ private RecordAttributes getAttributes(ResultSet rs) throws JsonProcessingExcept
         for (int j = 1; j <= metaData.getColumnCount(); j++) {
           String columnName = metaData.getColumnName(j);
           if (columnName.equals(primaryKeyColumn)) {
-            attributes.putAttribute(primaryKeyColumn, rs.getString(primaryKeyColumn));
+            attributes.putAttribute(primaryKeyColumn, rs.getString(j));
```
ashanhol, line 938: normally I'm fine with i and j as variable names for for loops but in this case I think it'd be easier to follow if we actually used the variable name "index" in both locations.

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
nawatts: Renamed to "columnIndex" in https://github.com/DataBiosphere/terra-workspace-data-service/pull/503/commits/057af3a2a67be7cb7db31f46b12969e714c102fc.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -906,8 +906,22 @@ public RecordRowMapper(RecordType recordType, ObjectMapper objectMapper, UUID in
 
     @Override
     public Record mapRow(ResultSet rs, int rowNum) throws SQLException {
+      ResultSetMetaData metaData = rs.getMetaData();
+
+      int primaryKeyColumnIndex = -1;
```
jladieu, line 911: I think it'd be worth embedding some of your excellent PR description here near the code to explain why we're getting the `columnName` by index.  I'd be in favor of the entire block being inline, but realize maybe you don't wanna do that.  Alternately you could put a pointer to the PR in here.  But either way I want to bring your justification for this change front and center because it makes a lot of sense but isn't intuitive.  Great job laying it out!

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -906,8 +906,22 @@ public RecordRowMapper(RecordType recordType, ObjectMapper objectMapper, UUID in
...
+      for (int columnIndex = 1; columnIndex <= metaData.getColumnCount(); columnIndex++) {
+        String columnName = metaData.getColumnName(columnIndex);
+        if (columnName.equals(primaryKeyColumn)) {
+          primaryKeyColumnIndex = columnIndex;
+          break;
+        }
+      }
+      if (primaryKeyColumnIndex == -1) {
+        throw new RuntimeException("Primary key column not found");
```
jladieu, line 920: Can you include the expected (but missing) `primaryKey` name in the error message?

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -918,10 +932,10 @@ private RecordAttributes getAttributes(ResultSet rs) throws JsonProcessingExcept
         ResultSetMetaData metaData = rs.getMetaData();
         RecordAttributes attributes = RecordAttributes.empty(primaryKeyColumn);
 
-        for (int j = 1; j <= metaData.getColumnCount(); j++) {
-          String columnName = metaData.getColumnName(j);
+        for (int columnIndex = 1; columnIndex <= metaData.getColumnCount(); columnIndex++) {
```
jladieu, line 935: Thank you so much for renaming `i` and `j`

jladieu (approved):
LGTM with some small suggestions.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/503/commits/29995ed35c3534e7e83cc429c1c3ca64f6f8607e.

PR #507, opened 2024-02-12
AJ-1563 Write Rawls JSON for imports in cWDS.

jladieu:
Make import-service generate Rawls JSON (without doing anything with it) in cWDS instead of writing it using the `RecordService`.


service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
```diff
@@ -42,7 +42,7 @@
 import org.springframework.web.server.ResponseStatusException;
 
 @Service
-public class RecordService implements BatchWriteService.RecordSink {
```
jladieu, line 45: I had to back away from having `RecordService` implement the `RecordSink` API due to problems with configuring it as a spring `@Bean`.

The particular issue that made this problematic is that `RecordService` was filling the roles of both the `RecordService` bean _and_ the `RecordSink` bean.   When running in the cWDS `deploymentMode`, the `RecordService` bean was still present (even though its `RecordSink` bean was disabled), and still eligible to be considered as the `RecordSink` bean, even though I intended for the `RawlsRecordSink` to fulfill this purpose.

To work around this, I made two explicit `RecordSink` beans (`{WDS,Rawls}RecordSink`) and one or the other is chosen.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobDataPlaneE2ETest.java
```diff
@@ -2,14 +2,8 @@
```
jladieu, line 2: I renamed the existing `PfbQuartzJobE2ETest` to clarify it's intended to test the `DataPlane` side effects.  This makes space at the top level namespace to also have a sibling `ControlPlane` test.

While making this rename, I noticed the setup code to simulate a PFB import was repeated, and needed also for the new E2E test file, so there's a lot of red diff here where this logic got DRYed up in a new method in `PfbTestSupport`.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
calypsomatic: I thought there was a way to tell Spring which bean to use.  Either with a `@Primary` annotation and/or with some kind of config.

davidangb (commented):
a few initial comments inline, overall this is looking on track to me!

service/src/main/java/org/databiosphere/workspacedataservice/service/RawlsRecordSink.java
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * {@link RecordSink} implementation that produces Rawls-compatible JSON using {@link Model}
+ * serialization classes.
+ *
+ * <p>TODO(AJ-1585): integrate with storage to write JSON output to the appropriate bucket
+ *
+ * <p>TODO(AJ-1586): integrate with pubsub to notify Rawls when JSON is ready to be processed
+ */
+@Service
```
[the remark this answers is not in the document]
jladieu: Hey, good catch...this is actually what was causing the issues I described in [comment](https://github.com/DataBiosphere/terra-workspace-data-service/pull/507#discussion_r1486444513)!  By removing this `@Service` I eliminated the source of a superfluous Bean config.

See 84e95695c22621938f9343f20f6c7f0205a792d6, which also gets rid of `WdsRecordSink` and re-inlines those methods into `RecordService` as I had initially intended.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
jladieu: Yes!  I tried `@Primary` too but was still hitting problems.  Turns out it was something silly that @davidangb noticed: https://github.com/DataBiosphere/terra-workspace-data-service/pull/507#discussion_r1488014879

Anyway, I've put this back the way I intended after eliminating the superfluous `@Service` annotation which was the source of all my problems with the bean config.

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteServiceConfig.java
```diff
@@ -1,12 +1,24 @@
...
 
 @Configuration
 public class BatchWriteServiceConfig {
   @Bean
-  BatchWriteService.RecordSink recordSink(RecordService recordService) {
-    return recordService;
+  @ConditionalOnProperty(
+      name = "twds.data-import.batch-write-record-sink",
+      havingValue = "wds",
+      matchIfMissing = true)
```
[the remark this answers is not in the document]
jladieu: I do agree with failing faster/earlier for mis-config, but I sort of sidestepped _this particular_ thing by adopting the `Properties` config approach.  That said, I still want to figure out how to fail fast over there :)

service/src/main/java/org/databiosphere/workspacedataservice/service/RawlsRecordSink.java
```diff
@@ -0,0 +1,104 @@
...
+      return Stream.concat(
+          Stream.of(new RemoveAttribute(name), new CreateAttributeValueList(name)),
+          values.stream().map(value -> new AddListMember(name, value)));
+    }
+
+    return Stream.of(new AddUpdateAttribute(name, attributeValue));
+  }
+
+  private String getAttributeName(RecordType recordType, String name) {
+    String fileType = "pfb"; // TODO: make dynamic, provided by context
```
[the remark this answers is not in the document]
jladieu: Added a couple unit tests in 4f3625931c79641e69cbf56f8a95b902bf480035, which also led me to move the prefix logic a bit closer to where I think it could end up living, which is in the construction of the `RawlsRecordSink`.  I'm thinking maybe in the same place we create the `RecordSource` we might do something similar for creating format-aware `RecordSink`.

README.md
```diff
@@ -37,24 +37,30 @@ Instead use the `run_postgres.sh` script to set up a docker container running po
...
 
-Other profiles that are available are: 
+Other profiles that are available are:
+
 - alpha
 - staging
 - dev
 - prod
 - bee
+- data-plane OR control-plane
```
jladieu, line 57: This is all I added in the README, the rest are auto-formatting changes from IDEA.

service/src/test/java/org/databiosphere/workspacedataservice/config/TwdsPropertiesDataPlaneTest.java
```diff
@@ -12,10 +13,22 @@
 // the "data-plane" profile enforces validity of twds.instance.workspace-id, so we need to set that
 @SpringBootTest(properties = {"twds.instance.workspace-id=00ddba11-0000-0000-0000-000000000000"})
 @ActiveProfiles("data-plane")
-class TenancyPropertiesDataPlaneTest {
```
jladieu, line 15: I was worried about the sprawl of `Test` files we'd generate while adding additional sub-configs, so I went ahead and renamed this (and its sibling test files) with the generic `Twds` prefix and shoved my additional tests in there.

README.md
davidangb: thinking out loud, for a future PR: should we add validation at startup that one or the other is set? Is there a valid use case (tests?) to have neither set?

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobControlPlaneE2ETest.java
```diff
@@ -0,0 +1,199 @@
...
+                .findFirst()
+                .orElseThrow())
+        .isEqualTo(expected);
+  }
+
+  private Stream<? extends AttributeOperation> filteredOps(Entity entity, Op operation) {
+    return entity.operations().stream().filter(op -> op.op() == operation);
+  }
+
+  private List<Model.Entity> getRecordedEntitiesSerde() {
```
calypsomatic, line 174: ???? is Serde??

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobControlPlaneE2ETest.java
```diff
@@ -0,0 +1,199 @@
...
+    testSupport.executePfbImportQuartzJob(collectionId, minimalDataPfb);
+    var entities = getRecordedEntitiesSerde();
+    assertThat(entities.size()).isEqualTo(1);
+    var entity = entities.get(0);
+    assertThat(entity.name()).isEqualTo("HG01101_cram");
+    assertThat(entity.entityType()).isEqualTo("submitted_aligned_reads");
+
+    int totalAttributes = 19;
+    int nullAttributes = 3;
+    int expectedAttributes = totalAttributes - nullAttributes;
```
calypsomatic, line 114: Bit of a nit, but it seems weird to set these variables (`totalAttributes` and `nullAttributes`) to an exact number and use them only to determine another variable.  Is the idea that we could change these numbers in the future?  We could just have a comment explaining where the number for `expectedAttributes` comes from

service/src/test/java/org/databiosphere/workspacedataservice/service/RawlsRecordSinkTest.java
```diff
@@ -0,0 +1,27 @@
...
+import org.junit.jupiter.api.Test;
+import org.springframework.boot.test.context.SpringBootTest;
+
+@SpringBootTest
+class RawlsRecordSinkTest {
+
+  @Test
+  void getAttributeName_prependsPrefix() {
+    var recordSink = RawlsRecordSink.withPrefix("prefix");
+    RecordType recordType = RecordType.valueOf("widget");
```
calypsomatic, line 15: 😸

davidangb (commented):
comments inline, nothing big/blocking, this is close!

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobControlPlaneE2ETest.java
jladieu: I originally had a comment here but whenever there's an explanatory comment that could be expressed as compiler-checked code, I try to express it as such, since then it won't ever drift from the implementation it's describing :)

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobControlPlaneE2ETest.java
jladieu: It's a weird word that exists as an abbreviation for for `serialize/deserialize` which is a bit long winded to put into a method name.  I could put in an explanatory comment (subject to the pitfalls I mention [below](https://github.com/DataBiosphere/terra-workspace-data-service/pull/507#discussion_r1489634096)) but I'm hoping this is one of those terms that once you know it, it's self-evident everywhere else you see it.

service/src/main/java/org/databiosphere/workspacedataservice/rawls/Model.java
```diff
@@ -0,0 +1,125 @@
+package org.databiosphere.workspacedataservice.rawls;
+
+import static org.databiosphere.workspacedataservice.rawls.Model.Op.ADD_LIST_MEMBER;
+import static org.databiosphere.workspacedataservice.rawls.Model.Op.ADD_UPDATE_ATTRIBUTE;
+import static org.databiosphere.workspacedataservice.rawls.Model.Op.CREATE_ATTRIBUTE_ENTITY_REFERENCE_LIST;
+import static org.databiosphere.workspacedataservice.rawls.Model.Op.CREATE_ATTRIBUTE_VALUE_LIST;
+import static org.databiosphere.workspacedataservice.rawls.Model.Op.REMOVE_ATTRIBUTE;
```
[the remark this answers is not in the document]
jladieu: Yeah so I didn't even look upstream at Rawls, I just copied verbatum from import-service.  Thanks for dropping the link; I think it'd make sense to link out to that and keep this list in sync.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobControlPlaneE2ETest.java
calypsomatic: No that's fine I'll just reluctantly learn something new ;)

README.md
jladieu: Thinking too...I think this is where I net out:
- we should ideally require one or the other to be set and crash if missing
- (maybe provide an offramp/ignore ability for tests that _really_ don't want one but don't want to crash because of it)
- have the default be `data-plane` for tests until control plane becomes enough of a center of gravity to compete with this default assumption.

service/src/main/java/org/databiosphere/workspacedataservice/service/RawlsRecordSink.java
```diff
@@ -0,0 +1,108 @@
...
+
+  private Stream<? extends AttributeOperation> toOperations(String name, Object attributeValue) {
+    if (attributeValue instanceof List<?> values) {
+      return Stream.concat(
+          Stream.of(new RemoveAttribute(name), new CreateAttributeValueList(name)),
+          values.stream().map(value -> new AddListMember(name, value)));
+    }
+
+    return Stream.of(new AddUpdateAttribute(name, attributeValue));
+  }
```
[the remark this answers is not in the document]
jladieu: Yeah...I think you're right here.   I don't think we'd want more "E2E" style tests for this, so I think I'd want to handle this as a `RawlsRecordSinkTest` that covers these bits of logic by probing the existing public API methods, rather than cracking these open to get at them in unit tests.  I'll see if I can get something like this in today at least as a starting point to make it easy to build on in subsequent changes.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobControlPlaneE2ETest.java
jladieu: I added a comment anyway...I think my worries about it drifting are silly.

service/src/main/java/org/databiosphere/workspacedataservice/config/TwdsProperties.java
```diff
@@ -7,10 +7,19 @@
...
 @ConfigurationProperties(prefix = "twds")
 public class TwdsProperties {
+  DataImportProperties dataImport;
   InstanceProperties instance;
   String startupToken;
   TenancyProperties tenancy;
 
+  public DataImportProperties dataImport() {
+    return dataImport;
+  }
```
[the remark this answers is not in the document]
jladieu: Thanks, good catch!

service/src/main/java/org/databiosphere/workspacedataservice/rawls/Model.java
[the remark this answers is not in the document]
jladieu: Fixed by adding it in and linking out to the Rawls code.

service/src/main/java/org/databiosphere/workspacedataservice/rawls/Model.java
```diff
@@ -0,0 +1,125 @@
...
+          RemoveAttribute,
+          CreateAttributeValueList,
+          CreateAttributeEntityReferenceList {
+    Op op();
+  }
+
+  // Concrete subclasses of AttributeOperation
+  public record AddUpdateAttribute(
+      @JsonProperty("attributeName") String attributeName,
+      @JsonProperty("addUpdateAttribute") Object addUpdateAttribute)
```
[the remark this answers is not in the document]
jladieu: Ooh nice one.  I didn't like how chatty these were getting, but I think these `@JsonProperty` crept in while I was struggling with Jackson before I had the `@JsonSubTypes` stuff working.

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteServiceConfig.java
```diff
@@ -1,12 +1,54 @@
...
 public class BatchWriteServiceConfig {
+  private static final Logger logger = LoggerFactory.getLogger(BatchWriteServiceConfig.class);
+
+  private final TwdsProperties twdsProperties;
+  private RecordService recordService;
+
+  @Autowired
+  public BatchWriteServiceConfig(TwdsProperties twdsProperties) {
+    this.twdsProperties = twdsProperties;
+  }
```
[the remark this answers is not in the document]
jladieu: Removed.

service/src/main/java/org/databiosphere/workspacedataservice/service/RawlsRecordSink.java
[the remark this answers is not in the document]
jladieu: Saving this for a followup commit which I'll push after this most recent round of fixen.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobControlPlaneE2ETest.java
```diff
@@ -0,0 +1,199 @@
...
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.core.io.Resource;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.ActiveProfiles;
+
+/**
+ * Tests for PFB import that execute "end-to-end" - that is, they go through the whole process of
+ * parsing the PFB, creating tables in Postgres, inserting rows, and then reading back the
+ * rows/counts/schema from Postgres.
+ */
```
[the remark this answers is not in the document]
jladieu: Fixed by adjusting them to reflect the current state and including pointers to the tickets which will cause the message to become outdated again :D

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobControlPlaneE2ETest.java
```diff
@@ -0,0 +1,199 @@
...
+    assertThat(json)
+        .contains(
+            "[{\"name\":\"HG01101_cram\",\"entityType\":\"submitted_aligned_reads\",\"operations\":[");
+
+    // check a sample "Op" attribute for expected "op", "attributeName", and
+    // "addUpdateAttribute" attributes.
+    assertThat(json)
+        .contains(
+            "{\"op\":\"AddUpdateAttribute\",\"attributeName\":\"pfb:error_type\",\"addUpdateAttribute\":\"file_size\"}");
+  }
```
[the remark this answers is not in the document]
jladieu: Great idea...I've checked in expected values JSON...I could use a sanity check that what I've put in there doesn't have any blatant errors.

jladieu:
I resolved most of the feedback and marked corresponding comments as *resolved*.  

Still TODO, but which I'd love to not hold up further :eyes::

- [ ] Add more unit tests to improve coverage of `RawlsRecordSink`, will do in another commit or if it gets tough in a followup PR
- [ ] Do a better job of enforcing the assumption that `data-plane` or `control-plane` profile should be set.  Followup ticket: [AJ-1629](https://broadworkbench.atlassian.net/browse/AJ-1629)

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteServiceConfig.java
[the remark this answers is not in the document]
jladieu: Hopefully we can address this as part of [AJ-1629](https://broadworkbench.atlassian.net/browse/AJ-1629)

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/rawls/Model.java
```diff
@@ -0,0 +1,133 @@
...
+  /** Deserialize into concrete subclasses based on the value of the "op" attribute in the JSON. */
+  @JsonTypeInfo(use = Id.NAME, property = "op")
+  @JsonSubTypes({
+    @Type(value = AddUpdateAttribute.class, name = "AddUpdateAttribute"),
+    @Type(value = AddListMember.class, name = "AddListMember"),
+    @Type(value = RemoveAttribute.class, name = "RemoveAttribute"),
+    @Type(value = CreateAttributeValueList.class, name = "CreateAttributeValueList"),
+    @Type(
+        value = CreateAttributeEntityReferenceList.class,
+        name = "CreateAttributeEntityReferenceList")
```
davidangb, line 71: I don't think we actually need `RemoveListMember`, but if we are gonna model it here, need to add it to `@JsonSubTypes`

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobControlPlaneE2ETest.java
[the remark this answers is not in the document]
davidangb: They look correct to me on the surface; I didn't try to verify all the individual values are correct, but structure looks good.

davidangb (approved):
one last tweak, giving this a proactive :+1: assuming nothing else changes!

service/src/main/java/org/databiosphere/workspacedataservice/rawls/Model.java
jladieu: Thanks, pushed an amendment with this included.  I agree we don't need it but I biased toward completeness and consistency with Rawls.

jladieu:
@davidangb @calypsomatic I added another commit which includes enhancements to the unit tests and reset the reviews.  Please take hopefully one last look! :)

calypsomatic (approved):
Nice testing!

yuliadub:
I think I have mostly internalized these changes but it would be super helpful to do a walkthrough of this PR in a it more detail, maybe sometime next week? (or during future office hours)

davidangb (approved):
thank you!

PR #508, opened 2024-02-12
[AJ-1611] Support top level arrays in JSON fields

nawatts:
Currently, WDS does not support top level arrays in JSON fields. This adds support for them.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -722,7 +722,10 @@ public class RecordDao {
         return LocalDateTime.parse(sVal, DateTimeFormatter.ISO_LOCAL_DATE_TIME);
       }
     }
-    if (attVal instanceof Map<?, ?>) {
+    // There are two possibilities for JSON data types...
+    // One, the value is a Map or List that needs to be serialized to a JSON string.
+    // Two, the value is a String that's already JSON encoded.
+    if (typeMapping == DataTypeMapping.JSON && !(attVal instanceof String)) {
       try {
         return objectMapper.writeValueAsString(attVal);
       } catch (JsonProcessingException e) {
@@ -983,8 +986,7 @@ public class RecordDao {
         return getArrayValue(pgArray.getArray(), typeMapping);
       }
       if (typeMapping == DataTypeMapping.JSON) {
-        return objectMapper.readValue(
-            object.toString(), new TypeReference<Map<String, Object>>() {});
+        return objectMapper.readValue(object.toString(), new TypeReference<Object>() {});
       }
       return object;
     }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -7,7 +7,6 @@ import static org.databiosphere.workspacedataservice.service.model.ReservedNames
 import static org.databiosphere.workspacedataservice.service.model.exception.InvalidNameException.NameType.ATTRIBUTE;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Maps;
@@ -986,7 +985,7 @@ public class RecordDao {
         return getArrayValue(pgArray.getArray(), typeMapping);
       }
       if (typeMapping == DataTypeMapping.JSON) {
-        return objectMapper.readValue(object.toString(), new TypeReference<Object>() {});
+        return objectMapper.readValue(object.toString(), Object.class);
       }
       return object;
     }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
@@ -21,6 +21,7 @@ import static org.databiosphere.workspacedataservice.service.model.DataTypeMappi
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.collect.Streams;
 import com.google.mu.util.stream.BiStream;
 import java.math.BigDecimal;
 import java.math.BigInteger;
@@ -203,7 +204,13 @@ public class DataTypeInferer {
 
   public boolean isValidJson(String val) {
     JsonNode jsonNode = parseToJsonNode(val);
-    return jsonNode != null && jsonNode.isObject();
+    // Arrays of primitive types should be inferred as ARRAY_OF_STRING, ARRAY_OF_NUMBER, etc.
+    // Arrays should only be inferred as JSON if they contain an object or array.
+    return jsonNode != null
+        && (jsonNode.isObject()
+            || (jsonNode.isArray()
+                && Streams.stream(jsonNode.elements())
+                    .anyMatch(childNode -> childNode.isObject() || childNode.isArray())));
   }
 
   public boolean isArray(String val) {
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
@@ -204,13 +204,21 @@ public class DataTypeInferer {
 
   public boolean isValidJson(String val) {
     JsonNode jsonNode = parseToJsonNode(val);
+
+    if (jsonNode == null) {
+      return false;
+    }
+
+    boolean isObject = jsonNode.isObject();
+
     // Arrays of primitive types should be inferred as ARRAY_OF_STRING, ARRAY_OF_NUMBER, etc.
     // Arrays should only be inferred as JSON if they contain an object or array.
-    return jsonNode != null
-        && (jsonNode.isObject()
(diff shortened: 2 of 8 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -722,7 +722,7 @@ private Object getValueForSql(Object attVal, DataTypeMapping typeMapping) {
         return LocalDateTime.parse(sVal, DateTimeFormatter.ISO_LOCAL_DATE_TIME);
       }
     }
-    if (attVal instanceof Map<?, ?>) {
```
ashanhol, line 725: Is Map<?,?> how JSON was being parsed before?

nawatts: Previously, there was an assumption that JSON attributes contained JSON objects (represented in Java as Maps). This PR breaks that assumption, since now JSON attributes can contain JSON arrays (represented in Java as Lists).

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -722,7 +722,10 @@ private Object getValueForSql(Object attVal, DataTypeMapping typeMapping) {
...
+    // There are two possibilities for JSON data types...
+    // One, the value is a Map or List that needs to be serialized to a JSON string.
+    // Two, the value is a String that's already JSON encoded.
```
nawatts, line 727: This took me a long time to figure out today. My original change broke specifying a JSON attribute value as a JSON encoded string and the only test that caught it was TsvDownloadTest. This PR adds a test RecordControllerMockMvcTest to more clearly test this behavior.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -983,8 +986,7 @@ private Object getAttributeValueForType(Object object, DataTypeMapping typeMappi
         return getArrayValue(pgArray.getArray(), typeMapping);
       }
       if (typeMapping == DataTypeMapping.JSON) {
-        return objectMapper.readValue(
-            object.toString(), new TypeReference<Map<String, Object>>() {});
+        return objectMapper.readValue(object.toString(), new TypeReference<Object>() {});
```
jladieu, line 989:
```suggestion
        return objectMapper.readValue(object.toString(), Object.class);
```
I think since you're no longer deserializing into a class with a generic, you don't need to use `TypeReference` anymore which is there to overcome type erasure when deserializing back into generic Java types.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -3052,4 +3056,61 @@ void testUpdateRelationArrayDataType() throws Exception {
...
+  }
+
+  static Stream<Arguments> jsonValues() {
+    return Stream.of(
+        Arguments.of(Map.of("value", "foo"), Map.of("value", "foo")),
+        Arguments.of("{\"value\":\"foo\"}", Map.of("value", "foo")),
+        Arguments.of(
+            List.of(Map.of("value", "foo"), Map.of("value", "bar")),
+            List.of(Map.of("value", "foo"), Map.of("value", "bar"))),
+        Arguments.of(
```
jladieu, line 3112: Thanks for adding more tests :)

Would you consider a few more samples that involve non-string type values? (eg: numbers, dates, etc.)

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -203,7 +204,13 @@ private JsonNode parseToJsonNode(String val) {
 
   public boolean isValidJson(String val) {
     JsonNode jsonNode = parseToJsonNode(val);
-    return jsonNode != null && jsonNode.isObject();
+    // Arrays of primitive types should be inferred as ARRAY_OF_STRING, ARRAY_OF_NUMBER, etc.
+    // Arrays should only be inferred as JSON if they contain an object or array.
+    return jsonNode != null
+        && (jsonNode.isObject()
+            || (jsonNode.isArray()
```
jladieu, line 211: Maybe extract everything after the `||` into a private (or heck, maybe even public?) method '

`isValidJsonArray(jsonNode)`

jladieu (approved):

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/508/commits/4ea722d16833e5d3f6cc387ff3869e4842421638.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
nawatts: I kept it in one method to avoid parsing the string twice. But restructured a bit in https://github.com/DataBiosphere/terra-workspace-data-service/pull/508/commits/0261e6cba9ecccbf2298868636afeaccaa9cfa28.

ashanhol (approved):

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
jladieu: Ahh okay, your changes look fine, but I think my suggestion still stands with the intent of shrinking the logic footprint in this method... you can still extract a method without re-parsing the JSON, it just would definitely be `private` in that case since I don't think it'd be useful externally but it'd still shrink the logic down to:

```
  return jsonNode.isObject() || isValidJsonArray(jsonNode);
}
...

// no additional parsing needed
private boolean isValidJsonArray(JsonNode jsonNode) {
  return jsonNode.isArray()
            && Streams.stream(jsonNode.elements())
                .anyMatch(childNode -> childNode.isObject() || childNode.isArray());
}
```

davidangb (commented):
From reading the PR, is it correct that arrays-of-json are stored using a (scalar) `jsonb` column in the db? In other words, is this PR changing what is allowed in the JSON type, as opposed to creating a new ARRAY_OF_JSON type?

That's better than we had before, but it still means we don't have first-class support for arrays of json, i.e. you can't see in the UI how many elements are in that array. I am ok merging this PR because it advances us but I would want to check with Product if we can close out the tickets.

jladieu (approved):

nawatts:
> From reading the PR, is it correct that arrays-of-json are stored using a (scalar) jsonb column in the db? In other words, is this PR changing what is allowed in the JSON type, as opposed to creating a new ARRAY_OF_JSON type?

Correct. I think my vote would be to leave JSON as a special case / escape hatch for flexible data. Arrays are valid JSON, so I think it would be confusing to have a separate type for arrays of JSON. It would persist the current definition of "JSON" as a limited subset of JSON.

Also, before this PR, you could still put arrays in JSON fields, just not at the root level. For example,
```json
 {"list": [{"value": "foo"}, {"value": "bar"}, {"value": "baz"}]}
```
So even if there were two types, I think the line between "JSON" and "array of JSON" would be blurry and further cause for confusion.

> That's better than we had before, but it still means we don't have first-class support for arrays of json, i.e. you can't see in the UI how many elements are in that array.

You can! Because Entity Service's allows mixed types in a column, the UI must only look at the field value. There's a UI bug interfering with the display of values inline in the cell, but it does show the number of items and clicking on that shows a modal with the list of values.

Screenshot 2024-02-13 at 10 23 14 AM
Screenshot 2024-02-13 at 10 23 19 AM

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
nawatts: I think I'll still resist... I think splitting up into multiple methods muddles the distinction between "JSON" the WDS type and "JSON" the text format. As is, the method interfaces refer to the WDS type and only the internals of the `isValidJson` method refer to the text format.

davidangb:
> Arrays are valid JSON, so I think it would be confusing to have a separate type for arrays of JSON: it would persist the current definition of "JSON" as a limited subset of JSON.

well, `[1,2,3]` is also valid JSON, but we have a separate `ARRAY_OF_NUMBER` class to handle it.

To me, this is really a choice between:
1. We have first-class support for arrays of any of our scalar types (array of number, array of boolean, array of json)
2. We have a "JSON" datatype that allows you to persist anything syntactically valid which does not have its own datatype (arrays of arrays, arrays of mixed datatypes, array of json)

Consider a future ticket like https://broadworkbench.atlassian.net/browse/AJ-1439. If we wanted only the first 10 items in an array, we can implement that ticket for arrays of numbers, arrays of strings, etc by writing SQL along the lines of `select mycolumn[1:10] from mytable…` - but we'd need a separate implementation for arrays of json, or we just don't support getting the first 10 items out of an array of json.

Similarly, if we have Cromwell workflows that require an array input, we'd need special handling to identify which JSON columns actually contain arrays, while other columns are clearly labeled ARRAY_OF_NUMBER, ARRAY_OF_STRING, etc

I am honestly ok with either approach. I just want to elevate the choice of which approach we should have to include Product and any other engineers who aren't looking at this PR, so we're all clear on the implications.

I'll also reiterate: this PR eliminates a bug/error, so I think it's valuable even if we decide we want future work on the array-of-json front.

nawatts:
Alright, I'm convinced. I see no downside to your approach and I certainly don't want to deal with a migration later.

Although, from this discussion, I've come to the conclusion that "JSON" is a bad name for this type. Maybe we could rename it to "Object" or "Map"?

PR #509, opened 2024-02-12
AJ-1591: startup validation

davidangb:
Some more prefactoring to help out eventually with #489 

In this PR:
* create a new "startup" package
* move InstanceInitializer* classes into the startup package
* expand our ConfigurationProperties classes to include the top-level `twds` property hierarchy and, importantly, the `twds.instance.workspace-id` property (which is populated from the `$WORKSPACE_ID` env var)
* add a new `StartupConfig` class in the startup package, which is responsible for performing some on-startup validations including validating the `twds.instance.workspace-id` property

NOT in this PR to save scope:
* I did not attempt to consolidate InstanceInitializer stuff into StartupConfig, but I think that would be possible if we wanted to do it
* I did not attempt to update any other classes that read from or require the `twds.instance.workspace-id` property. For instance, the `PfbQuartzJob` bean will fail if `$WORKSPACE_ID` is not a valid UUID, and I didn't try to change that
* I did not build out classes for ALL properties under `twds`; I figured we could do that as we need them

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
@@ -1,10 +1,8 @@
 package org.databiosphere.workspacedataservice;
 
 import com.microsoft.applicationinsights.attach.ApplicationInsights;
-import org.databiosphere.workspacedataservice.config.TwdsProperties;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
-import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cache.annotation.EnableCaching;
 import org.springframework.retry.annotation.EnableRetry;
 import org.springframework.transaction.annotation.EnableTransactionManagement;
@@ -19,7 +17,6 @@ import org.springframework.transaction.annotation.EnableTransactionManagement;
 @EnableRetry
 @EnableTransactionManagement
 @EnableCaching
-@EnableConfigurationProperties(TwdsProperties.class)
 public class WorkspaceDataServiceApplication {
 
   public static void main(String[] args) {
--- a/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
@@ -1,8 +1,10 @@
 package org.databiosphere.workspacedataservice;
 
 import com.microsoft.applicationinsights.attach.ApplicationInsights;
+import org.databiosphere.workspacedataservice.config.TwdsProperties;
 import org.springframework.boot.SpringApplication;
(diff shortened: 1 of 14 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
```diff
@@ -19,7 +17,6 @@
 @EnableRetry
 @EnableTransactionManagement
 @EnableCaching
-@EnableConfigurationProperties(WdsTenancyProperties.class)
```
davidangb, line 22: I realized that by adding `@Configuration` to a property class, we don't need to list the class(es) here, so I removed it.

service/src/test/java/org/databiosphere/workspacedataservice/config/TenancyPropertiesDataPlaneTest.java
```diff
@@ -0,0 +1,33 @@
+package org.databiosphere.workspacedataservice.config;
```
davidangb, line 1: This was renamed from WdsTenancyPropertiesDataPlaneTest; unfortunately github is showing this as a delete of the old file and a brand new file

ashanhol (approved):
seems good, looks like just moving stuff around to accommodate the new startup/property validation

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/startup/StartupConfig.java
```diff
@@ -0,0 +1,62 @@
...
+
+/** */
+@Component
+public class StartupConfig {
+
+  private static final Logger logger = LoggerFactory.getLogger(StartupConfig.class);
+
+  private final TwdsProperties twdsProperties;
+
+  public StartupConfig(TwdsProperties twdsProperties) {
```
jladieu, line 19: I imagine `TwdsProperties` becoming a burrito of stuff and eventually not wanting to pull it all in here.

Is it possible to inject just the `TenancyProperties` and `InstanceProperties` directly here?

jladieu: (No changes needed here, asking for food for thought)

jladieu (approved):
One question (no changes needed, but food for thought), but LGTM!

service/src/main/java/org/databiosphere/workspacedataservice/startup/StartupConfig.java
davidangb: The way I have it currently set up, it is not possible to inject `TenancyProperties` or `InstanceProperties`. To do this, we'd have to:
* remove `TenancyProperties` and `InstanceProperties` as member variables of `TwdsProperties`
* promote `TenancyProperties` and `InstanceProperties` to being first-class spring beans annotated with `@ConfigurationProperties(prefix = "twds.tenancy")` and `@ConfigurationProperties(prefix = "twds.instance")`, respectively

Then, any other bean that wants properties would have to inject whichever of `TwdsProperties`, `TenancyProperties`, and/or `InstanceProperties` that it needs

PR #510, opened 2024-02-12
AJ-1563 Clean up internal APIs

jladieu:
This is prefactoring for [AJ-1563](https://broadworkbench.atlassian.net/browse/AJ-1563).  Some of the internal APIs used by `BatchWriteService` can be tightened up with a little bit of refactoring in internal method signatures.

The commits are self-contained and might be easier to review individually.

[AJ-1563]: https://broadworkbench.atlassian.net/browse/AJ-1563?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
@@ -4,8 +4,10 @@ import static org.databiosphere.workspacedataservice.recordstream.TwoPassRecordS
 import static org.databiosphere.workspacedataservice.recordstream.TwoPassRecordSource.ImportMode.RELATIONS;
 
 import bio.terra.common.db.WriteTransaction;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMultimap;
 import com.google.common.collect.Multimaps;
+import com.google.mu.util.stream.BiStream;
 import java.io.Closeable;
 import java.io.IOException;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -148,34 +149,35 @@ private BatchWriteResult consumeWriteStream(
         // time we've seen this type, calculate a schema from its records and update the record type
         // as necessary. Then, write the records into the table.
         for (RecordType recType : groupedRecords.keySet()) {
-          List<Record> rList = groupedRecords.get(recType).asList();
```
jladieu, line 151: Creating a new instance variable `rList` to house the List version of the collection is what tipped me off to start this commit in the first place.  The key change that is unlocked by the refactoring is not having to actually call `asList()` here before passing the collection down.  And the nice thing is that the ordering is still respected later on if the collection ever needs to be iterated on.

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -148,42 +149,51 @@ private BatchWriteResult consumeWriteStream(
...
-                recType,
-                typesSeen.get(recType),
-                info.operationType(),
-                recordsForType,
-                primaryKey);
-            // update the result counts
-            result.increaseCount(recType, recordsForType.size());
-          }
-        }
+        OperationType opType = info.operationType();
```
jladieu, line 152: BiStream needs to operate on [effectively final](https://www.baeldung.com/java-effectively-final) local variables, which `info` is not because it gets changed by the earlier for loop.  However, the bistream didn't _need_ info and just needed `operationType()` - so doing this assignment here satisfies the requirement by making an effectively final `opType` variable.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -59,7 +61,8 @@ public Map<String, DataTypeMapping> inferTypes(RecordAttributes updatedAtts) {
     return result;
   }
 
-  public Map<String, DataTypeMapping> inferTypes(List<Record> records) {
+  @VisibleForTesting
+  Map<String, DataTypeMapping> inferTypes(Collection<Record> records) {
```
nawatts, line 65: Should this method be marked `VisibleForTesting`? Isn't it used by BatchWriteService?

nawatts (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -840,7 +841,7 @@ private String getInsertColList(List<String> existingTableSchema) {
   }
 
   @SuppressWarnings("squid:S2077")
-  public void batchDelete(UUID instanceId, RecordType recordType, List<Record> records) {
+  public void batchDelete(UUID instanceId, RecordType recordType, Collection<Record> records) {
```
davidangb, line 844: I am a bit hesitant to move from `List` to `Collection`, as these APIs are dependent on ordering of the incoming Records. If a caller used an unordered collection this could result in unexpected results.

davidangb (changes requested):
:+1: for the refactoring inside `BatchWriteService`.

I don't like moving method signatures from `List` to `Collection`, though, as it suggests that ordering is not relevant - ordering is critical in the input arguments.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
jladieu: Hah, good call...I removed this after all.

This commit ended up being way more underwhelming than I'd hoped...at first I was closing up a bunch of these methods but then realized the leaky part of the API is thus because of its close relationship with `TsvSerializer` which wouldn't let me privatize much at all unfortunately.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
jladieu: I punted the commit that switched out `List` for `Collection` although I think I'd still make the case that implicit ordering is sufficient, and in fact, all we can rely on.  As long as we don't mess with the order after it's handed to us we should be good.

I couldn't figure out where the strict ordering mattered in the underlying implementation (which I was squinting at while switching over from `List` -> `Collection`, so I'd like to hear more to better understand.

nawatts: I think because of side effects. For `batchDelete`, I could imagine the order matters if you're trying to delete a referenced record and the record referencing it in the same batch. You'd have to delete the referenced record first.

davidangb: Consider a payload to the `batchWriteRecords` API of:
```json
[
  {
    "operation": "upsert",
    "record": {
      "id": "1",
      "type": "mytype",
      "attributes": {
        "stringAttr": "foo"
      }
    }
  },
  {
    "operation": "upsert",
    "record": {
      "id": "1",
      "type": "mytype",
      "attributes": {
        "stringAttr": "bar"
      }
    }
  }
]
```
After submitting this payload, what would you expect to be the value of the "stringAttr" attribute on record "mytype:1"? Would it be "foo" or "bar"?

davidangb (approved):
thank you for switching back to `List`, :+1:

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
jladieu: Thanks for the example and helping me understand!  This makes sense to me...the last one should be recorded last, so "bar".

PR #511, opened 2024-02-13
AJ-1592 rename "instance" to "collection"

calypsomatic:
The term "instance" has been horribly confusing since the inception of WDS.  This PR is the first step in renaming "instance" to the hopefully less confusing "collection".  This PR only renames classes and variables in the code base.  It leaves for later PRs:
- Renaming APIs, parameters to APIs, or any other code at the controller level
- database tables and columns
- `twds.instance.` variables in `applicaton.yml` - I think these should be ok to rename without having any effects elsewhere, but I left it off this PR so I can verify

Reminder:


service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
```diff
@@ -28,35 +28,36 @@ public HttpSamDao(
   }
 
   /**
-   * Check if the current user has permission to create a "wds-instance" resource in Sam.
-   * Implemented as a check for write permission on the workspace which will contain this instance.
+   * Check if the current user has permission to create a collection by writing to the "workspace"
```
calypsomatic, line 31: Updated the comments in this file to be more accurate, since the "wds-instance" resource in Sam doesn't exist.  Hopefully the new comments are clear and accurate.

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamDao.java
```diff
@@ -21,40 +21,42 @@ public interface SamDao {
   String ACTION_READ = "read";
 
   /**
-   * Check if the current user has permission to create a "wds-instance" resource in Sam
+   * Check if the current user has permission to create a collection associated with a Sam workspace
```
calypsomatic, line 24: Comments updated here too.

service/src/main/java/org/databiosphere/workspacedataservice/dao/CollectionDao.java
```diff
@@ -0,0 +1,16 @@
+package org.databiosphere.workspacedataservice.dao;
```
ashanhol, line 1: Metacomentary (happy to move this conversation to slack if needed)
Do we think there'll be any confusion given that collection is a data structure? There was a PR changing lists to collections so if we think well be using java.util.collections this wording might get confusing

calypsomatic: I was told that the new term is now "Collection" and although I didn't love it, I did appreciate not having to make the decision myself.  So I will go with whatever everybody else decides, but we have to settle it before this PR goes through or else we'll have to go through all this again!

ashanhol: Actually @jladieu, share your java wisdom with me, why have you been more eager to use collections over lists recently? Do you think it'll be confusing if we see more collections (datastructure) pop up in the code?

davidangb (commented):
modulo the failing unit tests, this looks good to me. I think we'll see a long tail of stragglers that need renaming over time - given that, this is a good first pass and I am not too concerned if we missed anything.

service/src/main/java/org/databiosphere/workspacedataservice/dao/CollectionDao.java
calypsomatic: As it stands, I don't think there's any plain `Collection` class, just ids and daos and services and so forth.  Should all of those should be `WdsCollection{something}`?

davidangb (approved):

ashanhol (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dao/CollectionDao.java
jladieu: This is already merged, but wanted to reply to @ashanhol ...

My last couple attempts/suggestions to use `Collection` instead of `List` is coming out of an interest in trying to make our APIs relaxed to inputs _when possible_ ([robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) aka Postel's law/principle.

I'm not worried about those efforts resulting in a ton of sudden collisions/confusions, and in fact might be able to head them off by using an even more generic interface `Enumerable` :shrug: where possible?

PR #513, opened 2024-02-13
AJ-1556: streamline github workflows

davidangb:
Cleans up our GHAs. In this PR:

1. Update `actions/setup-java` to `v4` wherever it is out of date
2. Update `actions/checkout` to `v4` wherever it is out of date
3. Add `gradle/actions/setup-gradle@v3` wherever we use gradle. See https://github.com/marketplace/actions/build-with-gradle for more info; this action was not available when 
4. Remove `setup-java`'s gradle cache in favor of the one in `setup-gradle`
5. Remove explicit caching of gradle stuff via `actions/cache` in favor of the one in `setup-gradle`
6.  Change "Upload Test Reports" steps to only upload reports if tests failed, to save time … does anyone look at these?

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -11,13 +11,15 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v4
+      - name: Checkout sources
+        uses: actions/checkout@v4
       - name: Set up JDK
-        uses: actions/setup-java@v3
+        uses: actions/setup-java@v4
         with:
           java-version: '17'
           distribution: 'temurin'
-          cache: 'gradle'
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v3
 
       - name: Build all projects without running tests
         run: ./gradlew --build-cache build -x test
@@ -50,11 +52,13 @@ jobs:
         run: psql -h 127.0.0.1 -U postgres -f ./local-dev/local-postgres-init.sql
 
       - name: Set up JDK
-        uses: actions/setup-java@v3
+        uses: actions/setup-java@v4
         with:
           java-version: '17'
           distribution: 'temurin'
-          cache: 'gradle'
+
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v3
 
       - name: Cache SonarCloud packages
         uses: actions/cache@v3
@@ -62,13 +66,6 @@ jobs:
           path: ~/.sonar/cache
           key: ${{ runner.os }}-sonar
           restore-keys: ${{ runner.os }}-sonar
-
-      - name: Cache Gradle packages
-        uses: actions/cache@v3
-        with:
-          path: ~/.gradle/caches
-          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
-          restore-keys: ${{ runner.os }}-gradle
 
       - name: Build, assemble, and test
         id: build-test
@@ -80,10 +77,3 @@ jobs:
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: ./gradlew --build-cache sonar
-
-      - name: Upload Test Reports
-        if: always()
-        uses: actions/upload-artifact@v3
-        with:
-          name: Test Reports
-          path: service/build/reports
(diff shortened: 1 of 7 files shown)


.github/workflows/run-e2e-tests.yaml
```diff
@@ -40,48 +40,48 @@ jobs:
...
         uses: ./.github/actions/landing-zone-attach
         with:
-            run_name: "${{ env.ATTACH_BP_TO_LZ_RUN_NAME }}"
-            bee_name: ${{ env.BEE_NAME }}
-            token: '${{ env.TOKEN }}'
-            project_name: ${{needs.params-gen.outputs.project-name }}
+          run_name: "${{ env.ATTACH_BP_TO_LZ_RUN_NAME }}"
+          bee_name: ${{ env.BEE_NAME }}
+          token: '${{ env.TOKEN }}'
+          project_name: ${{needs.params-gen.outputs.project-name }}
```
nawatts, line 75: Super nit, but as long as you're formatting...

```suggestion
          project_name: ${{ needs.params-gen.outputs.project-name }}
```

Same below

.github/workflows/build-and-test.yml
```diff
@@ -80,10 +77,3 @@ jobs:
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: ./gradlew --build-cache sonar
-
-      - name: Upload Test Reports
-        if: always()
```
nawatts, line 85: Might be nice to upload reports on test failure, which would speed up the happy path but still leave them available for debugging.

nawatts (approved):

.github/workflows/run-e2e-tests.yaml
```diff
@@ -40,48 +40,48 @@ jobs:
...
         continue-on-error: true
         uses: ./.github/actions/landing-zone-attach
         with:
-            run_name: "${{ env.ATTACH_BP_TO_LZ_RUN_NAME }}"
-            bee_name: ${{ env.BEE_NAME }}
-            token: '${{ env.TOKEN }}'
-            project_name: ${{needs.params-gen.outputs.project-name }}
+          run_name: "${{ env.ATTACH_BP_TO_LZ_RUN_NAME }}"
+          bee_name: ${{ env.BEE_NAME }}
+          token: '${{ env.TOKEN }}'
```
yuliadub, line 74: stupid question incoming: is there a difference between ' and ", I cant seem to figure it out

davidangb: I doubt there's a difference, I bet this is the result of copy/paste (and I didn't change any of it, just left as-is)

.github/workflows/run-e2e-tests.yaml
```diff
@@ -40,48 +40,48 @@ jobs:
...
 
       - name: dispatch to terra-github-workflows
         id: FirstAttemptCreateBee
         continue-on-error: true
         uses: ./.github/actions/create-bee
         with:
-            bee_name: ${{ env.BEE_NAME }}
-            token: '${{ env.TOKEN }}'
+          bee_name: ${{ env.BEE_NAME }}
+          token: '${{ env.TOKEN }}'
```
calypsomatic, line 51: Another weird discrepancy, why does line 51 have quotes but 50 doesn't?

.github/workflows/run-e2e-tests.yaml
nawatts: Double quotes allow escape sequences, single quotes only allow printable characters.

https://yaml.org/spec/1.2.2/#single-quoted-style
https://yaml.org/spec/1.2.2/#double-quoted-style

calypsomatic (approved):
Would mildly prefer standardizing use of ' / " / no quote at all while you're at it, otherwise looks great.

.github/workflows/run-e2e-tests.yaml
davidangb: fixed

.github/workflows/build-and-test.yml
davidangb: I added these back in, with a condition to only run if tests failed; good call.

davidangb:
> Would mildly prefer standardizing use of ' / " / no quote at all while you're at it, otherwise looks great.

Given that there is actually a behavior difference between these (https://github.com/DataBiosphere/terra-workspace-data-service/pull/513#discussion_r1488440268), I don't really want to take it on in here.

.github/workflows/run-e2e-tests.yaml
yuliadub: @nawatts thank you 🙏

PR #514, opened 2024-02-13
Some additional rename suggestions and converting instanceid to collectionid

ashanhol:
Some additional rename suggestions and converting instanceid to collectionid 

-----------
Reminder:


davidangb (changes requested):
we can't rename anything in `apis-v1.yaml`. Doing so will cascade changes to our generated clients - you can see in this PR that some of the generated code was updated. This will break anyone using those clients.

To rename anything in `apis-v1.yaml` we need to create new v2 or v3 (etc) APIs, deprecate the v1 APIs, and fully remove the v1 APIs after some time.

service/src/main/resources/static/swagger/apis-v1.yaml
```diff
@@ -189,15 +189,15 @@ components:
         - status
         - created
         - updated
-        - instanceId
+        - collectionId
       properties:
```
calypsomatic, line 193: Ooh yeah I'm trying not to change anything in the APIs yet, even though it's obnoxious.

service/src/test/java/org/databiosphere/workspacedataservice/controller/JobControllerMockMvcTest.java
```diff
@@ -61,7 +61,7 @@ void smokeTestJobsInInstanceV1() throws Exception {
     // return a test job from the mocked JobDao
     UUID jobId1 = UUID.randomUUID();
     UUID jobId2 = UUID.randomUUID();
-    InstanceId instanceId = new InstanceId(UUID.randomUUID());
+    CollectionId collectionId = new CollectionId(UUID.randomUUID());
```
calypsomatic, line 64: Also deliberately leaving out the controller layer - since I associate that with the APIs I was going to leave this layer for a separate PR

ashanhol: This isn't actually renaming anything in the controller, just converting any use of InstanceId to CollectionId. But that rename also had to happen for the generated job model (which is why i modified the v1 api). I can go through and see if I can walk it back but if they're attached we might want to wait to rename the InstanceId.java

PR #515, opened 2024-02-14
[AJ-1611] Support top level arrays in JSON fields

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1611

Currently, WDS does not support top level arrays in JSON fields. This adds support for them by implementing an `ARRAY_OF_JSON` type.

Replacement for #508 based on https://github.com/DataBiosphere/terra-workspace-data-service/pull/508#issuecomment-1941879831.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -1035,7 +1035,8 @@ public class RecordDao {
       String[] jsonArray = (String[]) object;
       Object[] result = new Object[jsonArray.length];
       for (int i = 0; i < jsonArray.length; i++) {
-        result[i] = objectMapper.readValue(jsonArray[i], Object.class);
+        result[i] =
+            objectMapper.readValue(jsonArray[i], new TypeReference<Map<String, Object>>() {});
       }
       return result;
     }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -767,6 +767,18 @@ public class RecordDao {
           .map(Boolean::parseBoolean)
           .toList()
           .toArray(new Boolean[0]);
+      case ARRAY_OF_JSON -> attVal.stream()
+          .map(
+              el -> {
+                try {
+                  return objectMapper.writeValueAsString(el);
+                } catch (JsonProcessingException e) {
+                  LOGGER.error("Could not serialize array element to json string", e);
+                  throw new RuntimeException(e);
+                }
+              })
+          .toList()
+          .toArray(new String[0]);
       default -> throw new IllegalArgumentException("Unhandled array type " + typeMapping);
     };
   }
@@ -989,11 +1001,14 @@ public class RecordDao {
       return object;
     }
 
-    private Object getArrayValue(Object object, DataTypeMapping typeMapping) {
+    private Object getArrayValue(Object object, DataTypeMapping typeMapping)
+        throws JsonProcessingException {
       if (typeMapping == DataTypeMapping.ARRAY_OF_DATE_TIME) {
         return convertToLocalDateTime(object);
       } else if (typeMapping == DataTypeMapping.ARRAY_OF_DATE) {
         return convertToLocalDate(object);
+      } else if (typeMapping == DataTypeMapping.ARRAY_OF_JSON) {
+        return convertToJson(object);
       }
       return object;
     }
@@ -1012,6 +1027,15 @@ public class RecordDao {
       LocalDate[] result = new LocalDate[tzArray.length];
       for (int i = 0; i < tzArray.length; i++) {
         result[i] = tzArray[i].toLocalDate();
+      }
+      return result;
+    }
+
+    private Object[] convertToJson(Object object) throws JsonProcessingException {
+      String[] jsonArray = (String[]) object;
+      Object[] result = new Object[jsonArray.length];
+      for (int i = 0; i < jsonArray.length; i++) {
+        result[i] = objectMapper.readValue(jsonArray[i], Object.class);
       }
       return result;
     }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/tsv/TsvDeserializer.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/tsv/TsvDeserializer.java
@@ -136,6 +136,7 @@ public class TsvDeserializer extends StdDeserializer<RecordAttributes> {
         return objectMapper.readValue(val, new TypeReference<Map<String, Object>>() {});
       } catch (JsonProcessingException jpe) {
         // this shouldn't happen; if inferer.isValidJson(val) passes, so should the .readValue
+        LOGGER.error("Unable to parse validated JSON: {}", val);
         return val;
       }
     }
@@ -220,6 +221,7 @@ public class TsvDeserializer extends StdDeserializer<RecordAttributes> {
         return objectMapper.readValue(on.toString(), new TypeReference<Map<String, Object>>() {});
       } catch (JsonProcessingException jpe) {
         // this shouldn't happen since we've already parsed the array into JSON
+        LOGGER.error("Unable to parse array element JSON: {}", on);
         return on.toString();
       }
     }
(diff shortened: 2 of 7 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -3052,4 +3056,70 @@ void testUpdateRelationArrayDataType() throws Exception {
...
+  static Stream<Arguments> jsonValues() {
+    return Stream.of(
+        Arguments.of(
+            Map.of("string", "foo", "number", 1, "boolean", true),
+            Map.of("string", "foo", "number", BigInteger.valueOf(1), "boolean", Boolean.TRUE)),
+        // TODO: Is this case (creating a JSON attribute by passing a JSON encoded string)
+        // intentionally supported?
+        Arguments.of(
+            "{\"string\":\"foo\",\"number\":1,\"boolean\":true}",
+            Map.of("string", "foo", "number", BigInteger.valueOf(1), "boolean", Boolean.TRUE)),
```
nawatts, line 3115: After looking at how other types behave, I'm not sure that this is intentionally supported.

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -767,6 +767,18 @@ Object[] getListAsArray(List<?> attVal, DataTypeMapping typeMapping) {
...
           .toList()
           .toArray(new Boolean[0]);
+      case ARRAY_OF_JSON -> attVal.stream()
+          .map(
+              el -> {
+                try {
+                  return objectMapper.writeValueAsString(el);
+                } catch (JsonProcessingException e) {
+                  LOGGER.error("Could not serialize array element to json string", e);
+                  throw new RuntimeException(e);
```
davidangb, line 777: thinking out loud … if a user ever ended up with invalid data that couldn't be serialized, this would prevent them from seeing anything at all, any time the invalid row was in their query. Should this return `el.toString` instead, or even return the string "ERROR" here? I'm waffling on this, since if we swallow the error, an end user might never notice the problem and run Cromwell analyses on this …

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -1015,6 +1030,15 @@ private LocalDate[] convertToLocalDate(Object object) {
...
     }
+
+    private Object[] convertToJson(Object object) throws JsonProcessingException {
+      String[] jsonArray = (String[]) object;
+      Object[] result = new Object[jsonArray.length];
+      for (int i = 0; i < jsonArray.length; i++) {
+        result[i] = objectMapper.readValue(jsonArray[i], Object.class);
+      }
+      return result;
+    }
```
davidangb, line 1041: Is there anything more specific this could return other than `Object[]`? I believe that `readValue` will create these as a `Map<String, Object>` - is it feasible to return `Map<String, Object>[]`?

If this is onerous or will be a big effort, I wouldn't hold the PR on this.

service/src/main/java/org/databiosphere/workspacedataservice/tsv/TsvDeserializer.java
```diff
@@ -214,6 +215,14 @@ public Object arrayElementToObject(JsonNode element) {
     if (element instanceof BooleanNode bn) {
       return bn.asBoolean();
     }
+    if (element instanceof ObjectNode on) {
+      try {
+        return objectMapper.readValue(on.toString(), new TypeReference<Map<String, Object>>() {});
+      } catch (JsonProcessingException jpe) {
+        // this shouldn't happen since we've already parsed the array into JSON
+        return on.toString();
```
davidangb, line 223: agree this shouldn't happen - log an error here (for Sentry to pick up) or if you're feeling super ambitious, save an observability event - let's have some way of tracking if this ever happens.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
davidangb: I bet this is a (potentially undesired) side effect of TSV parsing

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
nawatts: I'm leaning towards fail loudly here. I think this shouldn't ever happen, since the data must have been parsed as JSON when uploaded.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
nawatts: 🤔 I think I need to add some more test cases and possibly validation. I'm not sure if this will currently allow uploading an array of mixed types or an array of arrays, etc.

Stricter validation would be more consistent with defining "JSON" as "map"/"object".

nawatts: I think this reinforces my opinion that "JSON" is a bad name for this type.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -767,6 +767,18 @@ Object[] getListAsArray(List<?> attVal, DataTypeMapping typeMapping) {
...
+              el -> {
+                try {
+                  return objectMapper.writeValueAsString(el);
+                } catch (JsonProcessingException e) {
+                  LOGGER.error("Could not serialize array element to json string", e);
+                  throw new RuntimeException(e);
+                }
+              })
+          .toList()
+          .toArray(new String[0]);
```
yuliadub, line 781: a bit confused on why this is this new String[0], as in array of size 0 or am I reading this wrong?

nawatts: Admittedly, I was cargo culting the other switch cases, but it looks like this determines the type of the returned array.

> Returns an array containing all of the elements in this list in proper sequence (from first to last element); the runtime type of the returned array is that of the specified array. If the list fits in the specified array, it is returned therein. Otherwise, a new array is allocated with the runtime type of the specified array and the size of this list.
> https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/List.html#toArray(T%5B%5D)

service/src/main/java/org/databiosphere/workspacedataservice/tsv/TsvDeserializer.java
davidangb: this ^ is  the only change keeping me from a 👍 , hopefully can just log this error

nawatts: 👍 I am planning to address this (and other comments), but haven't gotten a chance to revisit this PR lately.

nawatts: Added error logs in https://github.com/DataBiosphere/terra-workspace-data-service/pull/515/commits/5a65d77a69fbd3dd619aa9097311d4efaec632af.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/515/commits/c1e5779fa8f2b3df12820f25b9c21a258dd3766a along with a test to check that lists of mixed types don't get stored as JSON array.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -3052,4 +3056,118 @@ void testUpdateRelationArrayDataType() throws Exception {
...
+                    "/{instanceId}/records/{version}/{recordType}/{recordId}",
+                    instanceId,
+                    versionId,
+                    recordType,
+                    recordId))
+            .andExpect(status().isOk())
+            .andReturn();
+    RecordResponse record = fromJson(mvcSingleResult, RecordResponse.class);
+
+    List<String> expectedValue = List.of("{value=foo}", "{value=bar}", "baz");
```
nawatts, line 3170: I suspect this is not the intended behavior here (I think the first two elements should be encoded as JSON), but this is the current behavior (not introduced in this PR).

davidangb (approved):
ty!

PR #519, opened 2024-02-14
AJ-1591: job permissions

davidangb:
Well, this ended up being bigger than I thought it would be. In this PR:
* new methods in `CollectionService`: `canReadCollection` and `canWriteCollection`. These methods determine the workspace associated with a given collection, then check permissions on that workspace.
* get-job and list-jobs now call `CollectionService.canReadCollection` to check their permissions.
* extensive new test coverage for permissions, both control-plane and data-plane, in `JobServiceControlPlaneTest` and `JobServiceDataPlaneTest`
* renamed methods in `SamDao` from `.CollectionPermission` to `…WorkspacePermission` to be clearer about what they're actually doing

The work in this PR uncovered a bunch of outstanding work. To prevent this PR from ballooning, I am suggesting we tackle the outstanding work in future PRs:
- [ ] [AJ-1630](https://broadworkbench.atlassian.net/browse/AJ-1630): throw an error if the deployment does not support virtual collections and the specified collection is not found in the db
- [ ] [AJ-1631](https://broadworkbench.atlassian.net/browse/AJ-1631): use `CollectionService.canWriteCollection` in the code path to create an import job; currently this path does not respect the collection::workspace association
- [ ] [AJ-1632](https://broadworkbench.atlassian.net/browse/AJ-1632): exceptions thrown by get-job and list-jobs should return a neutral message "collection not found or you do not have permission to see it" instead of different errors for not-found and no-permission ([why?](https://apihandyman.io/hands-off-that-resource-http-status-code-401-vs-403-vs-404/))
- [ ] [AJ-1633](https://broadworkbench.atlassian.net/browse/AJ-1633): other methods in SamDao need to accept a workspaceId argument instead of relying on the default `$WORKSPACE_ID` value
- [ ] no ticket yet: all permission checks in all remaining APIs should use CollectionService instead of direct calls to SamDao. Since cWDS doesn't use any of these remaining APIs this can wait a bit.
- [ ] maybe: deprecate direct calls to SamDao instead of going through CollectionService? Since we now want to route all permissions through `CollectionService.canReadCollection` and `canWriteCollection`, should we make it clear the direct Sam methods are hands-off? This may be premature until we tackle the previous bullet point.

[AJ-1631]: https://broadworkbench.atlassian.net/browse/AJ-1631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/publish-app-version.yml
+++ b/.github/workflows/publish-app-version.yml
@@ -1,0 +1,57 @@
+name: Publish new app version to terra-helmfile
+on:
+  workflow_call:
+    inputs:
+      new-tag:
+        required: true
+        type: string
+      jiraId:
+        required: true
+        type: string
+    secrets:
+      BROADBOT_TOKEN:
+        required: true
+
+jobs:
+  create-terra-helmfile-pr:
+    name: Create terra-helmfile PR for latest WDS version
+    runs-on: ubuntu-latest
+    steps:
+      - name: Look for AJ (Analysis Journeys) Jira ID in the manual workflow trigger message
+        id: find-jira-id
+        run: |
+          set +e
+          JIRA_ID=$(echo "${{ inputs.jiraId }}" | grep -iEo -m 1 'AJ-[0-9]+' | head -1 || '')
+          if [[ -z "$JIRA_ID" ]]; then
+            echo "JIRA_ID missing, PR to terra-helmfile will not be created" 1>&2
+            exit 1;
+          else
+            echo "JIRA_ID=${JIRA_ID}" >> $GITHUB_OUTPUT
+          fi
+          set -e
+
+      - name: Clone terra-helmfile
+        uses: actions/checkout@v4
+        with:
+          repository: broadinstitute/terra-helmfile
+          token: ${{ secrets.BROADBOT_TOKEN }} # Has to be set at checkout AND later when pushing to work
+          path: terra-helmfile
+
+      - name: Create terra-helmfile PR for latest WDS version
+        env:
+          BROADBOT_TOKEN: ${{ secrets.BROADBOT_TOKEN }}
+          GH_TOKEN: ${{ secrets.BROADBOT_TOKEN }}
+          JIRA_ID: ${{ steps.find-jira-id.outputs.JIRA_ID }}
+          HELM_NEW_TAG: ${{ inputs.new-tag }}
+        run: |
+          set -e
+          cd terra-helmfile
+          HELM_CUR_TAG=$(grep "/terra-workspace-data-service:" charts/wds/values.yaml | sed "s,.*/terra-workspace-data-service:,,")
+          git checkout -b ${JIRA_ID}-auto-update-${HELM_NEW_TAG}
+          [[ -n "$HELM_CUR_TAG" && -n "$HELM_NEW_TAG" ]]
+          sed -i "s/$HELM_CUR_TAG/$HELM_NEW_TAG/" charts/wds/values.yaml
+          git config --global user.name "broadbot"
+          git config --global user.email "broadbot@broadinstitute.org"
+          git commit -am "${JIRA_ID}: Auto update WDS version to $HELM_NEW_TAG"
+          git push -u origin ${JIRA_ID}-auto-update-${HELM_NEW_TAG}
+          gh pr create --title "${JIRA_ID}: auto update WDS version to $HELM_NEW_TAG" --body "${JIRA_ID} helm chart update" --label "automerge"
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -5,15 +5,10 @@ on:
       new-tag:
         required: true
         type: string
-      jiraId:
-        required: true
-        type: string
     secrets:
       ACR_SP_PASSWORD:
         required: true
       ACR_SP_USER:
-        required: true
-      BROADBOT_TOKEN:
         required: true
 
 env:
@@ -35,18 +30,6 @@ jobs:
         with:
           java-version: '17'
           distribution: 'temurin'
-
-      - name: Look for AJ (Analysis Journeys) Jira ID in the manual workflow trigger message
-        id: find-jira-id
-        run: |
-          set +e
-          JIRA_ID=$(echo "${{ inputs.jiraId }}" | grep -iEo -m 1 'AJ-[0-9]+' | head -1 || '')
-          if [[ -z "$JIRA_ID" ]]; then
-            echo "JIRA_ID=missing" >> $GITHUB_OUTPUT
-          else
-            echo "JIRA_ID=${JIRA_ID}" >> $GITHUB_OUTPUT
-          fi
-          set -e
 
       - name: Set commit short hash
         id: setHash
@@ -115,33 +98,3 @@ jobs:
           --username ${{ secrets.ACR_SP_USER }} \
           --password-stdin
           docker push --all-tags ${{ steps.acr-image-name.outputs.name }}
-
-      - name: Clone terra-helmfile
-        uses: actions/checkout@v4
-        with:
-          repository: broadinstitute/terra-helmfile
-          token: ${{ secrets.BROADBOT_TOKEN }} # Has to be set at checkout AND later when pushing to work
-          path: terra-helmfile
-
-      - name: Create terra-helmfile PR for latest WDS version
-        env:
-          BROADBOT_TOKEN: ${{ secrets.BROADBOT_TOKEN }}
-          GH_TOKEN: ${{ secrets.BROADBOT_TOKEN }}
-        run: |
-          set -e
-          JIRA_ID=${{ steps.find-jira-id.outputs.JIRA_ID }}
-          if [[ $JIRA_ID == "missing" ]]; then
-            echo "JIRA_ID missing, PR to terra-helmfile will not be created"
-            exit 0;
-          fi
-          cd terra-helmfile
-          HELM_CUR_TAG=$(grep "/terra-workspace-data-service:" charts/wds/values.yaml | sed "s,.*/terra-workspace-data-service:,,")
-          HELM_NEW_TAG=${{ inputs.new-tag }}
-          git checkout -b ${JIRA_ID}-auto-update-${HELM_NEW_TAG}
-          [[ -n "$HELM_CUR_TAG" && -n "$HELM_NEW_TAG" ]]
-          sed -i "s/$HELM_CUR_TAG/$HELM_NEW_TAG/" charts/wds/values.yaml
-          git config --global user.name "broadbot"
-          git config --global user.email "broadbot@broadinstitute.org"
-          git commit -am "${JIRA_ID}: Auto update WDS version to $HELM_NEW_TAG"
-          git push -u origin ${JIRA_ID}-auto-update-${HELM_NEW_TAG}
-          gh pr create --title "${JIRA_ID}: auto update WDS version to $HELM_NEW_TAG" --body "${JIRA_ID} helm chart update" --label "automerge"
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -7,7 +7,7 @@ on:
         # Friendly description to be shown in the UI instead of 'name'
         description: 'Jira ID'
         # Default value if no value is explicitly provided
-        default: 'Required: Include latest Jira ID'
+        default: 'AJ-###'
         # Input has to be provided for the workflow to run
         required: true
         # The data type of the input
@@ -53,11 +53,9 @@ jobs:
     uses: ./.github/workflows/publish-docker.yml
     with:
       new-tag: ${{ needs.tag-job.outputs.new_tag }}
-      jiraId: ${{ inputs.jiraId }}
     secrets:
       ACR_SP_PASSWORD: ${{ secrets.ACR_SP_PASSWORD }}
       ACR_SP_USER: ${{ secrets.ACR_SP_USER }}
-      BROADBOT_TOKEN: ${{ secrets.BROADBOT_TOKEN }}
   release-job:
     needs: tag-job
     permissions:
@@ -77,3 +75,11 @@ jobs:
           config-name: release-drafter-config.yml
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  publish-app-version-job:
+    needs: [tag-job, docker-image-job]
+    uses: ./.github/workflows/publish-app-version.yml
+    with:
+      new-tag: ${{ needs.tag-job.outputs.new_tag }}
+      jiraId: ${{ inputs.jiraId }}
+    secrets:
+      BROADBOT_TOKEN: ${{ secrets.BROADBOT_TOKEN }}


service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
```diff
@@ -28,30 +28,25 @@ public class CollectionService {
...
-    } catch (Exception e) {
-      LOGGER.warn("WORKSPACE_ID could not be parsed into a UUID: {}", workspaceIdProperty);
-    }
-    if (workspaceUuid == null) {
-      workspaceId = null;
-    } else {
-      workspaceId = WorkspaceId.of(workspaceUuid);
+    this.twdsProperties = twdsProperties;
+    // stash the workspace id from config into a member var
+    if (twdsProperties.getInstance() != null
```
calypsomatic, line 47: should this method be renamed `getCollection`?

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceControlPlaneTest.java
```diff
@@ -0,0 +1,202 @@
...
+    UUID jobId = UUID.randomUUID();
+    // job not found
+    when(jobDao.getJob(jobId))
+        .thenThrow(new EmptyResultDataAccessException("unit test intentional error", 1));
+
+    // Act / assert
+    Exception actual = assertThrows(MissingObjectException.class, () -> jobService.getJob(jobId));
+
+    // Assert
+    assertThat(actual.getMessage()).startsWith("Job");
```
calypsomatic, line 62: What is this message supposed to be and is this unit test assert going to remain valid if we reword it?

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
davidangb: hah no - this is named "instance" to match the config property hierarchy `twds.instance`, which holds `twds.instance.workspace-id`. It's yet another overload of the word "instance". So we could certainly rename it but it wouldn't be renamed to "collection"

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceControlPlaneTest.java
davidangb: It matches the error message defined here: https://github.com/DataBiosphere/terra-workspace-data-service/blob/62f162c49286b9d22a24b8faeeabfbd9702458e9/service/src/main/java/org/databiosphere/workspacedataservice/service/model/exception/MissingObjectException.java#L10 and specifically checks that the thrown exception is `new MissingObjectException("Job")`. I don't feel strongly about keeping this assertion, if you worry it's fragile.

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
calypsomatic: I was actually unsure about that - `twds.instance.workspace-id` and `twds.instance.source-workspace-id` are NOT supposed to be collection?

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceControlPlaneTest.java
calypsomatic: Just making sure it's testing what it's supposed to be testing

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
```diff
@@ -97,24 +97,13 @@ public boolean hasWriteCollectionPermission(String token) {
...
-        collectionId,
-        ACTION_READ);
-    RestCall<Boolean> samFunction =
-        () ->
-            samClientFactory
-                .getResourcesApi(token)
-                .resourcePermissionV2(SamDao.RESOURCE_NAME_WORKSPACE, collectionId, ACTION_READ);
-    return restClientRetry.withRetryAndErrorHandling(
-        samFunction, "Sam.hasReadCollectionPermission");
+  public boolean hasReadWorkspacePermission(String workspaceId, String token) {
```
jladieu, line 105: Why you are to ignore the `workspaceId` now?  Assuming this is intended, can you explain with an inline comment?

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
```diff
@@ -82,13 +82,13 @@ private boolean hasPermission(String action, String loggerHint, String token) {
...
-  public boolean hasWriteCollectionPermission() {
-    return hasWriteCollectionPermission(null);
+  public boolean hasWriteWorkspacePermission() {
+    return hasWriteWorkspacePermission(null);
   }
 
   @Override
-  public boolean hasWriteCollectionPermission(String token) {
-    return hasPermission(ACTION_WRITE, "Sam.hasWriteCollectionPermission", token);
+  public boolean hasWriteWorkspacePermission(String token) {
```
jladieu, line 90: I noticed that this doesn't take a `workspaceId`.  If we had finer grained permissions, could this actually be called `hasCreateWorkspacePermission`?  If so, I wonder if it'd be worth naming it that anyway, even if the actual permission used is `ACTION_WRITE`.

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
jladieu, line 1: I think it'd be fine to mark these methods with `@deprecated` pre-emptively to prevent additional calls that don't go through `CollectionService`, and to provide an easy trailmarker to drive the migration work when we actually pull that ticket.

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
```diff
@@ -139,9 +142,22 @@ public WorkspaceId getWorkspaceId(CollectionId collectionId) {
...
+        throw new CollectionException("Expected a virtual collection");
+      }
     } catch (EmptyResultDataAccessException e) {
-      // swallow not-found errors; it's valid for rows to not exist in the case of
-      // virtual collections
+      // does this deployment allow for virtual collections?
+      // TODO AJ-1630: enable this error for data-plane WDS, which does not allow virtual
+      // collections
+      /*
+      if (twdsProperties.getTenancy() == null
```
ashanhol, line 156: I'm a little confused, why break out this into another ticket if it's already written? What's the extra work thats going into AJ-1630?

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
davidangb: I don't understand the question, I don't think it's ignoring anything? I will give an answer and hope it addresses your concern:

The previous code accepted a `collectionId` and then checked permission on a workspace whose workspaceId was the collectionId. This code didn't perform any lookups to the db to find the workspace associated with this collection; it just assumed workspaceId == collectionId.

The new code accepts a `workspaceId` and checks permissions on a workspace of that id. Elsewhere in `CollectionService`, we have more robust logic to determine the appropriate workspaceId for any given collectionId.

I don't have a code comment inline because I believe this method now works as expected: it's named `hasReadWorkspacePermission`, it accepts a workspaceId, it checks permissions on that workspaceId.

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
davidangb: I am totally fine with renaming these, but outside of this PR? I'm not adding anything here, I'm just using the existing config and method

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
davidangb: Oooh, I agree, this should accept a workspaceId. The `hasCreateCollectionPermission` and `hasDeleteCollectionPermission` methods need that argument too. AJ-1633

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceBaseTest.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.List;
+import java.util.UUID;
+import java.util.stream.IntStream;
+import org.databiosphere.workspacedataservice.generated.GenericJobServerModel;
+import org.databiosphere.workspacedataservice.shared.model.CollectionId;
+
+public abstract class JobServiceBaseTest {
+
+  List<String> allStatuses =
+      Arrays.stream(GenericJobServerModel.StatusEnum.values())
```
jladieu, line 15: nit: import `StatusEnum` (and other various inner classes) directly so the code can refer to them by their simple, non-namespaced name.  I wouldn't go so far as to directly import the Enum values (eg: refer only to `RUNNING` or `DATA_IMPORT`, but losing the `GenericJobServerModel` prefix everywhere will make this easier to read.

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceControlPlaneTest.java
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  /** requested job exists; its collection also exists. In the control plane, this is an error */
+  @Test
+  void collectionExists() {
+    // Arrange
+    UUID jobId = UUID.randomUUID();
+    CollectionId collectionId = CollectionId.of(UUID.randomUUID());
```
jladieu, line 70: Might it be worth providing a shorthand `{Workspace,Collection}Id.random()`?

I'm debating whether adding test-centric code to the prod classes is yucky or if there'd be a sensible reason to provide random generation on the prod codepath too?

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceDataPlaneTest.java
```diff
@@ -0,0 +1,260 @@
...
+    when(samDao.hasReadWorkspacePermission(getEnvWorkspaceId().toString())).thenReturn(true);
+    // return some jobs when listing this collection
+    when(jobDao.getJobsForCollection(eq(collectionId), any()))
+        .thenReturn(makeJobList(collectionId, 2));
+
+    // Act
+    List<GenericJobServerModel> actual = jobService.getJobsForCollection(collectionId, allStatuses);
+
+    // Assert
+    // this is verifying permissions only; only smoke-testing correctness of the result
```
ashanhol, line 202: if you want you can point to JobControllerTest as testing real results

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceBaseTest.java
```diff
@@ -0,0 +1,35 @@
...
+          .map(GenericJobServerModel.StatusEnum::toString)
+          .toList();
+
+  GenericJobServerModel makeJob(UUID jobId, CollectionId collectionId) {
```
jladieu, line 19: Maybe let `makeJob` generate the `jobId` and then have the calling code just get it from the returned model?

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceDataPlaneTest.java
```diff
@@ -0,0 +1,260 @@
...
+  }
+
+  /** Collection exists, associated with a non-default workspace */
+  @Test
+  void listJobsNonDefaultCollection() {
+    // Arrange
+    CollectionId collectionId = CollectionId.of(UUID.randomUUID());
+    WorkspaceId workspaceId = WorkspaceId.of(UUID.randomUUID());
+    // collection exists and is associated with a non-default workspace
+    when(collectionDao.getWorkspaceId(collectionId)).thenReturn(workspaceId);
```
ashanhol, line 236: Trying to understand this test. This is testing the scenario where we have a collection's workspaceId be different than `twds.instance.workspace-id`, right? How would we get into this situation?

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceControlPlaneTest.java
```diff
@@ -0,0 +1,202 @@
...
+    when(collectionDao.getWorkspaceId(collectionId))
+        .thenThrow(new EmptyResultDataAccessException("unit test intentional error", 1));
+    // user has permission to the workspace with the same id as the collection
+    when(samDao.hasReadWorkspacePermission(collectionId.toString())).thenReturn(false);
+
+    // Act / assert
+    Exception actual = assertThrows(AuthorizationException.class, () -> jobService.getJob(jobId));
+
+    // Assert
+    assertThat(actual.getMessage()).endsWith("this job.\"");
```
jladieu, line 128: nit: could you rephrase as a `contains(<substring>)`?

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceDataPlaneTest.java
```diff
@@ -0,0 +1,260 @@
...
+    UUID jobId = UUID.randomUUID();
+    // job not found
+    when(jobDao.getJob(jobId))
+        .thenThrow(new EmptyResultDataAccessException("unit test intentional error", 1));
+
+    // Act / assert
+    Exception actual = assertThrows(MissingObjectException.class, () -> jobService.getJob(jobId));
+
+    // Assert
+    assertThat(actual.getMessage()).startsWith("Job");
```
jladieu, line 66: nit: could you rephrase as a `contains(<substring>)`?

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceDataPlaneTest.java
```diff
@@ -0,0 +1,260 @@
...
+    // $WORKSPACE_ID workspace
+    when(collectionDao.getWorkspaceId(collectionId)).thenReturn(workspaceId);
+    // user has permission to that workspace
+    when(samDao.hasReadWorkspacePermission(workspaceId.toString())).thenReturn(true);
+
+    // Act / assert
+    Exception actual = assertThrows(CollectionException.class, () -> jobService.getJob(jobId));
+
+    // Assert
+    assertThat(actual.getMessage()).startsWith("Found unexpected workspaceId for collection");
```
jladieu, line 131: nit: could you rephrase as a `contains(<substring>)`?

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceDataPlaneTest.java
```diff
@@ -0,0 +1,260 @@
...
+    when(collectionDao.getWorkspaceId(collectionId))
+        .thenThrow(new EmptyResultDataAccessException("unit test intentional error", 1));
+    // user has permission to that workspace
+    when(samDao.hasReadWorkspacePermission(workspaceId.toString())).thenReturn(true);
+
+    // Act / assert
+    Exception actual = assertThrows(MissingObjectException.class, () -> jobService.getJob(jobId));
+
+    // Assert
+    assertThat(actual.getMessage()).startsWith("Collection");
```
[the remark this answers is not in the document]
jladieu: sorry...just dropped this comment in a bunch of places before I realized it was going to be spammy; if you aren't opposed to this treatment I'd suggest the same for all the `startsWith`/`endsWith`

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceBaseTest.java
```diff
@@ -0,0 +1,35 @@
+package org.databiosphere.workspacedataservice.service;
```
ashanhol, line 1: I wonder if it might be worth making this more accessible to other tests. This is super useful functionality for any time we just need to make some jobs. Just a thought

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceControlPlaneTest.java
davidangb: I had the same thought when writing the tests, but landed on it being a confusing method to have in runtime code, so I skipped it

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceDataPlaneTest.java
```diff
@@ -0,0 +1,260 @@
...
+    // Assert
+    assertThat(actual.getMessage()).startsWith("Job");
+  }
+
+  /** requested job exists; its collection uses the default workspace and user has permission */
+  @Test
+  void defaultCollectionWithPermission() {
+    // Arrange
+    UUID jobId = UUID.randomUUID();
+    CollectionId collectionId = CollectionId.of(UUID.randomUUID());
```
jladieu, line 74: I realize we deliberately added `CollectionId` to improve our typing posture, but whenever you've got a variable whose name is also its type, I think it's a reasonable candidate to declare as a `var collectionId`, especially in unit tests where terseness can aid scannability.

Not completely convinced this is the right tradeoff, BTW, but I've been dabbling with it and wonder what your thoughts are.

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceDataPlaneTest.java
```diff
@@ -0,0 +1,260 @@
...
+    assertThat(actual.getMessage()).startsWith("Collection");
+  }
+
+  // ==================================================
+  // ========== tests for getJobsForCollection ========
+  // ==================================================
+
+  /** Collection does not exist */
+  @Test
+  @Disabled("AJ-1630") // TODO AJ-1630
```
jladieu, line 167: neat :)

jladieu (approved):
Awesome test coverage!

I had all optional suggestions/questions, pre-emptive LGTM.

I agree with your followup tickets.

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceDataPlaneTest.java
ashanhol: I guess this tests behavior in CollectionService.getWorkspaceId specifically ("if the WORKSPACE_ID env var is set and a row exists in sys_wds.instance, confirm the row's workspace_id column matches the env var. If it does, return its value. If it doesn't match, throw an error.")

davidangb: We shouldn't have this situation under normal conditions. This is testing an edge case, but I figured it's worth the coverage.

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceDataPlaneTest.java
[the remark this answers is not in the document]
davidangb: I prefer `startsWith`/`endsWith` since they are stricter than just a `contains` … what's the reasoning for preferring `contains`?

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceDataPlaneTest.java
davidangb: My very personal preference is to have the type explicitly listed … but if folks prefer the terseness of `var`, I am game for starting to use it!

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceControlPlaneTest.java
```diff
@@ -0,0 +1,202 @@
+package org.databiosphere.workspacedataservice.service;
```
ashanhol, line 1: I don't see a way we're discriminating running this class only for cWDS. Is that ok because the "virtual collection" behavior is completely mocked?

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
davidangb: my hesitation here is that these methods should be considered deprecated … unless we're calling them from CollectionService. So I guess I'd prefer to leave this work for outside of this PR when we can really figure out what's best to do

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceBaseTest.java
davidangb: Sure, if other tests want to use these methods, we could just rename this to something like `JobTestSupport` and any test that wants it can extend it

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceBaseTest.java
```diff
@@ -7,21 +7,19 @@
...
+import org.databiosphere.workspacedataservice.generated.GenericJobServerModel.StatusEnum;
 import org.databiosphere.workspacedataservice.shared.model.CollectionId;
 
 public abstract class JobServiceBaseTest {
 
-  List<String> allStatuses =
-      Arrays.stream(GenericJobServerModel.StatusEnum.values())
-          .map(GenericJobServerModel.StatusEnum::toString)
-          .toList();
+  List<String> allStatuses = Arrays.stream(StatusEnum.values()).map(StatusEnum::toString).toList();
```
ashanhol, line 15: [just a note] it seems really easy to mix up GenericJobServerModel.StatusEnum and StatusEnum, I managed to find and correct it myself in my own PR.

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceControlPlaneTest.java
davidangb: It's the `@ActiveProfiles(profiles = {"mock-sam", "control-plane"})` annotation at the top of the test

ashanhol: ah I missed the "control-plane" one thank you

ashanhol (approved):
Thank you for test robustness! I feel like what this PR pointed out most to me is how tricky it is to keep track of all the different behavior nuances between WDS and cWDS

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
```diff
@@ -52,7 +52,8 @@ public GenericJobServerModel createImport(
     collectionService.validateCollection(instanceUuid);
 
     // validate write permission
-    boolean hasWriteInstancePermission = samDao.hasWriteCollectionPermission();
+    // TODO AJ-1631: this must use collectionService.canWriteCollection() instead
+    boolean hasWriteInstancePermission = samDao.hasWriteWorkspacePermission();
```
yuliadub, line 56: the reason for this is because a user may only have access to a single collection in a workspace that may have more then one collection, right?

davidangb: In the data plane, it's not an issue until/unless we support multi-tenancy. In cWDS, we need this to run through `collectionService.canWriteCollection()` or else it will always fail, since `$WORKSPACE_ID` won't be set.

Additionally, I'm trying to get all code running through a single forward-looking path (`CollectionService`) instead of calling SamDao directly - but that's just code organization.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -99,7 +99,7 @@ public void validateAndPermissions(UUID collectionId, String version) {
     validateVersion(version);
     collectionService.validateCollection(collectionId);
 
-    boolean hasWriteCollectionPermission = samDao.hasWriteCollectionPermission();
+    boolean hasWriteCollectionPermission = samDao.hasWriteWorkspacePermission();
```
yuliadub, line 102: I am bit confused by this pattern where the boolean is called "hasWriteCollectionPermission" but the check is "hasWriteWorkspacePermission". Do we then assume that collection permission=workspace permission?

yuliadub:
I am a bit thrown by concept of "virtual" collections - are we saying a virtual collection (today) wont actually exists since workspaceId is null in cWDS? but in the future we will have concepts of collections in cWDS as parts of rawls get migrated over. Or are virtual collections actually used for something in cWDS as of now?

davidangb:
> I am a bit thrown by concept of a "virtual" collections (or maybe its just the name and my brain 😄 ) - are we saying a virtual collection (today) wont actually exists since workspaceId is null in cWDS? but in the future we will have concepts of collections in cWDS as parts of rawls get migrated over. Or are virtual collections actually used for something in cWDS as of now?

@yuliadub in cWDS, collections will be "virtual" because they will have no corresponding row in the ~`sys_wds.instance`~ `sys_wds.collection` table. Since cWDS will only be used for imports, not to actually persist data tables, the "collection" concept there is tricky. A collection is a group of data tables - but in GCP, Rawls doesn't allow this. So, in GCP we really do assume 1 collection per workspace, where the collection id = workspace id, and therefore we don't need to store the collection row in `sys_wds.collection`.

My TODO is to document this in an understandable way!

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
jladieu: The source of my confusion is that I just don't see anywhere in this method where we consume the provided `workspaceId`.

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceDataPlaneTest.java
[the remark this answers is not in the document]
jladieu: Sorry for the late explanation here...no worries that you've shipped already.

I start from the assumption that error messages are kinda like UI code.  More subject to change on a whim than regular code.  Consequently, I try to craft error assertions that assert that the message is describing what it should _correctly_ but _as loosely as possible_ so that the assertion doesn't become a [change detector test](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html).  Using contains, regex, and/or case insensitivity are all ways to relax the assertion so that it doesn't fail on minor text changes, but if used judiciously will still allow you to capture and verify the intent.

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
jladieu: For discoverability & posterity, this is gets fixed in https://github.com/DataBiosphere/terra-workspace-data-service/pull/594/files#r1507670234

PR #520, opened 2024-02-14
AJ-1592 rename instance to collection in DB

calypsomatic:

Migrates the `sys_wds.instance` table to be `sys_wds.collection` and renames the `instance_id` column in the `sys_wds.job` table to be `collection_id`.  Also renames a few comments and variables that were missed in previous PR.

I tried out this version of WDS in a BEE and it deployed and worked with no problems.  Then I attempted to patch a running WDS to this version while importing (since importing uses the job table which was renamed, although assorted other actions also use the instance/collection table).  I did see some errors where sql could not find the referenced table/column WHILE the replicas were switching over, but everything worked as expected within ~1 minute once WDS was fully upgraded.  So I believe that it should work fine but whenever we do an upgrade that includes this or any later version, we can expect a minute or few of downtime/unreliable behavior for users.

Reminder:


jladieu (approved):
Well done!

davidangb (approved):
woohoo!

calypsomatic:
Adding a note to the [upgrade playbook](https://docs.google.com/document/d/1GB4A9hq-_H4nf94K8fZ9Ke4EnH5J3n2tMADDzirAcNk/edit) so we remember that upgrading will not be 100% seamless

PR #517, opened 2024-02-14
[AJ-1618] Split terra-helmfile PR out of publish-docker workflow

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1618

A bit of "prefactoring" for AJ-1618... This splits the creation of a terra-helmfile PR out of the publish-docker workflow into a separate publish-version workflow (suggestions for a better name welcome). This refines publish-docker to its stated purpose ("Publish Docker Images to GCR and ACR") and limits the BroadBot token to a smaller workflow.

The intent is to then add a second job to the publish-version workflow to notify Sherlock of a new cWDS version.

PR #521, opened 2024-02-14
[AJ-1420] Fail terra-helmfile PR job if no JIRA ticket provided

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1420

Running the `tag` workflow asks for the ID of a ticket. It uses that JIRA ticket ID to create a PR to terra-helmfile.

Currently, if no valid ticket ID is found, the `find-jira-id` step outputs "missing". If the terra-helmfile PR step sees that the JIRA ID is "missing", it prints a message and exits cleanly, which results in the job showing as successful.

This changes it so that if no ticket ID is found, the `find-jira-id` step fails, which would make the terra-helmfile PR job show as failed.

It also changes the default placeholder for the JIRA ID input field to "AJ-###".

### DIFF
--- a/.github/workflows/publish-app-version.yml
+++ b/.github/workflows/publish-app-version.yml
@@ -16,7 +16,6 @@ jobs:
   create-terra-helmfile-pr:
     name: Create terra-helmfile PR for latest WDS version
     runs-on: ubuntu-latest
-    needs: publish-docker-job
     steps:
       - name: Look for AJ (Analysis Journeys) Jira ID in the manual workflow trigger message
         id: find-jira-id
--- a/.github/workflows/publish-app-version.yml
+++ b/.github/workflows/publish-app-version.yml
@@ -24,7 +24,8 @@ jobs:
           set +e
           JIRA_ID=$(echo "${{ inputs.jiraId }}" | grep -iEo -m 1 'AJ-[0-9]+' | head -1 || '')
           if [[ -z "$JIRA_ID" ]]; then
-            echo "JIRA_ID=missing" >> $GITHUB_OUTPUT
+            echo "JIRA_ID missing, PR to terra-helmfile will not be created" 1>&2
+            exit 1;
           else
             echo "JIRA_ID=${JIRA_ID}" >> $GITHUB_OUTPUT
           fi
@@ -45,10 +46,6 @@ jobs:
           HELM_NEW_TAG: ${{ inputs.new-tag }}
         run: |
           set -e
-          if [[ $JIRA_ID == "missing" ]]; then
-            echo "JIRA_ID missing, PR to terra-helmfile will not be created"
-            exit 0;
-          fi
           cd terra-helmfile
           HELM_CUR_TAG=$(grep "/terra-workspace-data-service:" charts/wds/values.yaml | sed "s,.*/terra-workspace-data-service:,,")
           git checkout -b ${JIRA_ID}-auto-update-${HELM_NEW_TAG}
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-app-version.yml
```diff
@@ -16,15 +16,15 @@ jobs:
   create-terra-helmfile-pr:
     name: Create terra-helmfile PR for latest WDS version
     runs-on: ubuntu-latest
-    needs: publish-docker-job
```
nawatts, line 19: Unrelated. This was missed in #517 when moving this job from a different workflow file.

.github/workflows/tag.yml
```diff
@@ -7,7 +7,7 @@ on:
         # Friendly description to be shown in the UI instead of 'name'
         description: 'Jira ID'
         # Default value if no value is explicitly provided
-        default: 'Required: Include latest Jira ID'
+        default: 'AJ-###'
```
yuliadub, line 10: should we keep the "Required:' part?

yuliadub (approved):
thanks for doing this!

.github/workflows/publish-app-version.yml
```diff
@@ -16,15 +16,15 @@ jobs:
...
     steps:
       - name: Look for AJ (Analysis Journeys) Jira ID in the manual workflow trigger message
```
yuliadub, line 20: dont have to take place in this PR: but as I was looking at this, part of me started to wonder if we should do this check sooner (like right in tag job, as a first step before doing anything) and failing right away vs getting all the way here to find that Jira-ID is missing.... (we can then pass the JIRA Id as a param to other jobs that need it if we did get it) that way we wont make any unnecessary changes like bumping the version when we wont actually release anything.

.github/workflows/publish-app-version.yml
yuliadub: good catch!

.github/workflows/tag.yml
yuliadub: or add "required" to the description or something like that

nawatts: For what it's worth, this field is marked as required and so it gets the little red asterisk in the input popup.

Screenshot 2024-02-14 at 6 47 05 PM

.github/workflows/publish-app-version.yml
nawatts: Yeah, I was also wondering about that... I read the ticket as "fix the status of the terra-helmfile PR step". But in that case, is there a way to rerun the terra-helmfile job with an updated JIRA ID?

Or... maybe we change the JIRA ID input to a number and append that to "AJ-" in the terra-helmfile step. Assuming GitHub requires an input of the correct type, that would move the validation before the workflow submission.
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_callinputsinput_idtype

.github/workflows/tag.yml
yuliadub: ah okay, missed that somehow when I looked at it - that should be enough!

.github/workflows/publish-app-version.yml
nawatts: So it looks like GitHub will let you enter any text in a "number" field, so that won't work.

However, I think I've convinced myself to keep this the way it is. The JIRA ticket ID is only required for the terra-helmfile PR, and it's easy enough to create that PR manually if it's not automatically created because of a missing ticket ID.

.github/workflows/tag.yml
nawatts: Also, I was going by the acceptance criteria that you wrote! 😆 

> Set the placeholder ID text in the input field to be AJ-####"

calypsomatic (approved):

PR #522, opened 2024-02-15
[AJ-1618] Report new cWDS version to Sherlock and deploy to dev on tag

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1618

Add jobs to the `publish-app-version` workflow (run as part of the `tag` workflow) to report a new cWDS version to Sherlock and deploy it to dev.

Followed instructions in Google Doc linked from ticket.

### DIFF
--- a/.github/workflows/publish-app-version.yml
+++ b/.github/workflows/publish-app-version.yml
@@ -55,3 +55,24 @@ jobs:
           git commit -am "${JIRA_ID}: Auto update WDS version to $HELM_NEW_TAG"
           git push -u origin ${JIRA_ID}-auto-update-${HELM_NEW_TAG}
           gh pr create --title "${JIRA_ID}: auto update WDS version to $HELM_NEW_TAG" --body "${JIRA_ID} helm chart update" --label "automerge"
+
+  cwds-report-to-sherlock:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-app-version.yml
```diff
@@ -5,7 +5,7 @@ on:
       new-tag:
         required: true
         type: string
-      jiraId:
+      jira-id:
```
nawatts, line 8: Unrelated. Just making this consistent with other inputs, which use kebab instead of camel case.

yuliadub:
how are you thinking we test this? just run and see if it works I guess 😄 

also, should we consider being able to push an update to cWDS but not WDS?

nawatts:
> how are you thinking we test this? just run and see if it works I guess

Yeah, I don't know of another option for GitHub Actions.

> should we consider being able to push an update to cWDS but not WDS?

Without thinking much about it, I think I'd vote no. One of the motivations for cWDS was having consistent behavior / the same code in Google and Azure.

calypsomatic (approved):

ashanhol (approved):

yuliadub (approved):

PR #523, opened 2024-02-15
[AJ-1635] Configure Dependabot updates

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1635

Seeing that we've accumulated several Dependabot alerts, this sets up automated Dependabot updates using the same configuration that we use for [DataBiosphere/java-pfb](https://github.com/DataBiosphere/java-pfb/blob/main/.github/dependabot.yml).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
+    labels:
+      - "dependencies"
+      - "gradle"
+    open-pull-requests-limit: 10
+    reviewers:
+      - "@DataBiosphere/analysisjourneys"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
+version: 2
+updates:
```
calypsomatic, line 2: Does this get called by another GHA?  I don't really understand how it does anything

nawatts: This is a configuration file for Dependabot, not an Actions workflow.
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

calypsomatic: Gotcha

calypsomatic (approved):

ashanhol (approved):

PR #524, opened 2024-02-15
AJ-1632: not found exceptions

davidangb:
**_Reviewer: see comments inline for an important question about compatibility!_**

https://broadworkbench.atlassian.net/browse/AJ-1632 - the exceptions thrown by get-job and list-jobs APIs should not leak information about existence/non-existence of jobs and collections. For more context on why, see [here](https://apihandyman.io/hands-off-that-resource-http-status-code-401-vs-403-vs-404/).

I tested the following use cases manually, in addition to the unit tests in code:

## get job

| | Before this PR | After this PR |
|--------|--------|--------|
| job exists, I have permission to it | 200 with job response | 200 with job response |
| job exists, I do not have permission to it | 403 forbidden | 404 does not exist or you do not have permission to see it |
| job does not exist | 404 does not exist | 404 does not exist or you do not have permission to see it |
| job exists, job's collection has invalid workspace* | 500 unexpected workspace for collection | 500 unexpected workspace for collection |

* this corner case only happens if the db or the WDS deployment is misconfigured; we do not expect to see this in practice

## list jobs

| | Before this PR | After this PR |
|--------|--------|--------|
| collection does not exist | 200 with empty list | 404 does not exist or you do not have permission to see it |
| collection exists, I do not have permission to it | 403 forbidden | 404 does not exist or you do not have permission to see it |
| collection exists, I have permission to it | 200 with results | 200 with results |


service/src/main/resources/static/swagger/apis-v1.yaml
```diff
@@ -100,6 +102,8 @@ paths:
     get:
       summary: Get all jobs with a certain status under a particular instance.
       operationId: jobsInInstanceV1
+      security:
+        - bearerAuth: [ ]
```
davidangb, line 106: with the permission-checking added in #519, these APIs now need a token. The change here affects both the behavior of swagger UI _AND_ the behavior of our generated clients. 

So, without realizing it, I merged a technically-breaking change in #519: if a caller uses an older client to hit one of these APIs, the call will fail (assuming WDS is running code later than #519). Is it worth rolling back #519 and creating brand new get/list job APIs? That would also require updating Terra UI. Note that since Terra UI does NOT use a client, it won't see any problems.

service/src/test/java/org/databiosphere/workspacedataservice/controller/JobControllerMockMvcTest.java
```diff
@@ -65,6 +67,8 @@ void smokeTestJobsInCollectionV1() throws Exception {
     // set created and updated to now, but in UTC because that's how Postgres stores it
     OffsetDateTime time = OffsetDateTime.now(ZoneId.of("Z"));
 
+    when(collectionDao.collectionSchemaExists(collectionId.id())).thenReturn(true);
```
davidangb, line 70: now that list-jobs validates collection existence, this unit test needs this mock

service/src/main/resources/static/swagger/apis-v1.yaml
calypsomatic: I was expecting to update a lot of APIs and/or deprecate and replace with new APIs as we rename instance -> collection.  It's possible we could combine that breakingness into one fell swoop

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
```diff
@@ -25,9 +24,9 @@ public GenericJobServerModel getJob(UUID jobId) {
     try {
       GenericJobServerModel result = jobDao.getJob(jobId);
       if (!collectionService.canReadCollection(CollectionId.of(result.getInstanceId()))) {
-        // TODO AJ-1632: this should return a 404 a la "job does not exist or you do not have
-        //     permission to see it"
-        throw new AuthorizationException("Caller does not have permission to view this job.");
+        // if caller lacks permission, throw MissingObjectException, so we don't leak information
+        // about this job's existence or non-existence
+        throw new MissingObjectException("Job");
```
jladieu, line 29: I agree with obscuring the reason for the API, but I wonder if we've pushed this concern too far down.  Should the `Service` be doing the obfuscation?  Feels more like a concern the controller should handle, perhaps by capturing and rewriting `AuthorizationException` into `NotFoundExceptions`.

I guess my hunch here is that the Service layer's job should be to report exceptions accurately, without obscurity, under the assumption that _our_ code is going to react to it and do the right thing, and that then we should bake the policy of how to obscure errors in at the layer closest to the UI.

service/src/main/resources/static/swagger/apis-v1.yaml
jladieu: Oops...well, I kinda defer to your hunch on this.  Do you have any guess of how many folks use these APIs and could be impacted?

I guess I am assuming this is low traffic, and I think given we're changing the security posture for the better, it's worth causing breakage.  Do you really want clients to be able to hit this endpoint without auth after this goes live?

ashanhol:
Hold on, what is your reasoning for an empty collection list being a 404 over 200 -> empty list? I wrote the API under the impression we could eventually have a piece of UI that'd list jobs of certain status (or active/inactive if we want to group) and there could definitely be a valid case where there's no jobs running a certain way and we should display that. I feel like throwing a 404 would give the UI more work to do than just displaying an empty list (ie, no jobs).

service/src/main/resources/static/swagger/apis-v1.yaml
yuliadub: the chance will affect the python client- and today the client is used in broad tests at least, and other internal processes, not quite sure of the exact customer impact but we do have a notebook that shows how to use the client as well that would be impacted.

davidangb:
> Hold on, what is your reasoning for an empty collection list being a 404 over 200 -> empty list?

@ashanhol We might be talking about two different things here.

_If the collection exists_ but there are no jobs in that collection, this still returns `[]` just like it did before.

The change I made is that _if the specified collection doesn't exist_ we will now return a 404, where previously we returned `[]`.

This feels more accurate - it indicates to the user that the collection id they specified is invalid.

service/src/main/resources/static/swagger/apis-v1.yaml
davidangb: I think there's two different concerns being discussed here:
1. is it okay to sneak auth into these two APIs?
2. how do we go about renaming "instance" to "collection" in these APIs?

I am not at all trying to tackle the second one - though I agree that if we were to switch to new versions of these APIs _quickly_, we could just add auth to them and not try to sneak it into the v1 APIs.

davidangb: > Do you have any guess of how many folks use these APIs and could be impacted?

I suspect zero. As discussed when we were working on #498, import features are still feature-flagged and limited to Broad-internal users. The only notable consumer of these APIs is Terra UI, which doesn't use a generated client and therefore isn't vulnerable to this issue.

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
davidangb: Fair point - my thinking was to keep as much logic out of the controllers as possible, so they stay pretty thin. I'll take a shot at implementing something here.

ashanhol: For my learning- is there a reason to keep controllers thin? I know they're purpose is to route to the thing that's actually doing the work (ie the service) but is there really no code that should go there ever?

service/src/main/resources/static/swagger/apis-v1.yaml
ashanhol: I'm personally in favor of breaking things behind flags, that's kind of the point and disclaimer of a flag.

ashanhol (approved):

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
jladieu: I'm not sure there are hard and fast rules here but some projects adopt an architecture which results in controllers only having literally one line of code, a passthrough to a very beefy service layer.

I personally don't actively strive to make controller code slim, but I see it happening organically as a consequence of  pushing logic down (into a well-wrought business logic layer), or out (into thoughtful generalizations that take shape as aspects or project-wide concerns).

I see the controller layer as needing to manage the following responsibilities:
* authentication/authorization, if it's not managed already; usually in the form of cross-cutting request filters
* request routing
* acting as an adapter between the request protocol (eg: HTTP/JSON/protobuf) and an underlying business domain
* acting as a presenter of results _from_ the underlying business domain to the output protocol
* sometimes orchestrating across various services, doing parallelization and coordination if needed

And often times these responsibilites aren't done _by_ the controller, they are just composed and coreographed by the controller, and this can end up being a very lightweight handful of lines of code.

nawatts (approved):

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
davidangb: ok, I've made significant changes on this front. In the latest commits:
* added a new `AuthenticationMaskableException` (I thought about naming this `MaskableAuthenticationException` but I really wanted it to be alphabetically right next to `AuthenticationException`)
* `JobService` now throws `AuthenticationMaskableException` instead of `AuthenticationException`
* the `GlobalExceptionHandler` now catches all `AuthenticationMaskableException` that would be displayed to the user and rewrites it to look like a `MissingObjectException`.

So: the service layer can still throw an exception that indicates an auth problem; no need to add code to any of our controllers; centralized masking of these auth exceptions. We can adopt `MaskableAuthenticationException` elsewhere in the codebase if it makes sense to throw it.

service/src/main/resources/static/swagger/apis-v1.yaml
davidangb: ok, I am feeling good about making this change, thank you for all the feedback

service/src/main/java/org/databiosphere/workspacedataservice/controller/GlobalExceptionHandler.java
```diff
@@ -79,24 +81,51 @@ protected ResponseEntity<Object> createResponseEntity(
...
     if (!errorMessages.isEmpty()) {
-      errorBody.put(MESSAGE, errorMessages.get(errorMessages.size() - 1));
-    } else {
-      errorBody.put(MESSAGE, "Unexpected error: " + ex.getClass().getName());
+      errorMessage = errorMessages.get(errorMessages.size() - 1);
     }
+    return generateResponse(HttpStatus.BAD_REQUEST, errorMessage, servletRequest);
+  }
+
+  /**
```
jladieu, line 97: Slick!

jladieu (approved):

PR #536, opened 2024-02-15
AJ-1592 rename instance to collection in APIs

calypsomatic:

Renaming instances at the API level is a breaking change, since it will require new parameters and models.  This PR is marking all affected APIs as deprecated and replacing them with new APIs with the proper names, incrementing the versions.  APIs that are currently v0.2 are now v1; APIs that are currently v1 are now v2. 

Still to do:
- Finish updating APIs
- Make sure any generated models are updated to reflect the change
- Make sure unit tests are up-to-date

Questions:
- How to deal with [validateVersion](https://github.com/DataBiosphere/terra-workspace-data-service/blob/5eb3d314318f7bed97176c9aeadd6e819703b468/service/src/main/java/org/databiosphere/workspacedataservice/service/RecordUtils.java#L9)?  Should it just check that the version is any of our available versions?  Should it be specific to the controller/API it's coming from?  Do we need to validate version at all?
- Do we need a new `apis-v2.yaml` or should we rename `apis-v1.yaml` to be more inclusive?
- What about unit tests?  Do we leave all the `mockMvc` and related unit tests as-is and simply copy them for the new controllers, then delete the unit tests along with the deprecated controllers?
- New Controllers?  David suggested, instead of adding, say, `importV2` to `ImportController`, creating `ImportControllerV2` with the new import method, so that we can simply delete the entire old controller files in the future.  However, Spring doesn't like two controller beans with two methods that have the same signature; it gives the error:
> org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'requestMappingHandlerMapping' defined in class path resource [org/springframework/boot/autoconfigure/web/servlet/WebMvcAutoConfiguration$EnableWebMvcConfiguration.class]: Ambiguous mapping. Cannot map 'importControllerV2' method 
> org.databiosphere.workspacedataservice.controller.ImportControllerV2#importV2(UUID, ImportRequestServerModel)
> to {POST [/v2/{collectionUuid}/import], consumes [application/json], produces [application/json]}: There is already 'importController' bean method
> org.databiosphere.workspacedataservice.controller.ImportController#importV2(UUID, ImportRequestServerModel) mapped.

Did I do it wrong or is it just not an option to have two controllers with these two matching methods?

Follow-on work:
- Update terra-ui, example notebooks, documents, and other uses of clients to use the new APIs
- Communicate changes to users - notably CBAS, but any other users of notebooks or the clients, that they should start using the new APIs
- Set a date to remove deprecated APIs; after that date, they can be deleted

Reminder:


service/src/main/java/org/databiosphere/workspacedataservice/controller/CollectionController.java
```diff
@@ -0,0 +1,44 @@
...
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.DeleteMapping;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class CollectionController {
```
davidangb, line 16: since you defined the APIs in apis-v1.yaml, any controller we write should extend the generated controller stub. So here, this should extend `org.databiosphere.workspacedataservice.generated.CollectionsApi`. Once you've done that, just use IntelliJ to override the methods in the base class - no hand-writing of the routes!

PR #537, opened 2024-02-15
AJ-1563 Provide `@RawlsJsonConsumer` to collect emitted JSON.

jladieu:
This is refactoring work following [AJ-1563](https://broadworkbench.atlassian.net/browse/AJ-1563) and (hopefully) reducing friction for [AJ-1585](https://broadworkbench.atlassian.net/browse/AJ-1585) by providing a clearer offramp for the JSON emitted by `RawlsRecordSink`.

The commits are divided into prefactoring to mechanically reorganize related code with the final commit actually restructuring things to provide a `@RawlsJsonConsumer Consumer<String>` bean hook that can be implemented to capture the JSON.

[AJ-1563]: https://broadworkbench.atlassian.net/browse/AJ-1563?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AJ-1585]: https://broadworkbench.atlassian.net/browse/AJ-1585?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service/src/main/java/org/databiosphere/workspacedataservice/shared/model/CollectionId.java
```diff
@@ -3,8 +3,9 @@
 import java.util.UUID;
 
 /**
- * Model to represent a collection id, which currently is a UUID. Since we use UUIDs throughout our
- * code for multiple use cases, this wrapper class exists to help disambiguate between those UUIDs.
+ * RawlsModel to represent a collection id, which currently is a UUID. Since we use UUIDs throughout
```
nawatts, line 6: Were the comment changes in this directory intentional?

nawatts (approved):
Admittedly, didn't review super thoroughly. But the idea sounds good to me.

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/CollectionId.java
jladieu: Eww not at all.  Thanks for seeing this.  I'll revert them.

jladieu: Fixed in an amendment.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -1,49 +1,56 @@
...
+  @Target({ElementType.PARAMETER, ElementType.METHOD})
+  @Retention(RetentionPolicy.RUNTIME)
+  public @interface RawlsJsonConsumer {}
 
-  public static RawlsRecordSink withPrefix(String prefix) {
-    return new RawlsRecordSink(prefix);
+  RawlsRecordSink(
+      String attributePrefix,
+      ObjectMapper mapper,
+      @RawlsJsonConsumer Consumer<String> jsonConsumer) {
```
davidangb, line 50: I'm questioning if this should be a `Consumer<String>` or a `Consumer<Entity>`. If this remained typed as `Entity` then any consumers we plug in for testing or debugging could still work with `Entity` objects. The production-use consumer would encapsulate object-to-json serialization.

I could also see an argument for keeping the serialization here in RawlsRecordSink. So, this is thinking out loud and I wouldn't hold the PR for this, but do wanna raise the question.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RecordSinkFactory.java
```diff
@@ -1,31 +1,41 @@
...
 
-  @Bean
-  RecordSink recordSink() {
+  @Autowired(required = false) // jsonConsumer only required for RecordSinkMode.RAWLS
+  public void setJsonConsumer(@RawlsJsonConsumer Consumer<String> jsonConsumer) {
+    this.jsonConsumer = jsonConsumer;
+  }
+
+  // TODO(AJ-1589): make prefix assignment dynamic
+  public RecordSink buildRecordSink(String prefix) {
```
davidangb, line 38: again, thinking out loud, this comment does not block the PR: I notice this factory is a spring bean (via `@Component`), and we inject this factory where it is needed, then let the factory produce Sinks on demand. Spring offers this functionality in [prototype-scoped beans](https://docs.spring.io/spring-framework/reference/core/beans/factory-scopes.html#beans-factory-scopes-prototype) - if our Sinks were prototype-scoped beans, they'd gain all the Spring goodness like dependency injection, lifecycle hooks and events, etc.

Spring can also complicate things and managing our own factories is simple and understandable, so I see the benefit.

davidangb (approved):
two non-blocking design-level comments, syntax looks good!

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RecordSinkFactory.java
jladieu: I think you've mentioned Prototype scoped beans before when I did other factory stuff.

The key thing that keeps driving me toward using the Factory is that I couldn't readily find the idiomatic SpringBoot way of getting a partially injected object.  The key thing here is that the callsite knows the prefix, but SpringBoot does not, so I would love if I say to SpringBoot "Here's the value to use for prefix, can you inject everything else?"

I recall in my days of Guice, there was a way you could mark an attribute as `@Provided`, and that would automatically 🤯 generate a Factory / Factory method for you to use, but I haven't found the equivalent yet.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
jladieu: I could totally see this turning into a `Consumer<Entity>` or the like.  I think this was driven more by the ticket granularity (produce JSON) and the corresponding testing paradigm I took (copy the JSON-based tests over from import-service).  Going to leave as is but totally cool with us jiggling this around once the downstream stuff starts getting built out.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RecordSinkFactory.java
jladieu: FYI, I finally had a chance to come back around and look at this in more detail and I think I found what I was looking for...  Request-Scoped beans!

https://docs.spring.io/spring-framework/reference/core/beans/factory-scopes.html#beans-factory-scopes-request

I was experimenting with implementing this for BearerToken and put some scratch up in this [PR](https://github.com/DataBiosphere/terra-workspace-data-service/pull/595) for speculative consideration.

PR #540, opened 2024-02-16
AJ-1556 update java-pfb

calypsomatic:
I noticed we hadn't updated java-pfb in a while, it's been getting a number of security updates that we probably want to include.

Should we have a better process for keeping java-pfb up-to-date?  Can/Should we tie it to a major version instead of something specific?

Reminder:

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -83,7 +83,7 @@ dependencies {
     implementation 'com.squareup.okhttp3:okhttp' // required by Sam client
     implementation "bio.terra:datarepo-jakarta-client:1.570.0-SNAPSHOT"
     implementation "bio.terra:workspace-manager-client:0.254.983-SNAPSHOT"
-    implementation "bio.terra:java-pfb-library:0.15.0"
+    implementation "bio.terra:java-pfb-library:0.29.0"
     implementation project(path: ':client')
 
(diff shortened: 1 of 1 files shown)


jladieu (approved):

jladieu:
I'm not sure I understand your question about tying it to a major version instead of something specific?  Are you suggesting we import it with a more relaxed dependency wildcard so that it will pick up minor version bumps automatically somehow?  I'd be fine with that :shrug:

As for improving our process...I wonder if we could get Slack / Github plugin to yell at us whenever there's a new version of the java-pfb released (as seen on Github)?

calypsomatic:
> I'm not sure I understand your question about tying it to a major version instead of something specific? Are you suggesting we import it with a more relaxed dependency wildcard so that it will pick up minor version bumps automatically somehow? I'd be fine with that 🤷
> 
@jladieu  Right, some kind of wildcard or range, I've definitely seen that done.

nawatts:
Alternatively, we could do something like we do for terra-helmfile on WDS releases: when java-pfb is released, have the release automation open a PR against WDS.

davidangb (approved):

davidangb:
:+1: for relaxed version specifier !

PR #541, opened 2024-02-16
AJ-1563 Eliminate redundant `ImportMode` arg throughout.

jladieu:
`ImportMode` is already passed as part of creating the `RecordSource`, and so doesn't need to be explicitly passed down through the callstack.


jladieu:
Closed in favor of https://github.com/DataBiosphere/terra-workspace-data-service/pull/545 which more holistically reduces the arg list in a series of commits.

PR #543, opened 2024-02-16
Test enforces platform annotation for @RestController APIs

ashanhol:
https://broadworkbench.atlassian.net/browse/AJ-1562
Problem: "All APIs should be required at compile time to provide a platform annotation somehow; this will prevent unintentionally exposing APIs"

Solution: Unit test enforces all `@RestController`s have platform annotations `@ControlPlane` and/or `@DataPlane`

------------------
Reminder:

### DIFF
--- a/service/src/test/java/org/databiosphere/workspacedataservice/annotations/AnnotatedApisTest.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/annotations/AnnotatedApisTest.java
@@ -1,0 +1,39 @@
+package org.databiosphere.workspacedataservice.annotations;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.Arrays;
+import java.util.Set;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.config.BeanDefinition;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.context.annotation.ClassPathScanningCandidateComponentProvider;
+import org.springframework.core.type.filter.AnnotationTypeFilter;
+import org.springframework.web.bind.annotation.RestController;
+
+@SpringBootTest
+class AnnotatedApisTest {
+
+  @Test
+  void controllersMustHaveDeploymentAnnotation() {
+    // scanner that detects classes annotated with @RestController
+    ClassPathScanningCandidateComponentProvider scanner =
+        new ClassPathScanningCandidateComponentProvider(false);
+    scanner.addIncludeFilter(new AnnotationTypeFilter(RestController.class));
+    Set<BeanDefinition> restControllerComponents =
+        scanner.findCandidateComponents("org.databiosphere.workspacedataservice.controller");
+
+    for (BeanDefinition beanDefinition : restControllerComponents) {
+      // extract the class from the bean, then check its annotations for either a DataPlane or
+      // ControlPlane annotation
+      Class<?> cls = assertDoesNotThrow(() -> Class.forName(beanDefinition.getBeanClassName()));
+      assertTrue(
+          Arrays.stream(cls.getAnnotations())
+              .anyMatch(
+                  n ->
+                      n.annotationType().equals(DeploymentMode.DataPlane.class)
+                          || n.annotationType().equals(DeploymentMode.ControlPlane.class)));
+    }
+  }
+}
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/CapabilitiesController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/CapabilitiesController.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.IOException;
 import java.io.InputStream;
+import org.databiosphere.workspacedataservice.annotations.DeploymentMode.DataPlane;
 import org.databiosphere.workspacedataservice.generated.CapabilitiesApi;
 import org.databiosphere.workspacedataservice.generated.CapabilitiesServerModel;
 import org.slf4j.Logger;
@@ -13,6 +14,7 @@ import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.RestController;
 
 /** Controller for capabilities-related APIs */
+@DataPlane
 @RestController
 public class CapabilitiesController implements CapabilitiesApi {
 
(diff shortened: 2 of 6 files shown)


ashanhol:
Question: do we set the `env.wds.deploymentMode` property for our tests? As soon as I tagged the rest controllers the tests fail with 404s

service/src/test/java/org/databiosphere/workspacedataservice/annotations/AnnotatedApisTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+    for (BeanDefinition beanDefinition : restControllerComponents) {
+      // extract the class from the bean, then check its annotations for either a DataPlane or
+      // ControlPlane annotation
+      Class<?> cls = assertDoesNotThrow(() -> Class.forName(beanDefinition.getBeanClassName()));
+      assertTrue(
+          Arrays.stream(cls.getAnnotations())
+              .anyMatch(
+                  n ->
+                      n.annotationType().equals(DeploymentMode.DataPlane.class)
```
ashanhol, line 35: will change imports to get rid of `DeploymentMode.`

service/src/test/java/org/databiosphere/workspacedataservice/annotations/AnnotatedApisTest.java
```diff
@@ -0,0 +1,39 @@
...
+    scanner.addIncludeFilter(new AnnotationTypeFilter(RestController.class));
+    Set<BeanDefinition> restControllerComponents =
+        scanner.findCandidateComponents("org.databiosphere.workspacedataservice.controller");
+
+    for (BeanDefinition beanDefinition : restControllerComponents) {
+      // extract the class from the bean, then check its annotations for either a DataPlane or
+      // ControlPlane annotation
+      Class<?> cls = assertDoesNotThrow(() -> Class.forName(beanDefinition.getBeanClassName()));
+      assertTrue(
+          Arrays.stream(cls.getAnnotations())
```
jladieu, line 32: This is just for fun, but you can use some lambdas to make this even terser:

```
   // before the loop
   var requiredAnnotations =
        Set.of(DeploymentMode.DataPlane.class, DeploymentMode.ControlPlane.class);

    assertTrue(
          Arrays.stream(cls.getAnnotations())
              .map(Annotation::annotationType)
              .anyMatch(requiredAnnotations::contains));
```

service/src/test/java/org/databiosphere/workspacedataservice/annotations/AnnotatedApisTest.java
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.web.bind.annotation.RestController;
+
+@SpringBootTest
+class AnnotatedApisTest {
+
+  @Test
+  void controllersMustHaveDeploymentAnnotation() {
+    // scanner that detects classes annotated with @RestController
+    ClassPathScanningCandidateComponentProvider scanner =
+        new ClassPathScanningCandidateComponentProvider(false);
```
jladieu, line 22:
```suggestion
        new ClassPathScanningCandidateComponentProvider(/* useDefaultFilters= */ false);
```

jladieu (approved):
I like it!

I'll poke at the test failures to see if I can recommend any reasonable workarounds in lieu of [AJ-1629](https://broadworkbench.atlassian.net/browse/AJ-1629)

jladieu (commented):
I think the tests for `JobsController` are failing because `@DataPlane` and `@ControlPlane` are mutually exclusive.  If one says to enable the controller, the other says to disable it.  So I think you can't use both annotations.

I tried some workarounds, but I think the best option might just be to abandon `@DataPlane` and `@ControlPlane` in favor of the natively supported `@Profile({'data-plane'})` which can take multiple profiles (eg: (`@Profile({'data-plane', 'control-plane'})`

Or, if we like the `@DataPlane` and `@ControlPlane` annotations (I'm not at all attached to them now that I know about `@Profile`),  swapping out their implementation to essentially just be a shorthand for `@Profile({'data-plane'})` and then adding one for the 'both' case.

service/src/test/java/org/databiosphere/workspacedataservice/annotations/AnnotatedApisTest.java
```diff
@@ -0,0 +1,39 @@
...
+      Class<?> cls = assertDoesNotThrow(() -> Class.forName(beanDefinition.getBeanClassName()));
+      assertTrue(
+          Arrays.stream(cls.getAnnotations())
+              .anyMatch(
+                  n ->
+                      n.annotationType().equals(DeploymentMode.DataPlane.class)
+                          || n.annotationType().equals(DeploymentMode.ControlPlane.class)));
+    }
+  }
+}
```
davidangb, line 39: can I propose a more Spring-like solution which is hopefully simpler to read through and maintain?

The basics are:
1. `@Autowire` the Spring `ApplicationContext` as a member var of your test. When the test is a `@SpringBootTest`, then the application context contains all beans that Spring instantiated. Note that you'll want to ensure your test runs with _neither_ the data-plane nor control-plane profile (i.e. you can add `@ActiveProfiles("")`) to make sure you're finding all beans.
2. Use `ApplicationContext.getBeanNamesForAnnotation()` to find all `RestController` beans (and maybe for good measure also find all `Controller` beans?) in the context
3. Loop through all the beans you found in the previous step, and use `ApplicationContext.findAnnotationOnBean()` to look for any ControlPlane/DataPlane annotations on that bean
4. assert that each bean has either a ControlPlane/DataPlane annotation

davidangb: also, explore the various methods on `ApplicationContext` to see what else might be helpful!

davidangb:
> Or, if we like the @DataPlane and @ControlPlane annotations (I'm not at all attached to them now that I know about @Profile [kudos to @davidangb]), swapping out their implementation to essentially just be a shorthand for @Profile({'data-plane'}) and then adding one for the 'both' case.

I like this approach - basically, rewrite @DataPlane and @ControlPlane to be profile-driven instead of property-driven. Plus, it helps with the unit test :)

service/src/test/java/org/databiosphere/workspacedataservice/annotations/AnnotatedApisTest.java
davidangb: ah, you might need to use `@ActiveProfiles({ "control-plane", "data-plane" })` for such a test to work

service/src/test/java/org/databiosphere/workspacedataservice/annotations/AnnotatedApisTest.java
jladieu: I had one followup thought to this...the `assertTrue` should also print an error message that includes the offending controller if it doesn't pass, otherwise it'll be hard to tell which controller is missing an annotation :)

PR #545, opened 2024-02-16
AJ-1589 Simplify `BatchWriteService` internal APIs.

jladieu:
This is another round of refactoring to simplify the internal interfaces of `BatchWriteService` in prep for making some more tweaks to support [AJ-1589](https://broadworkbench.atlassian.net/browse/AJ-1589).

For easiest review, I recommend scanning the commits individually.

These commits largely attempt to:
- shrink the size of method argument lists the `RecordSink` API to eliminate cases where implementation classes just end up ignoring the arguments because they're not relevant
- shrink the overall size of `consumeWriteStream` to get it a little bit closer to fitting on a page.

[AJ-1589]: https://broadworkbench.atlassian.net/browse/AJ-1589?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
@@ -21,7 +21,6 @@ import org.databiosphere.workspacedataservice.recordsink.RawlsModel.Entity;
 import org.databiosphere.workspacedataservice.recordsink.RawlsModel.RemoveAttribute;
 import org.databiosphere.workspacedataservice.service.model.DataTypeMapping;
 import org.databiosphere.workspacedataservice.service.model.exception.BatchWriteException;
-import org.databiosphere.workspacedataservice.shared.model.OperationType;
 import org.databiosphere.workspacedataservice.shared.model.Record;
 import org.databiosphere.workspacedataservice.shared.model.RecordType;
 
@@ -57,24 +56,26 @@ public class RawlsRecordSink implements RecordSink {
       RecordType recordType,
       Map<String, DataTypeMapping> schema,
       List<Record> records,
-      String recordTypePrimaryKey) {
+      String primaryKey) {
     // This is a no-op for Rawls as the schema changes occur as a side effect of writeBatch
     return schema;
   }
 
   @Override
-  public void writeBatch(
+  public void upsertBatch(
       RecordType recordType,
-      Map<String, DataTypeMapping> schema,
-      OperationType opType,
+      Map<String, DataTypeMapping> schema, // ignored
       List<Record> records,
-      String primaryKey)
-      throws BatchWriteException, IOException {
-    // TODO: this method signature has a lot of unused arguments, can the interface be tidied up to
-    //  not require all these?
+      String primaryKey // ignored
+      ) throws BatchWriteException, IOException {
     ImmutableList.Builder<Entity> entities = ImmutableList.builder();
     records.stream().map(this::toEntity).forEach(entities::add);
     jsonConsumer.accept(mapper.writeValueAsString(entities.build()));
+  }
+
+  @Override
+  public void deleteBatch(RecordType recordType, List<Record> records) throws BatchWriteException {
+    throw new UnsupportedOperationException("RawlsRecordSink does not support deleteBatch");
   }
 
   private Entity toEntity(Record record) {
--- a/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
@@ -11,7 +11,6 @@ import java.lang.annotation.Target;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.UUID;
 import java.util.function.Consumer;
 import java.util.stream.Stream;
 import org.databiosphere.workspacedataservice.recordsink.RawlsModel.AddListMember;
@@ -55,7 +54,6 @@ public class RawlsRecordSink implements RecordSink {
 
   @Override
   public Map<String, DataTypeMapping> createOrModifyRecordType(
-      UUID collectionId,
       RecordType recordType,
       Map<String, DataTypeMapping> schema,
       List<Record> records,
@@ -66,7 +64,6 @@ public class RawlsRecordSink implements RecordSink {
 
   @Override
   public void writeBatch(
-      UUID collectionId,
       RecordType recordType,
       Map<String, DataTypeMapping> schema,
       OperationType opType,
--- a/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RecordSinkFactory.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RecordSinkFactory.java
@@ -22,7 +22,6 @@ public class RecordSinkFactory {
   private RecordService recordService;
   private RecordDao recordDao;
   private DataTypeInferer dataTypeInferer;
-
   private Consumer<String> jsonConsumer;
 
   public RecordSinkFactory(TwdsProperties twdsProperties, ObjectMapper mapper) {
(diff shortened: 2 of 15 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/recordsink/WdsRecordSink.java
jladieu, line 1: I realized I churned on this...at first I had made this its own class to work around `@Bean` config issues, but then put it back when I fixed those problems.

Now I wanted it as a standalone class to encapsulate `collectionId` which is only relevant for the `WdsRecordSink` scenario, so it's back again.

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -68,95 +58,77 @@ public BatchWriteResult batchWrite(
...
 
-        if (recordType != null && !Set.of(recordType).equals(groupedRecords.keySet())) {
-          throw new BadStreamingWriteRequestException(
-              "Record Type was specified as argument to BatchWriteService, "
-                  + "but actual records contained different record types. Cannot continue.");
+      // loop over all record types in this batch. For each record type, iff this is the first
+      // time we've seen this type, calculate a schema from its records and update the record type
+      // as necessary. Then, write the records into the table.
+      OperationType opType = info.operationType();
+      for (RecordType recType : groupedRecords.keySet()) {
```
jladieu, line 88: Another slight churn.  I had previously factored this out to make it into a `BiStream`, but doing so forced me to include exception handling for the calls to `recordSink`.  I decided to back up to a regular forloop (which doesn't require explicit exception handling) to take advantage of the fact that the calling code does a perfectly adequate job of exception handling for us.

service/src/test/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkTest.java
```diff
@@ -148,6 +148,15 @@ void renamesNameToIncludeRecordType() {
     assertThat(operation.attributeName()).isEqualTo("prefix:widget_name");
   }
 
+  @Test
+  void batchDeleteNotSupported() {
```
jladieu, line 152: New behavior, it'll fail fast if somehow a JSON request with a BatchDelete gets routed to Rawls.  Not that this would have ever happened...

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -55,31 +53,31 @@ public class RawlsRecordSink implements RecordSink {
...
-      UUID collectionId,
+  public void upsertBatch(
       RecordType recordType,
-      Map<String, DataTypeMapping> schema,
-      OperationType opType,
+      Map<String, DataTypeMapping> schema, // ignored
       List<Record> records,
-      String primaryKey)
-      throws BatchWriteException, IOException {
-    // TODO: this method signature has a lot of unused arguments, can the interface be tidied up to
```
jladieu, line 76: This is the TODO that spurred the changes to eliminate args from this API's arglists.

nawatts (approved):

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RecordSinkFactory.java
```diff
@@ -29,13 +35,23 @@ public void setRecordService(RecordService recordService) {
...
 
+  @Autowired(required = false) // RecordService only required for RecordSinkMode.WDS
+  public void setRecordDao(RecordDao recordDao) {
+    this.recordDao = recordDao;
+  }
+
+  @Autowired(required = false) // RecordService only required for RecordSinkMode.WDS
+  public void setDataTypeInferer(DataTypeInferer dataTypeInferer) {
+    this.dataTypeInferer = dataTypeInferer;
+  }
```
davidangb, line 46: nit: update the comment for these two setters

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RecordSinkFactory.java
```diff
@@ -44,19 +60,27 @@ public RecordSink buildRecordSink(String prefix) {
...
     }
 
     switch (twdsProperties.getDataImport().getBatchWriteRecordSink()) {
       case WDS -> {
-        return recordService;
+        return wdsRecordSink(collectionId);
       }
       case RAWLS -> {
-        return new RawlsRecordSink(prefix, mapper, jsonConsumer);
+        return rawlsRecordSink(prefix);
```
davidangb, line 71: when we get to the point of sending pub/sub messages to Rawls to trigger the import, I believe we'll need `collectionId` as an argument to `RawlsRecordSink`. Does that change the refactor at all? 

To be clear: I'm not suggesting you need to add it as an argument in this PR, just want to bring it up if it informs any of your refactoring.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -142,18 +141,16 @@ public int tsvUpload(
 
     TsvRecordSource recordSource =
         recordSourceFactory.forTsv(records.getInputStream(), recordType, primaryKey);
-    RecordSink recordSink = recordSinkFactory.buildRecordSink(/* prefix= */ "tsv");
+    RecordSink recordSink = recordSinkFactory.buildRecordSink(collectionId, /* prefix= */ "tsv");
```
davidangb, line 144: It's a little strange that data-plane WDS needs to pass a prefix "tsv", "json", etc - which is then promptly ignored. This isn't horrible, and I see why leaving it in makes the method signatures work. Maybe add an overload in `RecordSinkFactory` so you can have both of these?
```java
public RecordSink buildRecordSink(UUID collectionId);
public RecordSink buildRecordSink(UUID collectionId, String prefix);
```

service/src/test/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkTest.java
```diff
@@ -148,6 +148,15 @@ void renamesNameToIncludeRecordType() {
...
+    var thrown =
+        assertThrows(
+            UnsupportedOperationException.class,
+            () -> recordSink.deleteBatch(RecordType.valueOf("widget"), emptyList()));
```
davidangb, line 156: this is causing a Sonar complaint - rewrite it?

davidangb (commented):
comments inline!

service/src/test/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkTest.java
jladieu: Attempted a fix with c9304a527711696bfa0e70f87b9b1f0011a95e1a -- I'll have to see what Sonar says after it gets done reprocessing and I'll resolve this if it does the trick.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
jladieu: I agree... I _almost_ did what you suggested as part of this PR but backed off with the intent of deferring to the implementation PR for AJ-1589.  I'm not exactly sure what I want to do here, but I realize for the `tdr` prefix, it's also not just a hard and fast _always apply the prefix_ so my solution there might cause this to get reshuffled a bit anyway.  For now I am leaving it as is but have annotated the TODO to deal with the fact that prefix is optional in WDS mode.

service/src/test/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkTest.java
jladieu: Also, the sonar findings aren't really jumping out at me... because I look for the the GHA checkboxes being green as my indicator of "safe to proceed".  Can you share how you're spotting these?

nawatts:
For what it's worth, I see the Sonar issues more as code smells that are usually nice to address vs blocking issues.

For TODO comments, yeah, no issue merging that. I suppose the Sonar issue is a nice reminder in case the comment was left from the PR development vs a reminder for a future change.

PR #553, opened 2024-02-20
AJ-1589 Drop `TwoPass` interface

jladieu:
This intermediate interface is no longer needed as it previously existed as a check to support whether a two pass import mode was supported.

This is now handled by the caller of `BatchWriteService` depending on the context of the import.


service/src/main/java/org/databiosphere/workspacedataservice/recordsource/TwoPassRecordSource.java
```diff
@@ -1,14 +0,0 @@
-package org.databiosphere.workspacedataservice.recordsource;
-
-/**
- * Marker interface that denotes a RecordSource which runs in two passes. The first pass will upsert
- * base attributes, and the second pass will upsert relation attributes. This allows for data
- * streams where earlier records contain relations to later records.
- */
-public interface TwoPassRecordSource extends RecordSource {
-
-  enum ImportMode {
```
jladieu, line 10: This got pulled up into `RecordSource`

yuliadub (approved):

PR #554, opened 2024-02-20
AJ-1631: create-import permission change; test fixes

davidangb:
Changes in this PR:
* Change the create-import API to properly check permissions via `CollectionService` and throw `AuthenticationMaskableException` on permission denial
* Update unit tests for compatibility with the permission change
* Add new unit tests for control-plane/data-plane to validate the permission changes
* Introduce `NoopSchedulerDao` and the "noop-scheduler-dao" test profile, and use them in all tests that want job-scheduling to be a noop. This avoids needing to create `@MockBean`s and create noop mocks in each test.

_reviewers: note that the only runtime changes are in `ImportService.java`; all the rest of the changes are in unit tests_

### DIFF
--- a/service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceTest.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceTest.java
@@ -23,6 +23,7 @@ import java.util.stream.Stream;
 import org.broadinstitute.dsde.workbench.client.sam.ApiException;
 import org.broadinstitute.dsde.workbench.client.sam.api.GoogleApi;
 import org.broadinstitute.dsde.workbench.client.sam.api.ResourcesApi;
+import org.databiosphere.workspacedataservice.common.TestBase;
 import org.databiosphere.workspacedataservice.config.TwdsProperties;
 import org.databiosphere.workspacedataservice.dao.CollectionDao;
 import org.databiosphere.workspacedataservice.dao.JobDao;
@@ -54,7 +55,7 @@ import org.springframework.test.context.ActiveProfiles;
 @ActiveProfiles(profiles = {"mock-sam", "mock-collection-dao"})
 @DirtiesContext
 @SpringBootTest
-class ImportServiceTest {
+class ImportServiceTest extends TestBase {
 
   @Autowired ImportService importService;
   @Autowired CollectionDao collectionDao;
--- a/.github/workflows/release-python-client.yml
+++ b/.github/workflows/release-python-client.yml
@@ -88,6 +88,7 @@ jobs:
           export DATA_REPO_URL=http://localhost:9889
           export WORKSPACE_MANAGER_URL=http://localhost:9889
           export WORKSPACE_ID=123e4567-e89b-12d3-a456-426614174000
+          export SPRING_PROFILES_ACTIVE=data-plane
           ./gradlew --build-cache build -x test
           ./gradlew bootRun &
           count=20
--- a/README.md
+++ b/README.md
@@ -37,14 +37,16 @@ Instead use the `run_postgres.sh` script to set up a docker container running po
 
 #### Environment Variables
 
-For WDS to work properly, a spring profile needs to be set so that several environment variables are
-populated correctly.
+For WDS to work properly, several environment variables need to be set, and exactly one of the
+following active profiles must be set:
 
-To run WDS locally and therefore use the local application properties, set the profile the following
-way:
+- control-plane
+- data-plane
+
+To run WDS locally and therefore use the local application properties, set the profile as follows:
 
 ```bash
-export SPRING_PROFILES_ACTIVE=local
+export SPRING_PROFILES_ACTIVE=local,data-plane
 ```
 
 Other profiles that are available are:
@@ -59,9 +61,10 @@ Other profiles that are available are:
 You are unlikely to use prod and bee when running locally, those profiles are leveraged when WDS is
 deployed in Terra.
 
-If you would like to not use a profile (i.e. not set SPRING_PROFILES_ACTIVE variable), you can set
-the following environment variables manually. The variables that need to be set are described below.
-You can also add them to your `~/.zshrc` or similar shell profile.
+If you would like to not use the local profile, you can set some of the environment variables
+manually. The variables that need to be set are described below. You can also add them to your
+`~/.zshrc` or similar shell profile. However, in order for the app to run correctly, you still
+have to at least specify data-plane or control-plane.
 
 ##### SAM_URL
 
--- a/service/src/main/java/org/databiosphere/workspacedataservice/config/ConfigurationException.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/config/ConfigurationException.java
@@ -1,0 +1,12 @@
+package org.databiosphere.workspacedataservice.config;
+
+/** Thrown when there's a problem with the application configuration. */
+public class ConfigurationException extends RuntimeException {
+  public ConfigurationException(String message) {
+    super(message);
+  }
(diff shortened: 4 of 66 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceControlPlaneTest.java
```diff
@@ -0,0 +1,126 @@
...
+  @Autowired ImportService importService;
+  @MockBean CollectionDao collectionDao;
+  @MockBean SamDao samDao;
+  @MockBean SchedulerDao schedulerDao;
+
+  /* Collection does not exist, user has access */
+  @Test
+  void hasPermission() {
+    // ARRANGE
+    CollectionId collectionId = CollectionId.of(UUID.randomUUID());
```
jladieu, line 50: `doNothing().when(schedulerDao).schedule(any(Schedulable.class));` looks kinda like a telltale 'shove it aside to get it out of the way' mock.  If I'm right there, I think it'd be worth bumping up to the `BeforeEach` since it doesn't really pertain to the arrangements that the assertions care about.  (same with the other test case file)

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceTest.java
```diff
@@ -118,7 +133,7 @@ void persistsJobAsQueued(ImportRequestServerModel.TypeEnum importType) {
     // schedulerDao.schedule(), which returns void, returns successfully
     doNothing().when(schedulerDao).schedule(any(Schedulable.class));
     // create collection (in the MockCollectionDao)
-    UUID collectionId = UUID.randomUUID();
```
jladieu, line 121: Since the UUID is no longer random, I think it'd make sense to factor this out to happen once in the setup.

jladieu (commented):
Changes look good and the tests also look good.  I have some minor suggestions about DRYing up test setup.

This PR does highlight a couple concerns I'm starting to feel at the design level:

1) Having to spawn two different test case files for DataPlane vs. ControlPlane.  It's not immediately clear to me why the behavior of the code that's being changed is _different_ between the two deployment modes.  Is the behavior change something that we could mock more locally in the unit tests than to swap to the other profile which is pretty heavyweight?

2) `twdsProperties` is weaving its way throughout the code and test base (in this PR, as part of getting the `workspaceId` from the config), and with it bringing law of demeter violations (`twdsProperties.getInstance().getSomethingElse()`).  I like these properties classes, but I think we've sprouted an anti-pattern here which is also a bit infectious.  I think we need to figure out how to import properties without importing _ALL_ the properties.  Specifically, exposing just the `InstanceProperties` or `tenancyProperties` as `@Bean`s that can be autowired.  I think we can do this on top of what we've already built with TWDSProperties but haven't quite figured out how yet.  Wondering your thoughts on this?  Does this directionally seem correct?

jladieu:
@davidangb would you consider picking the `@DirtiesContext` fix out into its own standalone PR?  I think it would solve an immediate problem and probably doesn't require much discussion whereas I realize I just dropped a couple design level concerns into this one.

Alternately, I'd be happy to ship this one as is, but would love to keep discussing the design issues I raised.

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceDataPlaneTest.java
```diff
@@ -0,0 +1,162 @@
...
+  @MockBean CollectionDao collectionDao;
+  @MockBean SamDao samDao;
+  @MockBean SchedulerDao schedulerDao;
+
+  /* collection exists, workspace matches env var, user has access */
+  @Test
+  void userHasAccess() {
+    // ARRANGE
+    WorkspaceId workspaceId = WorkspaceId.of(twdsProperties.getInstance().getWorkspaceUuid());
+    CollectionId collectionId = CollectionId.of(UUID.randomUUID());
```
ashanhol, line 55: If you're randomly generating this every test, you could pull it out and just make it a global variable

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceDataPlaneTest.java
```diff
@@ -0,0 +1,162 @@
...
+    // schedulerDao.schedule(), which returns void, returns successfully
+    doNothing().when(schedulerDao).schedule(any(Schedulable.class));
```
ashanhol, line 57: I think Josh's schedule comment applies here too

ashanhol (approved):
Small refactoring comments but otherwise looks good

jladieu (approved):
Approved for merging as is, but I would welcome continuing the design conversations async if you want to get this landed in the meantime.

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
```diff
@@ -52,11 +52,15 @@ public GenericJobServerModel createImport(
     collectionService.validateCollection(collectionId);
 
     // validate write permission
-    // TODO AJ-1631: this must use collectionService.canWriteCollection() instead
-    boolean hasWriteWorkspacePermission = samDao.hasWriteWorkspacePermission();
+    boolean hasWriteWorkspacePermission =
```
yuliadub, line 55: nit: confusing naming - are we talking about collection or workspace?

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceControlPlaneTest.java
```diff
@@ -0,0 +1,115 @@
...
+    // collection dao says the collection does not exist
+    when(collectionDao.getWorkspaceId(collectionId))
+        .thenThrow(new EmptyResultDataAccessException("unit test intentional error", 1));
+    // sam dao says the user has write permission
+    when(samDao.hasWriteWorkspacePermission(collectionId.toString())).thenReturn(true);
+
+    // define the import request
+    URI importUri = URI.create("http://does/not/matter");
```
yuliadub, line 53: you define the same value throughout - can this be a class variable?

yuliadub (approved):
just 2 small comments, but neither are blocking

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceTest.java
davidangb: updated to move this to the class level.

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceDataPlaneTest.java
davidangb: updated to move this to the class level.

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceControlPlaneTest.java
davidangb: updated!

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
davidangb: heh, right now they're the same thing. I've renamed to `hasWriteCollectionPermission`.

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceControlPlaneTest.java
jladieu: Love it!

jladieu (approved):

PR #555, opened 2024-02-21
[AJ-1652] Handle primary keys specified in request body

nawatts:
PUT and PATCH requests to `/{instanceId}/records/{version}/{recordType}/{recordId}` specify the record's primary key value in the URL via the `recordId` path parameter and, for PUT requests, the `primaryKey` query parameter. However, they may also specify the primary key value in the attributes in the request body.

Currently, WDS doesn't handle this consistently.
- Creating a new record, whether or not the two primary key values match, the request succeeds.
- Updating an existing record, whether or not the two primary key values match, results in an SQL error.

This updates that endpoint to require that, if a primary key value is specified in the request body, it must match the one in the URL. It they do match, the request now succeeds instead of producing an SQL error.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
@@ -25,11 +25,11 @@ import org.databiosphere.workspacedataservice.service.model.RelationCollection;
 import org.databiosphere.workspacedataservice.service.model.RelationValue;
 import org.databiosphere.workspacedataservice.service.model.ReservedNames;
 import org.databiosphere.workspacedataservice.service.model.exception.BatchWriteException;
+import org.databiosphere.workspacedataservice.service.model.exception.ConflictingPrimaryKeysException;
 import org.databiosphere.workspacedataservice.service.model.exception.InvalidNameException;
 import org.databiosphere.workspacedataservice.service.model.exception.InvalidRelationException;
 import org.databiosphere.workspacedataservice.service.model.exception.MissingObjectException;
 import org.databiosphere.workspacedataservice.service.model.exception.NewPrimaryKeyException;
-import org.databiosphere.workspacedataservice.service.model.exception.ValidationException;
 import org.databiosphere.workspacedataservice.shared.model.Record;
 import org.databiosphere.workspacedataservice.shared.model.RecordAttributes;
 import org.databiosphere.workspacedataservice.shared.model.RecordRequest;
@@ -314,7 +314,7 @@ public class RecordService {
     String primaryKey = recordDao.getPrimaryKeyColumn(recordType, collectionId);
     if (incomingAttrs.containsAttribute(primaryKey)
         && incomingAttrs.getAttributeValue(primaryKey) != recordId) {
-      throw new ValidationException("Primary key in payload does not match primary key in URL");
+      throw new ConflictingPrimaryKeysException();
     }
     RecordAttributes allAttrs = singleRecord.putAllAttributes(incomingAttrs).getAttributes();
     Map<String, DataTypeMapping> typeMapping = inferer.inferTypes(incomingAttrs);
@@ -343,7 +343,7 @@ public class RecordService {
       if (primaryKey.isPresent()
           && attributesInRequest.containsAttribute(primaryKey.get())
           && attributesInRequest.getAttributeValue(primaryKey.get()) != recordId) {
-        throw new ValidationException("Primary key in payload does not match primary key in URL");
+        throw new ConflictingPrimaryKeysException();
       }
       RecordResponse response =
           new RecordResponse(recordId, recordType, recordRequest.recordAttributes());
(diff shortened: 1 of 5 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
```diff
@@ -310,21 +311,21 @@ public RecordResponse updateSingleRecord(
...
+    String primaryKey = recordDao.getPrimaryKeyColumn(recordType, collectionId);
+    if (incomingAttrs.containsAttribute(primaryKey)
+        && incomingAttrs.getAttributeValue(primaryKey) != recordId) {
+      throw new ConflictingPrimaryKeysException();
+    }
     RecordAttributes allAttrs = singleRecord.putAllAttributes(incomingAttrs).getAttributes();
     Map<String, DataTypeMapping> typeMapping = inferer.inferTypes(incomingAttrs);
     Map<String, DataTypeMapping> existingTableSchema =
-        recordDao.getExistingTableSchemaLessPrimaryKey(collectionId, recordType);
+        recordDao.getExistingTableSchema(collectionId, recordType);
```
nawatts, line 322: This was the key to avoiding SQL errors. `addOrUpdateColumnIfNeeded` needs to be aware of the existing primary key column.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
```diff
@@ -339,16 +340,25 @@ public ResponseEntity<RecordResponse> upsertSingleRecord(
     Map<String, DataTypeMapping> requestSchema = inferer.inferTypes(attributesInRequest);
     HttpStatus status = HttpStatus.CREATED;
     if (!recordDao.recordTypeExists(collectionId, recordType)) {
+      if (primaryKey.isPresent()
+          && attributesInRequest.containsAttribute(primaryKey.get())
+          && attributesInRequest.getAttributeValue(primaryKey.get()) != recordId) {
+        throw new ConflictingPrimaryKeysException();
+      }
```
davidangb, line 347: There is one corner case this doesn't handle:
* record type does not exist yet
* user does not specify the primary key column name, so therefore pk column defaults to "sys_name"
* user includes "sys_name" as an attribute in the request payload

I think this needs to be something like:
```java
      String calculatedPk = primaryKey.orElse(ReservedNames.RECORD_ID);
      if (attributesInRequest.containsAttribute(calculatedPk)
          && attributesInRequest.getAttributeValue(calculatedPk) != recordId) {
        throw new ConflictingPrimaryKeysException();
      }
```
( that snippet `primaryKey.orElse(ReservedNames.RECORD_ID)` is used in a few places, maybe should be a reusable method )

davidangb (commented):
one corner case inline, thanks for tackling this!

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
nawatts: Oo, nice catch!

nawatts: Handled this case in https://github.com/DataBiosphere/terra-workspace-data-service/pull/555/commits/a13432a9b63cc7639059d0d5738e2a29a6afc244.

yuliadub:
since you are looking at the primary key code, if you have a moment to check, Rob was able to upload a record without specifying the name of the primary key when using batch_insert from the python client (the primary key in that api is not required as a param) and the key ended up being called "sys_name" which is not a name that is allowed. (https://github.com/rtitle/terra-chess/blob/main/src/Prepare%20Data%20Table.ipynb?short_path=edd9f34#L306) not sure if this is exactly related but maybe?

nawatts:
That sounds like expected behavior? `sys_name` is the default name for the primary key column if none is specified.
https://github.com/DataBiosphere/terra-workspace-data-service/blob/0305a4094ea10a78920c0586d3198b6b5a5faf24/service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java#L388

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
yuliadub: aha David already mentioned the same thing 😄 should have read the comments first before adding mine, sorry about that!

yuliadub:
> That sounds like expected behavior? `sys_name` is the default name for the primary key column if none is specified.
> 
> https://github.com/DataBiosphere/terra-workspace-data-service/blob/0305a4094ea10a78920c0586d3198b6b5a5faf24/service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java#L388

ah okay, interesting! but when I download a file with sys_name and then try to reupload it via UI, it wont let me. So maybe then this is a terra-ui and wds behaviors not quite matching then?

yuliadub (approved):

nawatts:
Could be something related to TSV vs JSON APIs as well? Or new vs existing record type?

Your comments did prompt me to add a test case for including a `sys_name` field in the request to create a new record (https://github.com/DataBiosphere/terra-workspace-data-service/pull/555/commits/7dd008912ec40c87b421e16e96f90062b636dd16)

PR #557, opened 2024-02-21
AJ-1629 Make `data-plane` default profile

jladieu:
[AJ-1629](https://broadworkbench.atlassian.net/browse/AJ-1629) seeks to make WDS require either the `data-plane` or the `control-plane` profile to be specified in order to run.  This starts by making `data-plane` the default for test cases that extend `TestBase`.

In addition, since `data-plane` requires a `workspace-id` to be configured, it also defaults that value to a commonly used UUID, which eliminates that concern from several individual tests.

A `ConfigurationExceptionDetector` is also added to detect when a test fails due to a `ConfigurationException`, and provides a hint of how to resolve.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ export SAM_URL=http://localhost:9889
 To run WDS locally, you can either use the command line:
 
 ```bash
-./gradlew bootRun
+./gradlew bootRun --args='--spring.profiles.active=data-plane'
 ```
 
 Or, from Intellij, go
--- a/README.md
+++ b/README.md
@@ -37,14 +37,16 @@ Instead use the `run_postgres.sh` script to set up a docker container running po
 
 #### Environment Variables
 
-For WDS to work properly, a spring profile needs to be set so that several environment variables are
-populated correctly.
+For WDS to work properly, several environment variables need to be set, and exactly one of the
+following active profiles must be set:
 
-To run WDS locally and therefore use the local application properties, set the profile the following
-way:
+- control-plane
+- data-plane
+
+To run WDS locally and therefore use the local application properties, set the profile as follows:
 
 ```bash
-export SPRING_PROFILES_ACTIVE=local
+export SPRING_PROFILES_ACTIVE=local,data-plane
 ```
 
 Other profiles that are available are:
@@ -59,9 +61,10 @@ Other profiles that are available are:
 You are unlikely to use prod and bee when running locally, those profiles are leveraged when WDS is
 deployed in Terra.
 
-If you would like to not use a profile (i.e. not set SPRING_PROFILES_ACTIVE variable), you can set
-the following environment variables manually. The variables that need to be set are described below.
-You can also add them to your `~/.zshrc` or similar shell profile.
+If you would like to not use the local profile, you can set some of the environment variables
+manually. The variables that need to be set are described below. You can also add them to your
+`~/.zshrc` or similar shell profile. However, in order for the app to run correctly, you still
+have to at least specify data-plane or control-plane.
 
 ##### SAM_URL
 
@@ -137,7 +140,7 @@ export SAM_URL=http://localhost:9889
 To run WDS locally, you can either use the command line:
 
 ```bash
-./gradlew bootRun --args='--spring.profiles.active=data-plane'
+./gradlew bootRun
 ```
 
 Or, from Intellij, go
--- a/service/src/main/java/org/databiosphere/workspacedataservice/config/ConfigurationException.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/config/ConfigurationException.java
@@ -1,0 +1,12 @@
+package org.databiosphere.workspacedataservice.config;
+
+/** Thrown when there's a problem with the application configuration. */
+public class ConfigurationException extends RuntimeException {
+  public ConfigurationException(String message) {
+    super(message);
+  }
+
+  public ConfigurationException(String message, Throwable throwable) {
+    super(message, throwable);
+  }
+}
--- a/service/src/test/java/org/databiosphere/workspacedataservice/common/TestBase.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/common/TestBase.java
@@ -23,7 +23,6 @@ import org.springframework.test.context.TestPropertySource;
  * which will automatically provide a hint if it detects a test failure due to a {@link
  * ConfigurationException}.
  */
-@SpringBootTest
 @ActiveProfiles({"data-plane"})
 @TestPropertySource(
     properties = {
--- a/service/src/test/java/org/databiosphere/workspacedataservice/common/TestBase.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/common/TestBase.java
@@ -1,0 +1,35 @@
+package org.databiosphere.workspacedataservice.common;
+
+import org.databiosphere.workspacedataservice.config.ConfigurationException;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.TestPropertySource;
+
+/**
+ * Optional base class for any test class annotated with {@link SpringBootTest} to automatically
+ * enable the "data-plane" profile and a "workspace-id".
+ *
+ * <p>Subclasses can append additional active profiles by using the {@link ActiveProfiles}
+ * annotation.
+ *
+ * <p>To explicitly override the "data-plane" profile, subclasses can use
+ * {@code @ActiveProfiles(value={"some-profile"}, inheritProfiles=false)}.
+ *
+ * <p>To explicitly override the "workspace-id" that is configured, subclasses can use
+ * {@code @TestPropertySource(properties={"twds.instance.workspace-id=<...>"})}.
+ *
+ * <p>Note: extending this class also extends the {@link ConfigurationExceptionDetector} class,
+ * which will automatically provide a hint if it detects a test failure due to a {@link
+ * ConfigurationException}.
+ */
+@SpringBootTest
+@ActiveProfiles({"data-plane"})
+@TestPropertySource(
+    properties = {
+      // data-plane mode requires a workspace-id to be set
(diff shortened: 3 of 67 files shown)


jladieu:
This PR depends on a fix in https://github.com/DataBiosphere/terra-workspace-data-service/pull/560 so it is temporarily based off that branch rather than main.

service/src/test/java/org/databiosphere/workspacedataservice/common/TestBase.java
```diff
@@ -0,0 +1,35 @@
...
+ * which will automatically provide a hint if it detects a test failure due to a {@link
+ * ConfigurationException}.
+ */
+@SpringBootTest
+@ActiveProfiles({"data-plane"})
+@TestPropertySource(
+    properties = {
+      // data-plane mode requires a workspace-id to be set
+      // example uuid from https://en.wikipedia.org/wiki/Universally_unique_identifier
+      "twds.instance.workspace-id=123e4567-e89b-12d3-a456-426614174000",
```
jladieu, line 32: @ashanhol this is the part of the PR that I think might help fix the CI failures in https://github.com/DataBiosphere/terra-workspace-data-service/pull/558

README.md
```diff
@@ -137,7 +137,7 @@ export SAM_URL=http://localhost:9889
 To run WDS locally, you can either use the command line:
 
 ```bash
-./gradlew bootRun
+./gradlew bootRun --args='--spring.profiles.active=data-plane'
```
jladieu, line 140: I feel like this is slightly unfortunate in that everyone will need to start running these extra args when running locally, but I prefer this to having `data-plane` unintentionally chosen due to this param being missing in dev/staging/prod.

yuliadub: would it make sense instead to parse the profile variable at startup and just fail to start if data plane or control plane is missing? or would that be too much (philosophical question)

ashanhol: Since this is being added to the readme, we could also tell people to configure the environment variable like we do others.

service/src/main/java/org/databiosphere/workspacedataservice/config/ConfigurationException.java
```diff
@@ -0,0 +1,12 @@
+package org.databiosphere.workspacedataservice.config;
+
+/** Thrown when there's a problem with the application configuration. */
+public class ConfigurationException extends RuntimeException {
+  public ConfigurationException(String message) {
+    super(message);
```
ashanhol, line 6: Do we maybe want to add a logger message here since we're replacing those log messages with this exception?

service/src/test/java/org/databiosphere/workspacedataservice/common/TestBase.java
```diff
@@ -0,0 +1,35 @@
...
+ * Optional base class for any test class annotated with {@link SpringBootTest} to automatically
+ * enable the "data-plane" profile and a "workspace-id".
+ *
+ * <p>Subclasses can append additional active profiles by using the {@link ActiveProfiles}
+ * annotation.
+ *
+ * <p>To explicitly override the "data-plane" profile, subclasses can use
+ * {@code @ActiveProfiles(value={"some-profile"}, inheritProfiles=false)}.
+ *
+ * <p>To explicitly override the "workspace-id" that is configured, subclasses can use
```
ashanhol, line 19: these are awesome instructions

service/src/test/java/org/databiosphere/workspacedataservice/common/TestBase.java
ashanhol: ahhhhhhhhhhh yes ok i have my workspace id set as an environment variable

service/src/test/java/org/databiosphere/workspacedataservice/controller/ImportControllerMockMvcTest.java
```diff
@@ -16,7 +16,7 @@
 import org.springframework.test.context.ActiveProfiles;
 import org.springframework.test.web.servlet.MvcResult;
 
-@ActiveProfiles(profiles = {"mock-instance-dao", "mock-sam"})
+@ActiveProfiles(profiles = {"mock-collection-dao", "mock-sam"})
```
ashanhol, line 19: i see you sneaking a little rename in there 👀

not blocking but maybe don't since this PR already touches so many files?

service/src/test/java/org/databiosphere/workspacedataservice/controller/MockMvcTestBase.java
```diff
@@ -5,15 +5,18 @@
...
 import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
 import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.ActiveProfiles;
 import org.springframework.test.web.servlet.MockMvc;
 import org.springframework.test.web.servlet.MvcResult;
 
 @AutoConfigureMockMvc
 @SpringBootTest
-class MockMvcTestBase {
+@ActiveProfiles({"mock-sam"})
```
ashanhol, line 18: I noticed a lot of tests that extend MockMvcTestBase were manually adding mock-sam profiles. do you want to remove those now that you're adding it to the super class?

README.md
jladieu: @yuliadub I'm not entirely sure I understand your idea; I think the application will _currently_ fail to start if either of these are missing, it's just that it fails with an ugly stacktrace.

Areyou suggesting that before we start doing any springboot wiring and stuff that we check to see if this variable has been set first, and head it off by failing fast?

If that's the case, then IMO that's an interesting idea, although I might take it one step further, and just make it a full blown required command line arg (like, you have to run the executable with `control-plane` or `data-plane` -- and then we could programmatically _set_ the `ActiveProfile` based on that mode.

That said, for the purposes of this PR, I'm inclined to rely what spring-boot provides for this out of the box, and use documentation to encourage correct behavior, so I'm leaning toward just adjusting the documentation to suggest the environment variable as @ashanhol suggested.

service/src/main/java/org/databiosphere/workspacedataservice/config/ConfigurationException.java
jladieu: I don't think so -- I'd actively avoid logging AND throwing (here's random reference I pulled out of Google to substantiate my hunch here which is that you should log OR throw, but never both: https://rolf-engelhard.de/2013/04/logging-anti-patterns-part-ii/)

README.md
yuliadub: having an environment variable sounds good! 

I think the reason why I started thinking of the direction of somehow enforcing this is because I thought you were saying this could cause major issues if not used correctly. But if spring boot does the right thing and wont even let WDS start when the profile is missing, I think that is certainly sufficient (when I was playing with local/dev/prod profile it didnt seem like not having them actually made wds not start, but I know a few things have changed since then)

jladieu: Adjusted README to just include this detail as part of setting environment variable earlier in the doc.

service/src/test/java/org/databiosphere/workspacedataservice/controller/ImportControllerMockMvcTest.java
jladieu: Backed this out!

service/src/test/java/org/databiosphere/workspacedataservice/controller/MockMvcTestBase.java
jladieu: Sorry I put this in temporarily and then meant to back it out and forgot...I undid this change.

yuliadub (approved):
thank you for thinking though this!

README.md
jladieu: Just saw your second comment @yuliadub -- one of the goals of this ticket (although not necessarily covered by this PR which is mostly focused on the unit tests) is for app to crash without starting if one of the two profiles isn't specified.  I think if after the ticket is done we can still somehow get the app to start without a profile, there's more work to do =)

service/src/test/java/org/databiosphere/workspacedataservice/common/TestBase.java
```diff
@@ -0,0 +1,35 @@
...
+ * {@code @TestPropertySource(properties={"twds.instance.workspace-id=<...>"})}.
+ *
+ * <p>Note: extending this class also extends the {@link ConfigurationExceptionDetector} class,
+ * which will automatically provide a hint if it detects a test failure due to a {@link
+ * ConfigurationException}.
+ */
+@SpringBootTest
```
davidangb, line 26: is the `@SpringBootTest` annotation also overridable, and if so should there be instructions in the javadoc for it? I see for instance that `ConcurrentDataTypeChangesTest` (next in the diff) has `@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)` … does that work as expected?

davidangb: followup question: after reviewing the individual tests, I see we haven't removed `@SpringBootTest` from any of them. Does it make sense then to have that annotation both here on the base and on all tests?

davidangb (commented):
two questions inline, hopefully not big ones. Nice job on this, I think this will be helpful!

I also see 10 sonar findings - not your code I'm sure, but could you tackle them?

ashanhol (approved):

service/src/test/java/org/databiosphere/workspacedataservice/common/TestBase.java
jladieu: Thanks for raising the question...

So I did some minor experimentation (using the `properties` attribute of `@SpringBootTest`) and it looks like any subclass that provides its own `@SpringBootTest` will take precedence and overwrite the parent class.

If a subclass wants to do its own autowiring, it needs to annotate itself as `@SpringBootTest` anyway.  So at most, this annotation in the superclass becomes a convenience for a small minority of tests.

To that end, I opted not to do the exercise of stripping `@SpringBootTest` from all the classes that don't actually need it and just went ahead and removed it from the superclass.  Nothing complained :)

jladieu:
> I also see 10 sonar findings - not your code I'm sure, but could you tackle them?

Handled these in a separate PR: https://github.com/DataBiosphere/terra-workspace-data-service/pull/562

PR #558, opened 2024-02-21
Platform Annotations now Profile Based

ashanhol:
Prefactor for https://broadworkbench.atlassian.net/browse/AJ-1562

Relies on ~#557~ main

Our platform annotations were property based. I'm modifying the properties to be dependant on profiles (since we're using them anyway) and modifying the annotations to use those profiles too. 
Additionally - in the process of doing this we revealed CollectionInitalizer.java was using `@profile("!local")` to make sure it didn't run on startup while we were running certain integration tests. We reformatted it to instead use a twds.instance.run-on-startup property.  

-----------------------
Reminder:


service/src/main/resources/application-data-plane.yml
```diff
@@ -1,3 +1,6 @@
+env:
+  wds:
+    deploymentMode: data-plane
```
jladieu, line 3: What happens if you get rid of `deploymentMode` entirely?  I think if the annotations are driven by `@Profile` we _shouldn't_ need this anymore.

One note, I grepped around and I think we'd have to fix logback-spring.xml at least...but wondering what drove you to preserve this?

ashanhol: @davidangb and I discussed this offline and I believe he thought it was being used in multiple places still. 
If it is only used there, how easy is it to get xml to look at profile instead of property?

service/src/test/java/org/databiosphere/workspacedataservice/annotations/DataPlaneDeploymentModeTest.java
```diff
@@ -0,0 +1,26 @@
...
+@ActiveProfiles({"data-plane"})
+class DataPlaneDeploymentModeTest extends DeploymentModeTestBase {
+
+  @ParameterizedTest(name = "{0} is enabled for DeploymentMode data-plane only")
+  @MethodSource("dataPlaneConditionalBeans")
+  void beansEnabledForDataPlaneOnly(String beanName) {
+    assertNotNull(context.findAnnotationOnBean(beanName, DataPlane.class));
+  }
+
+  /* TODO: Enable when we have parameters
```
jladieu, line 19: I'm assuming you can't just uncomment it and this and let it no-op?  Does it fail when `controlPlaneConditionalBeans` returns an empty `Stream` for some reason?

Assuming it's necessary, I'm wondering if you could use `@Disabled("<reason>")` to natively disable the test instead.

jladieu:
I've also bumped into the test failures you're seeing on this PR, and might have a fix as part of turning on `data-plane` in unit tests...

service/src/test/java/org/databiosphere/workspacedataservice/annotations/DataPlaneDeploymentModeTest.java
ashanhol: Yeah, it fails with "" or no arguments. 
Happy to do comments or @Disabled("reason"), whichever you prefer

service/src/main/resources/application-control-plane.yml
```diff
@@ -1,9 +1,13 @@
+env:
+  wds:
+    deploymentMode: control-plane
```
davidangb, line 3: if we end up keeping this property (as opposed to [deleting it entirely](https://github.com/DataBiosphere/terra-workspace-data-service/pull/558/files#r1498089817)), could we move it under `twds` instead of under `env`?

The `env` section is for "values [that] are used more than once": https://github.com/DataBiosphere/terra-workspace-data-service/blob/1e67b3bd360acb7184658534115202907f2b7a8b/service/src/main/resources/application.yml#L1, i.e. reused inside other property definitions. This one doesn't match that.

service/src/test/java/org/databiosphere/workspacedataservice/annotations/DeploymentModeTestBase.java
```diff
@@ -0,0 +1,36 @@
...
+
+  @Autowired protected ApplicationContext context;
+
+  protected static Stream<Arguments> dataPlaneConditionalBeans() {
+    return Stream.of(Arguments.of("collectionInitializer"));
+  }
+
+  protected static Stream<Arguments> controlPlaneConditionalBeans() {
+    return Stream.of(Arguments.of());
+  }
```
davidangb, line 35: when each `Arguments` is singular, you can omit `Arguments` and just return the value, leads to slightly cleaner code:
```java
  protected static Stream<String> dataPlaneConditionalBeans() {
    return Stream.of("collectionInitializer");
  }

  protected static Stream<String> controlPlaneConditionalBeans() {
    return Stream.of();
  }
```

service/src/test/java/org/databiosphere/workspacedataservice/annotations/ControlPlaneDeploymentModeTest.java
```diff
@@ -0,0 +1,27 @@
...
+    assertThrows(NoSuchBeanDefinitionException.class, () -> context.getBean(beanName));
+  }
+
+  /*  TODO: Enable when we have parameters
+   @ParameterizedTest(name = "{0} is enabled for DeploymentMode control-plane only")
+   @MethodSource("controlPlaneConditionalBeans")
+   void beansEnabledForControlPlaneOnly(String beanName) {
+     assertNotNull(context.findAnnotationOnBean(beanName, ControlPlane.class));
+   }
+  */
```
davidangb, line 26: you could also use
```java
@Disabled("my reason for disabling this particular test")
```
to keep the code compiling, but ignore the test

service/src/main/resources/application-data-plane.yml
davidangb: I don't see this property used anywhere except in logback-spring.xml. Check out the existing usage of `<springProfile ...>` in that file. I think it's just a matter of using `<springProfile>` instead of `<springProperty>`.

service/src/test/java/org/databiosphere/workspacedataservice/annotations/ControlPlaneDeploymentModeTest.java
```diff
@@ -0,0 +1,27 @@
...
+
+@ActiveProfiles({"control-plane"})
+class ControlPlaneDeploymentModeTest extends DeploymentModeTestBase {
+
+  @ParameterizedTest(name = "{0} is disabled for DeploymentMode control-plane")
+  @MethodSource("dataPlaneConditionalBeans")
+  void beansDisabledForControlPlane(String beanName) {
+    // bean should not exist in this context if not tagged with @ControlPlane
+    assertThrows(NoSuchBeanDefinitionException.class, () -> context.getBean(beanName));
+  }
```
davidangb, line 18: this test is currently failing - do you know why?

ashanhol: It passes locally and Josh thinks he has a fix in his PR

service/src/main/resources/application-data-plane.yml
ashanhol: Ok I made the change in logback-spring.xml, can you confirm the behavior does what we want? I just want to make absolutely sure I don't bungle that. If it looks good Ill remove the properties

service/src/test/java/org/databiosphere/workspacedataservice/annotations/DataPlaneDeploymentModeTest.java
ashanhol: changed.

service/src/test/java/org/databiosphere/workspacedataservice/annotations/ControlPlaneDeploymentModeTest.java
ashanhol: changed.

service/src/main/resources/application-data-plane.yml
davidangb: To test this: start WDS with either the "data-plane" or "control-plane" profile and inspect the logging. Do you get JSON logging in the control plane and plaintext logging in the data plane? What about when you run with the "local" profile?

service/src/main/resources/logback-spring.xml
```diff
@@ -22,28 +22,26 @@
   </appender>
```
davidangb, line 22: line 8 above is now obsolete and should be removed … we no longer use the `<springProperty>`:
```xml
<springProperty scope="context" name="deploymentMode" source="env.wds.deploymentMode" defaultValue="data-plane" />
```

davidangb (commented):
one comment inline on logback-spring.xml. Nice job getting the `<springProfile>` to work! Just gotta get that test passing, which might be waiting on Josh's PR.

service/src/test/java/org/databiosphere/workspacedataservice/annotations/DeploymentModeTestBase.java
ashanhol: fixed

service/src/main/resources/logback-spring.xml
ashanhol: lol i had that line highlighted to be deleted in my ide 😅

service/src/test/java/org/databiosphere/workspacedataservice/annotations/ControlPlaneDeploymentModeTest.java
jladieu: Ohh...let me clarify...I don't think I have a fix for _this_ failing test in my PR...this one I'm struggling with on my branches as well.

I think this test strategy might be a fools' errand in the profile-based world, but open to being proven wrong!

ashanhol: well now i have different failing tests to look into that were passing locally so im looking into it all now

jladieu: Just to be clear, I didn't see this test failing in your earlier CI run, and I was actually hoping _you_ had already solved this and I could piggyback on that solution with another PR after yours 😆 

The failure I did see earlier (and which I think I fixed for) was related to `workspaceId` being missing.

service/src/main/resources/application-data-plane.yml
ashanhol: WDS
```
2024-02-21T17:03:35.612-08:00  INFO 18407 --- [           main] o.d.w.startup.StartupConfig              : single-tenant workspace id: eeeefc4f-f388-40fe-a923-fe928b8b60ec
2024-02-21T17:03:35.642-08:00  INFO 18407 --- [           main] o.s.b.a.b.JobLauncherApplicationRunner   : Running default command line with: []
```
cWDS
```
{"@timestamp":"2024-02-21T17:11:55.436411443-08:00","@version":"1","message":"Registering Ehcache MBean javax.cache:type=CacheStatistics,CacheManager=file./home/adinas/workbench/terra-workspace-data-service/service/build/resources/main/ehcache.xml,Cache=tokenResolution","logger_name":"org.ehcache.jsr107.Eh107CacheManager","thread_name":"main","level":"INFO","level_value":20000}
{"@timestamp":"2024-02-21T17:11:57.543262455-08:00","@version":"1","message":"Exposing 3 endpoint(s) beneath base path ''","logger_name":"org.springframework.boot.actuate.endpoint.web.EndpointLinksResolver","thread_name":"main","level":"INFO","level_value":20000}
```

service/src/test/java/org/databiosphere/workspacedataservice/annotations/ControlPlaneDeploymentModeTest.java
ashanhol: Conclusion: Bug: https://broadworkbench.atlassian.net/browse/AJ-1653
The behavior we defined with `@Profile("!local")` on CollectionInitalizer is not really correct with how we're wanting to use profiles, and doesn't allow us to supersede the behavior with our platform annotations. Disabled the test until that behavior is fixed.

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
```diff
@@ -28,7 +28,8 @@
     properties = {
       "twds.instance.workspace-id=123e4567-e89b-12d3-a456-426614174000",
       "twds.instance.source-workspace-id=10000000-0000-0000-0000-000000000111",
-      "twds.pg_dump.useAzureIdentity=false"
+      "twds.pg_dump.useAzureIdentity=false",
+      "twds.instance.run-on-startup=false"
```
ashanhol, line 32: new property added to use instead of `@Profile("!local")` for CollectionInitalizer/modify during tests

jladieu: nit: can you put this up as the first param before the other two `twds.instance` ones, and then do the same ordering in all the other spots too?

jladieu (approved):

service/src/main/resources/application.yml
```diff
@@ -119,6 +118,7 @@ twds:
     # Workspace Id for launching instance
     workspace-id: ${WORKSPACE_ID:}
     source-workspace-id: ${SOURCE_WORKSPACE_ID:}
+    run-on-startup: true
```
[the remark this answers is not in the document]
ashanhol: won't it have to be renamed anyway when we rename "instance" to "collection"?

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
```diff
@@ -28,7 +29,8 @@
...
```
davidangb, line 33: Both this test and `CollectionInitializerCloneTest` pass for me with `twds.instance.run-on-startup=true`. Given that, does this property need to exist at all? I apologize if my previous feedback wasn't clear - I am suggesting removing the conditional entirely, and always running `CollectionInitializer` for both local and !local profiles. This would simplify things overall.

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
```diff
@@ -40,6 +42,7 @@ class RestoreServiceIntegrationTest {
   @Value("${twds.instance.workspace-id:}")
   private String workspaceId;
 
+  @BeforeEach
   @AfterEach
```
davidangb, line 46: do we really need both `@BeforeEach` and `@AfterEach`? That strikes me as a code smell. If we truly need to clear entries both before and after, could you use `@BeforeEach` and `@AfterAll`? (or `@BeforeAll` and `@AfterEach`, should be equivalent). Note that to have a non-static method (which is required to use beans) in a beforeAll/afterAll you will also have to add `@TestInstance(TestInstance.Lifecycle.PER_CLASS)`.

In my local testing, I see everything pass if I use just `@BeforeEach`, don't need anything else. I haven't tried this extensively though.

(this comment applies to `CollectionInitializerCloneTest` too)

ashanhol: Josh just did this too in one of his PRs- I think we might have a deeper problem of having cleanup problems in our integration tests that might be worth separately investigating. 
I can change it to before each and after all, but we can't get by with no after <anything> unless we move all our integration tests to cleaning up before running

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
ashanhol: The reason CollectionInializer had the !local line was so we could control it for certain cloning test scenarios. Using !local was kind of the wrong way to do that so this is the (more correct) workaround. Those tests fail if we dont have a way to control CollectionInitalizer not automatically getting configured on startup.

service/src/main/resources/application.yml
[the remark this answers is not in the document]
ashanhol: changed to initialize-collection-on-startup

service/src/integrationTest/java/org/databiosphere/workspacedataservice/service/RestoreServiceIntegrationTest.java
ashanhol: changed

yuliadub (approved):
the start up property makes sense to me based on the debugging session that I crashed with you and Josh 😄

davidangb (approved):
:+1: for the PR as-is. I do think there is a more dependency-injection-friendly way to achieve this which would eliminate the need for the `twds.instance.initialize-collection-on-startup` property, but I don't want to block the PR on it. I may take a crack at it sometime to illustrate.

PR #559, opened 2024-02-21
AJ-1585 - Add ability to write to and read from a google bucket

yuliadub:
Adding a class into WDS that can write and read files from google storage. This is not yet integrated to write or read Rawls Jsons - but that would be the next step after this change is merged. 

Reminder:


service/src/main/resources/application-control-plane.yml
```diff
@@ -5,5 +5,8 @@ twds:
     allow-virtual-collections: true
     # Should this WDS deployment require a valid $WORKSPACE_ID env var?
     require-env-workspace: false
+  bucket:
+    name: ${SERVICE_GOOGLE_BUCKET:uknown}
```
jladieu, line 9: typonit: uknown -> unknown here and the next line :)

yuliadub: why is spelling SO HARD? haha thanks for spotting it!

service/build.gradle
```diff
@@ -124,6 +127,9 @@ dependencies {
     testImplementation 'au.com.dius.pact.consumer:junit5:4.6.7'
     testImplementation 'org.junit-pioneer:junit-pioneer:2.2.0'
 
+    // google bucket mock
+    implementation 'com.google.cloud:google-cloud-nio'
```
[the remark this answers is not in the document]
yuliadub: I swear I tried this at first and it didnt work for some reason, but now it works! changed

service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorage.java
```diff
@@ -0,0 +1,54 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.UUID;
+import org.springframework.beans.factory.annotation.Value;
+
+public class GcsStorage {
+  private Storage storage;
+
+  @Value("${twds.bucket.name}")
```
jladieu, line 17: This would be a good spot to use the new properties constructs @davidangb  & I have been working on, but there's an outstanding [PR](https://github.com/DataBiosphere/terra-workspace-data-service/pull/561) that will affect how you use this stuff.  After that PR, you would be able to inject a `DataImportProperties` bean as a parameter in the constructor (instead of the `bucketName` & `projectId`), and ask that for the bucket info.  I can work with you to get this going in a followup PR, or we can try to squish it into this one :)

service/src/test/java/org/databiosphere/workspacedataservice/storage/GcsStorageTest.java
```diff
@@ -0,0 +1,32 @@
...
+  @Test
+  public void testCreateandGetBlobSimple() {
+    GcsStorage storage = new GcsStorage(mockStorage, bucketName, projectId);
+    String initialString = "text";
+    InputStream targetStream = new ByteArrayInputStream(initialString.getBytes());
+    BlobId blobId = storage.createGcsFile(targetStream);
+
+    assertEquals(bucketName, blobId.getBucket());
+
+    var contents = storage.getBlobContents(blobId.getName());
```
jladieu, line 29:
```suggestion
    String contents = storage.getBlobContents(blobId.getName());
```
Normally, I'm completely in favor of using `var` in tests to keep them terser, but since `contents` doesn't really self-describe the type the way some other variable names do, I recommend using a strong type declaration here.

service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorage.java
```diff
@@ -0,0 +1,54 @@
...
+    var blob = storage.get(blobId);
+    var contents = blob.getContent();
+    return new String(contents, StandardCharsets.UTF_8);
+  }
+
+  public BlobId createGcsFile(InputStream contents) {
+    BlobId blobId = BlobId.of(bucketName, UUID.randomUUID().toString());
+    storage.create(BlobInfo.newBuilder(blobId).build(), contents);
+    return blobId;
+  }
```
davidangb, line 53: if you're able to switch to importing the starter package instead of importing storage libs directly, you'll get access to higher-level utils like `GoogleStorageResource`, which means you can write methods like this (based off [their example](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-samples/spring-cloud-gcp-storage-resource-sample/src/main/java/com/example/WebController.java)):

```java
  public String writeGcsFile(String filename, InputStream contents) throws IOException {
    // create the GCS Resource
    GoogleStorageResource gcsResource = new GoogleStorageResource(
        this.storage, String.format("gs://%s/%s", this.bucketName, filename));
    // write the data to the resource
    try (OutputStream os = gcsResource.getOutputStream()) {
      contents.transferTo(os);
    }
    return gcsResource.getBlobName();
  }
```

davidangb (commented):
I tried out importing the [GCS starter](https://github.com/GoogleCloudPlatform/spring-cloud-gcp) locally, and I was able to get it to work - so my comments are about that … it would entail a bit of a rewrite, though

service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorage.java
yuliadub: yeah I was looking at that but was wondering if it made sense to take a dependency on what is basically their example code? could they possible change it or stop supporting it so in the future we'd have to see that and make appropriate changes? plus the started package has a lot more functionality that we don't need. 

I also wasnt sure if it'd be okay to use that package since their open source license is not MIT, or we'd have to apply for a [license ](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-samples/spring-cloud-gcp-storage-resource-sample/src/main/java/com/example/WebController.java#L8C9-L8C52)to use it? (but maybe Broad already has that now that I think of it)

but if all those concerns are actually mute, happy to switch to the starter package!

yuliadub: I will say their example that uses GoogleStorageResource seems more confusing to me if I am being honest, vs the simple code that the create method has now. With that said though, I am no expect in the GCP space so will follow guidance from you around what makes the most sense when it comes to GCP stuff 😄

davidangb: The overall [`spring-cloud-gcp` package](https://github.com/GoogleCloudPlatform/spring-cloud-gcp), which includes the starters, looks stable and supported to me, and the Apache license is fine. I don't feel like using the starter package would have any dependency on the example code - I'd just use the example code as an inspiration.

I think your opinion on the code being confusing is a valid opinion; it's not really about GCP architecture, it's about code readability. In my personal opinion using a higher-level tested/maintained library has benefit, and to _me_ the code readability is just about the same … but I think it's a valid reason for you to push back.

yuliadub: I dont have very strong feelings about it though, the current implementation may look easier to me since I was the one to put it together (I think the starter code now looks to make more sense, once I have done work with buckets but when I first looked at it, it wasnt as readable, hence why I did not end up using it)

@jladieu, @ashanhol, @nawatts - what do you all think?

nawatts: Didn't look too closely, but agreed that [spring-cloud-gcp-storage](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/tree/main/spring-cloud-gcp-storage) looks like a proper library vs sample code (though sample code for using it is in the same monorepo).

But it doesn't look like there's all that much code here to simplify, so I don't have a strong opinion either way.

jladieu: If it's not a big lift, I'd say trying to use the spring-cloud package David recommended is a good idea.  While the actual code to accomplish the same thing isn't much different, I think preferring a dependency on a package focused on spring will potentially have better support & integration with spring going forward; thus it may get spring-integrated features, release / bugfix / dependency patches, etc.

jladieu:
FYI, Design feedback here:

tl;dr: I think your initial instinct to just turn around and call the GCS storage class (and pubsub) directly might be just fine rather than trying to tie into the `Consumer<String>` stuff I started building out.

We discussed in zoom how you might hang this off the end of the `RawlsRecordSink` by providing an implementation of `Consumer<String>` and using Spring's wiring to attach them together -- after thinking more about that approach, I just want to call out I don't feel married to it.  The driver for doing the approach I started proving out with `RawlsRecordSink` would be loose coupling between the `RecordSink` class and the downstreams that need it.  However, `RawlsRecordSink` exists for the Rawls integration, so I think a simpler/direct composition with GCS Storage and PubSub might be more appropriate, and the `Consumer<String>` stuff is probably overkill.  As long as there's a model by which we can test `RawlsRecordSink`'s JSON output in isolation, _how_ we integrate with GCS & PubSub doesn't _have_ to be so loosely coupled.

service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorage.java
yuliadub: sounds good, I went ahead and did the refactor!

service/build.gradle
```diff
@@ -91,6 +91,9 @@ dependencies {
     implementation "org.apache.parquet:parquet-avro:$parquet_mr_version"
     implementation "org.apache.parquet:parquet-hadoop:$parquet_mr_version"
 
+    implementation platform('com.google.cloud:libraries-bom:26.30.0')
+    implementation 'com.google.cloud:google-cloud-storage'
+
```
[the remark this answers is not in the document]
yuliadub: I had to keep these to get the nio library to work....just for the testImplementation though, still trying to look and see if there is another way but havent figured it out yet

service/src/test/java/org/databiosphere/workspacedataservice/storage/GcsStorageTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.contrib.nio.testing.LocalStorageHelper;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.UUID;
+import org.junit.jupiter.api.Test;
+
+public class GcsStorageTest {
```
davidangb, line 14: remove `public` (feedback from Sonar)

service/src/test/java/org/databiosphere/workspacedataservice/storage/GcsStorageTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+  private String bucketName = "test-bucket";
+
+  private String projectId = "test-projectId";
+
+  private final Storage mockStorage = LocalStorageHelper.getOptions().getService();
+
+  @Test
+  public void testCreateandGetBlobSimple() throws IOException {
```
davidangb, line 23: remove `public` (feedback from Sonar)

service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorage.java
```diff
@@ -0,0 +1,60 @@
...
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageOptions;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.charset.Charset;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.util.StreamUtils;
+
+public class GcsStorage {
```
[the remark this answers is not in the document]
yuliadub: I have opinion on if this should be a bean, but I'd like to handle that in a separate PR

service/src/test/java/org/databiosphere/workspacedataservice/storage/GcsStorageTest.java
yuliadub: why did sonar showup as green??? 😄 I didnt even think to look at it

service/build.gradle
```diff
@@ -124,6 +126,11 @@ dependencies {
     testImplementation 'au.com.dius.pact.consumer:junit5:4.6.7'
     testImplementation 'org.junit-pioneer:junit-pioneer:2.2.0'
 
+    // google bucket mock
+    testImplementation 'com.google.cloud:google-cloud-nio'
+    testImplementation platform('com.google.cloud:libraries-bom:26.30.0')
+    testImplementation 'com.google.cloud:google-cloud-storage'
```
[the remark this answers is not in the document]
yuliadub: ah of course it would be this simple...!

service/src/main/resources/application.yml
```diff
@@ -66,6 +66,11 @@ spring:
     jcache:
       config: classpath:ehcache.xml
 
+  cloud:
+    gcp:
+      storage:
+        enabled: false
```
ashanhol, line 72: is this something we could put in the control-plane/data-plane ymls specifically? I thought we'd want it enabled for control-plane?

yuliadub: this is a spring property that would need to be disabled on both since from my understanding this happens pre any of the our code that runs.

ashanhol: Ohhh is this an automatic spring thing that we have to configure because we've included the com.google.cloud:spring-cloud-gcp-starter-storage package?

ashanhol: i thought it was a thing you were doing ok. that makes more sense

service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorage.java
```diff
@@ -0,0 +1,58 @@
...
+  // under
+  public GcsStorage() throws IOException {
+    GoogleCredentials credentials = GoogleCredentials.getApplicationDefault();
+
+    StorageOptions storageOptions =
+        StorageOptions.newBuilder().setProjectId(projectId).setCredentials(credentials).build();
+    this.storage = storageOptions.getService();
+  }
+
+  // primary here for tests, but also allows this class to be used with values other than
```
ashanhol, line 32: [nit] did you mean "primarily"?

yuliadub: im really failing on the spelling this week 😢

service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorage.java
yuliadub: i'm thinking tackle it in the next PR vs growing this one!

ashanhol (approved):
dunno much about the nuances of google buckets but seems ok to me

davidangb (approved):
thank you!

service/src/main/resources/application-control-plane.yml
```diff
@@ -7,3 +7,6 @@ twds:
     require-env-workspace: false
   data-import:
     batch-write-record-sink: "rawls"
+    rawls-bucket:
+      name: ${SERVICE_GOOGLE_BUCKET:unknown}
+      projectId: ${SERVICE_GOOGLE_PROJECT:unknown}
```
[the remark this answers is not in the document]
jladieu: (No action needed for @yuliadub), but for my own understanding (contemplating similar things), is it easy to positionally shuffle things around in this file, or does it need to be coordinated with terra-helmfile somehow?

jladieu (approved):

service/src/main/resources/application-control-plane.yml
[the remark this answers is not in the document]
yuliadub: easy to shuffle, just the name of the env variable is what matters

PR #560, opened 2024-02-21
AJ-1649: add @DirtiesContext to tests that need it

davidangb:
add `@DirtiesContext` to more unit tests to ensure we don't create too many `DataSource` beans and therefore run out of Postgres connections

Out of scope: removing `@DirtiesContext` from any tests that don't need it. This could improve test timing, but I would like to do that work in a later PR.


jladieu (approved):
Thank you so much!

ashanhol (approved):

PR #561, opened 2024-02-21
AJ-1650 Expose TwdsProperties child props as Beans.

jladieu:
[AJ-1650](https://broadworkbench.atlassian.net/browse/AJ-1650): This allows classes to autowire only the properties they need, rather than getting _all_ the properties and having to traverse the hierarchy.

It also allows more rapid failure due to misconfigured beans (rather than having to tiptoe around the fact that some of the sub-properties of `TwdsProperties` _might_ be null, so we can be less defensive about null checking and can assume for the most part the beans will be present.

[AJ-1650]: https://broadworkbench.atlassian.net/browse/AJ-1650?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service/src/test/java/org/databiosphere/workspacedataservice/service/CollectionServiceTest.java
```diff
@@ -24,25 +19,17 @@
 
 @ActiveProfiles(profiles = {"mock-sam", "mock-collection-dao"})
 @DirtiesContext
-@SpringBootTest(
```
jladieu, line 27: This test started to fail with misconfigured beans and I had a hard time finding an explanation, so instead I just tried eliminating the one thing that made this test look not like the others, which was the very specific set of classes in the list here...this did the trick, paired with the fixes down below.

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
```diff
@@ -150,17 +149,15 @@ public WorkspaceId getWorkspaceId(CollectionId collectionId) {
       rowWorkspaceId = collectionDao.getWorkspaceId(collectionId);
       // as of this writing, if a deployment allows virtual collections, it is an error if the
       // collection DOES exist.
-      if (twdsProperties.getTenancy() != null
-          && twdsProperties.getTenancy().getAllowVirtualCollections()) {
+      if (tenancyProperties.getAllowVirtualCollections()) {
         throw new CollectionException("Expected a virtual collection");
       }
     } catch (EmptyResultDataAccessException e) {
       // does this deployment allow for virtual collections?
```
jladieu, line 156: This comment block is responsible for the one outstanding sonar warning: https://sonarcloud.io/project/issues?resolved=false&pullRequest=561&id=DataBiosphere_workspace-data-service&open=AY3N3VlKcgsFMb-SDeBv

However I'm leaving it in here as uncommenting is covered by the annotated ticket.  I'm hoping this PR at least makes that ticket a lot easier as fail-fast semantics seem to be readily happening now.

service/src/main/java/org/databiosphere/workspacedataservice/config/TenancyProperties.java
```diff
@@ -5,8 +5,8 @@
  * deployment.
  */
 public class TenancyProperties {
-  private Boolean allowVirtualCollections;
-  private Boolean requireEnvWorkspace;
+  private boolean allowVirtualCollections;
```
jladieu, line 8: Flipping to a primitive boolean is in response to an interesting Sonar issue about preferring primitives to boxed booleans: Per sonar java:S5411: Avoid using boxed "Boolean" types directly in
boolean expressions

I think the reasoning here is the boxed version of a Boolean exposes the code to nullability concerns, whereas the primitive can't be null.

service/src/main/java/org/databiosphere/workspacedataservice/config/TwdsProperties.java
```diff
@@ -1,21 +1,27 @@
...
+  private InstanceProperties instance;
+  private String startupToken;
+  private TenancyProperties tenancy;
 
   public DataImportProperties getDataImport() {
     return dataImport;
   }
 
+  @Bean
+  public DataImportProperties dataImportProperties() {
```
jladieu, line 21: I realize exposing the properties as `@Bean`s makes this file more verbose, and I suspect there is a way to accomplish this without adding a redundant set of getters, but I wanted to put this up as I think it makes progress preventing the propagation of TwdsProperties throughout the codebase.

Things I want to still try and which I believe should work, but am deferring for now would be:
- switch over to using a `record` instead of class to represent the properties, so that they are immutable, which allows us to...
- switch over to using constructor injection if possible, which might obviate the need for having setters and redundant getters

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresCollectionDao.java
```diff
@@ -6,33 +6,30 @@
...
   public PostgresCollectionDao(
-      NamedParameterJdbcTemplate namedTemplate, TwdsProperties twdsProperties) {
+      NamedParameterJdbcTemplate namedTemplate, InstanceProperties instanceProperties) {
     this.namedTemplate = namedTemplate;
-    // if we have a valid workspaceId, save it to a local var now
-    if (twdsProperties.getInstance() != null
-        && twdsProperties.getInstance().getWorkspaceUuid() != null) {
-      this.workspaceId = twdsProperties.getInstance().getWorkspaceUuid();
-    }
+    this.workspaceId = instanceProperties.getWorkspaceUuid();
```
jladieu, line 32: Deferring for now, but I think InstanceProperties would benefit from having a more strongly typed and null-safe support for `WorkspaceId` -- eg: rather than allowing `workspaceId` or `workspaceUuid` to be returned as null, provide a very strongly typed and simply interface of:

`@NonNull WorkspaceId getWorkspaceId()`

And hide the notion of a `workspaceIdString` or `workspaceUuid`

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RecordSinkFactory.java
```diff
@@ -54,20 +53,15 @@ public void setJsonConsumer(@RawlsJsonConsumer Consumer<String> jsonConsumer) {
   //   ignored for RecordSinkMode.WDS.  In this case, it might be worth adding support for omitting
   //   the prefix as part of supporting the prefix assignment.
   public RecordSink buildRecordSink(UUID collectionId, String prefix) {
-    if (twdsProperties.getDataImport() == null) {
```
jladieu, line 57: Can't instantiate this object without a legit `DataImportProperties`, so there's no reason to check anymore.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresCollectionDao.java
davidangb: I agree! My only concern here is that if the `$WORKSPACE_ID` env var has something invalid, we are able to throw/log a helpful error instead of a cryptic low-level Spring validation message.

service/src/test/java/org/databiosphere/workspacedataservice/service/CollectionServiceTest.java
davidangb: does it need a "data-plane" profile?

service/src/main/java/org/databiosphere/workspacedataservice/config/TwdsProperties.java
```diff
@@ -40,6 +51,11 @@ public TenancyProperties getTenancy() {
     return tenancy;
   }
 
+  @Bean
+  public TenancyProperties tenancyProperties() {
+    return getTenancy();
+  }
+
```
davidangb, line 58: Would you consider taking this even further? I'm envisioning: 
* do away with the `TwdsProperties` class entirely
* move the `@ConfigurationProperties` annotation down onto `DataImportProperties`, `InstanceProperties`, `TenancyProperties` and specify the right prefixes, e.g. `@ConfigurationProperties(prefix = "twds.dataImport")`, `@ConfigurationProperties(prefix = "twds.tenancy")`

The only place I see `TwdsProperties` used right now is in the `TwdsPropertiesControlPlaneTest` and `TwdsPropertiesDataPlaneTest` tests, and it should be easily deletable from there.

davidangb (commented):
One suggestion inline.

I'm also curious about the `CollectionServiceTest` test, I would think this PR shouldn't require changes there (except possibly requiring a profile)

service/src/test/java/org/databiosphere/workspacedataservice/service/CollectionServiceTest.java
jladieu: Yes, and it gets this from extending `TestBase`, so I think the failure was due to not getting _all_ the context required to bootstrap all the relevant dependency beans.

service/src/main/java/org/databiosphere/workspacedataservice/config/TwdsProperties.java
jladieu: I'm considering it, but let me make a case to hang onto it further...I actually _like_ that there's a single entrypoint for all this config stuff, and one place where an engineer to look to know how to configure it, and copy/paste from to add more.  I also like that the tests have this single starting point, from which we can verify all our assertions about how the props graph hangs together, because it also addresses my other concern which is proliferation of test classes that branch based on profile.  So for now, I feel like this solves for both problems:

- production code isn't tightly coupled due to everyone knowing about TwdsProperties
- test code isn't coupled either, as individual tests can autowire _just_ what they need
- test code for the config _is_ coupled, and uses this single entrypoint, because I think this is where we want to make front-and-center our assumptions about which profile we're running with and how it impacts all the downstream parts of the properties tree

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresCollectionDao.java
jladieu: Let's see if I can address this in the followup.  My hunch is that we can use Spring[ property validation](https://reflectoring.io/validate-spring-boot-configuration-parameters-at-startup/) to get some better messaging...although FWIW right now the messaging is pretty obvious but buried in a very deep stacktrace.

davidangb (approved):

service/src/test/java/org/databiosphere/workspacedataservice/service/CollectionServiceTest.java
```diff
@@ -24,25 +19,17 @@
...
-    })
+@SpringBootTest
 class CollectionServiceTest extends TestBase {
 
   @Autowired private CollectionService collectionService;
   @Autowired private CollectionDao collectionDao;
 
   private static final UUID COLLECTION = UUID.fromString("111e9999-e89b-12d3-a456-426614174000");
 
+  @BeforeEach
```
nawatts, line 30: Do we need to do this both before and after each test?

nawatts (approved):

service/src/test/java/org/databiosphere/workspacedataservice/service/CollectionServiceTest.java
jladieu: Unfortunately I found this to be necessary because there was some kind of pollution between the tests happening.  It's weird but I thought it was localized enough that it'd be ok for now and I didn't want to chase it down.

PR #564, opened 2024-02-22
AJ-1649: extraneous dirties context

davidangb:


yuliadub (approved):
I think it's worth it to make it really clear when texts can reuse context vs not now vs waiting or not doing it, and have that be consistent throughout.

jladieu (approved):
I love this, and I love the PR description :)

PR #566, opened 2024-02-22
AJ-1625: swagger UI for control plane vs data plane

davidangb:
* add a new `cwds-api-docs.yaml` OpenAPI spec which contains only the cWDS-enabled endpoints
* move `…resources/static/swagger/swagger-ui.html` to `…resources/swagger-ui-data-plane.html`, so it can't be served as a static resource anymore
* add `…resources/swagger-ui-control-plane.html`, with slightly different config than the data plane html
* add a new `SwaggerUiConfig` which reads the appropriate html file based on control-plane vs data-plane profile and stashes the html into a String
* add a new `SwaggerUiController` RestController which responds at /swagger/swagger-ui.html and responds with the stashed html
* add redirects from `/`, `/swagger`, and `/swagger/` to /swagger/swagger-ui.html for convenience


service/src/main/java/org/databiosphere/workspacedataservice/controller/SwaggerUiController.java
```diff
@@ -0,0 +1,51 @@
...
+
+    String content;
+
+    if (environment.matchesProfiles("control-plane")) {
+      // replace the title
+      content =
+          template.replace("${TITLE}", "cWDS API").replace("${APISPEC}", "cwds-api-docs.yaml");
+    } else {
+      content = template.replace("${TITLE}", "WDS API").replace("${APISPEC}", "openapi-docs.yaml");
+    }
```
davidangb, line 47: I thought about moving these strings into a config file (application-control-plane.yml, application-data-plane.yml), but … it felt like overkill.

Another option is to have two separate .html files, pre-populated with the correct values, and serve the appropriate one based on control-plane vs data-plane profiles. That would be more performant … but this does not feel like a particularly performance-sensitive spot.

service/src/main/resources/static/swagger/cwds-api-docs.yaml
```diff
@@ -0,0 +1,53 @@
...
+  /job/v1/instance/{instanceUuid}:
+    $ref: 'apis-v1.yaml#/paths/~1job~1v1~1instance~1{instanceUuid}'
+
+  ##############################
+  # General WDS Information APIs
+  ##############################
+  /status:
+    $ref: 'openapi-docs.yaml#/paths/~1status'
+  /version:
+    $ref: 'openapi-docs.yaml#/paths/~1version'
```
davidangb, line 53: as discussed over in https://github.com/DataBiosphere/terra-workspace-data-service/pull/572#discussion_r1501213706, we may want the Capabilities API at some point; we can add that later if we want it.

service/src/main/resources/swagger-ui-control-plane.html
```diff
@@ -0,0 +1,66 @@
...
+      plugins: [
+        SwaggerUIBundle.plugins.DownloadUrl,
+        cleanupPlugin
+      ],
+      layout: 'StandaloneLayout',
+      displayOperationId: true,       // show generated-client method names for each endpoint
+      defaultModelsExpandDepth: -1,   // hide the huge list of schemas under the routes
+      defaultModelRendering: 'model', // schema has the descriptions, unlike the example value
+      defaultModelExpandDepth: 2,     // affects the schema shown for a request or response
+      displayRequestDuration: true,   // show timing
```
davidangb, line 57: compared to the data plane HTML, this is missing: `docExpansion: 'none', filter: true`. Those features seemed not useful with such a small set of APIs in the control plane.

service/src/main/java/org/databiosphere/workspacedataservice/controller/SwaggerUiConfig.java
```diff
@@ -0,0 +1,40 @@
...
+
+  @DataPlane
+  @Bean(name = "swaggerHtml")
+  public String dataPlaneSwaggerHtml(
+      @Value("classpath:swagger-ui-data-plane.html") Resource swaggerUiDataPlaneResource)
+      throws IOException {
+    return StreamUtils.copyToString(
+        swaggerUiDataPlaneResource.getInputStream(), StandardCharsets.UTF_8);
+  }
+}
```
davidangb, line 40: I also considered having a single HTML file containing markers like `${TITLE}`, then performing string-replacement inside the controller. But when I got to the point of setting different flags in the swagger-ui javascript (e.g. removing `docExpansion: 'none', filter: true`), that got unweildy.

service/src/main/java/org/databiosphere/workspacedataservice/controller/SwaggerUiController.java
```diff
@@ -0,0 +1,39 @@
...
+public class SwaggerUiController {
+  /**
+   * holds the HTML content of the swagger-ui.
+   *
+   * @see SwaggerUiConfig
+   */
+  final String swaggerUiContent;
+
+  public SwaggerUiController(@Qualifier("swaggerHtml") String swaggerUiContent) {
+    this.swaggerUiContent = swaggerUiContent;
```
ashanhol, line 25: I think I'm missing something Spring Magic related. SwaggerUiConfig returns data or controller depending on the annotation but how does that hook up here?

service/src/main/resources/static/swagger/cwds-api-docs.yaml
```diff
@@ -0,0 +1,50 @@
+openapi: 3.0.3
+
+info:
+  title: cWDS API Spec
+  version: v1
+  description: |
+    cWDS - the (c)control-plane (W)orkspace (D)ata (S)ervice - is responsible for importing
```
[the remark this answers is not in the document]
davidangb: fixed :)

service/src/main/java/org/databiosphere/workspacedataservice/controller/SwaggerUiController.java
davidangb: `SwaggerUiConfig` produces a bean of type String. The bean's name is "swaggerHtml", as specified by:
```
@Bean(name = "swaggerHtml")
```

Here, Spring uses dependency injection to supply the String argument to this constructor. The `@Qualifier("swaggerHtml")` tells it to look for a bean named "swaggerHtml".

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/controller/SwaggerUiConfig.java
```diff
@@ -0,0 +1,40 @@
...
+  @Bean(name = "swaggerHtml")
+  public String controlPlaneSwaggerHtml(
+      @Value("classpath:swagger-ui-control-plane.html") Resource swaggerUiControlPlaneResource)
+      throws IOException {
+    return StreamUtils.copyToString(
+        swaggerUiControlPlaneResource.getInputStream(), StandardCharsets.UTF_8);
+  }
+
+  @DataPlane
+  @Bean(name = "swaggerHtml")
```
jladieu, line 33: ahh neat, I didn't know you could use `@Qualifier` with `@Bean(name=...)` -- I thought you had to use `@Qualifier` in both places!

service/src/main/resources/swagger-ui-control-plane.html
jladieu: Would you consider listing them in here with their default values, and commenting them with the same explanation inline?  I'm thinking of at some point doing a visual diff and it'd be helpful to aid knowing what configs are different and by what values.

jladieu (approved):
This looks great!  One optional suggestion about explicitly calling out the configs that are different, but LGTM either way.

service/src/main/resources/swagger-ui-control-plane.html
davidangb: added these back in!

service/src/main/java/org/databiosphere/workspacedataservice/controller/SwaggerUiController.java
ashanhol: oh wow thats really cool!

ashanhol (approved):
It's cool to see a UI spring example

PR #568, opened 2024-02-23
AJ-1649: remove extraneous comma

davidangb:
This should address intermittent errors in our e2e tests. We just had one last night, and the error message was:
```
--name: "wds-8017616954-1-dev," is not a valid environment name: environment name
must match regular expression \\A[a-z][a-z0-9]*(-[a-z0-9]+)*\\z"
```
if you look very carefully at the error message, you'll see that the name `"wds-8017616954-1-dev,"` _has a trailing comma_ which makes it invalid.

The reason this is so intermittent is that it only happens in the retry case of creating a bee.

Screenshot 2-23-2024 at 10 12 AM


nawatts (approved):

jladieu:
🤯

PR #569, opened 2024-02-23
[AJ-1638] Do not require statuses query parameter for list jobs endpoint

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1638

Currently, listing jobs without providing a `statuses` query parameter results in a 500 errors. This fixes that so that the query parameter is not required.

This also refactors `JobDao.getJobsForCollection` to make the `statuses` argument an `Optional`. I think that more clearly represents the intent of an optional statuses filter vs using an empty list to mean return all statuses.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/JobController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/JobController.java
@@ -3,6 +3,7 @@ package org.databiosphere.workspacedataservice.controller;
 import static org.databiosphere.workspacedataservice.generated.GenericJobServerModel.StatusEnum;
 
 import java.util.List;
+import java.util.Optional;
 import java.util.UUID;
 import org.databiosphere.workspacedataservice.generated.GenericJobServerModel;
 import org.databiosphere.workspacedataservice.generated.JobApi;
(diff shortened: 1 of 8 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/controller/JobController.java
```diff
@@ -49,7 +50,9 @@ public ResponseEntity<List<GenericJobServerModel>> jobsInInstanceV1(
       }
     }
     List<GenericJobServerModel> jobList =
-        jobService.getJobsForCollection(CollectionId.of(instanceUuid), statuses);
+        jobService.getJobsForCollection(
+            CollectionId.of(instanceUuid),
+            statuses == null ? Optional.empty() : Optional.of(statuses));
```
nawatts, line 55: @ashanhol this was the main thing: based on the error message in the ticket, `statuses` is null if the query parameter isn't provided.

ashanhol: Ahh gotcha, for some reason I thought the fix had something to do with the open API itself, not just handling the optional case. Thank you for the fix!!

ashanhol (approved):
thank you 🙏🙏

service/src/main/java/org/databiosphere/workspacedataservice/controller/JobController.java
davidangb, line 55: your call …. you could also do 
```java
Objects.requireNonNullElse(
                statuses, Arrays.stream(StatusEnum.values()).map(StatusEnum::getValue).toList()));
```
here, and pass all the possible statuses if the user didn't supply any. That way, you can avoid handling `Optional`s everywhere below the call. But I also see how using `Optional` makes the semantics clear.

PR #570, opened 2024-02-23
AJ-1629 Improve configuration fail-fast behavior

jladieu:
This PR's primary intention is to introduce fail-fast behavior to the Spring runtime when our assumptions about the deployment mode are violated.  This is best done as early as possible in the bean configuration lifecycle, and so happens as part of initializing `TwdsProperties`.  Due to the very central role of bean config in the codebase (and the way something completely unrelated can fail to run when beans aren't configured properly), and the fact that tenancy + `WORKSPACE_ID` + data-plane vs. control-plane are somewhat tightly coupled, this PR ended up absorbing much more scope than initially intended, from the following three tickets:

[AJ-1629](https://broadworkbench.atlassian.net/browse/AJ-1629): Require `control-plane` or `data-plane` to be set throughout. (_Also: Default to `data-plane` in tests_)).
[AJ-1630](https://broadworkbench.atlassian.net/browse/AJ-1630): Respect `allow-virtual-collections` config setting
[AJ-1650](https://broadworkbench.atlassian.net/browse/AJ-1650): Clean up properties classes

New fail-fast mechanics introduced:

`WORKSPACE_ID`:
- If in single-tenant mode, `WORKSPACE_ID` is required, and WDS should crash instead of allowing any code to run without it.  This is enforced by restricting all access to this environment variable through the `InstanceProperties.workspaceId()` accessor, which is designed to never return `null`.  In fact, `InstanceProperties` can't be resolved as a Bean at all _unless_ they have a valid `WorkspaceId`, so this is enforced in depth*** when injected via `InstanceProperties`.
- If in multi-tenant mode, `InstanceProperties` cannot be auto-configured, and so all beans that depend on it need to be marked as running in `@DataPlane` only.  *Note: this might need to be adjusted to look instead at tenancy-properties if we move to WDS in the LZ, see TODO below*
- A convenience `@Bean` for the WorkspaceId annotated as `@SingleTenant` is provided for direct injection, but it is also subject to the same constraints.

`allow-virtual-collections`:
- This is barely used at the moment and enforced by configuring mutually exclusive `@DataPlane` `PostgresCollectionDao` and `@ControlPlane` `VirtualWorkspaceIdDao` beans rather than honoring the `TenancyProperties`; however, in tandem with the `WORKSPACE_ID` enforcement, a bunch of logic around the virtual collections can be removed in favor of trusting that the presence/absence of `InstanceProperties` indicates whether we're in a singleTenant deployment or not.  I think this may be addressed by [AJ-1655](https://broadworkbench.atlassian.net/browse/AJ-1655)

TODO:
- [ ] The coupling between deploymentMode + tenancy + instance properties is probably exacerbated by using the `@DataPlane` and `@ControlPlane` annotations to selectively enable/disable beans.  We might be able to reduce this coupling by correctly honoring the booleans in tenancy and instance props, maybe with `@ConditionalOnProperty`?
- [ ] ***Direct access to the `workspace-id` using `@Value('twds.instance.workspace-id')` throughout the codebase still circumvents the validation and enforcement that it's non-null, non-blank, and a valid UUID, followup is required to eliminate these references, which will cause unexpected misbehavior in `control-plane` ([AJ-1656](https://broadworkbench.atlassian.net/browse/AJ-1656))

[AJ-1650]: https://broadworkbench.atlassian.net/browse/AJ-1650?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AJ-1656]: https://broadworkbench.atlassian.net/browse/AJ-1656?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AJ-1655]: https://broadworkbench.atlassian.net/browse/AJ-1655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service/src/main/resources/application-data-plane.yml
```diff
@@ -2,6 +2,10 @@ env:
   wds:
     deploymentMode: data-plane
 twds:
+  instance:
```
jladieu, line 5: This PR has led me to the conclusion that `instance` belongs as a child of `tenancy`, owing to the fact that it only has something reasonable to configure when a `WORKSPACE_ID` is present, which will only be true in the single-tenant (non-virtual) mode.  Ticket for this: https://broadworkbench.atlassian.net/browse/AJ-1655

service/src/test/java/org/databiosphere/workspacedataservice/config/TwdsPropertiesControlPlaneTest.java
```diff
@@ -16,7 +16,6 @@
 class TwdsPropertiesControlPlaneTest {
   @Autowired TwdsProperties twdsProperties;
   @Autowired DataImportProperties dataImportProperties;
-  @Autowired InstanceProperties instanceProperties;
```
jladieu, line 19: `instanceProperties` can no longer be wired to control-plane, since it has a hard dependency on `WORKSPACE_ID`

service/src/main/java/org/databiosphere/workspacedataservice/dao/WorkspaceIdDao.java
```diff
@@ -0,0 +1,14 @@
+package org.databiosphere.workspacedataservice.dao;
+
+import org.databiosphere.workspacedataservice.service.model.exception.MissingObjectException;
+import org.databiosphere.workspacedataservice.shared.model.CollectionId;
+import org.databiosphere.workspacedataservice.shared.model.WorkspaceId;
+
+public interface WorkspaceIdDao {
```
jladieu, line 7: I separated this interface out so that this bean can be provided in both `@ControlPlane` (`VirtualWorkspaceIdDao`) and `@DataPlane` (`PostgresCollectionDao`).

The alternative approach I started with was to extend `CollectionDao` with `VirtualCollectionDao` and it _only_ implemented the `getWorkspaceId` method, all other methods threw `UnsupportedOperationException`.

And the original approach that both of these options replace was to just have `CollectionService` deal with this concern as an internal implementation detail, with all the extra checks to verify that there's nothing in the database when there shouldn't be.  I thought this encapsulated approach might let us sidestep all those edge cases, as there will be no way to write any collections into the DB as the `PostgresCollectionDao` won't even be loaded into runtime.

service/src/main/java/org/databiosphere/workspacedataservice/config/TwdsProperties.java
```diff
@@ -9,54 +14,63 @@
...
-    return dataImport;
-  }
 
   @Bean
   public DataImportProperties dataImportProperties() {
-    return getDataImport();
+    return dataImport;
   }
 
-  public void setDataImport(DataImportProperties dataImport) {
```
jladieu, line 25: The setters can be package protected since they are only used by Spring.

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
```diff
@@ -118,66 +128,66 @@ public void validateCollection(UUID collectionId) {
...
-    }
+  public WorkspaceId getWorkspaceId(CollectionId collectionId) throws MissingObjectException {
+    // look up the workspaceId for this collection in the collection table; in the case of a virtual
+    // collection, this will just echo the same value as what was passed in
+    WorkspaceId workspaceId = workspaceIdDao.getWorkspaceId(collectionId);
 
-    // safety check: if we found a workspace id in the db, it indicates we are in a data-plane
-    // single-tenant WDS. Verify the workspace matches the $WORKSPACE_ID env var.
-    // we must remove this check in a future multi-tenant WDS.
-    if (rowWorkspaceId != null && !rowWorkspaceId.equals(workspaceId)) {
```
jladieu, line 169: we can remove the null check because `getWorkspaceId` now throws an exception if it's missing

service/src/main/java/org/databiosphere/workspacedataservice/startup/CollectionInitializerConfig.java
jladieu, line 1: Making `CollectionInitializerBean` a `@Component` so I could mark it as `@DataPlane` had the nice side effect of letting me get rid of this class which was just a pass-through.

service/src/test/java/org/databiosphere/workspacedataservice/startup/CollectionInitializerBeanTest.java
```diff
@@ -180,27 +157,12 @@ void sourceWorkspaceIDNotProvided() {
     assertFalse(cloneMode);
   }
 
-  @Test
```
jladieu, line 183: The failure modes tested by these tests are no longer possible due to the stronger typing of `isInCloneMode`, backed by the faster failure if the workspace ID isn't formatted correctly.

service/src/main/java/org/databiosphere/workspacedataservice/dao/WorkspaceIdDao.java
jladieu: As part of [AJ-1656](https://broadworkbench.atlassian.net/browse/AJ-1656) there will be at least several more places that are going to need `WorkspaceIdDao` but won't need the full `CollectionDao`, so after starting to scout out that effort I was glad to have separated out this interface, even though it feels a little like overkill for how it's used in this PR.

jladieu:
I'm going to take this back to the drawing board and try to shrink it's scope to improve reviewability and avoid risk of regressions.

jladieu:
Breakaway PR to clean up props and start introducing `WORKSPACE_ID` guarantees: https://github.com/DataBiosphere/terra-workspace-data-service/pull/583

service/src/test/java/org/databiosphere/workspacedataservice/annotations/AnnotatedApisTest.java
```diff
@@ -6,20 +6,28 @@
...
 import org.springframework.web.bind.annotation.RestController;
 
 @ActiveProfiles(
     value = {"control-plane", "data-plane"},
     inheritProfiles = false)
+@DirtiesContext
 @SpringBootTest
 class AnnotatedApisTest extends TestBase {
 
   @Autowired private ApplicationContext context;
```
jladieu, line 26: Trailmarker comment for crosslink.

service/src/test/java/org/databiosphere/workspacedataservice/annotations/AnnotatedApisTest.java
```diff
@@ -6,20 +6,28 @@
...
+  // Since we're running with both control-plane and data-plane profiles simultaneously, Spring
```
jladieu, line 27: Trailmarker comment for crosslink.

jladieu:
Closing for now, subsumed by smaller PRs.

PR #572, opened 2024-02-23
Enforce Platform for @RestControllers

ashanhol:
Ticket: https://broadworkbench.atlassian.net/browse/AJ-1562

This PR adds `@DataPlane` and or `@ControlPlane` to our `@RestController`s and adds a test to make sure at least one of those is on all of our `@RestController`s

-----------------------
Reminder:

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/CapabilitiesController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/CapabilitiesController.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.IOException;
 import java.io.InputStream;
+import org.databiosphere.workspacedataservice.annotations.DeploymentMode.DataPlane;
 import org.databiosphere.workspacedataservice.generated.CapabilitiesApi;
 import org.databiosphere.workspacedataservice.generated.CapabilitiesServerModel;
 import org.slf4j.Logger;
@@ -13,6 +14,7 @@ import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.RestController;
 
 /** Controller for capabilities-related APIs */
+@DataPlane
 @RestController
 public class CapabilitiesController implements CapabilitiesApi {
 
--- a/service/src/test/java/org/databiosphere/workspacedataservice/annotations/AnnotatedApisTest.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/annotations/AnnotatedApisTest.java
@@ -5,6 +5,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 import java.util.Optional;
 import org.databiosphere.workspacedataservice.annotations.DeploymentMode.ControlPlane;
 import org.databiosphere.workspacedataservice.annotations.DeploymentMode.DataPlane;
+import org.databiosphere.workspacedataservice.common.TestBase;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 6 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/annotations/AnnotatedApisTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+@ActiveProfiles({"control-plane", "data-plane"})
+@SpringBootTest
+class AnnotatedApisTest {
+
+  @Autowired private ApplicationContext context;
+
+  @Test
+  void controllersMustHaveDeploymentAnnotation() {
+    String[] beanNames = context.getBeanNamesForAnnotation(RestController.class);
```
ashanhol, line 23: Note: we're not using `@Controller` anywhere in our code (which makes sense because `@RestController` implements `@Controller`) so I've opted to only look for that.

service/src/main/java/org/databiosphere/workspacedataservice/controller/CapabilitiesController.java
```diff
@@ -13,6 +14,7 @@
 import org.springframework.web.bind.annotation.RestController;
 
 /** Controller for capabilities-related APIs */
+@DataPlane
 @RestController
 public class CapabilitiesController implements CapabilitiesApi {
```
nawatts, line 19: Since capabilities includes supported import formats, should this be ControlPlane as well?

nawatts (approved):

service/src/main/java/org/databiosphere/workspacedataservice/controller/CapabilitiesController.java
jladieu: I'd say yes, but I wonder if we should slim the list down accordingly though...with that in mind, leave as is and write a followup ticket?

ashanhol: Ah i was only going by things tagged jobs and import because that's what we were talking about exposing in the swagger (+1 to prioritizing updating swagger now). Were we planning on exposing capabilities in cWDS?

nawatts: Then again, I suppose it's not necessary. Since there's only one version of cWDS running and we know what it is, we don't really need this.

service/src/test/java/org/databiosphere/workspacedataservice/annotations/AnnotatedApisTest.java
```diff
@@ -0,0 +1,36 @@
...
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.context.ApplicationContext;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.web.bind.annotation.RestController;
+
+@ActiveProfiles({"control-plane", "data-plane"})
+@SpringBootTest
+class AnnotatedApisTest {
```
jladieu, line 17: Love this!

davidangb (commented):
looks good once tests pass! I agree that we may want the `/capabilities` API in the control plane, I'm fine leaving it as data-plane only until we decide we want it.

davidangb:
> looks good once tests pass!

If it's failing because of a missing workspace id, you can get one by extending `TestBase`

ashanhol:
@davidangb 

> If it's failing because of a missing workspace id, you can get one by extending TestBase

I saw, I'm trying (and failing) to get it to repro locally though to know if thats the issue for sure :/

PR #583, opened 2024-02-26
AJ-1650: Clean up properties.

jladieu:
This PR aims to clean up `TwdsProperties` and its children props classes to provide better fail-fast semantics and configuration for the `WORKSPACE_ID` environment variable:

- If in single-tenant mode, `WORKSPACE_ID` is required, and WDS should crash instead of allowing any code to run without it.  This is enforced by providing access to this environment variable through the `InstanceProperties.workspaceId()` accessor, which is designed to never return `null`.  In fact, `InstanceProperties` can't be resolved as a Bean at all _unless_ it has a valid `WorkspaceId`, so this can be used to enforced in depth when injected via `InstanceProperties`.
- A convenience `@Bean` for the `WorkspaceId` annotated as `@SingleTenant` is provided for direct injection, but still benefits from the same constraints.  Where possible, this is the easiest way to get access to `WorkspaceId`.
- A couple of example changes are provided in the code & tests where the prior approach of injecting `instanceProperties` needed to be updated, and demonstrate how we might stamp this out more broadly (eg: as part of [AJ-1656](https://broadworkbench.atlassian.net/browse/AJ-1656)).

Note: Direct access to the `workspace-id` using `@Value('twds.instance.workspace-id')` throughout the codebase still circumvents the validation and enforcement that it's non-null, non-blank, and a valid UUID. Followup is required to eliminate these references, which will probably cause unexpected misbehavior in `control-plane` ([AJ-1656](https://broadworkbench.atlassian.net/browse/AJ-1656))

[AJ-1656]: https://broadworkbench.atlassian.net/browse/AJ-1656?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service/src/main/java/org/databiosphere/workspacedataservice/config/InstanceProperties.java
```diff
@@ -1,68 +1,81 @@
...
 
-  public String getSourceWorkspaceId() {
-    return sourceWorkspaceId;
+  void setWorkspaceId(String workspaceId) {
+    if (!isNullOrEmpty(workspaceId)) {
+      try {
+        this.workspaceId = WorkspaceId.fromString(workspaceId);
+        validWorkspaceId = true;
+      } catch (Exception e) {
+        validWorkspaceId = false;
```
calypsomatic, line 55: Just double-checking - does `validWorkspaceId` need to be initialized to false?  Or set if `isNullOrEmpty(workspaceId)` returns false?

service/src/main/java/org/databiosphere/workspacedataservice/config/TenancyProperties.java
```diff
@@ -20,7 +20,7 @@ public boolean getAllowVirtualCollections() {
     return allowVirtualCollections;
   }
 
-  public void setAllowVirtualCollections(boolean allowVirtualCollections) {
+  void setAllowVirtualCollections(boolean allowVirtualCollections) {
```
ashanhol, line 23: who's calling the set functions that they don't need to be public?

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresCollectionDao.java
```diff
@@ -24,12 +26,15 @@ public class PostgresCollectionDao implements CollectionDao {
...
+  @Nullable private UUID workspaceId;
 
-  public PostgresCollectionDao(
-      NamedParameterJdbcTemplate namedTemplate, InstanceProperties instanceProperties) {
+  public PostgresCollectionDao(NamedParameterJdbcTemplate namedTemplate) {
     this.namedTemplate = namedTemplate;
-    this.workspaceId = instanceProperties.getWorkspaceUuid();
+  }
+
+  @Autowired(required = false) // control plane won't have workspaceId
```
ashanhol, line 35: so if I understand correctly, the `@nullable` allows us to have a null UUID without it throwing an exception?

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresCollectionDao.java
```diff
@@ -74,6 +79,10 @@ public void dropSchema(UUID collectionId) {
   @WriteTransaction
   @SuppressWarnings("squid:S2077") // since collectionId must be a UUID, it is safe to use inline
```
ashanhol, line 80: would UUID being null make this warning now valid?

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
```diff
@@ -30,24 +32,24 @@ public class CollectionService {
   private final ActivityLogger activityLogger;
   private final TenancyProperties tenancyProperties;
 
-  private WorkspaceId workspaceId;
+  @Nullable private WorkspaceId workspaceId;
```
ashanhol, line 35: Is there any way we can make this optional instead of nullable?

davidangb (commented):
minor comments inline, thank you!

service/src/main/java/org/databiosphere/workspacedataservice/config/TenancyProperties.java
jladieu: Spring container gets access to these through magic (reflection?) and so doesn't need them to be `public`

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresCollectionDao.java
ashanhol: (related to my other comment) why that approach over Optional? Seems like `@nullable` is riskier

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
jladieu: Yes totally!  I actually started by doing this, but the IDE squawked at me and told me it was a bad idea.  Which led me to Google "Why can't I make my attributes `Optional`?" and that opens up a whole can of worms 😆 

Here's some reading that I encountered which will do a better job than me of explaining:
* https://stackoverflow.com/questions/29033518/is-it-a-good-practice-to-use-optional-as-an-attribute-in-a-class
* https://dzone.com/articles/java-8-optional-usage-and-best-practices

service/src/main/java/org/databiosphere/workspacedataservice/config/InstanceProperties.java
jladieu: Boolean values default to false.  However, I think it makes sense to be as explicit as possible about this so I fixed in: a694898b7c49552941b7f2490d40255235f728c1

Also, your comment about setting it on the `else` made me think about this a bit...I think adding in an else clause here to be explicit would also imply that you could flip `InstanceProperties` back to an invalid state after being set up as valid.  I think this would be really bad, so I left the logic as is...which is that once `InstanceProperties` is considered 'valid' it can't be changed to be invalid.

Could also take this a step further and make it really immutable by only allowing you to call the setter once...but I am not too worried about that happening since this method is package protected.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresCollectionDao.java
jladieu: I went googling about this warning and found that it's now considered deprecated, so I didn't think too much about it.

I almost went and ripped all these `@SuppressWarnings` from codebase but held back because I didn't wanna bloat this PR.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresCollectionDao.java
jladieu: Answered in https://github.com/DataBiosphere/terra-workspace-data-service/pull/583#discussion_r1503249899

But from my reading on this...`Optional` can't exactly replace your concerns about null safety and you still have to think about it some way shape or form.   I'd _prefer_ to eliminate all `@Nullable` throughout the codebase, and actually hope by marking these as `@Nullable` I leave a little marker for some future cleanup, but I don't think swapping in `Optional` necessarily addresses the concern that a thing _might_ or _might not_ be provided.

service/src/main/java/org/databiosphere/workspacedataservice/config/InstanceProperties.java
```diff
@@ -1,68 +1,81 @@
...
  */
 public class InstanceProperties {
+  /**
+   * Shorthand annotation for @Qualifier("singleTenant"), used to mark the {@link WorkspaceId} bean
+   * used when running in single-tenant mode.
+   */
+  @Target({ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER})
+  @Retention(RetentionPolicy.RUNTIME)
+  @Qualifier("singleTenant")
+  public @interface SingleTenant {}
```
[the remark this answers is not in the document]
jladieu: Fixed in 75b2205a1b03b3430d7783a9ef5fe4c7dbc3ddaa

service/src/main/java/org/databiosphere/workspacedataservice/config/InstanceProperties.java
```diff
@@ -1,68 +1,81 @@
 package org.databiosphere.workspacedataservice.config;
 
-import java.util.UUID;
+import static com.google.common.base.Strings.isNullOrEmpty;
```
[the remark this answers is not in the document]
jladieu: Good call, thanks for pointing this out!  Fixed in 51c705acc6d93a5af5a827a9b57e756bdc1f9af7

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresCollectionDao.java
jladieu: I should point out...IMO a better/deeper way to fix this would be to disallow creating a `PostgresCollectionDao` _without_ a `WorkspaceId`, which I accomplished in another PR by marking the bean with `@DataPlane`.

But that's a much bigger change and doing that was actually one of the reasons https://github.com/DataBiosphere/terra-workspace-data-service/pull/570 became so unwieldy.  However, if you take a look at this file [over there](https://github.com/DataBiosphere/terra-workspace-data-service/blob/5b18fb122cd3c4120f73f593f0cb87f63289875a/service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresCollectionDao.java) you can see how it ends up looking :)

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresCollectionDao.java
davidangb: The comment "since collectionId must be a UUID, it is safe to use inline" really means: "since collectionId has already validated as a UUID - and UUIDs are limited to letters, numbers, and the dash character in a specific formation - we can be sure that this variable does not contain any sql-injection phrases. Therefore, we can use it in a sql statement safely."

davidangb (approved):
thank you!

calypsomatic (approved):

PR #584, opened 2024-02-26
[AJ-1560] Rename query parameter from statuses to status

nawatts:
Super nitpicky, but I noticed this while working on #569. Since the `statuses` query parameter for the list import jobs endpoint accepts a _single_ status, I think the parameter name should be singular as well (`status`). I would expect a plural parameter, like `statuses`, to accept multiple values (like a comma separated list or something like that).

Since this API hasn't been released yet, this isn't a breaking change.


ashanhol (changes requested):
https://swagger.io/docs/specification/serialization/
looks like the change we need is explode=false, then we'd be able to do `?statuses=thing1, thing2`

PR #585, opened 2024-02-26
AJ-1629: Fail with explanation on missing profile.

jladieu:
[AJ-1629](https://broadworkbench.atlassian.net/browse/AJ-1629): When starting WDS without a data-plane or control-plane profile, crash with a message explaining that they are needed, rather than letting Spring fail due to bean misconfiguration a bit later on.

To cause error:
```
./gradlew bootRun
```
Expect crash with stacktrace starting with:

`org.databiosphere.workspacedataservice.config.ConfigurationException: This deployment requires at least one of the following active profiles: [data-plane, control-plane]`

To make error not happen:
```
./gradlew bootRun --args='--spring.profiles.active=data-plane'
```

### DIFF
--- a/service/src/main/resources/application.yml
+++ b/service/src/main/resources/application.yml
@@ -126,6 +126,14 @@ twds:
     source-workspace-id: ${SOURCE_WORKSPACE_ID:}
   # short-lived credentials to use during cloning
   startup-token: ${STARTUP_TOKEN:}
+  # tenancy & data-import defaults match data-plane, but the source of truth for these configs
+  # should be application-data-plane.yml or application-control-plane.yml; they are provided here
+  # to allow Spring to boot without crashing and allow StartupConfig to run and check preconditions
+  tenancy:
+    allow-virtual-collections: false
+    require-env-workspace: true
+  data-import:
+    batch-write-record-sink: "wds"
 
   pg_dump:
     path: ${PGDUMP_PATH:/usr/bin/pg_dump}
--- a/service/src/main/java/org/databiosphere/workspacedataservice/startup/StartupConfig.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/startup/StartupConfig.java
@@ -1,5 +1,8 @@
 package org.databiosphere.workspacedataservice.startup;
 
+import static java.util.Arrays.stream;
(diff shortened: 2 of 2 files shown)


ashanhol:
This seems related to AJ-1629, include that link in your description?

service/src/main/resources/application.yml
```diff
@@ -126,6 +126,14 @@ twds:
     source-workspace-id: ${SOURCE_WORKSPACE_ID:}
   # short-lived credentials to use during cloning
   startup-token: ${STARTUP_TOKEN:}
+  # tenancy & data-import defaults match data-plane, but the source of truth for these configs
+  # should be application-data-plane.yml or application-control-plane.yml; they are provided here
```
ashanhol, line 130: So how would application-data/control-plane.yml override these?

davidangb: https://docs.spring.io/spring-boot/docs/3.2.3/reference/html/features.html#features.external-config has the doc:

> Spring Boot uses a very particular PropertySource order that is designed to allow sensible overriding of values. Later property sources can override the values defined in earlier ones. Sources are considered in the following order:

> Config data files are considered in the following order:
> 1. [Application properties](https://docs.spring.io/spring-boot/docs/3.2.3/reference/html/features.html#features.external-config.files) packaged inside your jar (application.properties and YAML variants).
> 2. [Profile-specific application properties](https://docs.spring.io/spring-boot/docs/3.2.3/reference/html/features.html#features.external-config.files.profile-specific) packaged inside your jar (application-{profile}.properties and YAML variants).

yuliadub: values in corresponding configs take precedence over those defined here

yuliadub: ha @davidangb you and I commented at the same time (although your answer is much more informative 😄, thank you! )

yuliadub (approved):
thank you!

ashanhol (approved):

davidangb (approved):

service/src/main/resources/application.yml
jladieu: Thanks for the link!

PR #586, opened 2024-02-26
AJ-1560: statuses param

davidangb:
Tweak swagger-ui so it generates the `statuses` parameter using a single comma-delimited value instead of multiple duplicate query params.

This *only* affects the urls that swagger-ui generates. It doesn't change what's allowed by the API, as illustrated by unit tests. But since end users sometimes rely on swagger-ui for API documentation, it's good to give them the structure we prefer.

Before:
Screenshot 2-26-2024 at 03 49 PM

After:
Screenshot 2-26-2024 at 03 52 PM

### DIFF
--- a/service/src/test/java/org/databiosphere/workspacedataservice/controller/JobControllerTest.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/controller/JobControllerTest.java
@@ -118,4 +118,28 @@ class JobControllerTest extends TestBase {
     assertEquals(StatusEnum.CREATED, jobList.get(0).getStatus());
     assertEquals(StatusEnum.CANCELLED, jobList.get(1).getStatus());
   }
+
+  @Test
+  void instanceJobsWithMultipleDelimitedStatuses() {
+    when(collectionDao.collectionSchemaExists(collectionId.id())).thenReturn(true);
+    assertDoesNotThrow(() -> jobDao.updateStatus(jobId, StatusEnum.CANCELLED));
+    HttpHeaders headers = new HttpHeaders();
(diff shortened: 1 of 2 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/controller/JobControllerTest.java
```diff
@@ -118,4 +118,28 @@ void instanceJobsWithMultipleStatuses() {
...
     assertEquals(StatusEnum.CANCELLED, jobList.get(1).getStatus());
   }
+
+  @Test
+  void instanceJobsWithMultipleDelimitedStatuses() {
+    when(collectionDao.collectionSchemaExists(collectionId.id())).thenReturn(true);
+    assertDoesNotThrow(() -> jobDao.updateStatus(jobId, StatusEnum.CANCELLED));
+    HttpHeaders headers = new HttpHeaders();
+    // ParameterizedTypeReference<List<GenericJobServerModel>> returnType = new
+    // ParameterizedTypeReference<List<GenericJobServerModel>>() {};
```
nawatts, line 128: Looks like these two lines were commented out code.

```suggestion
```

nawatts (approved):

service/src/test/java/org/databiosphere/workspacedataservice/controller/JobControllerTest.java
davidangb: good call; I blindly copy/pasted from the previous test. Removed from both.

ashanhol (approved):
thank you!

PR #587, opened 2024-02-27
AJ-1585 - Add logic that writes Rawls Json to bucket

yuliadub:
A few things at play:
- GcsStorage now uses Data Import config class 
- GcsStorage is a Bean, but I added it as an Optional one since the way RecordSink is currently set up, it needs to be created no matter if we are in app plane or data plane. However, we do not need it for data plane and it will always fail since there is no defautl credential wds would be running under. David proposed to potentially split the Record Sinks implementations instead of them being shared but I say lets tackle that in a different ticket and not block import e2e on that
- GcsStorage has a mock bean that is used just inside tests, it doesn't actually write anything to a real bucket, but that is how we test that things work as expected 
- Adjusted to list files in bucket in test since there would only be one file, per Josh's suggestion (~~)Since there is no easy was to propagate BlobName back to a test a this time, I ended up adding a new method that only will work for Rawls Record Sink that allows us to get the blob name to verify that all worked correctly (we really only need this for the test so far). I want to note that this wont necessarily work if two imports are happening at once (at least the way I understand it) -- but that is not the case for the test so it works? 🤷 there is a better/prettier way to do this, but it did not arrive into my brain today 😄  (~~)

Reminder:


service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
```diff
@@ -32,4 +32,16 @@ public RecordSinkMode getBatchWriteRecordSink() {
   void setBatchWriteRecordSink(String batchWriteRecordSink) {
     this.batchWriteRecordSink = RecordSinkMode.fromValue(batchWriteRecordSink);
   }
+
+  private String projectId;
```
jladieu, line 36: [no action needed for this PR] Whoa, you didn't need to add setters for these attributes to get them autowired by spring?  If that's the case, cool!  We can probably blow up all the setter methods in all our props files that exist for this purpose and aren't anything more than simple passthroughs.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -71,6 +85,12 @@ public void upsertBatch(
     ImmutableList.Builder<Entity> entities = ImmutableList.builder();
     records.stream().map(this::toEntity).forEach(entities::add);
     jsonConsumer.accept(mapper.writeValueAsString(entities.build()));
+
+    this.blobName = UUID.randomUUID().toString();
+    storage.createGcsFile(
```
jladieu, line 90: How about you drop the blobName from the method parameters here, and let `createGcsFile` take care of coming up with the name?

Then, you can just use the name returned by the call when generating the pubsub message in the next line, and don't need to set it on instance state.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RecordSink.java
```diff
@@ -31,4 +31,12 @@ void upsertBatch(
   /** Delete the given batch of records. */
   void deleteBatch(RecordType recordType, List<Record> records)
       throws BatchWriteException, IOException;
+
+  /**
+   * This method is only used in Control Plane WDS to get the name of the Google Bucket import has
+   * been written to
+   */
+  default String getBlobName() {
```
jladieu, line 39: I think you needed to add this to the interface for testability reasons but I don't think it belongs here, due to the issues you mention in your PR description.  I'll comment over there how you might avoid needing it.

service/src/test/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkTest.java
```diff
@@ -175,6 +184,14 @@ private List<Entity> doUpsert(Record record, Record... additionalRecords) {
           );
 
       assertThat(recordedJson.toString()).isNotNull();
+      var blobName = recordSink.getBlobName();
+      var text = storage.getBlobContents(blobName);
```
jladieu, line 188: As mentioned earlier (and in your PR description), having blobName be an instance variable on recordSink is problematic.  I think we should figure out a way to drop that from the interface.

Is there a way to to essentially `ls -l` the contents of a known location in the `Storage` API?  I am guessing you could get the blobName by doing the equivalent of:

1. `ls -l`
2. assert that there's exactly one item in the bucket
3. get the blobName of that item
4. use that blobName to read and check the contents

service/src/test/java/org/databiosphere/workspacedataservice/storage/GcsStorageTest.java
```diff
@@ -3,27 +3,30 @@
...
+@SpringBootTest
+@ActiveProfiles("control-plane")
 class GcsStorageTest {
 
-  private String bucketName = "test-bucket";
-
-  private String projectId = "test-projectId";
-
-  private final Storage mockStorage = LocalStorageHelper.getOptions().getService();
+  @Qualifier("mockGcsStorage")
```
jladieu, line 24: Nice use of `@Qualifier` to get the right instance in here...I would have just marked `mockGcsStorage` as `@Primary` but I think this is much more intention revealing and less likely to cause unexpected problems.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -45,10 +57,12 @@ public class RawlsRecordSink implements RecordSink {
   RawlsRecordSink(
       String attributePrefix,
       ObjectMapper mapper,
+      @Nullable GcsStorage storage,
```
jladieu, line 60: Cool, does marking it `@Nullable` allow Spring to happily proceed when the bean isn't available?  I've been accomplishing this by using a setter with `@Autowired(required=false)` but would like this more if it's working out.

service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorageConfig.java
```diff
@@ -0,0 +1,18 @@
...
+import org.databiosphere.workspacedataservice.config.DataImportProperties;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ControlPlane
+public class GcsStorageConfig {
+  @Bean
+  @Autowired(required = false)
```
jladieu, line 14: This `@Autowired(required=false)` looks unusual...not saying it's wrong...but usually I would expect that to be an annotation on a setter or constructor, not a getter for a `@Bean`.

jladieu: Ahhh I get it...I think I interpret this as follows:

- Provide `gcsStorageBean` even if `DataImportProperties` isn't autowired.
- `GcsStorage bean is fine with `DataProperties` being `null.

Does this jive with how you intended it to be used?

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
jladieu: I think I answered my own question here in https://github.com/DataBiosphere/terra-workspace-data-service/pull/587/files#r1506064298

With that in mind, since `storage` might be `null`, I think it'd make sense to check for null anywhere it's used (eg: in `upsertBatch`) and throw some kind of `RuntimeError`: (`ConfigurationException`?  `IllegalStateException`?  `UnsupportedOperationException`?) if it's not present.

jladieu (commented):
This is awesome!  Thanks for leaning into and experimenting with all the bean config stuff.  I have a few comments throughout for thought... probably the most major concern is the one you already highlighted in your PR description, which is that having the `blobName` persist on the instance state of `RecordSink` is probably not good.  I have an idea of how to avoid that but curious what you think.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
yuliadub: ah yes, great point!

service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorageConfig.java
yuliadub: yeah, that was my original thinking - since I was getting errors in data-plane that import properties were not there. But it seems that the other @Autowired(required=false) does the trick, so this one is not actually needed it seems after taking a closer look

service/src/test/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkTest.java
yuliadub: I like it! Thanks for giving me a solution here - I was way overengineering this 😄

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -23,6 +14,20 @@
 import org.databiosphere.workspacedataservice.service.model.exception.BatchWriteException;
 import org.databiosphere.workspacedataservice.shared.model.Record;
 import org.databiosphere.workspacedataservice.shared.model.RecordType;
+import org.databiosphere.workspacedataservice.storage.GcsStorage;
```
yuliadub, line 17: ughh intelij keeps doing weird stuff with my imports, not sure what is up.

yuliadub:
> This is awesome! Thanks for leaning into and experimenting with all the bean config stuff. I have a few comments throughout for thought... probably the most major concern is the one you already highlighted in your PR description, which is that having the `blobName` persist on the instance state of `RecordSink` is probably not good. I have an idea of how to avoid that but curious what you think.

@jladieu love your suggestion - refactored to use it!

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -45,10 +50,12 @@ public class RawlsRecordSink implements RecordSink {
   RawlsRecordSink(
       String attributePrefix,
       ObjectMapper mapper,
+      @Nullable GcsStorage storage,
```
davidangb, line 53: can `GcsStorage` actually be null here? I thought we didn't create a `RawlsRecordSink` until we actually needed one - and by the time we need one, `GcsStorage` should exist.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -35,6 +39,7 @@
 public class RawlsRecordSink implements RecordSink {
```
davidangb, line 39: nit: update/remove the "TODO(AJ-1585)" comment a few lines up?

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -71,6 +78,16 @@ public void upsertBatch(
     ImmutableList.Builder<Entity> entities = ImmutableList.builder();
     records.stream().map(this::toEntity).forEach(entities::add);
     jsonConsumer.accept(mapper.writeValueAsString(entities.build()));
+
+    if (storage != null) {
+      var blobName =
+          storage.createGcsFile(
+              new ByteArrayInputStream(mapper.writeValueAsBytes(entities.build())));
```
davidangb, line 85: nit: `var blobName` is unused, this line could just be `storage.createGcsFile(…`

service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorage.java
```diff
@@ -44,8 +48,9 @@ public InputStream getBlobContents(String blobName) throws IOException {
     return gcsResource.getInputStream();
   }
 
-  public String createGcsFile(String blobName, InputStream contents) throws IOException {
+  public String createGcsFile(InputStream contents) throws IOException {
     // create the GCS Resource
+    var blobName = UUID.randomUUID().toString();
```
davidangb, line 53: The way import service works now, this would be the jobId instead of a random UUID. This helps with debugging from a human perspective. BUUUUT, we don't have the jobId down here and it's probably a pain in the butt to get to it. Perhaps this can be a follow-on TODO.

service/src/main/resources/application-control-plane.yml
```diff
@@ -11,6 +11,5 @@ twds:
 
   data-import:
     batch-write-record-sink: "rawls"
-    rawls-bucket:
-      name: ${SERVICE_GOOGLE_BUCKET:unknown}
-      projectId: ${SERVICE_GOOGLE_PROJECT:unknown}
+    project-id: ${SERVICE_GOOGLE_PROJECT:unknown}
+    rawls-bucket-name: ${SERVICE_GOOGLE_BUCKET:unknown}
```
davidangb, line 15: I think better to set these as empty instead of the magic string "unknown":
```suggestion
    project-id: ${SERVICE_GOOGLE_PROJECT}
    rawls-bucket-name: ${SERVICE_GOOGLE_BUCKET}
```

service/src/test/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkTest.java
```diff
@@ -175,6 +185,21 @@ private List<Entity> doUpsert(Record record, Record... additionalRecords) {
...
+      var blobName = blobs.iterator().next().getName();
+      // check that the contents match the expected Json
+      var text = storage.getBlobContents(blobName);
+      String contents =
+          new BufferedReader(new InputStreamReader(text, StandardCharsets.UTF_8))
+              .lines()
+              .collect(Collectors.joining("\n"));
+      assertThat(contents).isEqualTo(recordedJson.toString());
+      // delete the blob so the next test can have a clean slate
+      storage.deleteBlob(blobName);
```
jladieu, line 202: consider doing this in a `finally` block after all the `catch` clauses, to guarantee it works even when there's an exception

jladieu (approved):
Couple small suggestions, but pre-emptively approved!  Great work!

service/src/test/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkTest.java
```diff
@@ -175,6 +185,21 @@ private List<Entity> doUpsert(Record record, Record... additionalRecords) {
...
+      // confirm there is only 1
+      assertThat(Iterables.size(blobs)).isEqualTo(1);
+      // get the name of the blob
+      var blobName = blobs.iterator().next().getName();
+      // check that the contents match the expected Json
+      var text = storage.getBlobContents(blobName);
+      String contents =
+          new BufferedReader(new InputStreamReader(text, StandardCharsets.UTF_8))
+              .lines()
+              .collect(Collectors.joining("\n"));
```
davidangb, line 199: as a shortcut, use the `org.springframework.util.StreamUtils` helper:
```java
String contents = StreamUtils.copyToString(text, StandardCharsets.UTF_8);
```

service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorage.java
```diff
@@ -55,4 +60,12 @@ public String createGcsFile(String blobName, InputStream contents) throws IOExce
...
     return gcsResource.getBlobName();
   }
+
+  public Iterable<Blob> getBlobsInBucket() {
+    return storage.list(this.bucketName).getValues();
+  }
+
+  public void deleteBlob(String blobName) {
+    storage.delete(this.bucketName, blobName);
+  }
```
davidangb, line 70: annotate these as `@VisibleForTesting` and/or add a comment that they should not be used for runtime code

davidangb (commented):
only minor comments inline, nice job on this!

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
yuliadub: was trying to make it helpful for when the pubsub stuff comes in, but i'll remove :)

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
yuliadub: it doesnt have to be null, my original thinking was to keep the Nullable there just in case something is ever not working as expected with profiles or something like that, but I see your point as well!

service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorage.java
yuliadub: they can be though? we dont use them today but if runtime code wanted to use them, they would be good to go

service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorage.java
yuliadub: added a note about his in RawlsRecordSink - maybe once we split them, RawlsRecordSink can get the JobId somehow and it wont impact WDS?

service/src/test/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkTest.java
yuliadub: aha, I knew there was an easier way to do this! Thank you!

davidangb (approved):
awesome, thank you!

PR #592, opened 2024-02-27
AJ-1586: split RecordSinkFactory into two implementations

davidangb:
Up for debate: should `RecordSinkFactory` be an interface, with two separate bean implementations?

The current `RecordSinkFactory` has a bunch of conditionals, including setting many of its member variables as optional. As Yulia is adding `GcsStorage` as another member variable, this expands even more and complicates the application context.

By splitting into two distinct implementations, each implementation can be more strict about what it requires.


service/src/main/java/org/databiosphere/workspacedataservice/recordsink/WdsRecordSinkFactory.java
```diff
@@ -0,0 +1,37 @@
...
+import org.databiosphere.workspacedataservice.dao.RecordDao;
+import org.databiosphere.workspacedataservice.service.DataTypeInferer;
+import org.databiosphere.workspacedataservice.service.RecordService;
+import org.springframework.context.annotation.Primary;
+import org.springframework.stereotype.Component;
+
+/** RecordSinkFactory implementation for the data plane */
+@DataPlane
+@Primary // marked as primary specifically so AnnotatedApisTest won't fail
+@Component
```
davidangb, line 14: I don't like having to set `@Primary` here. `AnnotatedApisTest` is the ONLY place where we set both the `control-plane` and the `data-plane` profiles. Under those conditions, Spring can't figure out if it should make a `RawslsRecordSinkFactory` or a `WdsRecordSinkFactory`, so I needed to set one as primary.

ashanhol: But `AnnotatedApisTest` doesn't touch this since it's not a `@RestController`? Is there a way we could modify `AnnotatedApisTest` so that it only includes the controller files?

jladieu (approved):
I like it!  I also agree having to do the`@Primary` is a bad smell...  I feel like I encountered this and came up with a workaround but can't remember what it was...I'll see if I can dig that up and share back here.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/WdsRecordSinkFactory.java
jladieu: Found my workaround for this: https://github.com/DataBiosphere/terra-workspace-data-service/pull/570/files#r1505159260

I like it because it brings the hack closer to `AnnotatedApisTest`

davidangb: > But AnnotatedApisTest doesn't touch this since it's not a @RestController? Is there a way we could modify AnnotatedApisTest so that it only includes the controller files?

Even though `AnnotatedApisTest` only _asserts_ on the RestControllers, Spring still creates the full application context, including a RecordSinkFactory. We could specify `@SpringBootTest(classes = {...})` or `@ContextConfiguration(classes = {…})` on it to tell Spring exactly which beans to consider in its application context, but that gets tricky due to dependencies. `RecordController` depends on `RecordOrchestratorService`, which depends on `RecordSinkFactory`, so we need to consider `RecordSinkFactory` somehow.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbTestSupport.java
```diff
@@ -18,9 +18,11 @@
 import org.quartz.JobExecutionContext;
 import org.quartz.JobExecutionException;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Lazy;
 import org.springframework.core.io.Resource;
 import org.springframework.stereotype.Component;
 
+@Lazy // only used by a few tests; don't instantiate when not needed
```
davidangb, line 25: this - and the same change in `TdrTestSupport` - are not required for this PR. As I was debugging bean dependencies, I noticed that `PfbTestSupport` and `TdrTestSupport` were being instantiated but not used in many places, so I figured I'd streamline them.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/WdsRecordSinkFactory.java
davidangb: > Found my workaround for this: https://github.com/DataBiosphere/terra-workspace-data-service/pull/570/files#r1505159260
> I like it because it brings the hack closer to AnnotatedApisTest

Bah, this solution did not work for me as-is, I think because `WdsRecordSinkFactory` and `RawlsRecordSinkFactory` are `@Component`s and are eagerly instantiated. The solution I reached is a little more complicated, but it still keeps it limited to `AnnotatedApisTest`, take a look?

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbTestSupport.java
jladieu: Nice!

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/WdsRecordSinkFactory.java
jladieu: Seems fine to me!  You mention that they are eagerly instantiated because they are `@Component`s...if they were `@Bean`s would they be lazy and thus not as exposed to this sort of conflict?

I have been assuming that `@Component` & `@Bean` were kinda interchangeable, but it'd be good to know if there are best practices around when to use one or the other.

davidangb:
#587 is more important than this PR, and they touch the same files - let's focus on that other PR first, then I'll rebase this one if we decide to move forward with it.

yuliadub:
looking closer at this, I think I like the split into to distinct Sinks - it seems like there are quite a few things that only Rawls or WDS needs and the split makes this easier to track vs having to disable a whole bunch of things when one or the other doesn't need it!

davidangb:
reverting to draft until I can properly rebase on top of #587

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/WdsRecordSinkFactory.java
jladieu: This is very interesting!  Thanks for sharing back.

jladieu (approved):

PR #593, opened 2024-02-27
AJ-1649: enable Gradle build scans

davidangb:
Enable Gradle [build scans](https://scans.gradle.com/). Hopefully, this will make working with dependency updates and other build duties easier.

You can navigate to the build scan results by:
1. Navigate to either of the `Build and Test / unit-tests` GHA jobs
2. Navigate to the Job Summary for that job
3. Scroll down and click on the "Build Scan PUBLISHED" badge

Alternately, we could surface the [job summary as a PR comment](https://github.com/gradle/actions/tree/main/setup-gradle#adding-job-summary-as-a-pull-request-comment). This will make it easier to find, but will also clutter up the PR. Which would you prefer? My personal opinion is that we won't need to look at the build scan for every PR, so it's ok to have it a bit hidden.

I've only turned on the build scan for the `test` job. We could turn it on other places too, but that seems like overkill right now.

WSM and TDR are both using build scans (I suspect others are too), and report positively about it.

---

Screenshot 2-27-2024 at 05 14 PM

---

Screenshot 2-27-2024 at 05 14 PM (1)

---


.github/workflows/build-and-test.yml
```diff
@@ -69,7 +69,7 @@ jobs:
 
       - name: Build, assemble, and test
         id: build-test
-        run: ./gradlew --build-cache test jacocoTestReport
+        run: ./gradlew --build-cache --scan test jacocoTestReport
```
ashanhol, line 72: Can you tell me a little more about build scans and when we'd want to use them?

ashanhol (approved):

.github/workflows/build-and-test.yml
davidangb: Check the first link in the PR description - it gives a pretty good explanation of what's in a build scan. You can also explore the build scan generated by this PR. In my particular case I have been using it to search for transitive dependencies and see what is pulling in those dependencies.

jladieu (approved):

yuliadub (approved):

calypsomatic (approved):

PR #594, opened 2024-02-27
AJ-1633 Split `SamDao` to wire up Auth for cWDS.

jladieu:
Prefactors and the final implementation for [AJ-1633](https://broadworkbench.atlassian.net/browse/AJ-1633), broken into discrete commits for easier review (one per bullet).  Each commit has its own detailed message and some inline comments to aid review, but here is a summary of each:

* Extract `SamAuthorizationDao` interface (as a superclass of `SamDao`) to encapsulate permission-checking concerns and adjust tests & bean config to take advantage.
* Migrate consumers of `SamAuthorizationDao` over to the new interface and eliminate the `extends` relationship from `SamDao`.
* Split out `HttpSamAuthorizationDao` to implement aforementioned interface and simplify `MisconfiguredSamDao` accordingly.
* Switch over all `@Autowired` and `@MockBean`s to use the `SamAuthorizationDaoFactory` instead of directly wiring `SamAuthorizationDao`.  This means we can get rid of the `@Bean` for `SamAuthorizationDao` and eliminate all the extra complexity from `SamConfig` that comes from having to worry about instantiating it without a proper `WORKSPACE_ID`
* Strip `BearerToken` out of the `SamAuthorizationDao` API, it's not needed!
* Strip `WorkspaceId` out of the `SamAuthorizationDao` API, it's taken care of completely by the factory.
* Finally finish wiring up `CollectionService` to use `SamAuthorizationDao` and derive the `WorkspaceId` for all methods that are used by cWDS endpoints.

Prior State:
* `SamDao` interface bore both permissions & other sam concerns.
* `SamDao` permissions API had a scattering of `BearerToken` and `WorkspaceId` args and corresponding overloads.
* `SamConfig` had to deal with potentially invalid values for `WORKSPACE_ID` and could produce a `MisconfiguredSamDao`.
* `SamDao` bore the concern of using `WORKSPACE_ID` to check permissions.

End State:
* `SamDao` has been split into two separate interfaces `SamAuthorizationDao` (concerned with permissions) and `SamDao` (everything else)
*  Each interface has an `Http` impl which implements the the API in the interface.
* `SamConfig` now configures `SamDaoFactory` instead of directly providing `SamAuthorizationDao` or a potentially misconfigured `SamAuthorizationDao`
* `HttpSamAuthorizationDao` is now created via `SamDaoFactory`, which injects the `WorkspaceId` concern on behalf of the clients that autowire it. (`ImportService`, `RecordOrchestratorService`, `CollectionService`)
* `BearerToken` concern is eliminated from `SamAuthorizationDao` and subclasses, as it was never utilized.
* Only one permission check method per permission.
* `CollectionService` now bears the concern of using `WORKSPACE_ID` to check permissions, or wiring through a `WorkspaceId` based on `CollectionId`.
* `JobService` and `CollectionService` both use derived `WorkspaceId` to check permissions.
* `RecordOrchestratorService` still does not for _reasons_ described in the last commit.  Addressed in followup https://github.com/DataBiosphere/terra-workspace-data-service/pull/602

[AJ-1633]: https://broadworkbench.atlassian.net/browse/AJ-1633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service/src/main/java/org/databiosphere/workspacedataservice/sam/SamAuthorizationDao.java
```diff
@@ -0,0 +1,56 @@
+package org.databiosphere.workspacedataservice.sam;
```
jladieu, line 1: This is all just moved from `SamDao`.  All these methods are going to operate in the context of a `WorkspaceId` which will be injected during creation of `HttpSamAuthorizationDao`, which will let us get rid of all the `WorkspaceId` arguments in the API.

The same can also be done with `BearerToken` (later on), which will leave us with only four methods on the interface.

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamDao.java
```diff
@@ -2,63 +2,12 @@
...
-
-  /**
-   * Check if the current user has permission to write to a workspace resource from Sam
-   *
-   * @return true if the user has permission
-   */
-  boolean hasWriteWorkspacePermission();
-
-  boolean hasWriteWorkspacePermission(WorkspaceId workspaceId);
+public interface SamDao extends SamAuthorizationDao {
```
jladieu: Having `SamDao` extend `SamAuthorizationDao` was done by the IDE; it allows all existing `SamDao` implementations to keep being used where the concerns are strictly limited to those methods that aren't on the `SamAuthorizationDao`.  It's a temporary measure that I will take out in a subsequent commit when I fully divide out the two interfaces.

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamConfig.java
```diff
@@ -38,6 +42,31 @@ public SamClientFactory getSamClientFactory() {
     return new HttpSamClientFactory(samUrl);
   }
 
+  @Bean
```
jladieu, line 45: Temporarily duplicating this bean configuration (I'll remove it immediately in the next commit).

This allows me to drop the shim of `SamDao` extending `SamAuthorizationDao`, as I don't intend for them to be interchangeable after this PR.

service/src/test/java/org/databiosphere/workspacedataservice/sam/SamAuthorizationDaoInvalidWorkspaceIdTest.java
```diff
@@ -24,27 +24,21 @@
...
-    assertFalse(samDao.hasDeleteCollectionPermission());
-    assertFalse(samDao.hasDeleteCollectionPermission(BearerToken.of("token")));
-    assertFalse(samDao.hasWriteWorkspacePermission());
-    assertFalse(samDao.hasWriteWorkspacePermission(WorkspaceId.of(randomUUID())));
+  void createsFailingDao() {
+    assertInstanceOf(MisconfiguredSamDao.class, samAuthorizationDao);
   }
 
   @Test
-  public void statusThrows() {
```
jladieu, line 47: `MisconfiguredSamDao` no longer has any of the non-permission concerns, so it now only implements the `SamAuthorizationDao` interface.  Consequently, there are no non-permission methods to test anymore.

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceControlPlaneTest.java
```diff
@@ -29,7 +29,7 @@
  *
  * @see ImportServiceTest
  */
-@ActiveProfiles({"control-plane", "noop-scheduler-dao"})
+@ActiveProfiles({"control-plane", "noop-scheduler-dao", "mock-sam"})
```
jladieu, line 32: Need to explicitly use `mock-sam` here since `MisconfiguredAuthorizationDao` won't take care of mocking out getting a pet token for us.

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceDataPlaneTest.java
```diff
@@ -31,7 +31,7 @@
  *
  * @see ImportServiceTest
  */
-@ActiveProfiles({"data-plane", "noop-scheduler-dao"})
+@ActiveProfiles({"data-plane", "noop-scheduler-dao", "mock-sam"})
```
jladieu, line 34: Same here, we don't get petToken mocking by accident & for free anymore via `MisconfiguredSamDao`, so we need to explicitly mock sam.

service/src/main/java/org/databiosphere/workspacedataservice/sam/MisconfiguredSamAuthorizationDao.java
```diff
@@ -1,96 +1,69 @@
...
-import org.broadinstitute.dsde.workbench.client.sam.model.SystemStatus;
 import org.databiosphere.workspacedataservice.shared.model.BearerToken;
 import org.databiosphere.workspacedataservice.shared.model.WorkspaceId;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * This class will be used instead of HttpSamDao if the WORKSPACE_ID env var contains a non-UUID
- * value, including being blank.
+ * This class will be used instead of HttpSamAuthorizationDao if a well-formed WORKSPACE_ID env var
```
jladieu, line 9: IMO this is of debatable value at this point and may be able to completely go away as part of completing this ticket.

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceTest.java
```diff
@@ -61,7 +60,6 @@ class ImportServiceTest extends TestBase {
   @Autowired ImportService importService;
   @Autowired CollectionDao collectionDao;
   @Autowired CollectionService collectionService;
-  @Autowired SamDao samDao;
```
jladieu, line 64: This was caught up in the rename but wasn't actually needed for `ImportServiceTest` to keep running, so I just dropped it.

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
```diff
@@ -57,44 +58,58 @@ public List<UUID> listCollections(String version) {
...
    * @param version WDS API version
    */
-  public void createCollection(UUID collectionId, String version) {
+  public void createCollection(WorkspaceId workspaceId, CollectionId collectionId, String version) {
     validateVersion(version);
 
     // check that the current user has permission on the parent workspace
-    boolean hasCreateCollectionPermission = samDao.hasCreateCollectionPermission();
+    boolean hasCreateCollectionPermission =
+        getSamAuthorizationDao(workspaceId).hasCreateCollectionPermission();
```
jladieu, line 85: This now wires `WorkspaceId` from the caller through to get the `SamAuthorizationDao`.  However currently, the only callers are coming through the deprecated endpoint that just assumes `workspaceId` (which is the `$WORKSPACE_ID` environment variable.

As we stand up alternate APIs that require a `WorkspaceId` in the URI, we can start passing that `WorkspaceId` down through the non-deprecated endpoint to get more multi-tenant friendly auth.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -104,7 +106,12 @@ public void validateAndPermissions(UUID collectionId, String version) {
     validateVersion(version);
     collectionService.validateCollection(collectionId);
 
-    boolean hasWriteWorkspacePermission = samDao.hasWriteWorkspacePermission();
+    // check that the caller has write permissions on the workspace associated with the collectionId
+    WorkspaceId workspaceId = collectionService.getWorkspaceId(CollectionId.of(collectionId));
```
jladieu, line 110: This is a key functional change where we switch over from using the old style (which relies on `WORKSPACE_ID`) to using the `getWorkspaceId` intermediate.

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
```diff
@@ -20,99 +19,10 @@ public class HttpSamDao implements SamDao {
...
-   *
-   * @return true if the user has permission
-   */
-  @Override
-  public boolean hasReadWorkspacePermission(WorkspaceId workspaceId) {
-    return hasReadWorkspacePermission(workspaceId, BearerToken.empty());
-  }
-
-  @Override
-  public boolean hasReadWorkspacePermission(WorkspaceId workspaceId, BearerToken token) {
```
jladieu, line 93: Note: the old implementation was just ignoring `WorkspaceId`, the new impl now passes it through.

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamAuthorizationDao.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  /**
+   * Check if the current user has permission to read a workspace resource from Sam.
+   *
+   * @return true if the user has permission
+   */
+  @Override
+  public boolean hasReadWorkspacePermission() {
+    return hasPermission(ACTION_READ, "Sam.hasReadWorkspacePermission", workspaceId);
```
jladieu, line 68: Contrasting with the red part of the diff: https://github.com/DataBiosphere/terra-workspace-data-service/pull/594/files#r1507663750, where the `workspaceId` is ignored.  This is a key change in functionality to serve the goal of this ticket.

jladieu:
Just reviewed Sonar warnings, which are all about TODOs or Deprecations.  I think I'm okay with these persisting.

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamAuthorizationDaoFactory.java
```diff
@@ -0,0 +1,28 @@
...
+    this.samClientFactory = samClientFactory;
+    this.restClientRetry = restClientRetry;
+  }
+
+  public SamAuthorizationDao getSamAuthorizationDao(WorkspaceId workspaceId) {
+    LOGGER.info(
+        "Sam integration will query type={}, resourceId={}, action={}",
+        SamAuthorizationDao.RESOURCE_NAME_WORKSPACE,
+        workspaceId,
+        SamAuthorizationDao.ACTION_WRITE);
```
calypsomatic, line 25: Doesn't `SamAuthorizationDao` check permissions other than `write`?  It seems like this statement is claimed the dao is always checking write permission, or am I misunderstanding?

jladieu: Yeah you're totally right, this log message doesn't make sense!  Good catch.

This log message is a direct copy/paste from `SamConfig`, so it's what is printed in prod now too, but I'll fix it here :)

jladieu: Pushed an amendment to fix this.

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamAuthorizationDaoFactory.java
```diff
@@ -0,0 +1,33 @@
...
+    this.samClientFactory = samClientFactory;
+    this.restClientRetry = restClientRetry;
+  }
+
+  public SamAuthorizationDao getSamAuthorizationDao(WorkspaceId workspaceId) {
+    LOGGER.info(
+        "Sam integration will query type={}, resourceId={}, actions={}",
+        SamAuthorizationDao.RESOURCE_NAME_WORKSPACE,
+        workspaceId,
+        Set.of(ACTION_READ, ACTION_WRITE, ACTION_DELETE));
```
jladieu, line 30: @calypsomatic now I print all the permissions

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamAuthorizationDao.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  // helpers for permission checks
+
+  // check for permission using the configured workspaceId
+  private boolean hasPermission(String action, String loggerHint) {
+    return hasPermission(action, loggerHint, workspaceId);
+  }
```
davidangb, line 76: I think that `hasCreateCollectionPermission`, `hasDeleteCollectionPermission`, `hasWriteWorkspacePermission`, `hasRadWorkspacePermission` should all either specify the `workspaceId` or not specify it - if those are all consistent, you can delete one or the other of these `hasPermission` methods

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamConfig.java
```diff
@@ -1,16 +1,19 @@
...
 
 /**
- * Bean creator for: - SamClientFactory, injecting the base url to Sam into that factory. - SamDao,
- * injecting the SamClientFactory into that dao.
+ * Bean creator for:
+ *
+ * <ul>
+ *   <li>SamClientFactory, injecting the base url to Sam into that factory.
+ *   <li>SamDao, injecting the SamClientFactory into that dao.
+ * </ul>
```
davidangb, line 16: nit: also `SamAuthorizationDaoFactory`

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
```diff
@@ -57,44 +58,58 @@ public List<UUID> listCollections(String version) {
...
   }
 
-  // TODO update this javadoc, also are there logical problems with the sam resource type?
-  // Also check the naming on these methods here
+  /** Migrate callers to use {@link #createCollection(WorkspaceId, CollectionId, String)} */
+  @Deprecated
+  public void createCollection(UUID collectionId, String version) {
+    if (workspaceId == null) {
+      throw new IllegalStateException("workspaceId must be set to create a collection.");
+    }
```
davidangb, line 66: could this clause check `TenancyProperties.allowVirtualCollections` instead of, or in addition to, `workspaceId == null`? (maybe "allowVirtualCollections" should be renamed to "requireVirtualCollections")

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
```diff
@@ -57,44 +58,58 @@ public List<UUID> listCollections(String version) {
...
+        user -> user.created().collection().withUuid(collectionId.id()));
   }
 
   public void deleteCollection(UUID collectionId, String version) {
     validateVersion(version);
     validateCollection(collectionId);
 
+    // TODO(AJ-1633): call getWorkspaceId(collectionId) to get the workspaceId, rather than relying
+    //   on the WORKSPACE_ID environment variable; test cases that rely on deleteCollection prevent
+    //   this change
```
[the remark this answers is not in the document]
davidangb: I suggest breaking this out into a separate PR/ticket - let's get this PR through and then figure out how to tackle the outstanding issue

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceControlPlaneTest.java
```diff
@@ -29,18 +34,26 @@
...
+  private final SamAuthorizationDao samAuthorizationDao = spy(MockSamAuthorizationDao.allowAll());
   private final URI importUri = URI.create("http://does/not/matter");
   private final ImportRequestServerModel importRequest =
       new ImportRequestServerModel(PFB, importUri);
 
+  @BeforeEach
+  void setup() {
+    when(samAuthorizationDaoFactory.getSamAuthorizationDao(any(WorkspaceId.class)))
+        .thenReturn(samAuthorizationDao);
+  }
```
davidangb, line 55: consider removing this from the `@BeforeEach`, and then setting mock expectations for the specific workspaceId used in each individual test, to ensure that we're checking permissions on the correct workspaceid.

davidangb (commented):
comments inline! I did add comments to individual commits, since you suggested reviewing commit-by-commit … therefore, some of my comments may be out of date

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
```diff
@@ -118,6 +129,7 @@ public void validateCollection(UUID collectionId) {
       return;
     }
     // else, check if this collection has a row in the collections table
+    // TODO(AJ-1662): wire workspaceId down through collectionDao
```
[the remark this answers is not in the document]
jladieu: When I started adding these TODOs, I was thinking in a future multi-tenancy mode, we'd probably want to scope all interactions with `CollectionDao` with an additional `WorkspaceId` predicate, assuming we'd have one in hand (eventually).  Then I realized the file was going to be covered with TODOs for this and backed off...but forgot and left this one behind!    I figure we can worry about that if/when we decide to do AJ-1662 and in the meantime will remove this one.

service/src/test/java/org/databiosphere/workspacedataservice/sam/SamDaoInvalidWorkspaceIdTest.java
```diff
@@ -0,0 +1,32 @@
...
+    })
+class SamDaoInvalidWorkspaceIdTest extends TestBase {
+
+  @Autowired SamDao samDao;
+
+  @Test
+  void statusThrows() {
+    assertThrows(RuntimeException.class, () -> samDao.getSystemStatus());
+  }
+}
```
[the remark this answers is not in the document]
jladieu: Removed this test entirely; it's passing but not for the right reason: 8416fcb85703d18113cd763c31978f5750a67f76 and the premise is no longer valid.

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamAuthorizationDao.java
```diff
@@ -0,0 +1,113 @@
...
+   */
+  @Override
+  public boolean hasDeleteCollectionPermission() {
+    return hasDeleteCollectionPermission(BearerToken.empty());
+  }
+
+  @Override
+  public boolean hasDeleteCollectionPermission(BearerToken token) {
+    return hasPermission(ACTION_DELETE, "Sam.hasDeleteCollectionPermission", token);
+  }
```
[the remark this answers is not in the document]
jladieu: Captured in a TODO: 0a9da2f22957fcf9f656e58c69f945c2c680c87e

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
```diff
@@ -120,11 +124,15 @@ public void validateCollection(UUID collectionId) {
   }
 
   public boolean canReadCollection(CollectionId collectionId) {
-    return samAuthorizationDao.hasReadWorkspacePermission(getWorkspaceId(collectionId));
+    return getSamAuthorizationDao(WorkspaceId.of(collectionId.id()))
```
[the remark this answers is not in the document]
jladieu: Yup, fixed in a subsequent commit.  Thanks for the attention to detail though!

My strategy was to make all the early commits pure non-functional refactorings, saving the actual functional changes for later on in the chain.

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamConfig.java
jladieu: Fixed in d01d81dfa7a247340a0713a22d9d8eb61b973c7f

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
```diff
@@ -94,7 +97,8 @@ public void deleteCollection(UUID collectionId, String version) {
     validateCollection(collectionId);
 
     // check that the current user has permission to delete the Sam resource
-    boolean hasDeleteCollectionPermission = samAuthorizationDao.hasDeleteCollectionPermission();
+    boolean hasDeleteCollectionPermission =
+        getSamAuthorizationDao().hasDeleteCollectionPermission();
```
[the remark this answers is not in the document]
jladieu: Deferred into a TODO, I might update the TODO with a fresh ticket number before done though.  Going to mark this comment as resolved and promise to track this in some external form.

jladieu:
> Instead of your (pseudo)code of collectionService.getWorkspaceId() followed by SamAuthorizationDao. hasWriteWorkspacePermission(), this should be just collectionService.canWriteCollection()

I'm so glad you saw this!  Doing this let me drop the `SamAuthorizationDaoFactory` entirely as a dependency of `RecordOrchestratorService`, entirely encapsulating that beneath `CollectionService`.  See: ca9a4cbfa8a7794561bc90f2aa421db5e338c9ce

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceDataPlaneTest.java
```diff
@@ -40,13 +45,21 @@ class ImportServiceDataPlaneTest {
...
   private final URI importUri = URI.create("http://does/not/matter");
   private final ImportRequestServerModel importRequest =
       new ImportRequestServerModel(PFB, importUri);
 
-  private final CollectionId collectionId = CollectionId.of(UUID.randomUUID());
+  private CollectionId collectionId; // initialized in setup to match the @SingleTenant WorkspaceId
+
+  @BeforeEach
+  void setup() {
+    collectionId = CollectionId.of(workspaceId.id());
```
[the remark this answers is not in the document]
jladieu: Fixed in b7723a575a5ee3adb5c523f52ef7df13617c78cb

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceDataPlaneTest.java
```diff
@@ -40,13 +45,21 @@ class ImportServiceDataPlaneTest {
...
+    when(samAuthorizationDaoFactory.getSamAuthorizationDao(any(WorkspaceId.class)))
+        .thenReturn(samAuthorizationDao);
```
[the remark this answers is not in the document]
jladieu: Fixed in 54c8116a47ee535f89667b0c6a6be09fa1a721d9

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceDataPlaneTest.java
```diff
@@ -39,14 +41,14 @@ class JobServiceDataPlaneTest extends JobServiceBaseTest {
...
+  private final SamAuthorizationDao samAuthorizationDao = spy(MockSamAuthorizationDao.allowAll());
 
   @BeforeEach
-  void beforeEach() {
-    reset(jobDao);
-    reset(collectionDao);
-    reset(samAuthorizationDao);
+  void setup() {
+    when(samAuthorizationDaoFactory.getSamAuthorizationDao(any(WorkspaceId.class)))
+        .thenReturn(samAuthorizationDao);
```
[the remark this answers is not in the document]
jladieu: Fixed in e593c010d417538738b5c7bb6a085410ad3ee216

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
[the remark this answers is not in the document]
jladieu: Pruned in a4899cf6dc83f9fdced54f0678da137382e6ded0

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamAuthorizationDao.java
jladieu: Fixed all three related comments in: 5126a9717a783badf0aa747ed6eded9fc28c6506

Thanks for noticing this!

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceControlPlaneTest.java
jladieu: Fixed in 566152e164758b697d5d4605a768bee363edd36b

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
jladieu: Fixed with 922535f29427c0f5c542f7df80cdc9cd829bbc2f

Also discovered that it's easy to mock out props objects with `@SpyBean`!  Might be able to cut down on some of our multiple `Test` files using this technique.

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamAuthorizationDao.java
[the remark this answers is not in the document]
jladieu: With this in mind...does it even make sense to have `hasCreateCollectionPermission` (which is identical in its implementation to `hasWriteWorkspacePermission` on `SamAuthorizationDao`?

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
davidangb: part of the benefit of using configuration classes instead of direct `@Value` references!

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceControlPlaneTest.java
```diff
@@ -116,4 +110,10 @@ void collectionExists() {
     assertThrows(
         CollectionException.class, () -> importService.createImport(collectionUuid, importRequest));
   }
+
+  private OngoingStubbing<Boolean> stubWriteWorkspacePermission(WorkspaceId workspaceId) {
+    when(samAuthorizationDaoFactory.getSamAuthorizationDao(workspaceId))
+        .thenReturn(samAuthorizationDao);
+    return when(samAuthorizationDao.hasWriteWorkspacePermission());
+  }
```
davidangb, line 118: I haven't tried this myself, so it may not work - you might also achieve this by declaring `@MockBean(reset = MockReset.NONE)`

jladieu: Not sure I understand the suggestion...I just pulled this out into a private method to make it read more fluently in the tests where it's used.

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamAuthorizationDao.java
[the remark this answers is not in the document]
jladieu: Thinking more about this (all for speculation and not for inclusion in this PR)...my hunch is `SamAuthorizationDao` methods should map 1:1 with the permission checks it's making to Sam, which I think is an argument for two methods only.

For the `canXyzCollection` methods (including `canCreate`), I think having them on `CollectionService` is a reasonable place right now, although I could also see a potential case for separating them out into their own `CollectionAuthorizationService` interface too :shrug:

I'm also imagining a future where every RESTful URL is nested under some RESTful convention:

`/workspace/<workspaceId>/collection/<collection-name>/`

In this case, I could see another similar pattern of of having `WorkspaceId` (and sometimes `CollectionId`) known up front, injected as request-scoped variables for auth and/or lookup concerns, and where they don't even need to show up as arguments in the `canXYZCollection` suite of methods.   Or, maybe even invocation of those methods is something that happens at the request-filter level before ever reaching controller code. 💭 💭 💭. Hearkening back to this experimental PR again: https://github.com/DataBiosphere/terra-workspace-data-service/pull/595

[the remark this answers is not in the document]
jladieu: Fixed the initial issue of this comment thread in https://github.com/DataBiosphere/terra-workspace-data-service/pull/602

jladieu:
> I suggest breaking this out into a separate PR/ticket - let's get this PR through and then figure out how to tackle the outstanding issue

Fixed the issue of needing `deleteCollection` to call `getWorkspaceId(collectionId)` in https://github.com/DataBiosphere/terra-workspace-data-service/pull/602

calypsomatic:
I think I'm fundamentally confused about what's going on here - there is no longer a `Sam[Authorization]Dao` for the entire WDS, but one is created by the factory as-needed for each call, or am I misreading it?  At least in the control plane we would have to create a new one per use if it's going to have a single `workspaceId` attached to it, or do we not check authorization for anything in cWDS and don't expect to?

jladieu:
> I think I'm fundamentally confused about what's going on here - there is no longer a Sam[Authorization]Dao for the entire WDS, but one is created by the factory as-needed for each call, or am I misreading it? At least in the control plane we would have to create a new one per use if it's going to have a single workspaceId attached to it, or do we not check authorization for anything in cWDS and don't expect to?

You've got it right.  The `Dao` gets created with information that is scoped to the request, so a new instance is factoried up just in time as its needed, primed with the `WorkspaceId` it needs to do auth checks.  We'll use it to check in both data-plane and control-plane.

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceControlPlaneTest.java
davidangb: ah, ignore this comment; I misread the code - my apologies!

calypsomatic (approved):

PR #595, opened 2024-02-28
WIP: experiment with request scoped bean, token holder

jladieu:
Inspired by: https://github.com/DataBiosphere/terra-workspace-data-service/pull/537#discussion_r1492773297

This is WIP/experimental, and I'm not planning to implement this with AJ-1633, but wanted to share a picture of what it might look like (if I could get it working properly).

This PR experiments with adding a "request-scoped" bean for `BearerTokenHolder`, which is populated by our existing `BearerTokenFilter`, and can be injected by Spring whenever we have a bean that _needs_ to know the BearerToken.

This is a more Spring-native approach to how we set and fetch the BearerToken in `TokenContextUtil`, as we can trust Spring to manage the thread state for us.  It also lets us avoid having to craft Factories just to provide a creation hook that we can use just in time.

What's demonstrated here is for BearerToken, but I could just as easily imagine having another RequestFilter that pays attention to the URI, extracts an instance identifier (or even a name!), and automatically fetches & populates the corresponding `WorkspaceId` for use with making Authorization checks.

And, I bet you could take it one step further, and codify Authorization as an up-front concern in your Controller classes, declaratively, using annotations, like:  `@Authorization(resource=WORKSPACE, action=WRITE)`, for example.


service/src/main/java/org/databiosphere/workspacedataservice/sam/SamConfig.java
```diff
@@ -43,8 +45,22 @@ SamClientFactory getSamClientFactory() {
   }
 
   @Bean
-  SamDaoFactory samDaoFactory(SamClientFactory samClientFactory, RestClientRetry restClientRetry) {
-    return new SamDaoFactory(samClientFactory, restClientRetry);
+  SamDaoFactory samDaoFactory(
+      SamClientFactory samClientFactory,
+      RestClientRetry restClientRetry,
+      BearerTokenHolder bearerTokenHolder) {
```
jladieu, line 51: The idea here is you can get a `BearerTokenHolder` injected for you, rather than having to grab it out of `TokenContextUtils` yourself when you need it.

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamConfig.java
```diff
@@ -43,8 +45,22 @@ SamClientFactory getSamClientFactory() {
...
+    return new SamDaoFactory(samClientFactory, restClientRetry, bearerTokenHolder);
+  }
+
+  /**
+   * A request-scoped bean that holds the BearerToken for the current request, set by {@link
+   * BearerTokenFilter}. Can be injected just like any other bean to get access to the token as part
+   * of bean configuration.
+   */
+  @Bean
+  @Scope(value = "request", proxyMode = ScopedProxyMode.TARGET_CLASS)
```
jladieu, line 61: This is the key to configuring this bean -- you mark it as request scoped, and it is able to generate a dynamic proxy class on your behalf, and inject that anywhere a bean demands a `BearerTokenHolder`.  That proxy class takes care of managing the interaction with your request-specific bean.

service/src/test/java/org/databiosphere/workspacedataservice/sam/StubBearerTokenHolder.java
```diff
@@ -0,0 +1,14 @@
+package org.databiosphere.workspacedataservice.sam;
+
+import org.springframework.boot.test.context.TestConfiguration;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Primary;
+
+@TestConfiguration
+public class StubBearerTokenHolder {
```
jladieu, line 8: This is a trick to stub it out for tests, which may not have a request context to initialize the `BearerTokenHolder`.

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamConfig.java
jladieu: I also bet using request-scoped beans you could completely _get rid of_ the SamDaoFactory, which exists purely to provide an injection-point for request-scoped information like `WorkspaceId` and `BearerToken`.

jladieu:
Closing for now, this is purely experimental.

PR #596, opened 2024-02-28
AJ-1586 cWDS sends pubsub to Rawls

calypsomatic:
https://broadworkbench.atlassian.net/browse/AJ-1586
After storing an import to a gcp bucket, cWDS will now send a pubsub message alerting rawls to the presence of data to upload.
Collects details needed for import into a new `ImportDetails` object; this is mostly needed for sending the information rawls needs in the pubsub, but helps a bit with the `RecordSinkFactory` interface.

For the time being, cWDS is using import-service's pubsub topic (will be updated to cWDS' own personal topic in https://broadworkbench.atlassian.net/browse/AJ-1677).  I struggled a bit with the `application.yaml` hierarchies and welcome suggestions on defining the project-id and topic in each yaml for each environment a little more cleanly.
Reminder:


service/src/main/resources/application.yml
```diff
@@ -69,6 +69,11 @@ spring:
     gcp:
       storage:
         enabled: false
+    #TODO arrange this in the right files
+      pubsub:
+        project-id: terra-importservice-dev
```
yuliadub, line 74: I pulled project-id out of bucket so it can be shared between bucket and pubsub, no need to add it again :) those live in the control-plane application yml

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDestinationDetails.java
```diff
@@ -0,0 +1,9 @@
+package org.databiosphere.workspacedataservice.dataimport;
+
+import java.util.UUID;
+
+public record ImportDestinationDetails(UUID jobId, String userEmail, UUID workspaceId) {
```
jladieu, line 5: Thinking out loud:

1) Structuring this record class:
* I think this should just become `ImportDetails` and house all the information gleaned from reading the `JobDataMap` that needs to be passed forward to the `RawlsRecordSink`.
* This should be created up front near the beginning of `executeInternal`
* Instead of housing `userEmail` it should just house the entire `UserInfo` object, which can then be used later on by `ActivityLogger` instead of having to go back out to `SamDao` to refetch.
* I don't think it should be possible to create an `empty()` one of these...is this just a shim that you plan to remove, or do you think we'll need it?

2) Since we have to look up UserInfo anyway, I wonder if the `UserInfo` object should be populated earlier, like in the ImportController, and serialized into the QuartzJob itself.  It might make it easier to troubleshoot if we could trace the user email all the way through.

calypsomatic: `empty()` was just a shim to satisfy the compiler with all the places that weren't actually using it.

service/src/main/resources/application-local.yml
```diff
@@ -1,6 +1,6 @@
 samurl: ${SAM_URL:https://sam.dsde-dev.broadinstitute.org/}
 datarepourl: https://jade.datarepo-dev.broadinstitute.org/
-workspacemanagerurl: https://workspace.dsde-dev.broadinstitute.org/
+workspacemanagerurl: ${WORKSPACE_MANAGER_URL:https://workspace.dsde-dev.broadinstitute.org/}
```
calypsomatic, line 3: Not particularly related to this PR, just ran into troubles with this when testing

service/src/test/resources/nginx.conf
```diff
@@ -41,5 +41,10 @@ http {
         location / {
                 root /usr/share/nginx/html; # Root for files
         }
+
+        	location ~ /register/user/ {
```
calypsomatic, line 45: When attempting to run through import locally, it would fail on WSM checks; this adds the relevant API url so that if you set WSM url to the nginx localhost you can get past that part.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -74,25 +82,52 @@ public void upsertBatch(
       ) throws BatchWriteException, IOException {
     ImmutableList.Builder<Entity> entities = ImmutableList.builder();
     records.stream().map(this::toEntity).forEach(entities::add);
-    jsonConsumer.accept(mapper.writeValueAsString(entities.build()));
+    // jsonConsumer is used primarily for testing; this line ensures that it does not break in real
+    // life
+    if (jsonConsumer != null) {
```
calypsomatic, line 87: `jsonConsumer` was null when I actually ran the app, so I needed to wrap it in this null check

service/src/main/resources/application-local.yml
```diff
@@ -13,6 +13,13 @@ spring:
   output:
     ansi:
       enabled: always
+  cloud:
+    gcp:
+      pubsub:
+        #TODO presumably developer will not have permissions here; what should we use?
+        # Using the import-service project and topic, will be updated to cWDS' own at a later date
+        project-id: import-service-dev
+        topic: import-service-notify-dev
```
yuliadub, line 22: could we pass the projectId/Topic via helm and just have these values in control plane config grabbing from enviorment variable, similar to how it is done with the storage bucket? that way all cWDS config lives with Terra helm.

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/ImportPubSub.java
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  // As noted in the PubSub superclass, this method encapsulates waiting for the pubsub library's
+  // CompletableFuture to complete. We could also return the CompletableFuture directly, but that
+  // would require callers to implement their own callbacks/waits/handlers. We could also create
+  // a separate `public CompletableFuture<String> publishAsync(...)` method so callers can
+  // choose between sync and async.
+  public String publishSync(String message) {
+    // PubSubTemplate.publish returns a future
+    CompletableFuture<String> publishFuture = this.pubSubTemplate.publish(fullTopicName, message);
```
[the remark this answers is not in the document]
calypsomatic: Hm it does [look like](https://github.com/broadinstitute/import-service/blob/a818658d0850818804235c5dc7f239c0de95c8cf/app/external/pubsub.py#L43) it's sent as attributes instead of data (I think the "data" of the message is `b''` and then `data` is the `attr` parameter).  I can make the change, but a) why is it like that? and b) presumably if I keep it the way it is, Rawls won't be able to read it?

service/src/main/resources/application.yml
calypsomatic: Turns out we need a different project id for pubsub

service/src/test/resources/nginx.conf
calypsomatic: Dang I must have forgotten what happened by the time I wrote this comment.  It looks like this is actually for the new `getUserEmail` that I added to put the user email in the pubsub message.

service/build.gradle
```diff
@@ -90,7 +91,10 @@ dependencies {
     implementation "org.apache.parquet:parquet-avro:$parquet_mr_version"
     implementation "org.apache.parquet:parquet-hadoop:$parquet_mr_version"
 
-    implementation 'com.google.cloud:spring-cloud-gcp-starter-storage:5.0.3'
+    implementation "com.google.cloud:spring-cloud-gcp-starter-storage:$spring_cloud_gcp_version"
+    implementation "com.google.cloud:spring-cloud-gcp-starter-pubsub:$spring_cloud_gcp_version"
+    implementation 'io.grpc:grpc-netty:1.62.2'
```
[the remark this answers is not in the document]
calypsomatic: I think it might not be required anymore with all the changes you added to make the tests not break!  Will verify and then either remove it or add the comment.

service/src/test/resources/nginx.conf
```diff
@@ -12,34 +12,43 @@ http {
...
         }
 
-        location ~ /api/workspaces/ {
-        	        types { } default_type "application/json";
-        			return 200 "{}";
-        		}
+        location ~ /register/user/ {
```
calypsomatic, line 38: Sorry for the noise, I just organized this file a bit so that we know what each API is for; the only real change in this PR is this `/register/user` was added for the Sam `getUserEmail` API when running locally

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
```diff
@@ -0,0 +1,12 @@
+package org.databiosphere.workspacedataservice.dataimport;
+
+import jakarta.annotation.Nullable;
+import java.util.UUID;
+
+public record ImportDetails(
+    @Nullable UUID jobId, @Nullable String userEmail, UUID collectionId, String prefix) {
```
jladieu, line 7: Is there any valid reason for these to ever be null?  If not, I think it'd be good to remove the `@Nullable` annotation here and fix upstream to prevent an `ImportDetails` from ever getting created with missing details.

jladieu: (I can help with this if it's a pain)

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/ImportPubSub.java
```diff
@@ -0,0 +1,34 @@
+package org.databiosphere.workspacedataservice.pubsub;
+
+import com.google.cloud.spring.pubsub.core.PubSubTemplate;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import org.databiosphere.workspacedataservice.annotations.DeploymentMode;
+
+@DeploymentMode.ControlPlane
```
jladieu, line 8: nit: I recommend using IDE to static import @DeploymentMode.ControlPlane so that you can just represent this as: `@ControlPlane`, which I think is WDS "common knowledge" enough without the `DeploymentMode` qualification.  This would also make it more consistent with the rest of the places we use the annotation.

In general, I usually try to use static imports to reduce inline code references down to the minimal level of qualification for the name to still be intention revealing.

So for another, maybe more interesting, example:

`BASE_ATTRIBUTES` <-- not really intention revealing, I don't know _what_ BASE_ATTRIBUTES actually is
`ImportMode.BASE_ATTRIBUTES` <-- I like this, I know it's talking about an import mode now, use this where I can
`RecordSource.ImportMode.BASE_ATTRIBUTES` <-- TMI, I don't need to know about its package structure

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
calypsomatic: Yes I specifically added @Nullable to satisfy sonar because all cases of record sinks now use `ImportDetails`, but only in `RawlsRecordSink` for PFBs does it have the jobId and userEmail to send to pubsub.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
jladieu: I understand this better now that I see where the code is at in this PR, sharing for the sake of explanation (no changes needed here)...

`RawlsRecordSink` is instantiated by `RawlsRecordSinkFactory` which isn't fixed to be null-safe stuff yet.  Therefore, it's able to create a `RawlsRecordSink` with a `null` `jsonConsumer`.

The right fix for this PR, without needing to change `RawlsRecordSinkFactory`, would be to mark `jsonConsumer` nullable in the constructor and in its declaration:

`@Nullable private final Consumer<String> jsonConsumer`

and

`RawlsRecordSink(...., @Nullable Consumer<String> jsonConsumer)`

If you want though, you can leave this alone, and I'll take care of it in a sweep I'm doing that will pick up `RawlsRecordSinkFactory` too.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -38,21 +41,26 @@ public class RawlsRecordSink implements RecordSink {
...
 
   /** Annotates a String consumer for JSON strings emitted by {@link RawlsRecordSink}. */
   @Target({ElementType.PARAMETER, ElementType.METHOD})
   @Retention(RetentionPolicy.RUNTIME)
   public @interface RawlsJsonConsumer {}
 
   RawlsRecordSink(
-      String attributePrefix,
       ObjectMapper mapper,
+      @RawlsJsonConsumer Consumer<String> jsonConsumer,
```
jladieu, line 54: Not necessarily related to this PR, but you can drop the `@RawlsJsonConsumer` annotation here, it should have been dropped when we added `RawlsRecordSinkFactory` and made that start doing the creation of `RawlsRecordSink`s

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -74,25 +82,51 @@ public void upsertBatch(
...
+    // jsonConsumer is used primarily for testing; this line ensures that it does not break in real
+    // life
+    if (jsonConsumer != null) {
+      jsonConsumer.accept(mapper.writeValueAsString(entities.build()));
+    }
 
-    // TODO(AJ-1661) - Would it make sense to propogate the JobId into here so that BlobName can be
-    // named
-    // with JobId as it is today with import service?
     if (storage != null) {
```
jladieu, line 91: I'm hoping the compiler is giving you a warning here telling you that this nullcheck isn't necessary...

calypsomatic: It is telling me it's unnecessary for both `storage` and `jsonConsumer`... except that `jsonConsumer` IS null sometimes

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -74,25 +82,51 @@ public void upsertBatch(
...
-
-    // AJ-1586 - the name of where in the bucket the file is inside blobName
   }
 
   @Override
   public void deleteBatch(RecordType recordType, List<Record> records) throws BatchWriteException {
     throw new UnsupportedOperationException("RawlsRecordSink does not support deleteBatch");
   }
 
+  private void publishToPubSub(UUID workspaceId, String user, UUID jobId, String upsertFile) {
```
jladieu, line 112: I recommend using `ImmutableMap.Builder` here to make the key-value pairs easier to discern:

```
Map<String, String> = new ImmutableMap.Builder<String, String>()
            .put("workspaceId", workspaceId.toString())
            .put("userEmail", user)
            ...
            ...
           .put("isCWDS", "true").build();
```

jladieu (approved):
A few minor suggestions, any of which you can defer if you want to merge this.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
jladieu: IMO I think we should just populate it completely in all the other spots too, and not leave it partially populated.  Okay if you want to punt this.

calypsomatic: Well in the other spots we just don't have the info, since it's used in synchronous import which doesn't have a job id, and anything that's not pubsub doesn't have a useremail.  So we would populate it with dummy data?  I went that route at first but it annoyed me.  I'll punt it and then you can secretly clean it up how you want next time you find a thin excuse in a PR.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -100,8 +101,12 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
     // Grab the manifest url from the job's data map
     JobDataMap jobDataMap = context.getMergedJobDataMap();
     URL url = getJobDataUrl(jobDataMap, ARG_URL);
+
+    // Collect details needed for import
     UUID targetCollection = getJobDataUUID(jobDataMap, ARG_COLLECTION);
 
+    ImportDetails importDetails = new ImportDetails(targetCollection, "tdr");
+
```
calypsomatic, line 109: I left this as the `targetCollection` instead of the `workspaceId`, sincehttps://github.com/DataBiosphere/terra-workspace-data-service/pull/622 kept `targetCollection` as the UUID passed into `importTables` - just verifying with @davidangb that that's the intended use.

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/ImportPubSub.java
```diff
@@ -0,0 +1,34 @@
+package org.databiosphere.workspacedataservice.pubsub;
+
+import com.google.cloud.spring.pubsub.core.PubSubTemplate;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import org.databiosphere.workspacedataservice.annotations.DeploymentMode.ControlPlane;
+
+@ControlPlane
```
davidangb, line 8: This probably doesn't need `@ControlPlane` any more, since bean creation is handled by `PubSubConfig`

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkFactory.java
```diff
@@ -20,9 +21,12 @@ public class RawlsRecordSinkFactory implements RecordSinkFactory {
 
   private final GcsStorage storage;
 
-  public RawlsRecordSinkFactory(ObjectMapper mapper, GcsStorage storage) {
+  private final PubSub pubSub;
+
+  public RawlsRecordSinkFactory(ObjectMapper mapper, GcsStorage storage, PubSub pubSub) {
     this.mapper = mapper;
     this.storage = storage;
+    this.pubSub = pubSub;
```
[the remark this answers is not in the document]
davidangb: heh Sonar is now giving a different warning about this, it wants:
```java
this.jsonConsumer = json -> {};
```

service/src/main/resources/application-control-plane.yml
```diff
@@ -8,9 +8,17 @@ twds:
 
   data-import:
     batch-write-record-sink: "rawls"
-    project-id: ${SERVICE_GOOGLE_PROJECT}
+    project-id: ${SERVICE_GOOGLE_PROJECT:broad-dsde-dev} #TODO is this a good default value?
```
davidangb, line 11: does it need a default? I'd rather see an error if it is not defined than try to use the wrong project. Same comment goes for setting a default on the topic name.

davidangb (approved):
There are a couple Sonar warnings that should be quick fixes. I have a few comments inline, none big.

I see errors about default credentials when starting with the `data-plane` profile. Do you think those are eliminatable? I am fine doing that as follow-on work, but it would be real nice to keep data-plane WDS clean.

Giving this a proactive :+1 since all my comments should be minor changes. Great work getting through this!

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/ImportPubSub.java
```diff
@@ -0,0 +1,34 @@
...
+  // As noted in the PubSub superclass, this method encapsulates waiting for the pubsub library's
+  // CompletableFuture to complete. We could also return the CompletableFuture directly, but that
+  // would require callers to implement their own callbacks/waits/handlers. We could also create
+  // a separate `public CompletableFuture<String> publishAsync(...)` method so callers can
+  // choose between sync and async.
+  public String publishSync(Map<String, String> message) {
+    // PubSubTemplate.publish returns a future
+    // Rawls expects the actual data to be in the headers rather than the payload
+    CompletableFuture<String> publishFuture =
+        this.pubSubTemplate.publish(fullTopicName, "b''", message);
```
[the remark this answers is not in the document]
calypsomatic: That's what it is in import service... I can set it to anything you want!

service/src/main/resources/application-control-plane.yml
calypsomatic: I think I was thinking about for local development since there wouldn't be a helmfile injecting the value, but I can just update the README to also tell you to make sure this is set

calypsomatic:
> I see errors about default credentials when starting with the `data-plane` profile. Do you think those are eliminatable? I am fine doing that as follow-on work, but it would be real nice to keep data-plane WDS clean.
> 

Argh... oh wait, that happens for me on `main` too... must be also from gcp storage?  I'm going to punt it for after this PR.

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/ImportPubSub.java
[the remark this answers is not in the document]
davidangb: oh heh … that's a python syntax for creating `bytes` instead of `string`! Since the python library used by Import Service expects bytes, `b''` means "the bytes that represent the empty string" - see https://www.geeksforgeeks.org/effect-of-b-character-in-front-of-a-string-literal-in-python/

PR #598, opened 2024-02-28
Talk about this test

jladieu:
Sharing this branch for discussion.


service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceDataPlaneTest.java
```diff
@@ -152,7 +152,7 @@ void collectionDoesNotExist() {
     when(collectionDao.getWorkspaceId(collectionId))
         .thenThrow(new EmptyResultDataAccessException("unit test intentional error", 1));
     // user has permission to that workspace
-    when(samDao.hasReadWorkspacePermission(workspaceId)).thenReturn(true);
+    when(samDao.hasReadWorkspacePermission(any(WorkspaceId.class))).thenReturn(true);
```
jladieu, line 155: If I switch this test case to stub more widely, it starts failing.

Is this intentional?  I'm having a hard time discerning the actual desired behavior here.

PR #599, opened 2024-02-28
AJ-1630: Throw error on missing collection

jladieu:
[AJ-1630](https://broadworkbench.atlassian.net/browse/AJ-1630): In the data-plane, when `twds.tenancy.allow-virtual-collections` is false, `CollectionService` should throw an error if a collection is missing.

### DIFF
--- a/service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceDataPlaneTest.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceDataPlaneTest.java
@@ -155,11 +155,11 @@ class JobServiceDataPlaneTest extends JobServiceBaseTest {
     when(samDao.hasReadWorkspacePermission(workspaceId)).thenReturn(true);
 
     // Act / assert
-    AuthenticationMaskableException actual =
-        assertThrows(AuthenticationMaskableException.class, () -> jobService.getJob(jobId));
+    MissingObjectException actual =
+        assertThrows(MissingObjectException.class, () -> jobService.getJob(jobId));
 
     // Assert
-    assertEquals("Job", actual.getObjectType());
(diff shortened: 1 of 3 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceDataPlaneTest.java
```diff
@@ -155,11 +155,11 @@ void collectionDoesNotExist() {
     when(samDao.hasReadWorkspacePermission(workspaceId)).thenReturn(true);
 
     // Act / assert
-    AuthenticationMaskableException actual =
-        assertThrows(AuthenticationMaskableException.class, () -> jobService.getJob(jobId));
+    MissingObjectException actual =
```
jladieu, line 158: I don't think we actually want an `AuthenticationMaskableException` here because the user _does_ have permission, the collection just doesn't exist.

calypsomatic (approved):

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceDataPlaneTest.java
davidangb: I think the end result here is fine - throwing `MissingObjectException` instead of `AuthenticationMaskableException`. But for clarity's sake: we haven't performed any permissions checks in this flow. Since the collection doesn't exist in the db, we couldn't determine the appropriate workspace id to check.

davidangb (approved):

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceDataPlaneTest.java
jladieu: Thanks for the clarification!

PR #600, opened 2024-02-29
AJ-1623 Fix AzureBlobStorage utf8 write

ashanhol:
Ticket: https://broadworkbench.atlassian.net/browse/AJ-1623

Problem: An "unexpected" character would cause restore to fail with error invalid byte sequence for encoding. However, we should absolutely be able to account for that character. 

Solution: Make sure our write stream to Azure is explicitly encoded in utf8. It uses "system default" otherwise, which seemed to be giving us issues. 

Test: https://github.com/broadinstitute/dsp-reusable-workflows/pull/26/

Original Workspace
image

Cloned Workspace with same uploaded tables
image

-----------------
Reminder:


service/src/main/java/org/databiosphere/workspacedataservice/storage/AzureBlobStorage.java
```diff
@@ -46,12 +48,14 @@ public void streamOutputToBlobStorage(
...
-                .getBlockBlobClient()
-                .getBlobOutputStream())) {
+    try (OutputStreamWriter blobOS =
+        new OutputStreamWriter(
+            new BufferedOutputStream(
+                blobContainerClient
+                    .getBlobClient(blobName)
+                    .getBlockBlobClient()
+                    .getBlobOutputStream()),
+            Charset.forName("UTF-8").newEncoder())) {
```
davidangb, line 58: nit: use `StandardCharsets.UTF_8.newEncoder()` instead

davidangb (commented):
nice! Is it possible to write a test for this case?

yuliadub (approved):
with e2e tests checking for this, looks good to me. Read a bunch of stuff on mocking blob storage but I am not confident that would actually give us enough confidence that it would work the same in a real blob storage in azure. (getting a test container in azure up and running to maintain for this one unit test doesnt seem worth it)

davidangb (approved):

jladieu (approved):

PR #602, opened 2024-03-01
AJ-1633: Finish wiring up `deleteCollection`

jladieu:
For [AJ-1633](https://broadworkbench.atlassian.net/browse/AJ-1633), branch based off https://github.com/DataBiosphere/terra-workspace-data-service/pull/594

* Stop checking `delete` workspace permission to determine if it's ok to delete a collection, instead check `write` workspace.
* Eliminate redundant `SamAuthorizationDao` methods which are subsumed by `SamAuthorizationDao#hasWriteWorkspacePermission`
* Adjust `SamPactTest` to no longer verify `delete` interaction and instead verify previously untested `read` interaction.
* Adjust a bunch of tests which were verifying the safety check behavior which is now removed.
* This allows wiring up `deleteCollection` to correctly look up the `WorkspaceId` for the collection and then check for `write` permission.
* Add `tenancy.enforce-collections-match-workspace-id` which is enabled on `data-plane` but can be disabled for stubborn tests that violate the rule it is checking.

[AJ-1633]: https://broadworkbench.atlassian.net/browse/AJ-1633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
@@ -16,8 +16,6 @@ import org.databiosphere.workspacedataservice.service.model.exception.Collection
 import org.databiosphere.workspacedataservice.service.model.exception.MissingObjectException;
 import org.databiosphere.workspacedataservice.shared.model.CollectionId;
 import org.databiosphere.workspacedataservice.shared.model.WorkspaceId;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.dao.EmptyResultDataAccessException;
 import org.springframework.http.HttpStatus;
@@ -34,8 +32,6 @@ public class CollectionService {
   private final TenancyProperties tenancyProperties;
 
   @Nullable private WorkspaceId workspaceId;
-
-  private static final Logger LOGGER = LoggerFactory.getLogger(CollectionService.class);
 
   public CollectionService(
       CollectionDao collectionDao,
@@ -88,11 +84,7 @@ public class CollectionService {
     validateVersion(version);
 
     // check that the current user has permission on the parent workspace
-    boolean hasCreateCollectionPermission =
-        getSamAuthorizationDao(workspaceId).hasCreateCollectionPermission();
-    LOGGER.debug("hasCreateCollectionPermission? {}", hasCreateCollectionPermission);
-
-    if (!hasCreateCollectionPermission) {
+    if (!canCreateCollection(workspaceId)) {
       throw new AuthorizationException("Caller does not have permission to create collection.");
     }
 
@@ -112,15 +104,8 @@ public class CollectionService {
     validateVersion(version);
     validateCollection(collectionId);
 
-    // TODO(AJ-1633): call getWorkspaceId(collectionId) to get the workspaceId, rather than relying
-    //   on the WORKSPACE_ID environment variable; test cases that rely on deleteCollection prevent
-    //   this change
     // check that the current user has permission to delete the Sam resource
-    boolean hasDeleteCollectionPermission =
-        getSamAuthorizationDao(workspaceId).hasDeleteCollectionPermission();
-    LOGGER.debug("hasDeleteCollectionPermission? {}", hasDeleteCollectionPermission);
-
-    if (!hasDeleteCollectionPermission) {
+    if (!canDeleteCollection(CollectionId.of(collectionId))) {
       throw new AuthorizationException("Caller does not have permission to delete collection.");
     }
 
@@ -147,6 +132,14 @@ public class CollectionService {
   }
 
   public boolean canWriteCollection(CollectionId collectionId) {
+    return getSamAuthorizationDao(getWorkspaceId(collectionId)).hasWriteWorkspacePermission();
+  }
+
+  private boolean canCreateCollection(WorkspaceId workspaceId) {
+    return getSamAuthorizationDao(workspaceId).hasWriteWorkspacePermission();
+  }
+
+  private boolean canDeleteCollection(CollectionId collectionId) {
     return getSamAuthorizationDao(getWorkspaceId(collectionId)).hasWriteWorkspacePermission();
   }
 
@@ -182,22 +175,6 @@ public class CollectionService {
         throw new MissingObjectException("Collection");
       }
     }
-
-    // safety check: if we found a workspace id in the db, it indicates we are in a data-plane
-    // single-tenant WDS. Verify the workspace matches the $WORKSPACE_ID env var.
-    // we must remove this check in a future multi-tenant WDS.
-    if (rowWorkspaceId != null && !rowWorkspaceId.equals(workspaceId)) {
-      // log the details, including expected/actual workspace ids
-      LOGGER.error(
-          "Found unexpected workspaceId for collection {}. Expected {}, got {}.",
-          collectionId,
-          workspaceId,
-          rowWorkspaceId);
-      // but, don't include workspace ids when throwing a user-facing error
-      throw new CollectionException(
-          "Found unexpected workspaceId for collection %s.".formatted(collectionId));
-    }
-
     return Objects.requireNonNullElseGet(rowWorkspaceId, () -> WorkspaceId.of(collectionId.id()));
   }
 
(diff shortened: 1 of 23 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
```diff
@@ -179,22 +176,6 @@ public WorkspaceId getWorkspaceId(CollectionId collectionId) {
         throw new MissingObjectException("Collection");
       }
     }
-
-    // safety check: if we found a workspace id in the db, it indicates we are in a data-plane
```
jladieu, line 183: Is it safe to remove this safety check yet, or do you think we need to put in more safeguards first?  My hunch is that since we're blocking the `@Deprecated` version of `createCollection` by disallowing creates for virtual collections & missing `$WORKSPACE_ID` that this should be safe now.

ashanhol: is there an observability way for us to check how often we'd be hitting the `CollectionException`? If so, why not keep the safety check in and monitor for a bit until we're sure?

service/src/test/java/org/databiosphere/workspacedataservice/service/CollectionServiceGetWorkspaceIdTest.java
```diff
@@ -74,13 +74,14 @@ void missingWorkspaceId() {
...
     // collection dao returns a value not equal to the workspace id set in $WORKSPACE_ID
+    WorkspaceId mismatchingWorkspaceId = WorkspaceId.of(UUID.randomUUID());
     when(mockCollectionDao.getWorkspaceId(any(CollectionId.class)))
-        .thenReturn(WorkspaceId.of(UUID.randomUUID()));
+        .thenReturn(mismatchingWorkspaceId);
 
     CollectionId collectionId = CollectionId.of(UUID.randomUUID());
     // expect getWorkspaceId to throw, since it found an unexpected workspace id
-    assertThrows(RuntimeException.class, () -> collectionService.getWorkspaceId(collectionId));
+    assertThat(collectionService.getWorkspaceId(collectionId)).isEqualTo(mismatchingWorkspaceId);
```
ashanhol, line 85: There's been a lot of changes so apologies if I'm a bit behind.  I know multitenant will allow mismatching workpsace/collection IDs, but single tenant does not. Where are we enforcing the matching case for single tenant?

service/src/test/java/org/databiosphere/workspacedataservice/service/CollectionServiceSamExceptionTest.java
```diff
@@ -38,19 +39,25 @@
...
- * with status code 401, they should throw AuthenticationException. - if Sam returns an ApiException
- * with status code 403, they should throw AuthorizationException. - if Sam returns an ApiException
- * with a well-known status code like 404 or 503, they should throw a RestException with the same
- * status code. - if Sam returns an ApiException with a non-standard status code such as 0, which
- * happens in the case of a connection failure, they should throw a RestException with a 500 error
- * code. - if Sam returns some other exception such as NullPointerException, they should throw a
- * RestException with a 500 error code.
+ * createCollection and deleteCollection to respond appropriately:
+ *
+ * <ul>
```
ashanhol, line 44: did this format automatically? if so, how did you get it to do that?

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceDataPlaneTest.java
```diff
@@ -141,5 +131,14 @@ void collectionDoesNotExist() {
     assertThrows(
         MissingObjectException.class,
         () -> importService.createImport(collectionUuid, importRequest));
+
+    verifyNoInteractions(samAuthorizationDaoFactory);
+    verifyNoInteractions(samAuthorizationDao);
+  }
+
+  private OngoingStubbing<Boolean> stubWriteWorkspacePermission(WorkspaceId workspaceId) {
```
ashanhol, line 139: :O i didn't know you could do this! That's so useful when you want some of the same whens but return different things in different cases

service/src/test/java/org/databiosphere/workspacedataservice/service/CollectionServiceGetWorkspaceIdTest.java
jladieu: Timely that you should raise this question :)

Linking a comment attached to the stanza of removals where I'm pulling that enforcement out: https://github.com/DataBiosphere/terra-workspace-data-service/pull/602#discussion_r1508431660

I think we need to remove that safeguard if we want both of the following to be simultaneously true:

- Tests can setup/teardown random collections as part of fixture setup
- CollectionService has internally consistent implementation for canDeleteCollection (look up WorkspaceId based on CollectionId when figuring out which workspace to use for auth)

service/src/test/java/org/databiosphere/workspacedataservice/service/CollectionServiceSamExceptionTest.java
jladieu: I wish...just toiled through it 😆

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceDataPlaneTest.java
jladieu: I know right?!  I discovered that as part of the big PR that preceded this one.  I like how it makes it easy to express multi-step stub chains as a nice, self-documenting single method call.

service/src/test/java/org/databiosphere/workspacedataservice/dao/MockCollectionDao.java
```diff
@@ -71,7 +72,7 @@ public void alterSchema(UUID sourceWorkspaceId, UUID workspaceId) {
 
   @Override
   public WorkspaceId getWorkspaceId(CollectionId instanceId) {
```
yuliadub, line 74: just cause this caught my eye - can you rename to collectionId instead of instanceId? (or do we even need it?)

jladieu:
I put the safety check back in, put some tests back the way they were before I removed it, and added `tenancy.enforce-collections-match-workspace-id` to disable the check in stubborn test(s).

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
jladieu: Added a kill switch which I only needed to use sparingly :shrug:

service/src/test/java/org/databiosphere/workspacedataservice/dao/MockCollectionDao.java
jladieu: 🦅 👁️

service/src/test/java/org/databiosphere/workspacedataservice/tsv/TsvErrorMessageTest.java
```diff
@@ -23,10 +23,16 @@
...
 import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.TestPropertySource;
 
 @ActiveProfiles(profiles = "mock-sam")
 @SpringBootTest
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@TestPropertySource(
+    properties = {
+      "twds.instance.workspace-id=123e4567-e89b-12d3-a456-426614174000",
+      "twds.tenancy.enforce-collections-match-workspace-id=false",
```
jladieu, line 34: This is the only test where I ended up needing the killswitch.  I was a bit surprised by this though because I encountered a whole bunch of these the first time I tried to make this change.  This leads me to the suspicion that there is/was some kind of cross-test pollution, but I couldn't repro it on this go-around :shrug:

service/src/main/resources/application.yml
```diff
@@ -132,6 +132,7 @@ twds:
   tenancy:
     allow-virtual-collections: false
     require-env-workspace: true
+    enforce-collections-match-workspace-id: true
```
davidangb, line 135: is it necessary to have this in both `application.yml` and `data-plane|control-plane.yml`? If the default is `true`, we could remove it from the data plane

davidangb (approved):
Minor comments inline, giving this a :+1:.

I don't like having the killswitch on the one test that needs it; that feels like we're doing something wrong. BUT, I also want to try to get away from refactoring/cleanup and focus on the cWDS MVP, so I am content to leave it as-is and save that for a future ticket.

service/src/main/resources/application.yml
jladieu: I don't think it _should_ be, but for reasons of wanting to just get this merged, I am just gonna leave it alone.

Based on my comment in `application.yml` attached to `tenancy` and `data-import`, I think these are purposefully duplicated, although the fact that I felt like I needed to does seem to point to a problem we can look at on a rainy day...

calypsomatic (approved):

jladieu:
FYI, I immediately encountered a problematic test after merging this, which leads me to think there's some kind of test pollution/ordering/seeding problem which necessitated introducing (and using) the kill switch from this PR: https://github.com/DataBiosphere/terra-workspace-data-service/pull/624

PR #603, opened 2024-03-01
AJ-1621: configure Sentry to distinguish between control-plane and data-plane

davidangb:
* tweak properties files to populate reasonable tags to send to Sentry
* refactor SentryInitializer a bit so we don't send any empty tags
* don't attempt to start Application Insights if we don't have a connection string

My test control-plane alert: [here](https://broad-institute.sentry.io/issues/5027707131/events/8c75ce7db7af488aae55cbe72405a7df/?project=4504299946835968&query=is%3Aunresolved+deploymentMode%3Acontrol-plane&referrer=next-event&statsPeriod=30d&stream_index=0)

My test data-plane alert: [here](https://broad-institute.sentry.io/issues/5027707131/events/a9e6708a15794eff9323192a027c6461/?project=4504299946835968&query=is%3Aunresolved+deploymentMode%3Acontrol-plane&referrer=previous-event&statsPeriod=30d&stream_index=0)


service/src/main/resources/application-control-plane.yml
```diff
@@ -1,6 +1,3 @@
-env:
-  wds:
-    deploymentMode: control-plane
```
davidangb, line 3: this is no longer widely needed, but I need to send it as a tag to Sentry, so I moved it under the sentry section

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
```diff
@@ -35,6 +34,12 @@ public class SentryInitializer {
   @Value("${sentry.mrg}")
   String mrg;
 
+  @Value("${sentry.env}")
+  String terraEnv;
+
+  @Value("${sentry.deploymentMode}")
+  String deploymentMode;
+
```
davidangb, line 42: I did not refactor all of this to use a config class instead of reading properties directly - do you think it's worth the refactor?

jladieu: I'm happy either way you wanna go.

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
```diff
@@ -43,7 +48,12 @@ public class SentryInitializer {
 
   @Bean
   public SmartInitializingSingleton initialize() {
-    String env = urlToEnv(samurl);
+    String env;
+    if (StringUtils.isNotBlank(terraEnv)) {
```
jladieu, line 52: nit: static import `isNotBlank` to shrink the code footprint?

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
```diff
@@ -52,11 +62,33 @@ public SmartInitializingSingleton initialize() {
...
+    if (StringUtils.isNotBlank(workspaceId)) {
+      tags.put("workspaceId", workspaceId);
+    }
+    // MRG, used in data plane
+    if (StringUtils.isNotBlank(mrg)) {
+      tags.put("mrg", mrg);
+    }
+    // deploymentMode, should be present in both control plane and data plane
+    if (StringUtils.isNotBlank(deploymentMode)) {
+      tags.put("deploymentMode", deploymentMode);
```
jladieu, line 87: Could you get this by scanning `@ActiveProfiles` for one or the other?  Or...maybe `StartupConfig` could do this for us (as a side effect of validating that a profile is set) and expose the deploymentMode as a @Bean?  That way we can still get rid of the `deploymentMode` in the yml.

jladieu: Oh I just got down to where this is teased off into its own `Sentry` subconfig.  I'm totally good with leaving this as is too now that it's not a global property.

jladieu (approved):
Couple small thoughts, everything optional.  Thanks for doing this!

service/src/main/resources/application-control-plane.yml
```diff
@@ -13,3 +10,9 @@ twds:
     batch-write-record-sink: "rawls"
     project-id: ${SERVICE_GOOGLE_PROJECT}
     rawls-bucket-name: ${SERVICE_GOOGLE_BUCKET}
+
+sentry:
+  mrg: # intentionally empty
+  releasename: "cWDS"
+  env: ${TERRA_ENV:}
+  deploymentMode: control-plane
```
calypsomatic, line 18: Just double-checking - nothing else was relying on `deploymentMode` living under `env.wds`?

davidangb: nope! It's passing tests, and I did a text search; nothing is using it. We discussed this back in https://github.com/DataBiosphere/terra-workspace-data-service/pull/558/files#r1498089817 (might have to expand that comment thread)

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
davidangb: yeah at this point it's sentry-specific, just a tag to help distinguish any alerts we get

calypsomatic (approved):

PR #609, opened 2024-03-02
AJ-1586: pubsub needs to handle futures

davidangb:
This is a PR into #596, not into `main`.

The pubsub library works in futures, not in synchronous calls. This is why tests look like they run forever. This PR handles the future and waits for it to complete, turning it into a synchronous call.

Note that the `ImportPubSubTest` is now failing, because publishing fails. Previously we never got the notification that publishing was failing.

### DIFF
--- a/service/src/test/java/org/databiosphere/workspacedataservice/pubsub/MockPubSub.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/pubsub/MockPubSub.java
@@ -2,7 +2,7 @@ package org.databiosphere.workspacedataservice.pubsub;
 
 public class MockPubSub implements PubSub {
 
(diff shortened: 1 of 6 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/pubsub/MockPubSub.java
```diff
@@ -2,7 +2,7 @@
 
 public class MockPubSub implements PubSub {
 
-  public void publish(String message) {
-    // no-op
+  public String publishSync(String message) {
+    return this.getClass().getName();
```
calypsomatic, line 6: Just curious, why return the classname?

calypsomatic (approved):

PR #610, opened 2024-03-02
AJ-1634: reduce log volume, refactor logging config

davidangb:
[AJ-1634](https://broadworkbench.atlassian.net/browse/AJ-1634)

Multiple logging-related changes in this PR:
1. Reduce logging volume from DEBUG to INFO most everywhere
2. Configure `DispatcherServlet` to continue to log at DEBUG so we get request URLs and response codes
3. Elevate a couple log messages to INFO from DEBUG so they're still visible
4. Reorganize `logback-spring.xml` to reduce duplication and nesting
5. When running with the `local` profile, re-enable the default Spring boot color-coded logs for developer goodness

NOT in this PR: filtering out /status/liveness and /status/readiness from the logs. This is somewhat involved, and I hope to tackle that in a follow-on PR. Is this worth doing?

`SPRING_PROFILES_ACTIVE=dev,control-plane gw bootRun`: 
dev,control-plane

`SPRING_PROFILES_ACTIVE=dev,data-plane gw bootRun`:
dev,data-plane

`SPRING_PROFILES_ACTIVE=local,control-plane gw bootRun`
and
`SPRING_PROFILES_ACTIVE=local,data-plane gw bootRun`:
local

[AJ-1634]: https://broadworkbench.atlassian.net/browse/AJ-1634?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/sam/SamAuthorizationDaoFactory.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/sam/SamAuthorizationDaoFactory.java
@@ -23,7 +23,7 @@ public class SamAuthorizationDaoFactory {
   }
 
   public SamAuthorizationDao getSamAuthorizationDao(WorkspaceId workspaceId) {
-    LOGGER.info(
+    LOGGER.debug(
         "Sam integration will query type={}, resourceId={}, actions={}",
         SamAuthorizationDao.RESOURCE_NAME_WORKSPACE,
         workspaceId,
--- a/service/src/main/resources/application-local.yml
+++ b/service/src/main/resources/application-local.yml
@@ -7,3 +7,14 @@ twds:
   instance:
     # Workspace Id for launching instance
     workspace-id: ${WORKSPACE_ID:123e4567-e89b-12d3-a456-426614174000}
+
+# enable color-coding for logs
+spring:
+  output:
+    ansi:
+      enabled: always
+# use the MDC "requestId" as the log correlation pattern. If/when we use Micrometer tracing to
+# create correlation ids, this should be removed.
+logging:
+  pattern:
+    correlation: "%X{requestId} "
--- a/service/src/main/resources/logback-spring.xml
+++ b/service/src/main/resources/logback-spring.xml
@@ -40,8 +40,7 @@
       <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
         <level>${CONSOLE_LOG_THRESHOLD}</level>
       </filter>
-      <encoder class="net.logstash.logback.encoder.LogstashEncoder"/>
-      <charset>${CONSOLE_LOG_CHARSET}</charset>
+      <encoder class="net.logstash.logback.encoder.LogstashEncoder" charset="UTF-8"/>
     </appender>
   </springProfile>
 
--- a/service/src/main/resources/logback-spring.xml
+++ b/service/src/main/resources/logback-spring.xml
@@ -1,10 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration>
-  
+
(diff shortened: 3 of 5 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/sam/SamAuthorizationDaoFactory.java
```diff
@@ -23,7 +23,7 @@ public SamAuthorizationDaoFactory(
   }
 
   public SamAuthorizationDao getSamAuthorizationDao(WorkspaceId workspaceId) {
-    LOGGER.info(
+    LOGGER.debug(
```
davidangb, line 26: In the recent refactors of Samstuff, this message is now logged for *every* Sam request, so I have moved it to debug.

service/src/main/resources/application-local.yml
```diff
@@ -7,3 +7,14 @@ twds:
...
+# enable color-coding for logs
+spring:
+  output:
+    ansi:
+      enabled: always
+# use the MDC "requestId" as the log correlation pattern. If/when we use Micrometer tracing to
+# create correlation ids, this should be removed.
+logging:
+  pattern:
+    correlation: "%X{requestId} "
```
davidangb, line 20: The `spring.output.ansi.enabled` default is `detect`, but that seems to have a problem in MacOS/Terminal, so I've just set it to `always`. Since this only takes effect when running with the "local" profile, this seems safe.

service/src/main/resources/logback-spring.xml
```diff
@@ -1,82 +1,83 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration>
-  
+
```
davidangb, line 3: this file has a lot of diffs. A high-level summary of changes:
* local profile and non-local profile now use the same `<logger>` definitions for consistency. This means the local profile includes the `Sentry` appender-ref, but since `SentryInitializer` turns off sentry for the local env, this is a non-issue.
* only include "org/springframework/boot/logging/logback/console-appender.xml" for the local profile. That include defines the Spring default `CONSOLE` appender. In non-local profiles we overrode that, which means before this PR we defined it as the spring default and then immediately redefined it.
* flatten the nested `<springProfile name="!local"><springProfile name="control-plane">` clauses into what I hope is more readable single clause `<springProfile name="!local &amp; control-plane">`
* and finally the point of this PR: set most `<logger>`s to INFO instead of DEBUG

yuliadub (approved):

calypsomatic (approved):
> NOT in this PR: filtering out /status/liveness and /status/readiness from the logs. This is somewhat involved, and I hope to tackle that in a follow-on PR. Is this worth doing?

Yes I would *love* to have those filtered out!

service/src/main/resources/logback-spring.xml
```diff
@@ -1,82 +1,83 @@
...
-        </filter>
-        <encoder class="net.logstash.logback.encoder.LogstashEncoder"/>
+  <!-- when running locally, use the Spring Boot default logging pattern. Note we also include
+         a logging.pattern.correlation override in application-local.yml. -->
+  <springProfile name="local">
+    <include resource="org/springframework/boot/logging/logback/console-appender.xml"/>
+  </springProfile>
+
+  <!-- in the data plane, log in plaintext (with no colors) -->
+  <springProfile name="!local &amp; data-plane">
```
jladieu, line 30: Not for this PR, but might it make sense to have an explicit config or profile dictating which appender to use?

eg:

`@Profile("json-logs")`, `@Profile("plaintext-logs")`

or:

`some.env.logging.default.format=plaintext|json`

jladieu (approved):
Nice cleanup!

PR #618, opened 2024-03-04
AJ-1649: update banner

davidangb:
adds a [custom banner](https://www.baeldung.com/spring-boot-custom-banners)


calypsomatic (approved):
Aren't you supposed to be taking more time OFF

jladieu (approved):

PR #620, opened 2024-03-05
AJ-1673: pfb import must use collection's workspaceid

davidangb:
[AJ-1673](https://broadworkbench.atlassian.net/browse/AJ-1649)

for PFB imports, when linking snapshots, use the workspaceId returned from CollectionService instead of the $WORKSPACE_ID env var. In the control plane, that env var will be missing.

Furthermore, inside the linkSnapshots code, ensure that WorkspaceManagerDao accepts a workspaceId argument on each method; WorkspaceManagerDao should not use the $WORKSPACE_ID env var either.

This is an interim fix to unblock other work.

Still todo: implement the same for TDR snapshot import, which likely means consolidating/reusing code. This PR is already kinda big; I'd like to tackle that separately.

[AJ-1673]: https://broadworkbench.atlassian.net/browse/AJ-1673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJob.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJob.java
@@ -210,7 +210,7 @@ public class PfbQuartzJob extends QuartzJob {
     // list existing snapshots linked to this workspace
     WsmSnapshotSupport wsmSnapshotSupport =
         new WsmSnapshotSupport(workspaceId.id(), wsmDao, restClientRetry, activityLogger);
-    wsmSnapshotSupport.linkSnapshots(snapshotIds);
+    wsmSnapshotSupport.linkSnapshots(workspaceId, snapshotIds);
   }
 
   private UUID maybeUuid(String input) {
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJob.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJob.java
@@ -28,15 +28,17 @@ import org.databiosphere.workspacedataservice.recordsource.RecordSource.ImportMo
 import org.databiosphere.workspacedataservice.recordsource.RecordSourceFactory;
 import org.databiosphere.workspacedataservice.retry.RestClientRetry;
 import org.databiosphere.workspacedataservice.service.BatchWriteService;
+import org.databiosphere.workspacedataservice.service.CollectionService;
 import org.databiosphere.workspacedataservice.service.model.BatchWriteResult;
 import org.databiosphere.workspacedataservice.service.model.exception.PfbParsingException;
+import org.databiosphere.workspacedataservice.shared.model.CollectionId;
 import org.databiosphere.workspacedataservice.shared.model.RecordType;
+import org.databiosphere.workspacedataservice.shared.model.WorkspaceId;
 import org.databiosphere.workspacedataservice.workspacemanager.WorkspaceManagerDao;
 import org.quartz.JobDataMap;
 import org.quartz.JobExecutionContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
 /** Shell/starting point for PFB import via Quartz. */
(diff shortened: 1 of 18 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJob.java
```diff
@@ -89,6 +91,9 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
     URL url = getJobDataUrl(jobDataMap, ARG_URL);
     UUID targetCollection = getJobDataUUID(jobDataMap, ARG_COLLECTION);
 
+    // determine the workspace for the target collection
+    WorkspaceId workspaceId = collectionService.getWorkspaceId(CollectionId.of(targetCollection));
```
yuliadub, line 95: is this under the assumption that collectionId = workspaceId (since we only support a single colelction)? could that ever not be true in the future?

jladieu (approved):
Looks good to unblock cWDS progress!

General design thought, I think this calls out for the same treatment as `SamAuthorizationDao` which is to configure `WSMDao` up front with a scoping `WorkspaceId`, at creation time, and drop it from all the method signatures.

The design signal I look for to trigger such a change is any time you have an API where the majority of methods take in the same variable or list of variables, they could probably be encapsulated instead as an instance variable.  Okay to punt this though!

yuliadub:
> Looks good to unblock cWDS progress!
> 
> General design thought, I think this calls out for the same treatment as `SamAuthorizationDao` which is to configure `WSMDao` up front with a scoping `WorkspaceId`, at creation time, and drop it from all the method signatures.
> 
> The design signal I look for to trigger such a change is any time you have an API where the majority of methods take in the same variable or list of variables, they could probably be encapsulated instead as an instance variable. Okay to punt this though!

I like that (i.e. configuring WSMDao up front with a scoping WorkspaceId)

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJob.java
davidangb: It will absolutely be true that collectionId != workspaceId in the future. For instance, as soon as we support workspace data/reference data in (data-plane) WDS, we'll have multiple collections per workspace; necessarily some of those collections will have collectionId != workspaceId.

Higher-level, we want to just do away with any assumptions. We should always just ask `CollectionService` to tell us the appropriate workspaceId for any given collection. Then, all logic is encapsulated in `CollectionService.getWorkspaceId` and callers never have to think about it.

yuliadub: got it! for a moment I forgot how collectionService.getWorkspaceId() actually worked, I was just looking around to remind myself 😄

yuliadub (approved):
I'd say it makes sense to apply Josh's suggestion now -- but this also LGTM if we want to push this in as is.

service/src/test/java/org/databiosphere/workspacedataservice/pact/WsmPactTest.java
```diff
@@ -143,8 +145,10 @@ void linkSnapshotForPolicy_conflict(MockServer mockServer) {
     var wsmDao = buildWsmDao(mockServer);
     var snapshotModel = buildSnapshotModel();
 
+    WorkspaceId workspaceId = WorkspaceId.of(WORKSPACE_UUID);
```
davidangb, line 148: this awkward extra var makes Sonar happy

service/src/test/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDaoTest.java
```diff
@@ -85,10 +88,11 @@ void testErrorThrown() throws ApiException {
         new SnapshotModel().name("test snapshot").id(UUID.randomUUID());
     given(mockReferencedGcpResourceApi.createDataRepoSnapshotReference(any(), any()))
         .willThrow(new ApiException(statusCode, "Intentional error thrown for unit test"));
+    WorkspaceId wid = WorkspaceId.of(workspaceId);
```
davidangb, line 91: this awkward extra var makes Sonar happy

jladieu (approved):

service/src/test/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDaoTest.java
jladieu: I really dislike the sonar warning that makes us do this, but I totally understand why so can't do anything other than grumble.

PR #623, opened 2024-03-06
AJ-1669: Fix warnings near `BatchWriteService`

jladieu:
This is cleanup in prep for making changes to `BatchWriteService` and `RecordSink` to only write one file out to storage at the end of an entire batch for [AJ-1669](https://broadworkbench.atlassian.net/browse/AJ-1669).

Eliminates most compiler warnings about nullability, as well as other minor syntax issues in these classes and all their adjacent collaborators.

The process I took to end up with this scope is as follows:
1) I started at `RawlsRecordSink` and `BatchWriteService` and fixed warnings there.
2) Nullability problems can be fixed locally but need to be traced upward/downward to ensure they are fixed at the root, so I searched backward for references to these two classes and fixed warnings in those files.
3) I opportunistically fixed other easy warnings in touched files, but left a few behind which would be annoying, harm readability, or were too much of a detour to fix correctly.
4) Go back to step 1 and recurse until all referencing classes are fixed.

Got lucky here with this process as the tree of dependents wasn't that broad, but admittedly this process could turn into a fanout nightmare for other more central dependencies.

Lessons learned: 
* It's much easier & safer to mark something as `@Nullable` to essentially accept the status quo and defer until later than it is to deal with deeply making something `@NonNull`.
* Marking code as `@Nullable` is a way of signaling future developers and the IDE to exercise caution, and so should be used liberally where needed to push the safety checks outward.  It's also a breadcrumb for future improvement (most `@Nullable` markings indicate an opportunity for improvement).
* Making something satisfy `@NonNull` provides benefits to surrounding code but requires care and should be done with intention.
* Can't trust nullability of all surrounding code, like our model classes, which are _not_ annotated as null-safe and don't give the compiler any hints to that effect.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/WsmSnapshotSupport.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/WsmSnapshotSupport.java
@@ -191,6 +191,7 @@ public class WsmSnapshotSupport {
                 tableModel -> identifyPrimaryKey(tableModel.getPrimaryKey())));
   }
 
+  // snapshotKeys comes from generated code which doesn't guarantee null safety
   String identifyPrimaryKey(@Nullable List<String> snapshotKeys) {
     if (snapshotKeys != null && snapshotKeys.size() == 1) {
       return snapshotKeys.get(0);
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/WsmSnapshotSupport.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/WsmSnapshotSupport.java
@@ -27,6 +27,7 @@ import org.databiosphere.workspacedataservice.shared.model.WorkspaceId;
 import org.databiosphere.workspacedataservice.workspacemanager.WorkspaceManagerDao;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.lang.Nullable;
 
 public class WsmSnapshotSupport {
 
@@ -122,6 +123,7 @@ public class WsmSnapshotSupport {
    * @param resourceDescription the WSM object in which to find a snapshotId
    * @return the snapshotId if found, else null
    */
+  @Nullable
   protected UUID safeGetSnapshotId(ResourceDescription resourceDescription) {
     ResourceAttributesUnion resourceAttributes = resourceDescription.getResourceAttributes();
     if (resourceAttributes != null) {
@@ -189,7 +191,7 @@ public class WsmSnapshotSupport {
                 tableModel -> identifyPrimaryKey(tableModel.getPrimaryKey())));
   }
 
-  String identifyPrimaryKey(List<String> snapshotKeys) {
+  String identifyPrimaryKey(@Nullable List<String> snapshotKeys) {
     if (snapshotKeys != null && snapshotKeys.size() == 1) {
       return snapshotKeys.get(0);
     }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
@@ -149,14 +149,13 @@ public class TdrManifestQuartzJob extends QuartzJob {
     result
         .entrySet()
         .forEach(
-            entry -> {
-              activityLogger.saveEventForCurrentUser(
-                  user ->
-                      user.upserted()
-                          .record()
-                          .withRecordType(entry.getKey())
-                          .ofQuantity(entry.getValue()));
-            });
+            entry ->
+                activityLogger.saveEventForCurrentUser(
+                    user ->
+                        user.upserted()
+                            .record()
+                            .withRecordType(entry.getKey())
+                            .ofQuantity(entry.getValue())));
     // delete temp files after everything else is completed
     // Any failed deletions will be removed if/when pod restarts
     fileDownloadHelper.deleteFileDirectory();
@@ -225,9 +224,7 @@ public class TdrManifestQuartzJob extends QuartzJob {
                       // generate the HadoopInputFile
                       InputFile inputFile = HadoopInputFile.fromPath(hadoopFilePath, configuration);
                       var result = importTable(inputFile, importTable, importMode, importDetails);
-                      if (result != null) {
-                        combinedResult.merge(result);
-                      }
+                      combinedResult.merge(result);
                     } catch (IOException e) {
                       throw new TdrManifestImportException(e.getMessage(), e);
                     }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
@@ -1,5 +1,7 @@
 package org.databiosphere.workspacedataservice.recordsink;
 
+import static java.util.Objects.requireNonNull;
+
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
@@ -16,7 +18,6 @@ import java.util.Objects;
 import java.util.UUID;
 import java.util.function.Consumer;
 import java.util.stream.Stream;
-import org.databiosphere.workspacedataservice.config.ConfigurationException;
 import org.databiosphere.workspacedataservice.dataimport.ImportDetails;
 import org.databiosphere.workspacedataservice.pubsub.PubSub;
 import org.databiosphere.workspacedataservice.recordsink.RawlsModel.AddListMember;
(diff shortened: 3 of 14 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -143,14 +143,13 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
...
         .forEach(
-            entry -> {
-              activityLogger.saveEventForCurrentUser(
-                  user ->
-                      user.upserted()
-                          .record()
-                          .withRecordType(entry.getKey())
-                          .ofQuantity(entry.getValue()));
-            });
+            entry ->
```
jladieu, line 146: This was just a warning about being able to change this to a lambda expression (drop the curlies)

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobDataPlaneE2ETest.java
```diff
@@ -73,7 +77,7 @@ class PfbQuartzJobDataPlaneE2ETest {
   @BeforeEach
   void beforeEach() {
     collectionId = UUID.randomUUID();
-    collectionService.createCollection(collectionId, "v0.2");
+    collectionService.createCollection(workspaceId, CollectionId.of(collectionId), "v0.2");
```
jladieu, line 80: This is just a migration away from the deprecated overload of `createCollection`

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceTest.java
```diff
@@ -141,12 +135,12 @@ void persistsJobAsQueued(ImportRequestServerModel.TypeEnum importType) {
     // schedulerDao.schedule(), which returns void, returns successfully
     doNothing().when(schedulerDao).schedule(any(Schedulable.class));
     // create collection (in the MockCollectionDao)
-    collectionService.createCollection(defaultCollectionId, VERSION);
+    collectionService.createCollection(workspaceId, defaultCollectionId(), VERSION);
```
jladieu, line 138: these are all migrations away from the deprecated overload of `createCollection`

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/WdsSnapshotSupportTest.java
```diff
@@ -46,7 +45,7 @@ class WsmSnapshotSupportTest extends TestBase {
 
   @ParameterizedTest(name = "paginates through results when WSM has {0} references")
   @ValueSource(ints = {0, 1, 49, 50, 51, 99, 100, 101, 456})
-  void paginateExistingSnapshots(int wsmCount) throws ApiException {
```
jladieu, line 49: unused/unthrown exception

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/WsmSnapshotSupport.java
```diff
@@ -189,7 +191,7 @@ public Map<RecordType, String> identifyPrimaryKeys(List<TableModel> tables) {
                 tableModel -> identifyPrimaryKey(tableModel.getPrimaryKey())));
   }
 
-  String identifyPrimaryKey(List<String> snapshotKeys) {
+  String identifyPrimaryKey(@Nullable List<String> snapshotKeys) {
```
jladieu, line 194: This one was tricky; it looked like it should have been okay to treat this as `@NonNull` but it received its value from `tableModel` which is generated code and doesn't make the claim of being a non-nullable API, so we can't (and shouldn't) trust it as such.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -34,8 +35,6 @@
 /**
  * {@link RecordSink} implementation that produces Rawls-compatible JSON using {@link RawlsModel}
  * serialization classes.
- *
- * <p>TODO(AJ-1586): integrate with pubsub to notify Rawls when JSON is ready to be processed
```
jladieu, line 38: Just noticed this TODO was still hanging out but this is already done!

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobControlPlaneE2ETest.java
```diff
@@ -50,8 +50,10 @@
 
 /**
  * Tests for PFB import that execute "end-to-end" - that is, they go through the whole process of
- * parsing the PFB, and generating the JSON that (will eventually be) stored in a bucket (AJ-1585)
- * and communicated to Rawls via pubsub (AJ-1586).
+ * parsing the PFB, and generating the JSON that gets stored in a bucket and communicated to Rawls
+ * via pubsub.
+ *
+ * <p>TODO(AJ-1669): Add coverage for bucket storage & pubsub
```
jladieu, line 56: Since I saw the other defunct TODO I took a quick look around and found that this Javadoc also needed an update.  I included a TODO to add some coverage to this E2E test here to verify expected behavior.  Hoping to cover it in a later PR for this ticket.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/WsmSnapshotSupport.java
```diff
@@ -122,6 +123,7 @@ protected ResourceList listAllSnapshots(int pageSize) {
    * @param resourceDescription the WSM object in which to find a snapshotId
    * @return the snapshotId if found, else null
    */
+  @Nullable
```
yuliadub, line 126: this means the return of this function is `@Nullable?`, correct?

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -85,20 +84,16 @@ public void upsertBatch(
     records.stream().map(this::toEntity).forEach(entities::add);
     jsonConsumer.accept(mapper.writeValueAsString(entities.build()));
 
-    if (storage != null) {
```
yuliadub, line 88: storage can be null (although it shouldn't happen, it'd be helpful to know if/when it does) -- is there a reason why you removed this?

yuliadub:
curious what other think about this - everywhere where we are adding `@Nullable?` I think it'd be helpful to have a comment that summarizes when it is okay for that to be nullable vs when we expect it to have a value. I think that can be helpful in the future when debugging issues since I can see myself being confused unless I was looking at the full context.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/WsmSnapshotSupport.java
jladieu: Yeah this is saying that this function _may_ return `null`, or more importantly, it can't make a guarantee that it will always return non-null :)

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
jladieu: The purpose of introducing `@NonNullApi` and `@NonNullFields` is if you annotate your code comprehensively (and follow the IDE's warnings), you don't need to code defensively against nullness because the IDE will prevent you from authoring code that lets nulls slip through.  It lets you eliminate null checks like this.  But you can't trust it until you've traced back all the creation paths and ensure that all the code leading up here is null safe, which is why I searched backward for warnings and addressed them as I found them.

There _are_ [gotchas](https://github.com/DataBiosphere/terra-workspace-data-service/pull/623#discussion_r1515339036), but if you follow `storage` back through its instantiation path, we are the only callers of the constructor, and all the callsites are observing null safety.  There shouldn't be any way for it to ever be `null` (anymore).  Under normal circumstances, Spring will crash before it instantiates a `RawlsRecordFactory` without all of its dependencies.

jladieu:
> curious what other think about this - everywhere where we are adding `@Nullable`? I think it'd be helpful to have a comment that summarizes when it is okay for that to be nullable vs when we expect it to have a value. I think that can be helpful in the future when debugging issues since I can see myself being confused unless I was looking at the full context.

This is a really good idea.  I'll take a pass back through this PR and give some explanation where I've sprinkled `@Nullable`; ideally, eventually, `@Nullable` becomes an exception rather than the norm, and documenting _why_ will make it easier to undo as well :)

calypsomatic (approved):

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobDataPlaneE2ETest.java
davidangb: NOT for this PR - we should be careful about false confidence in the `createCollection(workspaceId …` method; it does not write the passed `workspaceId` to the db, so it still writes the `$WORKSPACE_ID` env var to the db for each collection row just like the deprecated method. We'll have to solve that before supporting full multi-tenancy but definitely not needed as part of the cWDS "write one file per import" effort

davidangb (approved):

jladieu:
>Either way, @yuliadub wdyt of the comments I added in? Is there enough info in there to be useful? Do you think the ones that are missing comments would benefit from similar treatment or do you think it's reasonable enough to infer why they're @Nullable from the context in the other spots?

Chatted offline with Yulia and got her ok to proceed on this PR.

PR #624, opened 2024-03-06
Disable workspaceId check for `LogStatementTest`

jladieu:
This test is failing because it is creating/encountering collections that have a workspaceId that mismatches the `WORKSPACE_ID`.  Disabling this check allows these tests to keep passing (and verifying what they are actually interested in), without having to redesign them for the concern of the safety check.


calypsomatic (approved):

jladieu:
Added a TODO to https://broadworkbench.atlassian.net/browse/AJ-1682 to not require these kill switches.

davidangb (approved):

PR #626, opened 2024-03-07
AJ-1666 Report PR images to sherlock

calypsomatic:
https://broadworkbench.atlassian.net/browse/AJ-1666
Before this PR:
The `publish-docker` workflow:
1. requires the input `input-tag`
2. is only called by the `tag` workflow
3. tags the new docker image with both the `input-tag` sent from `tag` and the github short hash.

The `tag` workflow, which is called manually for release:
1. discovers the new semver tag (e.g. 0.2.117)
2. passes the new tag to the `publish-docker` workflow
4. passes the new tag to sherlock to report the existence of the new docker image.

After this PR:
The `publish-docker` workflow:
1. does not require the input `input-tag`
2. is called both by the `tag` workflow and upon PR
3. If `input-tag` is present:
-  tags the new docker image with both the semver `input-tag` sent from `tag` and the github short hash
- reports to sherlock the existence of the new docker image with the semver tag

If `input-tag` is not present:
-  tags the new docker image only with the github short hash
- reports to sherlock the existence of the new docker image using the github short hash

The `tag` workflow:
1. discovers the new semver tag
2. passes the new tag to the `publish-docker` workflow
3. No longer calls report-to-sherlock, as this is now accomplished in `publish-docker`.

Reminder:

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -114,6 +114,14 @@ jobs:
           --password-stdin
           docker push --all-tags ${{ steps.acr-image-name.outputs.name }}
 
+  verify-new-version:
+    runs-on: ubuntu-latest
+    needs: publish-docker-job
+    steps:
+      - name: Verify new version
+        run: |
+          echo "The newVersion is: ${{ needs.publish-docker-job.outputs.newVersion }}"
+
   # Report the new version of cWDS to Sherlock.
   # This enables Sherlock to gather metrics and generate changelogs for deployments.
   cwds-report-to-sherlock:
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -87,8 +87,8 @@ jobs:
 
       - name: Add version tag to GCR
         run: |
-          if [ -n "${{ inputs.new-tag }}" ]; then
-            docker image tag  ${{ steps.gcr-image-name.outputs.name }}:${{ inputs.new-tag }}
+          if [ ${{ inputs.new-tag != '' }} ]; then
+            docker image tag ${{ steps.gcr-image-name.outputs.name }}:${{ inputs.new-tag }}
           fi
           docker image tag ${{ steps.gcr-image-name.outputs.name }}:${{ steps.setHash.outputs.git_short_sha }} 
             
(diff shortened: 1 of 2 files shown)


jladieu:
Thank you for the writeup describing the before/after state.  One question I still have: can you explain what the presence or absence of inputTag actually means?  Under what circumstances do we omit it or provide it, and who does the providing?

davidangb (approved):
did not test it myself, but looks great, thank you!

.github/workflows/publish-docker.yml
```diff
@@ -38,6 +42,10 @@ jobs:
           echo $git_short_sha
           echo "git_short_sha=${git_short_sha}" >> $GITHUB_OUTPUT
 
+      - name: Determine new version
+        id: newVersion
+        run: echo "newVersion=${{ inputs.new-tag || steps.setHash.outputs.git_short_sha }}" >> "$GITHUB_OUTPUT"
```
jladieu, line 47: What does a value of `newVersion` look like?  I'm assuming it's something like `0.1.2wiopiwoipwo`.

Do we use that convention anywhere else?  I wonder if there should be a delimiter in between the tag & sha?

calypsomatic:
> Thank you for the writeup describing the before/after state. One question I still have: can you explain what the presence or absence of inputTag actually means? Under what circumstances do we omit it or provide it, and who does the providing?

@jladieu The inputTag is the semvar generated upon calling the `tag` workflow for a release - e.g. 0.2.117 when previously we had 0.2.116.  So we only have this when we call the 'tag' workflow and are releasing a new version.  Whereas, on a PR, we haven't updated the version and have no new semvar to pass into the `publish-docker` workflow, leaving only the git commit hash to distinguish the new docker image.  Is that clearer?

jladieu (approved):

.github/workflows/publish-docker.yml
calypsomatic: It's either the semvar - e.g. 0.2.117 - or the git hash - e.g. 4b57b53.  The docker images are always tagged with the git hash.  This `newVersion` variable is only used to report to sherlock what tags are present on docker images.  For releases we normally care about the semver, but for development and deploying to our BEEs, git hashes for our PRs are what we want to know about.  I can also rename this variable if that helps.

PR #628, opened 2024-03-07
AJ-1669: Reuse `RecordSink` for full import.

jladieu:
For [AJ-1669](https://broadworkbench.atlassian.net/browse/AJ-1669): this makes an entire import job use the same `RecordSink` instance for all batches.  The goal here is to make the `RawlsRecordSink` capable of handling the entire import job with a single GCS file & pubsub, rather than triggering a new file & pubsub message for each batch.

In a subsequent change, the `RecordSink` will be given an additional method to finalize the import, which is what will be called at the very end of a batch to do all the post-processing.


calypsomatic (approved):
Looks good although without reviewing the rest of the code base it seems weird that there weren't any PFB tests that needed changes.  Are they just written different/better?

jladieu:
One goal I have when teasing out these refactorings ahead of time is to make changes that _in theory_ aren't supposed to change actual functionality.   I think the gap here with not triggering any test failures is nothing is currently verifying the current multi-batch behavior at the moment, which is that it would trigger multiple GCS writes and pubsubs.

Toward the end of this ticket, I plan to add a test that I expect to fail on multi-batches initially (checking that _only_ one file/pubsub happens).  I want to use that to drive the actual functional behavior change which will be to collapse the writes into one.

PR #629, opened 2024-03-08
AJ-1687: control-plane import jobs should stay running after they finish in quartz

davidangb:
When Rawls checks cWDS for status of a given import job, cWDS reports it has succeeded, so Rawls won't write anything.  This PR leaves cWDS jobs in RUNNING so Rawls will see them as safe.

In this PR:
* adds a new `twds.data-import.succeed-on-completion` property, set to `true` in the data plane and `false` in the control plane
* pipes the property through the `DataImportProperties` config class
    * used IntelliJ to restructure/reformat the `DataImportProperties` class
* reads the `DataImportProperties.isSucceedOnCompletion()` inside `QuartzJob`, and only mark the job as `SUCCESS` when true

_NOT_ in this PR:
* awaiting/reading a pub/sub message from Rawls, so we can finally mark the job as succeeded. After this PR, jobs will seem to stay in running forever according to WDS, but at least we'll be writing data to the Rawls-powered data tables.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
@@ -2,6 +2,54 @@ package org.databiosphere.workspacedataservice.config;
 
 /** Properties that dictate how data import processes should behave. */
 public class DataImportProperties {
+  private RecordSinkMode batchWriteRecordSink;
+  private String projectId;
+  private String rawlsBucketName;
+  private boolean succeedOnCompletion;
+
+  /** Where to write records after import, options are defined by {@link RecordSinkMode} */
+  public RecordSinkMode getBatchWriteRecordSink() {
+    return batchWriteRecordSink;
+  }
+
+  void setBatchWriteRecordSink(String batchWriteRecordSink) {
+    this.batchWriteRecordSink = RecordSinkMode.fromValue(batchWriteRecordSink);
+  }
+
+  public String getProjectId() {
+    return projectId;
+  }
+
+  void setProjectId(String projectId) {
+    this.projectId = projectId;
+  }
+
+  public String getRawlsBucketName() {
+    return rawlsBucketName;
+  }
+
+  void setRawlsBucketName(String rawlsBucketName) {
+    this.rawlsBucketName = rawlsBucketName;
+  }
+
+  /**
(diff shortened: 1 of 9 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
```diff
@@ -43,13 +24,49 @@ void setProjectId(String projectId) {
...
+
+    static RecordSinkMode fromValue(String value) {
+      for (RecordSinkMode mode : RecordSinkMode.values()) {
+        if (mode.value.equals(value)) {
+          return mode;
+        }
+      }
+      throw new RuntimeException("Unknown RecordSinkMode value: %s".formatted(value));
+    }
+  }
```
davidangb, line 71: a lot of the diff here is just IntelliJ-based reformatting, putting all vars at the top, for instance

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
```diff
@@ -230,6 +238,30 @@ void observationLogsFailure() throws org.quartz.JobExecutionException {
...
+
+    String randomToken = RandomStringUtils.randomAlphanumeric(10);
+    UUID jobUuid = UUID.randomUUID();
+    JobExecutionContext mockContext = setUpTestJob(randomToken, jobUuid.toString());
+
+    // this test requires a TestObservationRegistry
+    TestObservationRegistry testObservationRegistry =
+        assertInstanceOf(TestObservationRegistry.class, observationRegistry);
+
+    // execute the TestableQuartzJob, then confirm observation recorded failure
```
calypsomatic, line 256: Nit, but I'm confused by this comment.  What's failing that needs to be recorded?

calypsomatic (approved):

yuliadub (approved):
will we need to do anything special about stuck in "running" cwds jobs when we add the logic to mark them as succeeded once we get the signal from Rawls? (these would only be test ones, if any at all, so maybe also a non issue altogether)

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
davidangb: good catch; that was a copy/paste error. Fixed!

jladieu (approved):
Nice!

PR #631, opened 2024-03-08
AJ-1680 Use Rawls instead of WSM for control plane imports

calypsomatic:
https://broadworkbench.atlassian.net/browse/AJ-1680

Changes in this PR:
- Separate `SnapshotSupport` into control plane and data plane versions
  - Creates a `SnapshotSupport` abstract class.  
  - Everything from the previous `WsmSnapshotSupport` that doesn't directly interact with the wsmdao now lives in `SnapshotSupport`.
  - `WsmSnapshotSupport` and the new `RawlsSnapshotSupport` call their respective services when needed
 - Creates `SnapshotSupportFactory` interface to determine which `SnapshotSupportFactory` to use based on whether we are in control-plane or data-plane  
   - `WsmSnapshotSupportFactory` and `RawlsSnapshotSupportFactory` build the required `SnapshotSupport`
   - Other classes use `SnapshotSupportFactory` to get their `SnapshotSupport` instead of building it directly
- Creates a new `RawlsClient` class to talk to Rawls

Todo:

- [x] Make sure bearer token (and any other headers) are set correctly in `RawlsClient`
- [x] Make ticket or otherwise follow up with contract tests for Rawls
- [x] Make sure `SnapshotSupport` classes are properly tested
- [x] Deal with logging statements in `SnapshotSupport` classes
- [x] Make sure control-plane profile is used appropriately so we don't have RawlsClients and things we don't need in the data plane
- [x] Make sure rawlsUrl is populated appropriately, possibly needs a terra-helmfile pr?

Not done in this PR, reserved for a follow-on:
- Contain workspacemanager and azureblob storage to data plane
- Deal appropriately with retries in the Rawls client
- Verify objectmapper use by the rawls resttemplate

Reminder:


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/RawlsSnapshotSupport.java
```diff
@@ -0,0 +1,46 @@
...
+import bio.terra.workspace.model.ResourceList;
+import java.util.UUID;
+import org.databiosphere.workspacedataservice.activitylog.ActivityLogger;
+import org.databiosphere.workspacedataservice.rawls.RawlsClient;
+import org.databiosphere.workspacedataservice.retry.RestClientRetry;
+import org.databiosphere.workspacedataservice.shared.model.WorkspaceId;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class RawlsSnapshotSupport extends SnapshotSupport {
```
jladieu, line 12: Contract tests would be a good candidate to verify this class, but it's a very big lift to get a new integration off the ground so I'm not suggesting you take that on in this PR.

calypsomatic:
> 1. The Rawls interactions will need to distinguish snapshot references with "purpose:policy" vs. those without. Most importantly, any legacy snapshot references did NOT link policies, so we have to ignore them when considering which references to link.

Is this different from what we do with WSM right now?  If so, what exactly needs to be done differently?  When enumerating references I dismiss those *without* "purpose:policy" to make sure I DO add a link?  Or when/where am I making this distinction?

> 2. Related, when creating references we'll need to add "purpose:policy"; this is blocked by WOR-1567

That ticket isn't in progress yet.  Do I just have to sit on this PR until then or can I do some incremental work and fill this in later?

> 3. We probably could get away with only talking to Rawls to create new references, but still talking to WSM to _list_ the references (and treating 404s equivalent to the empty list). This could be a bad idea and not worth it, but it would consolidate a bit of code.

Seems like it would un-consolidate more code than it would consolidate if in some cases I need to use BOTH WSM and Rawls instead of just having separate beans.

> 4. We _could_ manually run openapi-generator against the Rawls openapi spec to make a client; is that worthwhile?

A quick attempt just to see what would happen if I did that make it seem like it requires a lot of massaging to get into a useable state, unless I just don't know what I'm doing.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/RawlsSnapshotSupport.java
calypsomatic: Added https://broadworkbench.atlassian.net/browse/AJ-1697 to make sure it gets done

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/SnapshotSupport.java
```diff
@@ -18,109 +16,27 @@
...
 import org.databiosphere.workspacedataservice.service.model.exception.RestException;
 import org.databiosphere.workspacedataservice.shared.model.RecordType;
-import org.databiosphere.workspacedataservice.shared.model.WorkspaceId;
-import org.databiosphere.workspacedataservice.workspacemanager.WorkspaceManagerDao;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.lang.Nullable;
 
-public class WsmSnapshotSupport {
+public abstract class SnapshotSupport {
```
calypsomatic, line 26: This class now contains the vast majority of the code from `WsmSnapshotSupport`.  Only methods that actively require a dao or client were named abstract and left for implementing classes.  So there's not a lot of actual code changes here, but I did end up moving things around a bit as I was deciding what needed to go where, which accounts for most of the activity here.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/SnapshotSupportTest.java
```diff
@@ -1,100 +1,27 @@
...
 @DirtiesContext
 @SpringBootTest
-class WsmSnapshotSupportTest extends TestBase {
-
-  @MockBean JobDao jobDao;
-  @MockBean WorkspaceManagerDao wsmDao;
-  @MockBean ActivityLogger activityLogger;
-  @Autowired RestClientRetry restClientRetry;
-
-  @ParameterizedTest(name = "paginates through results when WSM has {0} references")
```
calypsomatic, line 46: This test has been moved to both the new `WsmSnapshotSupportTest` and `RawlsSnapshotSupportTest`. Am undecided about whether to leave it like that or adapt it to continue to live in this test, and also about what other implementation-specific tests would be useful.

service/src/main/java/org/databiosphere/workspacedataservice/rawls/RawlsClient.java
```diff
@@ -0,0 +1,95 @@
...
+import org.springframework.http.HttpMethod;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Component;
+import org.springframework.web.client.RestClientResponseException;
+import org.springframework.web.client.RestTemplate;
+import org.springframework.web.util.UriComponentsBuilder;
+
+@Component
+@ControlPlane
+public class RawlsClient {
```
calypsomatic, line 26: Turns out WSMDao is used for `AzureBlobStorage`, needed for cloning, so although Rawls is only used in the controlplane, WSM is needed for both.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/RawlsSnapshotSupport.java
```diff
@@ -0,0 +1,43 @@
...
+    this.workspaceId = workspaceId;
+    this.rawlsClient = rawlsClient;
+    this.activityLogger = activityLogger;
+    this.restClientRetry = restClientRetry;
+  }
+
+  protected void linkSnapshot(UUID snapshotId) {
+    RestClientRetry.VoidRestCall voidRestCall =
+        (() -> rawlsClient.createSnapshotReference(workspaceId.id(), snapshotId));
+    restClientRetry.withRetryAndErrorHandling(voidRestCall, "Rawls.createSnapshotReference");
```
jladieu, line 31: supernit: consider inlining `voidRestCall`

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/SnapshotSupport.java
```diff
@@ -133,16 +49,21 @@ protected UUID safeGetSnapshotId(ResourceDescription resourceDescription) {
         try {
           return UUID.fromString(snapshotIdStr);
         } catch (Exception e) {
+          // TODO The only thing this method requires from the implementing class is workspaceId
```
jladieu, line 52: I'm not sure I understand the comment...are you referring to the workspaceId that gets logged as part of the warning message?

IMO, I think this method should just crash hard in this circumstance to prevent garbage data from coming in, in which case you wouldn't need to log.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/SnapshotSupport.java
```diff
@@ -161,28 +82,36 @@ public void linkSnapshots(Set<UUID> snapshotIds) {
...
       }
     }
   }
 
-  // TODO: reevaluate moving this method to TdrManifestQuartzJob (see comment on PR #449)
+  protected abstract void linkSnapshot(UUID snapshotId);
+
+  /**
+   * Query for the full list of referenced snapshots in this workspace, then return the list of
+   * unique snapshotIds from those references. Calls Rawls or WSM depending on the implementation of
```
jladieu, line 105: nit: interface comments that describe assumptions introduced by implementors can easily get out of date when those assumptions change, but it's hard to know that this comment would need to be changed if you're making changes in the subclass only; would you consider dropping the extra sentence referring to Rawls/WSM?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/SnapshotSupport.java
```diff
@@ -191,29 +120,67 @@ public Map<RecordType, String> identifyPrimaryKeys(List<TableModel> tables) {
                 tableModel -> identifyPrimaryKey(tableModel.getPrimaryKey())));
   }
 
-  // snapshotKeys comes from generated code which doesn't guarantee null safety
+  public Multimap<RecordType, RelationshipModel> identifyRelations(
```
jladieu, line 123: `identifyRelations` looks like it might only be used by `TdrManifestQuartzJob`, maybe remove it from here and make it a private static method on that class?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/SnapshotSupport.java
```diff
@@ -191,29 +120,67 @@ public Map<RecordType, String> identifyPrimaryKeys(List<TableModel> tables) {
...
+        relationshipModel -> RecordType.valueOf(relationshipModel.getFrom().getTable()));
+  }
+
+  /**
+   * Given a ResourceList, find all the valid ids of referenced snapshots in that list
+   *
+   * @param resourceList the list in which to look for snapshotIds
+   * @return the list of unique ids in the provided list
+   */
+  protected List<UUID> extractSnapshotIds(ResourceList resourceList) {
```
jladieu, line 136: This looks like it's only used by `existingPolicySnapshotIds`; consider making it `private`?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/SnapshotSupport.java
calypsomatic: Right, we can't log the workspaceId in this error anymore.  Since I'm only moving this code from a previous class, I'm not super sure how important it is that we have this rigamarole just for logging the error.  We could crash and burn instead, or possible just log without the workspaceId

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/SnapshotSupport.java
calypsomatic: Good point

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/SnapshotSupport.java
```diff
@@ -191,29 +120,67 @@ public Map<RecordType, String> identifyPrimaryKeys(List<TableModel> tables) {
...
+   * @param pageSize how many references to return in each paginated request
+   * @return the full list of snapshot references in this workspace
    */
-  public Multimap<RecordType, RelationshipModel> identifyRelations(
-      List<RelationshipModel> relationshipModels) {
-    return Multimaps.index(
-        relationshipModels,
-        relationshipModel -> RecordType.valueOf(relationshipModel.getFrom().getTable()));
+  protected ResourceList listAllSnapshots(int pageSize) {
+    final AtomicInteger offset = new AtomicInteger(0);
```
jladieu, line 159: It looks like you're using a type that's designed for multi-threading, but I can't tell how the code that uses it is multithreaded...is something about the while block multithreaded that requires this?

If not, could you just use a basic integer and drop the `final`, since you'd need to re-assign it?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/SnapshotSupport.java
```diff
@@ -191,29 +120,67 @@ public Map<RecordType, String> identifyPrimaryKeys(List<TableModel> tables) {
...
+    final int hardLimit = 10000; // under no circumstances return more than this many snapshots
```
jladieu, line 160: Consider punting `hardLimit` out to be a `private static final` constant, maybe with a more intention revealing name like `MAX_SNAPSHOTS_HARD_LIMIT`

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/SnapshotSupport.java
calypsomatic: Ah yeah there was a comment to this effect that may have gotten lost in the shuffle, maybe I'll just go ahead and do that

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/WsmSnapshotSupport.java
```diff
@@ -0,0 +1,43 @@
...
+      RestClientRetry restClientRetry,
+      ActivityLogger activityLogger) {
+    this.workspaceId = workspaceId;
+    this.wsmDao = wsmDao;
+    this.restClientRetry = restClientRetry;
+    this.activityLogger = activityLogger;
+  }
+
+  protected ResourceList enumerateDataRepoSnapshotReferences(int offset, int pageSize) {
+    RestClientRetry.RestCall<ResourceList> restCall =
```
jladieu, line 30: nit: consider inlining `restCall` and `voidRestCall` below since they're a mouthful to declare

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/WsmSnapshotSupport.java
```diff
@@ -0,0 +1,43 @@
...
+        (() -> wsmDao.enumerateDataRepoSnapshotReferences(workspaceId.id(), offset, pageSize));
+    return restClientRetry.withRetryAndErrorHandling(
+        restCall, "WSM.enumerateDataRepoSnapshotReferences");
+  }
+
+  protected void linkSnapshot(UUID snapshotId) {
+    RestClientRetry.VoidRestCall voidRestCall =
+        (() -> wsmDao.linkSnapshotForPolicy(workspaceId, new SnapshotModel().id(snapshotId)));
+    restClientRetry.withRetryAndErrorHandling(voidRestCall, "WSM.createDataRepoSnapshotReference");
```
jladieu, line 39: should the hint here be `WSM.linkSnapshot`?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/RawlsSnapshotSupport.java
jladieu, line 31: Should the hint here be `Rawls.linkSnapshot`?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/SnapshotSupport.java
```diff
@@ -161,28 +82,36 @@ public void linkSnapshots(Set<UUID> snapshotIds) {
...
+   * listAllSnapshots
+   *
+   * @param pageSize how many references to return in each paginated request
+   * @return the list of unique ids for all pre-existing snapshot references
+   */
+  public List<UUID> existingPolicySnapshotIds(int pageSize) {
+    return extractSnapshotIds(listAllSnapshots(pageSize));
+  }
+
   public Map<RecordType, String> identifyPrimaryKeys(List<TableModel> tables) {
```
jladieu, line 115: Is this also only used by `TdrManifestQuartzJob`?  Maybe relocate this over there too as a private method?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/WsmSnapshotSupport.java
calypsomatic: Maybe?  `wsmDao.linkSnapshotForPolicy` calls `WSM.createDataRepoSnapshotReference` so that hint is correct, depending on what level you want the hint to be at

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/RawlsSnapshotSupport.java
calypsomatic: Looks like these hints are referring to the underlying method that's being called in the actual dao/client, which is presumably the one that is being retried so I think it makes sense to keep the hints that way

service/src/main/java/org/databiosphere/workspacedataservice/rawls/RawlsClient.java
```diff
@@ -0,0 +1,95 @@
...
+
+      ResponseEntity<ResourceList> response =
+          restTemplate.exchange(
+              builder.toUriString(),
+              HttpMethod.GET,
+              new HttpEntity<>(getAuthedHeaders()),
+              ResourceList.class);
+      return response.getBody();
+    } catch (RestClientResponseException e) {
+      LOGGER.warn("Error retrieving snapshot references for workspace {}", workspaceId, e);
```
jladieu, line 54: Consider getting rid of the warning and just include the helpful details from the message in the message of the thrown `RawlsException`.  (Why? don't log and throw: https://rolf-engelhard.de/2013/04/logging-anti-patterns-part-ii/)

service/src/main/java/org/databiosphere/workspacedataservice/rawls/RawlsClient.java
```diff
@@ -0,0 +1,95 @@
...
+  // key-value pair to the reference’s properties
+  public void createSnapshotReference(UUID workspaceId, UUID snapshotId) {
+    try {
+      restTemplate.exchange(
+          getSnapshotsUrl(workspaceId),
+          HttpMethod.POST,
+          new HttpEntity<>(new SnapshotModel().id(snapshotId), getAuthedHeaders()),
+          DataRepoSnapshotResource.class);
+    } catch (RestClientResponseException e) {
+      LOGGER.warn(
```
jladieu, line 69: Consider getting rid of the warning and just include the helpful details from the message in the message of the thrown `RawlsException`.  (Why? don't log and throw: https://rolf-engelhard.de/2013/04/logging-anti-patterns-part-ii/)

service/src/main/java/org/databiosphere/workspacedataservice/rawls/RawlsClient.java
```diff
@@ -0,0 +1,95 @@
...
+  // Get the user's token from the context and attach it to headers
+  private HttpHeaders getAuthedHeaders() {
+    HttpHeaders headers = new HttpHeaders();
+    BearerToken token = TokenContextUtil.getToken();
+
+    if (token.nonEmpty()) {
+      LOGGER.debug("setting access token for rawls request");
+      headers.setBearerAuth(Objects.requireNonNull(token.getValue()));
+    } else {
+      LOGGER.warn("No access token found for rawls request.");
```
jladieu, line 87: Could we just crash in this case?  Would we ever want to let a through a non-authed request?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/SnapshotSupport.java
calypsomatic: Hm this one is used in a test and isn't as straightforward to extract

service/src/main/java/org/databiosphere/workspacedataservice/rawls/RawlsClient.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+    return headers;
+  }
+
+  private String getSnapshotsUrl(UUID workspaceId) {
+    return rawlsUrl + "/api/workspaces/" + workspaceId + "/snapshots/v2";
```
jladieu, line 93: So, I saw this line and was going to suggest redoing it without concatenating strings...and was going to suggest

`return "%s/api/workspaces/%s/snapshots/v2".formatted(rawlsUrl, workspaceId)`

But I didn't know why I thought that might be better, so I tested my hunch by asking ChatGPT to critique this line of code, and it dropped this bombshell on me:

`UriComponentsBuilder.fromHttpUrl(rawlsUrl)
    .pathSegment("api", "workspaces", workspaceId, "snapshots", "v2")
    .toUriString();`

It's reasoning:
```
This line of code constructs a URL by concatenating strings, including a base URL (`rawlsUrl`), static path segments, and a dynamic segment (`workspaceId`). While it's a common and straightforward approach, there are several aspects to consider for improvement or potential issues:

### 1. Hardcoding and Maintainability
- **Hardcoding Paths**: The path segments are hardcoded, which can make maintenance and updates more challenging, especially if the base path for snapshots changes or if you need to construct similar URLs elsewhere in your application.
- **Solution**: Consider using a constant for repeated segments or a centralized method for constructing such URLs.

### 2. URL Encoding
- **URL Encoding Issues**: The `workspaceId` is directly concatenated into the URL without explicit encoding. If `workspaceId` contains characters that are not URL-safe (e.g., spaces, special characters), it could lead to malformed URLs or security vulnerabilities.
- **Solution**: Use `URLEncoder.encode(workspaceId, "UTF-8")` to ensure that `workspaceId` is safely encoded.

### 3. String Concatenation Efficiency
- **String Concatenation**: While not a significant issue in this single line, repeated string concatenation (especially in loops) can be inefficient because strings in Java are immutable, leading to the creation of multiple string objects.
- **Solution**: For more complex or dynamic URL constructions, consider using `StringBuilder` to append strings efficiently.

### 4. Lack of URL Validation
- **Validation**: There's no validation that `rawlsUrl` is a well-formed URL or that it does not already include a trailing slash, which could lead to incorrect URL formation (e.g., double slashes or missing slashes).
- **Solution**: Validate the format of `rawlsUrl` and ensure proper handling of slashes. Alternatively, use `java.net.URL` or `java.net.URI` for more robust URL handling and manipulation.

### 5. Flexibility and Scalability
- **Scalability**: As your application grows and requires more complex URL constructions or query parameters, this approach might become cumbersome and error-prone.
- **Solution**: Consider using a URL building library like `UriComponentsBuilder` from Spring Framework, which provides a more flexible and error-resistant way to construct URLs with query parameters and path variables.

### Improved Example with UriComponentsBuilder (Spring)

Assuming you're working in a Spring environment, here's how you could refactor the code using `UriComponentsBuilder` for better flexibility and safety:

```java
import org.springframework.web.util.UriComponentsBuilder;

// Assuming workspaceId is URL-encoded as needed
String url = UriComponentsBuilder.fromHttpUrl(rawlsUrl)
    .pathSegment("api", "workspaces", workspaceId, "snapshots", "v2")
    .toUriString();

return url;
```
This approach addresses hardcoding, ensures proper handling of path segments (including encoding), and improves maintainability and readability. It also seamlessly supports adding query parameters if needed in the future.
```

TIL....

Comment (calypsomatic) [reply service/src/main/java/org/databiosphere/workspacedataservice/rawls/RawlsClient.java:93]: Nice!

File: service/src/test/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/WsmSnapshotSupportTest.java [hunk 19]
```diff
@@ -0,0 +1,91 @@
...
+
+  @ParameterizedTest(name = "paginates through results when WSM has {0} references")
+  @ValueSource(ints = {0, 1, 49, 50, 51, 99, 100, 101, 456})
+  void paginateExistingSnapshots(int wsmCount) {
+    int testPageSize = 50; // page size to use during this test
+
+    List<ResourceDescription> mockResources = new ArrayList<>();
+    // generate the full list of snapshots as known by our mock WSM
+    for (int i = 0; i < wsmCount; i++) {
+      ResourceDescription resourceDescription = new ResourceDescription();
```
jladieu, line 47: nit: would you consider factoring out lines 47:53:

```
private ResourceDescription buildResourceDescription(int index) {
  ...
}
```

That would slim down the test body so it's a bit easier to scan.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/WsmSnapshotSupportTest.java
```diff
@@ -0,0 +1,91 @@
...
+      dataRepoSnapshotAttributes.setInstanceName("index: " + i);
+      resourceAttributesUnion.setGcpDataRepoSnapshot(dataRepoSnapshotAttributes);
+      resourceDescription.setResourceAttributes(resourceAttributesUnion);
+      mockResources.add(resourceDescription);
+    }
+    // configure the mock to return the appropriate page of snapshots
+    when(wsmDao.enumerateDataRepoSnapshotReferences(any(), anyInt(), anyInt()))
+        .thenAnswer(
+            invocation -> {
+              int offset = invocation.getArgument(1);
```
jladieu, line 60: similar to above...could you extract

```
private ResourceList stubPage(List<ResourceDescription> resources, <whatever type invocation is> invocation)
```

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/WsmSnapshotSupportTest.java
```diff
@@ -0,0 +1,91 @@
...
+            .listAllSnapshots(testPageSize);
+
+    // assert total size of all results
+    assertEquals(wsmCount, actual.getResources().size());
+    // assert that the "snapshot instance name" (not "WDS instance") is unique in all our results
+    // i.e. we did not return the same snapshot more than once
+    List<String> snapshotInstanceNames =
+        actual.getResources().stream()
+            .map(res -> res.getResourceAttributes().getGcpDataRepoSnapshot().getInstanceName())
+            .distinct()
```
jladieu, line 83: Doesn't adding this `distinct()` force de-duping?  I gather from the comment above this that we want to make sure dupes aren't happening...if that's the case, wouldn't we want to let dupes go through and detect/fail if they're encountered?

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/WsmSnapshotSupportTest.java
```diff
@@ -0,0 +1,91 @@
...
+            .toList();
+    assertEquals(wsmCount, snapshotInstanceNames.size());
+    // assert the number of requests made to WSM to generate the list
+    double expectedInvocations = Math.floor((double) wsmCount / testPageSize) + 1;
```
jladieu, line 87: fancy!

maybe cast `expectedInvocations` to an int here, so you don't have to cast it inside the `times()` method call

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/SnapshotSupport.java
jladieu: I'd support either of those approaches, but lean toward crashing unless there's some kind of historical reason for wanting to support partial/weird behavior.

jladieu (approved):
I took a pass through and had a bunch of small stuff, but nothing super critical.  Preemptively approving but happy to take another look later if you want.

service/src/main/java/org/databiosphere/workspacedataservice/rawls/RawlsClient.java
```diff
@@ -39,7 +39,8 @@ public RawlsClient(String rawlsUrl, RestTemplate restTemplate) {
   public ResourceList enumerateDataRepoSnapshotReferences(UUID workspaceId, int offset, int limit) {
     try {
       UriComponentsBuilder builder =
-          UriComponentsBuilder.fromHttpUrl(getSnapshotsUrl(workspaceId))
+          UriComponentsBuilder.fromHttpUrl(rawlsUrl)
```
jladieu, line 42: Cool!! I didn't even realize we were already using `UriComponentsBuilder` here!!

service/src/main/java/org/databiosphere/workspacedataservice/rawls/RawlsClient.java
jladieu: Ahh I didn't even realize one of the callsites was already using `UriComponentsBuilder`

I think your should keep the helper, but instead of returning a String, just return the `UriComponentsBuilder`; use it in the one spot, and finalize it to a string in the other.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/SnapshotSupport.java
davidangb: Crashing here is fine with me, I agree this is over-complicated code just to log an error. Who wrote that, anyway? (me)

I'd be fine with either of:
* uncomment the commented-out logging statement, and remove the `for workspace %s` part of it
* delete the commented-out logging statement and `resourceId` logic entirely

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/SnapshotSupport.java
davidangb: `AtomicInteger` was used in the original `WsmSnapshotSupport` code this was copied from - there, it was a workaround to using the var inside a lambda - it got around the "Variable used in lambda expression should be final or effectively final" problem (yes there are other ways to get around that). But, I don't think it's an issue here and thus could be a simple `int`

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/WsmSnapshotSupportTest.java
davidangb: This does de-dupe. If de-duping removed duplicates, it would cause the next assertion `assertEquals(wsmCount, snapshotInstanceNames.size());` to fail - that's the test

service/src/main/java/org/databiosphere/workspacedataservice/rawls/RawlsClient.java
davidangb: but … why is `AzureBlobStorage` required in the control plane? Even if it's needed for cloning, we don't do anything with cloning in the control plane (if this scope creeps this PR, ok to solve later)

service/src/main/java/org/databiosphere/workspacedataservice/rawls/RawlsClient.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  // TODO: (AJ-1705) Add cloning instructions COPY_REFERENCE and a purpose=policy
+  // key-value pair to the reference’s properties
+  public void createSnapshotReference(UUID workspaceId, UUID snapshotId) {
+    try {
+      restTemplate.exchange(
+          UriComponentsBuilder.fromHttpUrl(rawlsUrl)
+              .pathSegment("api", "workspaces", workspaceId.toString(), "snapshots", "v2")
+              .toUriString(),
```
davidangb, line 66: nit: keep `enumerateDataRepoSnapshotReferences` and `createSnapshotReference` parallel - either inline the builder in both or create the builder as a separate var in both. This is just style, feel free to push back if you have a preference otherwise

service/src/main/java/org/databiosphere/workspacedataservice/rawls/RawlsClientConfig.java
```diff
@@ -0,0 +1,27 @@
...
+  private String rawlsUrl;
+
+  @Bean
+  public RawlsClient rawlsClient(RestTemplate restTemplate) {
+    return new RawlsClient(rawlsUrl, restTemplate);
+  }
+
+  @Bean
+  public RestTemplate restTemplate(RestTemplateBuilder builder) {
+    return builder.build();
```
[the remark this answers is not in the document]
calypsomatic: Hm, I don't know how to really verify what the active `ObjectMapper` is.  [This article](https://dzone.com/articles/configuring-a-custom-objectmapper-for-spring-restt) has you add a custom `ObjectMatter` to the front of a *list* of `ObjectMappers`, which supposedly means it takes precedence.  If I inspect the `RestTemplate` in the debugger, I can see that it has a few `MappingJackson2HttpMessageConverter`s, and then if I add in those magic incantations it has yet another one at the front of the list.  Not confident about how or where down the line I can put in a test to verify.

service/src/main/java/org/databiosphere/workspacedataservice/rawls/RawlsClient.java
```diff
@@ -0,0 +1,88 @@
...
+          HttpMethod.POST,
+          new HttpEntity<>(new SnapshotModel().id(snapshotId), getAuthedHeaders()),
+          DataRepoSnapshotResource.class);
+    } catch (RestClientResponseException e) {
+      throw new RawlsException(e);
+    }
+  }
```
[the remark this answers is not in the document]
calypsomatic: I don't really understand how `RestClientRetry` works; what I did was preserve the use of it up a level, in the `SnapshotSupport` classes.  Both `WorkspaceManagerDao` AND `WsmSnapshotSupport` have a `RestClientRetry`, with `WsmSnapshotSupport` wrapping calls to `WSMDao` in a retry, even though the method being called also uses `RestClientRetry`.   Is it necessary to have it on both levels?

service/src/main/java/org/databiosphere/workspacedataservice/rawls/RawlsClient.java
calypsomatic: Yeah, it looks like maybe a whole thing to tease it out from the control plane.  I'll leave it for now.

service/src/main/java/org/databiosphere/workspacedataservice/rawls/RawlsClient.java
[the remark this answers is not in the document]
davidangb: Oooooh hmm, thanks for calling that out.

For the data plane: I think we should NOT add retries at both the `WsmSnapshotSupport` and `WorkspaceManagerDao` level; we should pick one. I think most everywhere else we do it at the DAO level, so probably we should take it out of `WsmSnapshotSupport`.

For the control plane: adding `RestClientRetry` into `RawlsSnapshotSupport` will … do nothing in the current codebase. It is set up to retry when it encounters a `RestRetryableException`, as configured [here](https://github.com/DataBiosphere/terra-workspace-data-service/blob/165212e64e5a5a61c4cc2a206e703463b427c077/service/src/main/java/org/databiosphere/workspacedataservice/retry/RestClientRetry.java#L34C18-L34C40). This relies on extracting the status code out of the exceptions thrown by the rest clients, which happens [here](https://github.com/DataBiosphere/terra-workspace-data-service/blob/165212e64e5a5a61c4cc2a206e703463b427c077/service/src/main/java/org/databiosphere/workspacedataservice/retry/RestClientRetry.java#L140-L162).

You may be able to get retries working for Rawls by simply adding:
```java
    // Rawls
    if (e instanceof org.springframework.web.client.RestClientResponseException restTemplateException) {
      return restTemplateException.getStatusCode().value();
    }
```
to the implementation of `RestClientRetry.extractResponseCode`

davidangb (approved):
Giving this a proactive :+1:, thank you for sticking with it!

For follow-on work:
* why is AzureBlobStorage required in the control plane, and would solving that allow us to also not require WsmDao in the control plane?
* deal with retries for Rawls, your call if you want to include that in this PR

PR #632, opened 2024-03-08
AJ-1669: Add `finalizeBatchWrite` to `RecordSink`

jladieu:
For [AJ-1669](https://broadworkbench.atlassian.net/browse/AJ-1669): this provides a callback hook that can be used to perform any necessary steps _after_ all batches are complete.  Its intended use is to allow `RawlsRecordSink` to finish writing to the underlying GCS Storage, and send a pubsub message downstream just once after all batches are complete.

Right now it is basically a no-op, but it could be used at this point to consolidate the activity logging logic, for example.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJob.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJob.java
@@ -132,13 +132,15 @@ public class PfbQuartzJob extends QuartzJob {
 
     // This is HTTP connection #2 to the PFB.
     logger.info("Importing tables and rows from this PFB...");
-    withPfbStream(url, stream -> importTables(stream, recordSink, BASE_ATTRIBUTES));
+    BatchWriteResult result =
+        withPfbStream(url, stream -> importTables(stream, recordSink, BASE_ATTRIBUTES));
 
     // This is HTTP connection #3 to the PFB.
     logger.info("Updating tables and rows from this PFB with relations...");
-    withPfbStream(url, stream -> importTables(stream, recordSink, RELATIONS));
-    // TODO(AJ-1669): Add and use a finalize callback on RecordSink here to take care of post
-    //   processing steps
+    result.merge(withPfbStream(url, stream -> importTables(stream, recordSink, RELATIONS)));
+
+    // Commit results, publish to downstream systems, etc.
+    recordSink.finalizeBatchWrite(result);
     // TODO(AJ-1453): save the result of importTables and persist them to the job
   }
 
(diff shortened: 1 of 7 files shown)


calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -84,6 +85,13 @@ public void deleteBatch(RecordType recordType, List<Record> records) throws Batc
     throw new UnsupportedOperationException("RawlsRecordSink does not support deleteBatch");
   }
 
+  @Override
+  public void finalizeBatchWrite(BatchWriteResult result) {
+    // currently a no-op
+    // TODO(AJ-1669): do pubsub here, maybe do GCS cleanup here (assuming a file was reused
+    //   throughout)
+  }
+
```
davidangb, line 94: Thinking out loud: I wonder if would make more sense for this to accept a functional interface, so finalization code could easily be defined (and tested) independently.

No changes requested for this PR … when it comes time to actually implement the finalization code, consider it then!

davidangb (approved):
Depending on which of this PR and #633 merges first, I expect some minor conflicts

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
jladieu: I'll try this out when I start using the callback.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJob.java
```diff
@@ -132,13 +132,15 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
...
-    withPfbStream(url, stream -> importTables(stream, recordSink, BASE_ATTRIBUTES));
+    BatchWriteResult result =
+        withPfbStream(url, stream -> importTables(stream, recordSink, BASE_ATTRIBUTES));
 
     // This is HTTP connection #3 to the PFB.
     logger.info("Updating tables and rows from this PFB with relations...");
-    withPfbStream(url, stream -> importTables(stream, recordSink, RELATIONS));
-    // TODO(AJ-1669): Add and use a finalize callback on RecordSink here to take care of post
-    //   processing steps
+    result.merge(withPfbStream(url, stream -> importTables(stream, recordSink, RELATIONS)));
```
jladieu, line 140: In hindsight, merging in this way actually might be buggy...records with relations will get double-counted in the results, unless we make `BatchWriteResult` smart enough to distinguish between first-pass and second-pass incrementing (maybe we should?)

PR #633, opened 2024-03-08
AJ-1589: attribute prefixing

davidangb:
Updates the attribute-prefixing behavior in cWDS to match Import Service behavior.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
@@ -2,11 +2,15 @@ package org.databiosphere.workspacedataservice.dataimport;
 
 import jakarta.annotation.Nullable;
 import java.util.UUID;
+import org.databiosphere.workspacedataservice.recordsink.RawlsAttributePrefixer.PrefixStrategy;
 
 public record ImportDetails(
-    @Nullable UUID jobId, @Nullable String userEmail, UUID collectionId, String prefix) {
+    @Nullable UUID jobId,
+    @Nullable String userEmail,
+    UUID collectionId,
+    PrefixStrategy prefixStrategy) {
 
-  public ImportDetails(UUID collectionId, String prefix) {
-    this(null, null, collectionId, prefix);
+  public ImportDetails(UUID collectionId, PrefixStrategy prefixStrategy) {
+    this(null, null, collectionId, prefixStrategy);
   }
 }
--- a/service/src/test/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkPrefixingTest.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkPrefixingTest.java
@@ -1,0 +1,215 @@
+package org.databiosphere.workspacedataservice.recordsink;
+
+import static com.google.common.collect.MoreCollectors.onlyElement;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.collect.Iterables;
+import java.io.IOException;
+import java.io.StringWriter;
+import java.nio.charset.StandardCharsets;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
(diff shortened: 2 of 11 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkPrefixingTest.java
```diff
@@ -0,0 +1,214 @@
+package org.databiosphere.workspacedataservice.recordsink;
```
davidangb, line 1: This new test adds coverage all the way through RawlsRecordSink … but it duplicates `RawlsAttributePrefixerTest` a lot. Is it worth having both?

jladieu: Amazing test coverage!!  ...just curious to know your process, did you start with this one and add the other one after, or did this evolve after the unit was done?

Regarding whether it's worth having both, I could go either way, but probably leaning towards pulling back on the coverage at this level.  Here are my thoughts to do with what ye will...

I personally really like outside-in tests like this one for high-level & e2e validation but to your point, I guess when they start to overlap in breadth with the unit level as this one does it might be good to pull back and slim down the sheer number of cases.

I'm thinking about the relative 'cost' of each one of these tests, as they involve a call out to GCS storage (which at some point I'm betting we'll want to cut back on or stub out somehow).    Possibly a way to still have a nice E2E spotcheck but reduce the overall cost of having tests would be to just incorporate an assertion for each prefix type into the corresponding existing E2E tests.  In this way, you get that extra mile of coverage without incurring a bunch of extra 'cost' from extra GCS storage events.

If you did do that, you could also include a comment near such an assertion with pointer back down to the unit test for detailed edge cases.

jladieu (approved):
I love this!

Approved as is, with some optional thoughts for your consideration.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
```diff
@@ -2,11 +2,15 @@
...
 public record ImportDetails(
-    @Nullable UUID jobId, @Nullable String userEmail, UUID collectionId, String prefix) {
+    @Nullable UUID jobId,
+    @Nullable String userEmail,
+    UUID collectionId,
+    PrefixStrategy prefixStrategy) {
 
-  public ImportDetails(UUID collectionId, String prefix) {
-    this(null, null, collectionId, prefix);
+  public ImportDetails(UUID collectionId, PrefixStrategy prefixStrategy) {
```
jladieu, line 13: No changes needed here if you'd prefer to keep as is, but: maybe since `PrefixStrategy` is additive behavior, and all the places we use this overload pass in `PrefixStrategy.NONE`, I think it might make sense to just make this constructor take one arg, and assume the `NONE` strategy.

jladieu:
Looks like I threw some conflicts in front of you with https://github.com/DataBiosphere/terra-workspace-data-service/pull/632 as you predicted.  Hopefully nothing too gross...

service/src/test/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkPrefixingTest.java
```diff
@@ -0,0 +1,205 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.util.StreamUtils;
+
+/**
+ * Smoke-tests a few paths fof attribute prefixing behavior via a full RawlsRecordSink. More
```
calypsomatic, line 37: Nit: "fof"

davidangb: I frixed this fypo

service/src/test/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkPrefixingTest.java
davidangb: > I'm thinking about the relative 'cost' of each one of these tests, as they involve a call out to GCS storage

Note that these tests use `@Qualifier("mockGcsStorage")`, which creates - via `LocalStorageHelper` - "an in-memory storage configuration for testing". So I agree with there's a lot of redundancy, but at least these are pretty light as far as storage operations go

calypsomatic (approved):

jladieu (approved):

PR #643, opened 2024-03-11
AJ-1586: send full bucketname + file in pubsub message

davidangb:
In the pubsub message that cWDS sends to Rawls, we must include the bucket name as well as the file path.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
@@ -98,7 +98,7 @@ public class RawlsRecordSink implements RecordSink {
             .put("workspaceId", workspaceId.toString())
             .put("userEmail", user)
             .put("jobId", jobId.toString())
-            .put("upsertFile", upsertFile)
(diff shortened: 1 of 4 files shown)


nawatts (approved):

jladieu (approved):

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -98,7 +98,7 @@ private void publishToPubSub(UUID workspaceId, String user, UUID jobId, String u
             .put("workspaceId", workspaceId.toString())
             .put("userEmail", user)
             .put("jobId", jobId.toString())
-            .put("upsertFile", upsertFile)
+            .put("upsertFile", storage.getBucketName() + "/" + upsertFile)
```
nawatts, line 101: Seems kind of odd that this isn't a URL, but 🤷

davidangb: argggg I agree! But it's what Import Service sends and what Rawls expects … from reading code I think Rawls would actually fail if this were a real URL

PR #645, opened 2024-03-12
[AJ-1635] Configure Dependabot to group minor and patch version updates

nawatts:
A couple of times now, I've noticed @davidangb and @jladieu combine multiple Dependabot updates into one PR. Instead of doing that manually, we can have Dependabot do it for us.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups


calypsomatic (approved):
Genius!

PR #646, opened 2024-03-12
AJ-1669: Write and publish only once.

jladieu:
For [AJ-1669](https://broadworkbench.atlassian.net/browse/AJ-1669): Rather than writing a Blob for every batch and every record type, record all writes for a whole batch into one Blob and send a single Pubsub message.
 
The PR currently contains 4 commits.  3 prefactoring & 1 actual change.

The prefactorings are done up front to ease review and reduce the footprint of the final actual change.

Each commit has a description to explain intent and is intended to be easier to review than the whole, so I recommend taking a look at each separately.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
@@ -44,6 +44,7 @@ import org.databiosphere.workspacedataservice.service.BatchWriteService;
 import org.databiosphere.workspacedataservice.service.CollectionService;
 import org.databiosphere.workspacedataservice.service.model.BatchWriteResult;
 import org.databiosphere.workspacedataservice.service.model.TdrManifestImportTable;
+import org.databiosphere.workspacedataservice.service.model.exception.DataImportException;
 import org.databiosphere.workspacedataservice.service.model.exception.TdrManifestImportException;
 import org.databiosphere.workspacedataservice.shared.model.CollectionId;
 import org.databiosphere.workspacedataservice.shared.model.RecordType;
@@ -161,9 +162,8 @@ public class TdrManifestQuartzJob extends QuartzJob {
                               .record()
                               .withRecordType(entry.getKey())
                               .ofQuantity(entry.getValue())));
-    } catch (IOException e) {
-      // TODO: better exception handling?
-      throw new RuntimeException(e);
+    } catch (DataImportException e) {
+      throw new TdrManifestImportException(e.getMessage(), e);
     }
 
     // delete temp files after everything else is completed
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
@@ -132,39 +132,39 @@ public class TdrManifestQuartzJob extends QuartzJob {
     // get all the parquet files from the manifests
 
     FileDownloadHelper fileDownloadHelper = getFilesForImport(tdrManifestImportTables);
-    RecordSink recordSink = recordSinkFactory.buildRecordSink(importDetails);
+    try (RecordSink recordSink = recordSinkFactory.buildRecordSink(importDetails)) {
+      // loop through the tables to be imported and upsert base attributes
+      var result =
+          importTables(
+              tdrManifestImportTables,
+              fileDownloadHelper.getFileMap(),
+              ImportMode.BASE_ATTRIBUTES,
+              recordSink);
 
-    // loop through the tables to be imported and upsert base attributes
-    var result =
-        importTables(
-            tdrManifestImportTables,
-            fileDownloadHelper.getFileMap(),
-            ImportMode.BASE_ATTRIBUTES,
-            recordSink);
+      // add relations to the existing base attributes
+      result.merge(
+          importTables(
+              tdrManifestImportTables,
+              fileDownloadHelper.getFileMap(),
+              ImportMode.RELATIONS,
+              recordSink));
 
-    // add relations to the existing base attributes
-    result.merge(
-        importTables(
-            tdrManifestImportTables,
-            fileDownloadHelper.getFileMap(),
-            ImportMode.RELATIONS,
-            recordSink));
-
-    // activity logging for import status
-    // no specific activity logging for relations since main import is a superset
-    result
-        .entrySet()
-        .forEach(
-            entry ->
-                activityLogger.saveEventForCurrentUser(
-                    user ->
-                        user.upserted()
-                            .record()
-                            .withRecordType(entry.getKey())
-                            .ofQuantity(entry.getValue())));
-
-    // Commit results, publish to downstream systems, etc.
-    recordSink.finalizeBatchWrite(result);
+      // activity logging for import status
+      // no specific activity logging for relations since main import is a superset
+      result
+          .entrySet()
+          .forEach(
+              entry ->
+                  activityLogger.saveEventForCurrentUser(
+                      user ->
+                          user.upserted()
+                              .record()
+                              .withRecordType(entry.getKey())
+                              .ofQuantity(entry.getValue())));
+    } catch (IOException e) {
+      // TODO: better exception handling?
+      throw new RuntimeException(e);
+    }
 
     // delete temp files after everything else is completed
     // Any failed deletions will be removed if/when pod restarts
--- a/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
@@ -39,12 +39,26 @@ public class RawlsRecordSink implements RecordSink {
   private final ImportDetails importDetails;
 
   RawlsRecordSink(
-      ObjectMapper mapper, GcsStorage storage, PubSub pubSub, ImportDetails importDetails) {
-    this.attributePrefixer = new RawlsAttributePrefixer(importDetails.prefixStrategy());
+      RawlsAttributePrefixer attributePrefixer,
+      ObjectMapper mapper,
+      GcsStorage storage,
+      PubSub pubSub,
+      ImportDetails importDetails) {
+    this.attributePrefixer = attributePrefixer;
     this.mapper = mapper;
     this.storage = storage;
     this.pubSub = pubSub;
     this.importDetails = importDetails;
+  }
+
+  public static RawlsRecordSink create(
+      ObjectMapper mapper, GcsStorage storage, PubSub pubSub, ImportDetails importDetails) {
+    return new RawlsRecordSink(
+        new RawlsAttributePrefixer(importDetails.prefixStrategy()),
+        mapper,
+        storage,
+        pubSub,
+        importDetails);
   }
 
   @Override
--- a/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
@@ -2,12 +2,14 @@ package org.databiosphere.workspacedataservice.recordsink;
 
 import static java.util.Objects.requireNonNull;
 
+import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.common.collect.ImmutableList;
+import com.google.cloud.spring.storage.GoogleStorageResource;
+import com.google.cloud.storage.Blob;
 import com.google.common.collect.ImmutableMap;
 import com.google.mu.util.stream.BiStream;
-import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -22,9 +24,11 @@ import org.databiosphere.workspacedataservice.recordsink.RawlsModel.CreateAttrib
 import org.databiosphere.workspacedataservice.recordsink.RawlsModel.Entity;
 import org.databiosphere.workspacedataservice.recordsink.RawlsModel.RemoveAttribute;
 import org.databiosphere.workspacedataservice.service.model.DataTypeMapping;
+import org.databiosphere.workspacedataservice.service.model.exception.DataImportException;
 import org.databiosphere.workspacedataservice.shared.model.Record;
 import org.databiosphere.workspacedataservice.shared.model.RecordType;
 import org.databiosphere.workspacedataservice.storage.GcsStorage;
+import org.springframework.util.function.ThrowingConsumer;
 
 /**
  * {@link RecordSink} implementation that produces Rawls-compatible JSON using {@link RawlsModel}
@@ -32,32 +36,47 @@ import org.databiosphere.workspacedataservice.storage.GcsStorage;
  */
 public class RawlsRecordSink implements RecordSink {
   private final RawlsAttributePrefixer attributePrefixer;
-  private final ObjectMapper mapper;
-  private final GcsStorage storage;
   private final PubSub pubSub;
   private final ImportDetails importDetails;
 
+  private final JsonWriter jsonWriter;
+  private final Blob blob;
+
   RawlsRecordSink(
       RawlsAttributePrefixer attributePrefixer,
-      ObjectMapper mapper,
-      GcsStorage storage,
+      JsonWriter jsonWriter,
       PubSub pubSub,
-      ImportDetails importDetails) {
+      ImportDetails importDetails,
+      Blob blob) {
     this.attributePrefixer = attributePrefixer;
-    this.mapper = mapper;
-    this.storage = storage;
+    this.jsonWriter = jsonWriter;
+
+    // TODO: These three instance variables are used exclusively for pubsub concerns; if we can
+    //   factor the pubsub responsibility out somehow, RawlsRecordSink can be a lot more focused.
     this.pubSub = pubSub;
     this.importDetails = importDetails;
+    this.blob = blob;
   }
 
+  /**
+   * Creates a {@link RawlsRecordSink} that writes a stream of JSON entities to the given {@link
+   * GcsStorage}.
+   *
+   * @param mapper used to generate the stream of JSON entities
+   * @param storage where the JSON stream will be written
+   * @param pubSub to notify upon JSON stream completion
+   * @param importDetails to use for generating the message to send to PubSub
+   */
   public static RawlsRecordSink create(
       ObjectMapper mapper, GcsStorage storage, PubSub pubSub, ImportDetails importDetails) {
+    String blobName = getBlobName(requireNonNull(importDetails.jobId()));
+    Blob blob = storage.createBlob(blobName);
     return new RawlsRecordSink(
         new RawlsAttributePrefixer(importDetails.prefixStrategy()),
-        mapper,
-        storage,
+        JsonWriter.create(storage.getOutputStream(blob), mapper),
         pubSub,
-        importDetails);
+        importDetails,
+        blob);
   }
 
   @Override
@@ -76,20 +95,8 @@ public class RawlsRecordSink implements RecordSink {
       Map<String, DataTypeMapping> schema, // ignored
       List<Record> records,
       String primaryKey // ignored
-      ) throws IOException {
-    ImmutableList.Builder<Entity> entities = ImmutableList.builder();
-    records.stream().map(this::toEntity).forEach(entities::add);
-
-    String upsertFileName =
-        storage.createGcsFile(
-            new ByteArrayInputStream(mapper.writeValueAsBytes(entities.build())),
-            requireNonNull(importDetails.jobId()));
-
-    publishToPubSub(
-        importDetails.collectionId(),
-        requireNonNull(importDetails.userEmail()),
-        importDetails.jobId(),
-        upsertFileName);
+      ) {
+    records.stream().map(this::toEntity).forEach(jsonWriter::writeEntity);
   }
 
   @Override
@@ -99,22 +106,8 @@ public class RawlsRecordSink implements RecordSink {
 
   @Override
   public void close() {
-    // currently a no-op
-    // TODO(AJ-1669): do pubsub here, maybe do GCS cleanup here (assuming a file was reused
-    //   throughout)
-  }
-
-  private void publishToPubSub(UUID workspaceId, String user, UUID jobId, String upsertFile) {
-    Map<String, String> message =
-        new ImmutableMap.Builder<String, String>()
-            .put("workspaceId", workspaceId.toString())
-            .put("userEmail", user)
-            .put("jobId", jobId.toString())
-            .put("upsertFile", storage.getBucketName() + "/" + upsertFile)
-            .put("isUpsert", "true")
-            .put("isCWDS", "true")
-            .build();
-    pubSub.publishSync(message);
+    jsonWriter.close();
+    publishToPubSub(importDetails);
   }
 
   private Entity toEntity(Record record) {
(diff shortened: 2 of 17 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -86,23 +85,32 @@ public void deleteBatch(RecordType recordType, List<Record> records) throws Batc
...
+  public void close() {
+    if (jsonWriter != null) {
+      try {
+        jsonWriter.close();
+        jsonWriter = null; // reset the writer; they aren't reusable
+      } catch (IOException e) {
+        throw new RuntimeException(e);
+      }
+    }
+    publishToPubSub(
```
jladieu, line 97: After switching the method to `close` this feels less like it belongs here and more like it belongs in an actual post-processing step.  I could almost see us doing some kind of observer pattern for this, but didn't want to blow up the scope of this and so am leaving it as is.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RecordSink.java
```diff
@@ -13,7 +13,7 @@
  * Implementations of this interface are responsible for modifying the schema and writing/deleting
  * batches of records.
  */
-public interface RecordSink {
+public interface RecordSink extends Closeable {
```
jladieu, line 16: Extending `Closeable` is key to ensuring `RecordSink#close()` gets called because it produces a warning when `RecordSink` is used without a corresponding try with resources block.  Without that warning, it's easy to accidentally initialize a `RecordSink` without calling close at some point.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -86,23 +85,32 @@ public void deleteBatch(RecordType recordType, List<Record> records) throws Batc
...
-            .put("isCWDS", "true")
-            .build();
-    pubSub.publishSync(message);
+  private void writeRecord(Record record) throws IOException {
+    getJsonWriter().writeEntity(toEntity(record));
+  }
+
+  private JsonWriter getJsonWriter() throws IOException {
+    if (jsonWriter == null) {
+      jsonWriter = new JsonWriter(destinationResource, mapper);
```
davidangb, line 110: any reason to do this lazily vs. creating it in the constructor and not worrying about the null check? Is it because `JsonWriter` is a pretty complex class?

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -128,4 +136,76 @@ private Stream<? extends AttributeOperation> toOperations(String name, Object at
...
+        throws IOException {
+      jsonGeneratorConsumer.accept(jsonGenerator);
+      jsonGenerator.flush();
+      blobWriter.write(ByteBuffer.wrap(byteStream.toByteArray(), 0, byteStream.size()));
+
+      // TODO: we should measure the actual buffer size so we can tweak accordingly, ideally we want
+      // to set a size that is bigger than the biggest chunk we ever see, but not so big as to be
+      // wasteful
+      byteStream.reset();
+    }
```
davidangb, line 203: Caveat: I don't have a concrete suggestion on how to do this better. But I worry about maintaining two separate things here - first, the `JsonGenerator` which writes to a `ByteArrayOutputStream`, and second the `WriteChannel` to GCS. On every write, you have to explicitly ask the `ByteArrayOutputStream` for its bytes, write those to the `WriteChannel`, and then flush/reset everything.

I am curious if there is a more integrated solution that alleviates the need to handle a lot of this ourselves. Possibly we could find or create a subclass of `DataOutput` that knows how to write to GCS, then create the generator pointing at that `DataOutput`? Or possibly a `Writer` - check out the various overrides of `createGenerator()`.

davidangb (commented):
design-level thoughts inline; I haven't done a line-level code review.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
davidangb: Do you need the `WriteChannel`? Could you do just:
```java
this.jsonGenerator = objectMapper.getFactory().createGenerator(resource.getOutputStream());
```
? I guess you lose the benefits of the writechannel.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
jladieu: I waffled a bunch on this and didn't like doing it this way so I'd love to figure out how to do what you're suggesting.

I started with this approach because I didn't think creating a `RawlsRecordSink` instance should have the side effect of putting stuff in GCS, which it would if I did this in the constructor.  This is because the first thing `JsonWriter` does when instantiated is write a "[" to the JSON stream, which would then get serialized to GCS.  Even this feels a bit off to me, now that I think of it.

I could change this around a bit and make it so `JsonWriter` doesn't immediately do that.  Maybe instead of lazily initializing the `jsonWriter` we could just lazily write that "[" out on the first invocation?  That would let `jsonWriter` be initialized in the constructor, but wouldn't result in creation of a Blob until a write occurs.

I think I'd still need to maintain some kind of boolean state in `JsonWriter` either way "writesActive = true" or something which dictates whether `close()` does something, and would probably need to make `JsonWriter` resettable or reusable somehow.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
jladieu: I also think this is gross and it felt weird when I was implementing it.   It was even worse before I pulled out the `JsonWriter` to at least encapsulate this stuff.  I'll think more about this.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -86,23 +85,32 @@ public void deleteBatch(RecordType recordType, List<Record> records) throws Batc
...
+      try {
+        jsonWriter.close();
+        jsonWriter = null; // reset the writer; they aren't reusable
+      } catch (IOException e) {
+        throw new RuntimeException(e);
+      }
+    }
+    publishToPubSub(
+        importDetails.collectionId(),
+        requireNonNull(importDetails.userEmail()),
```
ashanhol, line 99: I'm still getting used to the nullable stuff, why is this explicitly called out as require non null? aren't they all required to be non null unless marked with `@nullable`?

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -128,4 +136,76 @@ private Stream<? extends AttributeOperation> toOperations(String name, Object at
...
+      // well pick a reasonable starting point
+      this.byteStream = new ByteArrayOutputStream(DEFAULT_BUFFER_SIZE);
+      this.jsonGenerator = objectMapper.getFactory().createGenerator(byteStream);
+      this.blobWriter = resource.createBlob().writer();
+
+      // TODO: wire this up through props so we can override it in test only
+      // this is a hack to work around LocalStorageHelper, which fails when chunk size exceeded due
+      // to not being able to handle Content-Range header with an unspecified size; in prod
+      // this probably won't be an issue, so we should start with the default which is 16MiB
+      blobWriter.setChunkSize(64 * _1MiB);
```
ashanhol, line 185: is there a reason `64 * _1MiB` isn't a const like the ones you set above?

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -128,4 +136,76 @@ private Stream<? extends AttributeOperation> toOperations(String name, Object at
...
+      blobWriter.write(ByteBuffer.wrap(byteStream.toByteArray(), 0, byteStream.size()));
+
+      // TODO: we should measure the actual buffer size so we can tweak accordingly, ideally we want
+      // to set a size that is bigger than the biggest chunk we ever see, but not so big as to be
+      // wasteful
+      byteStream.reset();
+    }
+
+    public void close() throws IOException {
+      writeJson(JsonGenerator::writeEndArray); // write a "]" to end the array of entities
```
ashanhol, line 206: It's clear we need to encapsulate what we're writing in in square brackets, but I don't love the method of writing `[` on creation and `]` on close. Is there any way we can just pre/append the square brackets before we send the data?

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
ashanhol: When is close() called? I'm not familiar with `Closable` so I'm trying to understand the new flow. Instead of us calling `recordSink.finalizeBatchWrite` (which is writing multiple times), close() will get called in some sort of automatic process?

jladieu: This is now using a 'try with resources' block.  This is a good explainer: https://www.baeldung.com/java-try-with-resources (first 3 bullets are enough to get the gist).

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
jladieu: `importDetails.userEmail()` is marked `@Nullable` at its [declaration](https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java#L9)

While `publishToPubSub` expects to receive a `@NonNull String` for this argument (as is now the default expectation).

`Objects.requireNonNull` has the nice side effects of 

1. throwing an exception if given a `null` object
2. returning a `@NonNull` version of what it is given, thereby satisfying the `@NonNull` requirement of `publishToPubSub`

Since `publishToPubSub` won't work correctly if given `null`, I'm ok with this crashing right here instead of letting `null` through, so `requireNonNull` does the trick here.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
jladieu: Ehhhh not really, I was tweaking it a lot to try to figure out how high I needed it to go to get the test to pass.  I don't think this way of representing the value will survive its extraction to a config property where I'm pretty sure I'll be required to set a plain old integer value.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
jladieu: I agree this is gross.  I don't know of a better way than to attach to _some_ kind of lifecycle method.  Rather than having it be a side-effect of creation or destruction, I could add some kind of callback hooks (eg: `start()`/`stop()` on the `RecordSink` interface, but I didn't really think this was merited yet.

One possible thing to try: maybe we just don't need the start/stop square brackets (?) and Rawls will merrily still parse the whole file as an array?

ashanhol: side question that might help me understand this better- what is the JsonWriter (specifically its byte stream) used for? It doesn't seem obviously involved with publishToPubSub?

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
ashanhol: that's really cool

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
jladieu: I didn't end up with the exact snippet of your suggestion, but I got to having only one resource to manage.  I'm a lot more happy with the result: c36e08692bd0863b7520dbd48ebb3a69233e15cb

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkFactory.java
```diff
@@ -31,6 +31,6 @@ public RecordSink buildRecordSink(ImportDetails importDetails) {
   }
 
   private RecordSink rawlsRecordSink(ImportDetails importDetails) {
-    return new RawlsRecordSink(mapper, storage, pubSub, importDetails);
+    return RawlsRecordSink.create(mapper, storage, pubSub, importDetails);
```
jladieu, line 34: Here and the next two diffs: Instead of throwing exceptions out of constructors, use a factory
method and throw from that.  This lets the constructors become really
dumb and just set fields, which should improve unit testability.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -30,31 +30,54 @@
...
+   *
+   * @param mapper used to generate the stream of JSON entities
+   * @param storage where the JSON stream will be written
+   * @param pubSub to notify upon JSON stream completion
+   * @param importDetails to use for generating the message to send to PubSub
+   */
+  public static RawlsRecordSink create(
+      ObjectMapper mapper, GcsStorage storage, PubSub pubSub, ImportDetails importDetails) {
+    String blobName = getBlobName(requireNonNull(importDetails.jobId()));
+    return new RawlsRecordSink(
```
jladieu, line 75: Instead of throwing exceptions out of constructors, use a factory method and throw from that. This lets the constructors become really dumb and just set fields, which should improve unit testability.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -168,28 +167,65 @@ private static class JsonWriter implements Closeable {
...
-      writeJson(JsonGenerator::writeEndArray); // write a "]" to end the array of entities
-      jsonGenerator.close();
+    /**
+     * Terminate the JSON array and close the underlying {@link JsonGenerator}.
+     *
+     * @throws JsonWriteException if the underlying {@link JsonGenerator} fails to close
+     */
+    @Override
+    public void close() {
+      if (streamStarted) {
```
jladieu, line 209: This is still brittle and would probably be better handled as a state machine that progresses from INITIALIZED -> WRITING -> CLOSED, but I didn't want to start pulling that until determining if this is a reasonable way to defer writing to GCS without lazily initializing `JsonWriter`

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
jladieu: This is a bit of a speculative commit: fa4be2211f8b5471f9540cf2bc13467a61d11c6c to try to address the concerns of lazy initialization vs. having a full set of non-nullable objects, plus knock out the ergonomic concern of throwing IOExceptions everywhere, curious what you think...should I lean into it and add a simple state machine to make it less brittle, pivot, or abort?

jladieu:
I pushed some more commits to address some (not all) of the feedback.  I like all of the commits, but I'm unsure about the last one which felt more speculative.  Happy to roll that one back if it's too much but looking for a nudge.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -128,4 +136,114 @@ private Stream<? extends AttributeOperation> toOperations(String name, Object at
...
+    }
+
+    private static OutputStream getOutputStream(GoogleStorageResource resource) {
+      WriteChannel writeChannel = resource.createBlob().writer();
+      // TODO: wire chunkSize override through props so we can override it only in tests
+      // this is a hack to work around LocalStorageHelper, which fails when chunk size exceeded due
+      // to not being able to handle Content-Range header with an unspecified size; in prod
+      // this probably won't be an issue, so we should start with the default which is 16MiB
+      writeChannel.setChunkSize(64 * _1MiB);
+      return Channels.newOutputStream(writeChannel);
```
davidangb, line 235: nice, you found a way to keep the writechannel but still use an outputstream!

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
davidangb: I think the `streamStarted` boolean is good enough, state machine isn't needed (but state machine would achieve the same goal)

davidangb (commented):
yeah, I like this approach - centralizing writes through the `JsonGenerator`. Looks good to me design-wise, give a shout when it's ready for full code review!

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
ashanhol: that's a lot clearer, thank you

service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
```diff
@@ -1,31 +1,36 @@
 package org.databiosphere.workspacedataservice.config;
```
jladieu, line 1: All the changes in this file (and corresponding changes in classes that use the props here) ended up being optional in the end.   I did this thinking I'd need to use property injection to override chunkSize for the tests, but found an easier insertion point in `mockGcsStorage`.

I figured they were a net positive and kept them in, but if it's too much churn I can back them out.

service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorage.java
```diff
@@ -9,57 +11,56 @@
...
+    this.blobWriterChunkSize = blobWriterChunkSize;
   }
 
-  public InputStream getBlobContents(String blobName) throws IOException {
-    return getGcsResource(blobName).getInputStream();
+  GcsStorage(Storage storage, String bucketName) {
+    this(storage, bucketName, DEFAULT_CHUNK_SIZE);
   }
 
-  public String createGcsFile(InputStream contents, UUID jobId) throws IOException {
```
jladieu, line 52: This method basically got replaced by `getBlobOutputStream` and writing to that; in removing this I also updated the tests that were its only consumer to use the new write path

service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorage.java
```diff
@@ -75,4 +76,15 @@ public Iterable<Blob> getBlobsInBucket() {
   public void deleteBlob(String blobName) {
     storage.delete(this.bucketName, blobName);
   }
+
+  private GoogleStorageResource getGcsResource(String blobName) {
```
jladieu, line 80: Was able to make this private again after inlining `getBlobOutputStream`

service/src/main/resources/application-control-plane.yml
```diff
@@ -12,8 +12,6 @@ twds:
     succeed-on-completion: false
     project-id: ${SERVICE_GOOGLE_PROJECT}
     rawls-bucket-name: ${SERVICE_GOOGLE_BUCKET}
-
```
jladieu, line 15: Just remnants of my temporary edit to this file...I ended up backing the new configurable out and handling it as a default, but we could also call it out explicitly here if we wanted to.

service/src/test/java/org/databiosphere/workspacedataservice/config/TwdsPropertiesControlPlaneTest.java
```diff
@@ -36,7 +36,7 @@ void nonNullDataImport() {
 
   @Test
   void batchWriteRecordSink() {
-    assertEquals(RecordSinkMode.RAWLS, dataImportProperties.getBatchWriteRecordSink());
+    assertThat(dataImportProperties.getBatchWriteRecordSink()).hasValue(RecordSinkMode.RAWLS);
```
jladieu, line 39: Note: `RecordSinkMode` is no longer used since https://github.com/DataBiosphere/terra-workspace-data-service/pull/592, so I technically could just rip this stuff out instead.

Alternately, we could adjust `RecordSinkFactory` beans over to use the RecordSinkMode instead of `@DataPlane` or `@ControlPlane` annotation.  Leaving this comment here for discussion about this, but otherwise planning to leave this as is for the PR.

service/src/test/java/org/databiosphere/workspacedataservice/storage/GcsStorageTest.java
```diff
@@ -33,15 +34,31 @@ class GcsStorageTest {
   private GcsStorage storage;
 
   @Test
-  void testCreateandGetBlobSimple() throws IOException {
+  void testCreateAndGetBlobSimple() throws IOException {
```
jladieu, line 37: Switched over the existing unit test to use the invocation path actually taken by `RawlsRecordSink`, and added some extra coverage for the other public methods.

service/src/test/java/org/databiosphere/workspacedataservice/config/TwdsPropertiesControlPlaneTest.java
jladieu: This comment is irrelevent now, I backed this change out after I saw some pretty extensive changes going on in this file in another PR, so I'm leaving it out for now.

service/src/main/resources/application-control-plane.yml
jladieu: This comment is irrelevent now, I backed this change out after I saw some pretty extensive changes going on in this file in another PR, so I'm leaving it out for now.

service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
jladieu: This comment is irrelevent now, I backed this change out after I saw some pretty extensive changes going on in this file in another PR, so I'm leaving it out for now.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
jladieu: Ok this config got relocated and turned less magic in 7179710b0ea747d4e34079cf31d85e1eafdc8aaa

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -132,39 +133,38 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
...
-                activityLogger.saveEventForCurrentUser(
-                    user ->
-                        user.upserted()
-                            .record()
-                            .withRecordType(entry.getKey())
-                            .ofQuantity(entry.getValue())));
-
-    // Commit results, publish to downstream systems, etc.
-    recordSink.finalizeBatchWrite(result);
+    try (RecordSink recordSink = recordSinkFactory.buildRecordSink(importDetails)) {
```
jladieu, line 136: This chunk introduces a try { } catch { } block and is mostly indentation.

service/src/test/java/org/databiosphere/workspacedataservice/storage/GcsStorageTest.java
```diff
@@ -1,47 +1,153 @@
...
-    String contents =
-        StreamUtils.copyToString(storage.getBlobContents(newBlobName), Charset.defaultCharset());
-    assertThat(contents).isEqualTo(initialString);
+
+    // Act
+    try (OutputStream outputStream = storage.getOutputStream("testBlobName")) {
+      outputStream.write(initialString.getBytes());
+    }
+
+    // Act / Assert
```
calypsomatic, line 49: Nit: this looks like it's only Assert and not Act/Assert

service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorage.java
```diff
@@ -9,57 +13,75 @@
...
+    return getOutputStream(createBlob(name));
+  }
+
+  /**
+   * Creates a {@link Blob} with the given name in the configured bucket.
+   *
+   * @return the newly created Blob
+   */
+  public Blob createBlob(String name) {
+    return getGcsResource(name).createBlob();
```
davidangb, line 84: mark as `@VisibleForTesting`? I believe this is only used by tests, correct?

davidangb (approved):
one minor comment inline, proactive :+1: and thank you, this is an important one!

service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorage.java
jladieu: I ended up using this in `RawlsRecordSink#create`, because the `Blob` provided a nice handle on both the bucketName & the blobName, usable to generate the `upsertFile` for pubsub.

service/src/test/java/org/databiosphere/workspacedataservice/storage/GcsStorageTest.java
jladieu: I took a pass through the whole file to tidy this up in cd52e615fbe9f2172fd2e59a8b5b33caaa2ea6ab

yuliadub (approved):
LGTM

ashanhol (approved):
Looks good to me :)

PR #648, opened 2024-03-13
[AJ-1588] Validate URLs for file imports

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1588

Per security requirements, we can't import from arbitrary URLs. This restricts import URLs to HTTPS URLs pointing to files in GCS buckets, Azure blob storage containers, or S3 buckets. Additional allowed import services can be configured.

One complication is that some import tests use a file URL for a local resource. I added another configuration option (to be used only in tests) to allow different URL schemes (file for some unit tests and http for client tests).

### DIFF
--- a/.github/workflows/release-python-client.yml
+++ b/.github/workflows/release-python-client.yml
@@ -69,7 +69,7 @@ jobs:
           PGPASSWORD: postgres
         run: psql -h 127.0.0.1 -U postgres -f ./local-dev/local-postgres-init.sql
 
-      - name: Mock Sam via command-line docker run
+      - name: Mock APIs via command-line docker run
         run: |
           docker run -v ${{ github.workspace }}/service/src/test/resources/nginx.conf:/etc/nginx/nginx.conf:ro -v ${{ github.workspace }}/service/src/test/resources:/usr/share/nginx/html -p 9889:80 -d nginx:1.23.3
 
@@ -90,6 +90,7 @@ jobs:
           export WORKSPACE_ID=123e4567-e89b-12d3-a456-426614174000
           export SPRING_PROFILES_ACTIVE=data-plane
           ./gradlew --build-cache build -x test
+          # Allow http imports from localhost to test with files from resources served by the mock API server
           ./gradlew bootRun --args='--twds.data-import.allowed-hosts=localhost --twds.data-import.allowed-schemes=http' &
           count=20
           until $(curl --output /dev/null --silent --head --fail http://localhost:8080/status); do
--- a/.github/workflows/release-python-client.yml
+++ b/.github/workflows/release-python-client.yml
@@ -90,7 +90,7 @@ jobs:
           export WORKSPACE_ID=123e4567-e89b-12d3-a456-426614174000
           export SPRING_PROFILES_ACTIVE=data-plane
           ./gradlew --build-cache build -x test
-          ./gradlew bootRun &
+          ./gradlew bootRun --args='--twds.data-import.allowed-hosts=localhost --twds.data-import.allowed-schemes=http' &
           count=20
           until $(curl --output /dev/null --silent --head --fail http://localhost:8080/status); do
             printf '.'
--- a/service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
@@ -1,29 +1,31 @@
 package org.databiosphere.workspacedataservice.config;
 
+import static java.util.Arrays.stream;
+import static java.util.Collections.emptySet;
+
 import com.google.common.collect.Sets;
-import java.net.URI;
-import java.util.Arrays;
-import java.util.Collections;
 import java.util.Set;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import org.springframework.lang.Nullable;
 
 /** Properties that dictate how data import processes should behave. */
 public class DataImportProperties {
+  private static final Set<Pattern> defaultAllowedHosts =
+      Set.of(
+          Pattern.compile("storage\\.googleapis\\.com"),
+          Pattern.compile(".*\\.core\\.windows\\.net"),
+          // S3 allows multiple URL formats
+          // https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html
+          Pattern.compile("s3\\.amazonaws\\.com"), // path style legacy global endpoint
+          Pattern.compile(".*\\.s3\\.amazonaws\\.com") // virtual host style legacy global endpoint
+          );
   private RecordSinkMode batchWriteRecordSink;
   private String projectId;
   private String rawlsBucketName;
   private boolean succeedOnCompletion;
-  private final Set<AllowedHost> defaultAllowedHosts =
-      Set.of(
-          new AllowedHost("storage.googleapis.com"),
-          new AllowedHost("*.core.windows.net"),
-          // S3 allows multiple URL formats
-          // https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html
-          new AllowedHost("s3.amazonaws.com"), // path style legacy global endpoint
-          new AllowedHost("*.s3.amazonaws.com") // virtual host style legacy global endpoint
-          );
-  private Set<AllowedHost> allowedHosts = Collections.emptySet();
+
+  private Set<Pattern> allowedHosts = emptySet();
   private Set<String> allowedSchemes = Set.of("https");
 
   /** Where to write records after import, options are defined by {@link RecordSinkMode} */
@@ -73,15 +75,15 @@ public class DataImportProperties {
    * Accepted sources for imported files. This includes configured sources as well as default /
    * always allowed sources (GCS buckets, Azure storage containers, and S3 buckets).
    */
-  public Set<AllowedHost> getAllowedHosts() {
+  public Set<Pattern> getAllowedHosts() {
     return Sets.union(defaultAllowedHosts, allowedHosts);
   }
 
   public void setAllowedHosts(@Nullable String[] allowedHosts) {
     this.allowedHosts =
         allowedHosts == null
-            ? Collections.emptySet()
-            : Arrays.stream(allowedHosts).map(AllowedHost::new).collect(Collectors.toSet());
+            ? emptySet()
+            : stream(allowedHosts).map(Pattern::compile).collect(Collectors.toSet());
   }
 
   public Set<String> getAllowedSchemes() {
@@ -90,9 +92,7 @@ public class DataImportProperties {
 
   public void setAllowedSchemes(@Nullable String[] allowedSchemes) {
     this.allowedSchemes =
-        allowedSchemes == null
-            ? Collections.emptySet()
-            : Arrays.stream(allowedSchemes).collect(Collectors.toSet());
+        allowedSchemes == null ? emptySet() : stream(allowedSchemes).collect(Collectors.toSet());
   }
 
   /** Dictates the sink where BatchWriteService should write records after import. */
@@ -112,22 +112,6 @@ public class DataImportProperties {
         }
       }
       throw new RuntimeException("Unknown RecordSinkMode value: %s".formatted(value));
-    }
-  }
-
-  public static class AllowedHost {
-    private final String pattern;
-
-    public AllowedHost(String hostPattern) {
-      this.pattern = hostPattern;
-    }
-
-    public boolean matchesUrl(URI url) {
-      if (pattern.startsWith("*")) {
-        return url.getHost().endsWith(pattern.substring(1));
-      } else {
-        return url.getHost().equals(pattern);
-      }
     }
   }
 }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
@@ -24,6 +24,7 @@ public class DataImportProperties {
           new AllowedImportSource("*.s3.amazonaws.com") // virtual host style legacy global endpoint
           );
   private Set<AllowedImportSource> allowedImportSources = Collections.emptySet();
+  private boolean fileImportsAllowed = false;
 
   /** Where to write records after import, options are defined by {@link RecordSinkMode} */
   public RecordSinkMode getBatchWriteRecordSink() {
@@ -83,6 +84,15 @@ public class DataImportProperties {
             : Arrays.stream(allowedImportSources)
                 .map(AllowedImportSource::new)
                 .collect(Collectors.toSet());
+  }
+
+  /** Whether files may be imported from file: URLs. This should only be enabled for tests. */
+  public boolean fileImportsAllowed() {
+    return fileImportsAllowed;
+  }
+
+  public void setAllowFileImports(boolean allowFileImports) {
+    this.fileImportsAllowed = allowFileImports;
   }
 
   /** Dictates the sink where BatchWriteService should write records after import. */
--- a/service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
@@ -24,7 +24,7 @@ public class DataImportProperties {
           new AllowedHost("*.s3.amazonaws.com") // virtual host style legacy global endpoint
           );
   private Set<AllowedHost> allowedHosts = Collections.emptySet();
-  private boolean fileImportsAllowed = false;
+  private Set<String> allowedSchemes = Set.of("https");
 
   /** Where to write records after import, options are defined by {@link RecordSinkMode} */
   public RecordSinkMode getBatchWriteRecordSink() {
@@ -84,13 +84,15 @@ public class DataImportProperties {
             : Arrays.stream(allowedHosts).map(AllowedHost::new).collect(Collectors.toSet());
   }
 
-  /** Whether files may be imported from file: URLs. This should only be enabled for tests. */
-  public boolean fileImportsAllowed() {
-    return fileImportsAllowed;
+  public Set<String> getAllowedSchemes() {
+    return allowedSchemes;
   }
 
-  public void setAllowFileImports(boolean allowFileImports) {
-    this.fileImportsAllowed = allowFileImports;
+  public void setAllowedSchemes(@Nullable String[] allowedSchemes) {
+    this.allowedSchemes =
+        allowedSchemes == null
+            ? Collections.emptySet()
+            : Arrays.stream(allowedSchemes).collect(Collectors.toSet());
   }
 
   /** Dictates the sink where BatchWriteService should write records after import. */
--- a/service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
@@ -113,8 +113,8 @@ public class DataImportProperties {
     }
   }
 
-  public class AllowedHost {
-    private String pattern;
+  public static class AllowedHost {
+    private final String pattern;
 
     public AllowedHost(String hostPattern) {
       this.pattern = hostPattern;
--- a/service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
@@ -7,7 +7,6 @@ import com.google.common.collect.Sets;
 import java.util.Set;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
-import org.springframework.lang.Nullable;
 
 /** Properties that dictate how data import processes should behave. */
 public class DataImportProperties {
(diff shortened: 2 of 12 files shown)


nawatts:
Ugh, the Python client tests import from `http://localhost`. Back to drawing board...

.github/workflows/release-python-client.yml
```diff
@@ -90,7 +90,7 @@ jobs:
           export WORKSPACE_ID=123e4567-e89b-12d3-a456-426614174000
           export SPRING_PROFILES_ACTIVE=data-plane
           ./gradlew --build-cache build -x test
-          ./gradlew bootRun &
+          ./gradlew bootRun --args='--twds.data-import.allowed-hosts=localhost --twds.data-import.allowed-schemes=http' &
```
jladieu, line 93: If you base this off @yuliadub's https://github.com/DataBiosphere/terra-workspace-data-service/pull/650 you can  stick these overrides in `application-local.yml` instead

service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
```diff
@@ -69,4 +114,20 @@ static RecordSinkMode fromValue(String value) {
       throw new RuntimeException("Unknown RecordSinkMode value: %s".formatted(value));
     }
   }
+
+  public static class AllowedHost {
+    private final String pattern;
```
jladieu, line 119: The patterns you're using _kinda_ look like you're trying to use a Regex, but the way you're using them is just with string equality...do you want full regex support here?  If so, it might be better to use an actual regex and the `match` methods on them.

I wasn't sure how to do this in practice, so I asked ChatGPT to generate something for me that used this...and it came up with something that seemed pretty good, and even produced some unit tests.

Do with it what ye will.


```
    private final List<Pattern> allowedDomainPatterns;

    public DomainAllowlistChecker(List<String> allowedDomains) {
        // Convert each allowed domain into a regex pattern, assuming simple domain names for illustration
        this.allowedDomainPatterns = allowedDomains.stream()
                .map(domain -> domain.replace(".", "\\.").replace("*", ".*")) // Convert wildcards to regex
                .map(Pattern::compile)
                .collect(Collectors.toList());
    }

    public boolean isUrlAllowed(URI url) {
        String host = url.getHost();
        return allowedDomainPatterns.stream().anyMatch(pattern -> pattern.matcher(host).matches());
    }

    public static void main(String[] args) {
        List<String> allowedDomains = Arrays.asList("*.example.com", "allowed.org");
        DomainAllowlistChecker checker = new DomainAllowlistChecker(allowedDomains);

        // Test URLs
        System.out.println(checker.isUrlAllowed(URI.create("http://sub.example.com"))); // true
        System.out.println(checker.isUrlAllowed(URI.create("http://disallowed.com"))); // false
        System.out.println(checker.isUrlAllowed(URI.create("http://allowed.org"))); // true
    }
}
```

service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
```diff
@@ -1,11 +1,30 @@
...
+import java.util.stream.Collectors;
+import org.springframework.lang.Nullable;
+
 /** Properties that dictate how data import processes should behave. */
 public class DataImportProperties {
   private RecordSinkMode batchWriteRecordSink;
   private String projectId;
   private String rawlsBucketName;
   private boolean succeedOnCompletion;
+  private final Set<AllowedHost> defaultAllowedHosts =
```
jladieu, line 17: as these are constants, I'd recommend putting them at the top of the list of fields and using the convention:

`private static final Set<AllowedHost> DEFAULT_ALLOWED_HOSTS = ...`

service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
```diff
@@ -50,6 +69,32 @@ public void setSucceedOnCompletion(boolean succeedOnCompletion) {
...
 
+  /**
+   * Accepted sources for imported files. This includes configured sources as well as default /
+   * always allowed sources (GCS buckets, Azure storage containers, and S3 buckets).
+   */
+  public Set<AllowedHost> getAllowedHosts() {
+    return Sets.union(defaultAllowedHosts, allowedHosts);
+  }
+
+  public void setAllowedHosts(@Nullable String[] allowedHosts) {
```
jladieu, line 80: Is this `@Nullable` because Spring sends `null` to you when this property isn't configured?

service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
```diff
@@ -50,6 +69,32 @@ public void setSucceedOnCompletion(boolean succeedOnCompletion) {
...
+  }
+
+  public Set<String> getAllowedSchemes() {
+    return allowedSchemes;
+  }
+
+  public void setAllowedSchemes(@Nullable String[] allowedSchemes) {
+    this.allowedSchemes =
+        allowedSchemes == null
+            ? Collections.emptySet()
```
jladieu, line 94: Static imports can really tighten this method implementation up (same with the earlier one that looks similar).

```
import static java.util.Arrays.stream;
import static java.util.Collections.emptySet;
import static java.util.stream.Collectors.toSet;

...

public void setAllowedSchemes(@Nullable String[] allowedSchemes) {
        this.allowedSchemes =
        allowedSchemes == null ? emptySet() : stream(allowedSchemes).collect(toSet());
  }
```

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
```diff
@@ -136,4 +144,22 @@ protected Schedulable createSchedulable(
...
+
+    if (!dataImportProperties.getAllowedSchemes().contains(importUrl.getScheme())) {
+      throw new ValidationException(
+          "Files may not be imported from %s URLs.".formatted(importUrl.getScheme()));
+    }
+
+    // File URLs don't have a host to validate.
+    boolean isFileUrl = importUrl.getScheme().equals("file");
+    if (!isFileUrl
+        && dataImportProperties.getAllowedHosts().stream()
```
jladieu, line 159: I think instead of grabbing the `AllowedHosts` out of `dataImportProperties` it'd make more sense to create a new bean like `ImportSourceValidator` that itself has `DataImportProperties` injected.

public class ImportSourceValidator {

    private final Set<AllowedHosts> ...

    ImportSourceValidator(DataImportProperties properties) {
       this.allowedHosts = properties.getAllowedHosts();
    }

    public void validateXyz(URL url) {
      ...
    }

    // etc.
}

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
```diff
@@ -136,4 +144,22 @@ protected Schedulable createSchedulable(
...
+  private void validateImportRequest(ImportRequestServerModel importRequest) {
+    URI importUrl = importRequest.getUrl();
+
+    if (!dataImportProperties.getAllowedSchemes().contains(importUrl.getScheme())) {
+      throw new ValidationException(
+          "Files may not be imported from %s URLs.".formatted(importUrl.getScheme()));
+    }
+
+    // File URLs don't have a host to validate.
+    boolean isFileUrl = importUrl.getScheme().equals("file");
```
jladieu, line 157: Won't isFileUrl always be false?  I kinda assume from the prior conditional that it will just throw if the scheme is `file` before it ever reaches this.

jladieu: Oh I was assuming it was hardcoded to `https`, but I realize some of your tests you set the allowed-schemes to `file`...

I think maybe I don't understand under what circumstances the allowed schemes will change...

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceTest.java
```diff
@@ -229,6 +232,71 @@ void doesNotCreateJobWithoutPetToken(ImportRequestServerModel.TypeEnum importTyp
     verify(jobDao, times(0)).createJob(any());
   }
 
+  @ParameterizedTest(name = "for import type {0}, should require HTTPS URLs")
+  @EnumSource(TypeEnum.class)
```
jladieu, line 236: Normally I'm all for abbreviating fully-qualified class names, but in this case I think it's hard to tell what `TypeEnum` you're talking about since it's such a super generic class name, so I'd suggest fully qualifying it: `ImportRequestServerModel.TypeEnum`

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceTest.java
```diff
@@ -229,6 +232,71 @@ void doesNotCreateJobWithoutPetToken(ImportRequestServerModel.TypeEnum importTyp
...
+  void testRequiresHttpsImportUrls(TypeEnum importType) {
```
jladieu, line 237: Nit: I think conventionally when tests are annotated with `@Test` or `@ParameterizedTest` you can drop the `test` suffix and just go with a statement, like:

`requiresHttpsImportUrls`

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceTest.java
```diff
@@ -229,6 +232,71 @@ void doesNotCreateJobWithoutPetToken(ImportRequestServerModel.TypeEnum importTyp
...
+
+    // Act
+    URI importUri = URI.create("https://files.terra.bio/file");
+    ImportRequestServerModel importRequest = new ImportRequestServerModel(importType, importUri);
+    GenericJobServerModel createdJob =
+        importService.createImport(defaultCollectionId().id(), importRequest);
+
+    // Assert
+    GenericJobServerModel actual = jobDao.getJob(createdJob.getJobId());
+    assertEquals(GenericJobServerModel.StatusEnum.QUEUED, actual.getStatus());
```
jladieu, line 275: Supernit: This is opposite advice to my earlier comment on `TypeEnum`!  In this case I think `StatusEnum.QUEUED` is enough to describe the type you're checking, so if you adjust the imports (IntelliJ does this for me quickly, I can show you how), you can drop the `GenericJobServerModel.` prefix which is a mouthful.

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceTest.java
```diff
@@ -55,7 +57,7 @@
 
 @ActiveProfiles(profiles = {"mock-sam", "mock-collection-dao"})
```
jladieu, line 58:
```suggestion
@ActiveProfiles(profiles = {"mock-sam", "mock-collection-dao", "noop-scheduler-dao"})
```

If you do this, you won't have to stub out scheduler throughout.

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceTest.java
```diff
@@ -229,6 +232,71 @@ void doesNotCreateJobWithoutPetToken(ImportRequestServerModel.TypeEnum importTyp
     verify(jobDao, times(0)).createJob(any());
   }
 
+  @ParameterizedTest(name = "for import type {0}, should require HTTPS URLs")
+  @EnumSource(TypeEnum.class)
+  void testRequiresHttpsImportUrls(TypeEnum importType) {
+    // Arrange
+    // schedulerDao.schedule(), which returns void, returns successfully
+    doNothing().when(schedulerDao).schedule(any(Schedulable.class));
```
jladieu, line 240: You can drop this and the `MockBean schedulerDao` if you use `noop-scheduler-dao` to the list of `@ActiveProfiles`

.github/workflows/release-python-client.yml
nawatts: I'm not sure I agree with putting these in local. A few tests need configuration, but for local development, mostly I want the app to behave as it normally would in production.

service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
nawatts: I think I assumed that would be the case, but I'm not actually sure.

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
nawatts: Right, additional/other schemes can be configured. And file URLs are a bit of an outlier since they don't have a host, so we can't apply host validation to them.

service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
nawatts: Done in 5c441d33d99b09010031b6119886e5ee1d9f0c06.

service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
nawatts: Actually, it looks like the method doesn't get called if the configuration isn't present. Removed this in 874fafc1d935adbccd6dd84abcb9971ace3068c1.

service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
nawatts: Done in 5c441d33d99b09010031b6119886e5ee1d9f0c06.

service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
nawatts: I guess it might be easier to use regexes instead of requiring the configurer to understand a bespoke pattern. Switched to regexes in 5c441d33d99b09010031b6119886e5ee1d9f0c06.

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceTest.java
nawatts: Since this test is copied from existing tests, leaving this alone for now. I can clean up all tests in this class in a follow up PR.

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceTest.java
nawatts: Leaving this alone for now since it matches other tests in this class.

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceTest.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/648/commits/eaeed33d024824af888743fce69ff0a3ed454bd6. I'm conditioned by pytest.

calypsomatic (approved):

.github/workflows/release-python-client.yml
jladieu: 🤷 👍

service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
jladieu: supernit: constant naming convention `DEFAULT_ALLOWED_HOSTS`

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportSourceValidator.java
```diff
@@ -0,0 +1,34 @@
...
+import org.databiosphere.workspacedataservice.generated.ImportRequestServerModel;
+import org.databiosphere.workspacedataservice.service.model.exception.ValidationException;
+import org.springframework.stereotype.Component;
+
+@Component
+public class ImportSourceValidator {
+  private final DataImportProperties dataImportProperties;
+
+  public ImportSourceValidator(DataImportProperties dataImportProperties) {
+    this.dataImportProperties = dataImportProperties;
```
jladieu, line 14: Super nit: Now that I see how you're using the allowedHosts and allowedSchemas down below, consider assigning them to local instance variables for a tiny improvement in readability down below in the actual logic:

```
private final Set<String> allowedSchemes;
private final Set<Pattern> allowedHosts;


public ImportSourceValidator(DataImportProperties dataImportProperties) {
    this.allowedSchemes = dataImportProperties.getAllowedSchemes();
    this.allowedHosts = dataImportProperties.getAllowedHosts();
}
```

Then, down in `validateImportRequest` you don't have to traverse through the props to get at the allowlists which will make things a little bit more succinct.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportSourceValidator.java
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  public void validateImportRequest(ImportRequestServerModel importRequest) {
```
jladieu, line 17: Consider taking in a URI as an argument here instead, since none of the rest of the fields of `importRequest` are used by this method.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobDataPlaneE2ETest.java
```diff
@@ -48,7 +48,7 @@
  */
 @ActiveProfiles(profiles = {"mock-sam", "noop-scheduler-dao", "data-plane"})
 @DirtiesContext
-@SpringBootTest
+@SpringBootTest(properties = {"twds.data-import.allowed-schemes=file"})
```
jladieu, line 51: Might be worth an inline comment explaining that we're allowing `file` here because the test fixture data includes some `file` URLs, otherwise I'd wonder why this wasn't permitted by default.

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceDataPlaneTest.java
```diff
@@ -48,7 +48,8 @@ class ImportServiceDataPlaneTest {
   @MockBean SamAuthorizationDaoFactory samAuthorizationDaoFactory;
   private final SamAuthorizationDao samAuthorizationDao = spy(MockSamAuthorizationDao.allowAll());
 
-  private final URI importUri = URI.create("http://does/not/matter");
+  private final URI importUri =
+      URI.create("https://teststorageaccount.blob.core.windows.net/testcontainer/file");
```
jladieu, line 52: Just curious, how will this fail if you changed it back to the old URI?  Is the error self-explanatory enough that it'd be easy to figure out why?  If so, no change needed, but if it's opaque it might be worth including a comment explaining why this particular host is needed.

jladieu (approved):
Small suggestions at this point.  All optional. Great work!

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceDataPlaneTest.java
nawatts: The original would fail because imports from HTTP URLs aren't allowed by default. If it were changed to HTTPS, it would still fail the host validation because by default, only various cloud storage buckets are allowed.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobDataPlaneE2ETest.java
nawatts: Also, still thinking about adding a test profile with allowed import configuration after #650 merges.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportSourceValidator.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/648/commits/66f900b06186eb68c5c68c49bf54ad6e02e626de.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportSourceValidator.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/648/commits/cf598f3539aa0ed5a07f1e72e0086a9366f6f0cd.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobDataPlaneE2ETest.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/648/commits/3e187ef80f9a79f7e188cc2744490276bb990b15.

jladieu (approved):

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceTest.java
nawatts: I did try this, but it looks like the MockBean is still needed since some tests assert calls to schedulerDao.

PR #650, opened 2024-03-14
AJ-1681 - Clean up WDS env variables

yuliadub:
A profile config by default overwrites the contents of what exists in application.yaml if  any of the values collide. This PR removes setting certain variable urls via env vars from the main yaml, and moves that functionality to only be possible only when running a local profile. For all other profiles, hard coded values from the profile config should be used instead. 

Reminder:


jladieu (approved):

calypsomatic (approved):

service/src/main/resources/application-local.yml
```diff
@@ -1,7 +1,7 @@
 samurl: ${SAM_URL:https://sam.dsde-dev.broadinstitute.org/}
-datarepourl: https://jade.datarepo-dev.broadinstitute.org/
+datarepourl: ${DATA_REPO_URL:https://jade.datarepo-dev.broadinstitute.org/}
 workspacemanagerurl: ${WORKSPACE_MANAGER_URL:https://workspace.dsde-dev.broadinstitute.org/}
-leoUrl: https://leonardo.dsde-dev.broadinstitute.org/
+leoUrl: ${LEONARDO_URL:https://leonardo.dsde-dev.broadinstitute.org/}
```
ashanhol, line 4: I'm not familiar with this syntax- is this defining $LEONARDO_URL if it's not defined?

yuliadub: it will grab a value from $LEONARDO_URL env variable if it is not empty - but if it is, it will instead make the value "https://leonardo.dsde-dev.broadinstitute.org/"

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
```diff
@@ -89,24 +83,12 @@ Map<String, String> getTags() {
...
-    if (samUrl == null) {
-      return DEFAULT_ENV;
-    }
-    Matcher matcher = SAM_ENV_PATTERN.matcher(samUrl);
-    boolean found = matcher.find();
-    if (found) {
-      return matcher.group(1);
-    } else {
-      return DEFAULT_ENV;
+  boolean determineIfEnvIsMonitored(String env) {
```
nawatts, line 86: Super nit, but a few alternative suggestions for function names: 

```suggestion
  boolean isEnvMonitored(String env) {
```
or
```suggestion
  boolean shouldMonitorEnv(String env) {
```

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
```diff
@@ -40,26 +36,24 @@ public class SentryInitializer {
...
 
   @Bean
-  public SmartInitializingSingleton initialize() {
+  public SmartInitializingSingleton initialize(Environment environment) {
     String env;
     if (StringUtils.isNotBlank(terraEnv)) {
       env = terraEnv;
     } else {
-      env = urlToEnv(samurl);
+      env = String.join(",", environment.getActiveProfiles());
```
nawatts, line 49: It seems like this allows WDS running in local environments, etc. to be configured to use Sentry. Is that actually what we want? It seems like we'd only want Sentry in live environments where `TERRA_ENV` is set?

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
nawatts: Alternatively, since the `env` passed is stored in a property of this class, does this method need to take an argument?

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
yuliadub: I followed the behavior that was there before, but I am also fine with just doing it when TERRA_ENV is set.

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
yuliadub: env only exists in the context of the constructor at this time.

ashanhol (approved):

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
```diff
@@ -40,26 +36,24 @@ public class SentryInitializer {
...
     }
 
     return () ->
         Sentry.init(
             options -> {
               options.setEnvironment(env);
```
davidangb, line 55: in cases where WDS is started with multiple profiles - which should be always, e.g. "prod,data-plane" - this will set the Sentry "environment" to the full comma-delimited string. This is not ideal - it will will make filtering/classifying issues in Sentry more difficult. I'd much prefer to leave the Sentry "environment" as one of dev|staging|prod, and add any other relevant profiles to their own tags.

My suggestion: use the `sentryEnvironments` List to return a single dev|staging|prod value based on $TERRA_ENV and/or the active profiles, and use that single value to set the Sentry environment. If there are any other profiles you want to include as Sentry tags, add them explicitly as a new tag in the `getTags()` method. Note that we already set a tag of "deploymentMode" for control-plane|data-plane.

It might help to visualize how these are set in the Sentry UI:
Screenshot 3-19-2024 at 09 08 AM (1)

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
davidangb: I believe $TERRA_ENV is only set in the control plane, so the else clause that inspects `environment.getActiveProfiles()` is useful for the data plane.

davidangb, line 49: In the else clause that calls `environment.getActiveProfiles()`, this turns the String array into a comma-delimited string, and later in determineIfEnvIsMonitored you're checking for substrings of that comma-delimited string. This breaks if a profile value itself contains a comma (which we don't have now, admittedly). I'd prefer to leave the data as an array, e.g.

```java
    String[] envs;
    if (StringUtils.isNotBlank(terraEnv)) {
      envs = new String[]{terraEnv};
    } else {
      envs = environment.getActiveProfiles();
    }
```
and
```java
  boolean determineIfEnvIsMonitored(String[] envs) {
    for (var env : envs) {
      if (sentryEnvironments.contains(env)) {
        return true;
      }
    }
    return false;
  }
```

You'd have to modify the `options.setEnvironment(…)` call too, but I think that's necessary - see my other comment

davidangb (changes requested):
see comments inline, I am worried this changes how we set tags in Sentry.

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
yuliadub: getActiveProfiles() actually reads the comma and returns it as a string[] without the comma from my testing, but I totally understand the tagging of env in Sentry being a comma separated list being an issue. Let me know if this revised way looks better

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
```diff
@@ -40,26 +36,24 @@ public class SentryInitializer {
...
   // actions
-  private static final List<String> environments = List.of("prod", "staging", "dev");
+  private static final List<String> sentryEnvironments = List.of("prod", "staging", "dev");
 
   @Bean
-  public SmartInitializingSingleton initialize() {
+  public SmartInitializingSingleton initialize(Environment environment) {
     String env;
     if (StringUtils.isNotBlank(terraEnv)) {
       env = terraEnv;
```
yuliadub, line 47: is it okay to trust TerraEnv here or would we still need to check if this is set to prod/staging/dev?

davidangb: We should still check if this is set to prod/staging/dev. `terraEnv` here can/will also reflect BEEs, but we don't want BEEs to send to Sentry to keep noise down.

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
```diff
@@ -89,24 +83,12 @@ Map<String, String> getTags() {
...
-    Matcher matcher = SAM_ENV_PATTERN.matcher(samUrl);
-    boolean found = matcher.find();
-    if (found) {
-      return matcher.group(1);
-    } else {
-      return DEFAULT_ENV;
+  String getSentryEnvironment(String[] profiles) {
+    for (var profile : profiles) {
+      if (sentryEnvironments.contains(profile)) {
+        return profile;
```
jladieu, line 89:
```suggestion
        return Optional.of(profile);
```

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
```diff
@@ -89,24 +83,12 @@ Map<String, String> getTags() {
...
+      }
     }
+    return "";
```
jladieu, line 92:
```suggestion
    return Optional.empty();
```

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
```diff
@@ -89,24 +83,12 @@ Map<String, String> getTags() {
...
-    if (samUrl == null) {
-      return DEFAULT_ENV;
-    }
-    Matcher matcher = SAM_ENV_PATTERN.matcher(samUrl);
-    boolean found = matcher.find();
-    if (found) {
-      return matcher.group(1);
-    } else {
-      return DEFAULT_ENV;
+  String getSentryEnvironment(String[] profiles) {
```
jladieu, line 86:
```suggestion
  Optional<String> getSentryEnvironment(String[] profiles) {
```

Would you consider returning an `Optional<String>` here?  I'll add some suggestions here and where it's called for how it might look.

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
```diff
@@ -40,26 +36,24 @@ public class SentryInitializer {
...
-      env = urlToEnv(samurl);
+      env = getSentryEnvironment(environment.getActiveProfiles());
     }
 
     return () ->
         Sentry.init(
             options -> {
               options.setEnvironment(env);
-              options.setDsn(environments.contains(env) ? dsn : "");
+              options.setDsn(env != "" ? dsn : "");
```
jladieu, line 56:
```suggestion
              options.setDsn(getSentryEnvironment(environment.getActiveProfiles()).orElse("");
```

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
```diff
@@ -40,26 +36,24 @@ public class SentryInitializer {
...
   // Environments we want to monitor on sentry - don't send errors from local, bees, or Github
   // actions
-  private static final List<String> environments = List.of("prod", "staging", "dev");
+  private static final List<String> sentryEnvironments = List.of("prod", "staging", "dev");
 
   @Bean
-  public SmartInitializingSingleton initialize() {
+  public SmartInitializingSingleton initialize(Environment environment) {
     String env;
     if (StringUtils.isNotBlank(terraEnv)) {
```
jladieu, line 46: If you like my suggestion to use an `Optional<String>` for `getSentryEnvironment` and want to incorporate it, then I think it might make sense to hide the `terraEnv` check shortcut in that method and drop this if/else here in this method entirely.

jladieu (approved):
I had an idea of how to tidy up some of the logic around sentryEnvironment using an `Optional`, but...the suggestion is, itself...entirely _optional_ if you just want to ship this :)

service/src/main/java/org/databiosphere/workspacedataservice/SentryInitializer.java
```diff
@@ -40,26 +37,21 @@ public class SentryInitializer {
...
-    }
+  public SmartInitializingSingleton initialize(Environment environment) {
+    var env = getSentryEnvironment(environment.getActiveProfiles()).orElse(DEFAULT_ENV);
 
     return () ->
         Sentry.init(
             options -> {
               options.setEnvironment(env);
-              options.setDsn(environments.contains(env) ? dsn : "");
+              options.setDsn(!env.equals(DEFAULT_ENV) ? dsn : "");
```
nawatts, line 54: Nit: this could be simplified by swapping the branches.

```suggestion
              options.setDsn(env.equals(DEFAULT_ENV) ? "" : dsn);
```

service/src/test/java/org/databiosphere/workspacedataservice/SentryInitializerTest.java
```diff
@@ -9,26 +9,26 @@
 
 class SentryInitializerTest {
 
-  private static Stream<Arguments> provideSamUrls() {
+  private static Stream<Arguments> provideProfiles() {
     /* Arguments are pairs of input Sam URL and expected environment parsed from that URL
```
nawatts, line 13: This comment is out of date.

nawatts (approved):

davidangb (approved):
yay thank you!

davidangb (approved):

PR #652, opened 2024-03-15
AJ-1695: (take 1) add public. qualifier to file/relation types

davidangb:
another draft, looking for early feedback. #653 is probably better than this one.


davidangb:
closing in favor of #653

PR #655, opened 2024-03-15
[AJ-1588] Configure allowed import sources

nawatts:
Configure allowed import sources to match import-service.

https://github.com/broadinstitute/import-service/blob/a818658d0850818804235c5dc7f239c0de95c8cf/app.yaml.ctmpl#L69

### DIFF
--- a/service/src/main/resources/application-prod.yml
+++ b/service/src/main/resources/application-prod.yml
@@ -2,3 +2,11 @@ samurl: https://sam.dsde-prod.broadinstitute.org/
 datarepourl: https://data.terra.bio/
 workspacemanagerurl: https://workspace.dsde-prod.broadinstitute.org/
 leoUrl: https://leonardo.dsde-prod.broadinstitute.org/
+
+twds:
+  data-import:
+    allowed-hosts:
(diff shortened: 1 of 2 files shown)


service/src/main/resources/application-prod.yml
```diff
@@ -2,3 +2,11 @@ samurl: https://sam.dsde-prod.broadinstitute.org/
...
 workspacemanagerurl: https://workspace.dsde-prod.broadinstitute.org/
 leoUrl: https://leonardo.dsde-prod.broadinstitute.org/
+
+twds:
+  data-import:
+    allowed-hosts:
+      - service\.prod\.anvil\.gi\.ucsc\.edu
+      - .*\.singlecell\.gi\.ucsc\.edu,
+      - .*gen3\.biodatacatalyst\.nhlbi\.nih\.gov
+      - .*service\.azul\.data\.humancellatlas\.org
```
nawatts, line 12: I don't know if we actually need the leading `.*` on these two, but this is matching what's in import-service.

calypsomatic (approved):

jladieu (approved):

PR #659, opened 2024-03-18
[AJ-1709] Configure project ID for PubSub

nawatts:
In office hours today, we discovered that the GCP PubSub library has its own project ID configuration and WDS fails to start if that configuration is not provided.

```
org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'pfbQuartzJob' defined in file [/Users/nwatts/workspace/terra-workspace-data-service/service/build/classes/java/main/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJob.class]: Unsatisfied dependency expressed through constructor parameter 4: Error creating bean with name 'rawlsRecordSinkFactory' defined in file [/Users/nwatts/workspace/terra-workspace-data-service/service/build/classes/java/main/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkFactory.class]: Unsatisfied dependency expressed through constructor parameter 2: Error creating bean with name 'applicationDefaultCredentialsPubSub' defined in class path resource [org/databiosphere/workspacedataservice/pubsub/PubSubConfig.class]: Unsatisfied dependency expressed through method 'applicationDefaultCredentialsPubSub' parameter 0: Error creating bean with name 'pubSubTemplate' defined in class path resource [com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.class]: Unsatisfied dependency expressed through method 'pubSubTemplate' parameter 0: Error creating bean with name 'pubSubPublisherTemplate' defined in class path resource [com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.class]: Unsatisfied dependency expressed through method 'pubSubPublisherTemplate' parameter 0: Error creating bean with name 'defaultPublisherFactory' defined in class path resource [com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.class]: Failed to instantiate [com.google.cloud.spring.pubsub.support.PublisherFactory]: Factory method 'defaultPublisherFactory' threw exception with message: The project ID can't be null or empty.
```

Our development instructions didn't explicitly cover this, but it worked for some developers because they had a default project set via `gcloud config`. Similarly, this works in the dev environment because it can get a project ID from the environment.

Setting `spring.cloud.gcp.pubsub.project-id` based on the `SERVICE_GOOGLE_PROJECT` environment variable (in https://github.com/DataBiosphere/terra-workspace-data-service/pull/659/commits/728a54ecfe80d52012a3df2c7491e01d445f55a3) made things work, but then we have the same project ID configured in two different places (one for Spring GCP config and one for our own TWDS data import config).

This changes everything to use Spring Cloud GCP's `GcpProjectIdProvider`. Spring Cloud GCP looks for configuration in several places, but this also sets `spring.cloud.gcp.pubsub.project-id` based on the `SERVICE_GOOGLE_PROJECT` environment variable so that our existing environment variable still works.
https://googlecloudplatform.github.io/spring-cloud-gcp/reference/html/index.html#project-id

We could also simplify by removing the `project-id` config and using `GOOGLE_CLOUD_PROJECT` instead of `SERVICE_GOOGLE_PROJECT` in development.

### DIFF
--- a/service/src/main/resources/application-control-plane.yml
+++ b/service/src/main/resources/application-control-plane.yml
@@ -10,15 +10,14 @@ twds:
   data-import:
     batch-write-record-sink: "rawls"
     succeed-on-completion: false
-    project-id: ${SERVICE_GOOGLE_PROJECT}
     rawls-bucket-name: ${SERVICE_GOOGLE_BUCKET}
 
 
 spring:
   cloud:
     gcp:
+      project-id: ${SERVICE_GOOGLE_PROJECT}
       pubsub:
-        project-id: ${SERVICE_GOOGLE_PROJECT}
         topic: ${RAWLS_NOTIFY_TOPIC}
         enabled: true
 
--- a/service/src/main/resources/application-control-plane.yml
+++ b/service/src/main/resources/application-control-plane.yml
@@ -18,6 +18,7 @@ spring:
   cloud:
     gcp:
       pubsub:
+        project-id: ${SERVICE_GOOGLE_PROJECT}
         topic: ${RAWLS_NOTIFY_TOPIC}
         enabled: true
 
--- a/service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
@@ -20,7 +20,6 @@ public class DataImportProperties {
           Pattern.compile(".*\\.s3\\.amazonaws\\.com") // virtual host style legacy global endpoint
           );
   private RecordSinkMode batchWriteRecordSink;
-  private String projectId;
   private String rawlsBucketName;
   private boolean succeedOnCompletion;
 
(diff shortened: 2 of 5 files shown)


davidangb (approved):

service/src/main/resources/application-control-plane.yml
```diff
@@ -18,6 +18,7 @@ spring:
   cloud:
     gcp:
       pubsub:
+        project-id: ${SERVICE_GOOGLE_PROJECT}
```
calypsomatic, line 21: Bit of a nit, but I didn't have this here because we were already setting it under `twds.project-id`.  If we have to have it twice, maybe for [pubsub config](https://github.com/DataBiosphere/terra-workspace-data-service/blob/165212e64e5a5a61c4cc2a206e703463b427c077/service/src/main/java/org/databiosphere/workspacedataservice/pubsub/PubSubConfig.java#L21) use the relevant reference?  Should be the same in the end, I guess unless for some reason we use different projects for pubsub v blob storage

ashanhol (approved):
thank you :)

service/src/main/resources/application-control-plane.yml
nawatts: https://github.com/DataBiosphere/terra-workspace-data-service/pull/659/commits/c47c3987be05e9bcaa77680f958cc67cfc85850c updates everything to use the same project ID configuration.

davidangb (approved):
thank you!

PR #662, opened 2024-03-20
[AJ-1687] Add handler for PubSub messages from Rawls

nawatts:

An import isn't complete until Rawls finishes importing the entities JSON file written by CWDS. Currently, when it is done, it notifies import service of the final job status via PubSub ([code](https://github.com/broadinstitute/rawls/blob/44f717ce2b7d076aebf577b64377c7c56e4e11f3/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala#L341-L344)). We want a similar set up for CWDS. This adds a handler for those PubSub messages.

See https://cloud.google.com/pubsub/docs/push#receive_push for the PubSub request format.

Adapted from import service: https://github.com/broadinstitute/import-service/blob/a818658d0850818804235c5dc7f239c0de95c8cf/app/status.py#L55-L106

Still to do here is authentication for the PubSub requests. PubSub will be configured to authenticate requests with CWDS' service account, but this PR does not include verification of that token. I wanted to see if the PubSub wiring works before dealing with that.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/PubSubController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/PubSubController.java
@@ -1,0 +1,48 @@
+package org.databiosphere.workspacedataservice.controller;
+
+import org.databiosphere.workspacedataservice.annotations.DeploymentMode.ControlPlane;
+import org.databiosphere.workspacedataservice.pubsub.JobStatusUpdate;
+import org.databiosphere.workspacedataservice.pubsub.PubSubMessage;
+import org.databiosphere.workspacedataservice.pubsub.PubSubRequest;
+import org.databiosphere.workspacedataservice.service.JobService;
+import org.databiosphere.workspacedataservice.service.model.exception.ValidationException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RestController;
+
+@ControlPlane
+@RestController
+public class PubSubController {
+  private static final Logger LOGGER = LoggerFactory.getLogger(PubSubController.class);
+  private final JobService jobService;
+
+  public PubSubController(JobService jobService) {
+    this.jobService = jobService;
+  }
+
+  @PostMapping("/pubsub/import-status")
+  public ResponseEntity<String> receiveImportNotification(@RequestBody PubSubRequest request) {
+    PubSubMessage message = request.message();
+    LOGGER.info(
+        "Received PubSub message: {}, published {}", message.messageId(), message.publishTime());
+    try {
+      JobStatusUpdate update = JobStatusUpdate.createFromPubSubMessage(request.message());
+      LOGGER.info(
+          "Received status update for job {}: {} -> {}",
+          update.jobId(),
+          update.currentStatus(),
+          update.currentStatus());
+      jobService.processStatusUpdate(update);
+      return new ResponseEntity<>(HttpStatus.OK);
+    } catch (ValidationException e) {
+      // Return a successful status for invalid updates to prevent PubSub from retrying the request.
+      // https://cloud.google.com/pubsub/docs/push#receive_push
+      LOGGER.error("Error processing status update: {}", e.getMessage());
+      return new ResponseEntity<>(HttpStatus.ACCEPTED);
+    }
+  }
+}
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/PubSubController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/PubSubController.java
@@ -35,7 +35,7 @@ public class PubSubController {
           "Received status update for job {}: {} -> {}",
           update.jobId(),
           update.currentStatus(),
-          update.currentStatus());
+          update.newStatus());
       jobService.processStatusUpdate(update);
       return new ResponseEntity<>(HttpStatus.OK);
     } catch (ValidationException e) {
--- a/service/src/main/java/org/databiosphere/workspacedataservice/pubsub/JobStatusUpdate.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/pubsub/JobStatusUpdate.java
@@ -1,0 +1,29 @@
+package org.databiosphere.workspacedataservice.pubsub;
+
+import java.util.Map;
+import java.util.UUID;
+import javax.annotation.Nullable;
+import org.databiosphere.workspacedataservice.generated.GenericJobServerModel.StatusEnum;
+
+public record JobStatusUpdate(
+    UUID jobId, StatusEnum currentStatus, StatusEnum newStatus, @Nullable String errorMessage) {
+  public static JobStatusUpdate createFromPubSubMessage(PubSubMessage message) {
+    Map<String, String> attributes = message.attributes();
+    UUID jobId = UUID.fromString(attributes.get("import_id"));
+    StatusEnum newStatus = rawlsStatusToJobStatus(attributes.get("new_status"));
+    StatusEnum currentStatus = rawlsStatusToJobStatus(attributes.get("current_status"));
+    String errorMessage = attributes.get("error_message");
+    return new JobStatusUpdate(jobId, currentStatus, newStatus, errorMessage);
+  }
+
+  private static StatusEnum rawlsStatusToJobStatus(String rawlsStatus) {
+    return switch (rawlsStatus) {
+      case "ReadyForUpsert" -> StatusEnum.RUNNING;
+      case "Upserting" -> StatusEnum.RUNNING;
+      case "Done" -> StatusEnum.SUCCEEDED;
+      case "Error" -> StatusEnum.ERROR;
+      default -> throw new RuntimeException(
+          "Unknown Rawls import status: %s".formatted(rawlsStatus));
+    };
+  }
+}
--- a/service/src/main/java/org/databiosphere/workspacedataservice/pubsub/JobStatusUpdate.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/pubsub/JobStatusUpdate.java
@@ -7,6 +7,10 @@ import org.databiosphere.workspacedataservice.generated.GenericJobServerModel.St
 
 public record JobStatusUpdate(
     UUID jobId, StatusEnum currentStatus, StatusEnum newStatus, @Nullable String errorMessage) {
+  public JobStatusUpdate(UUID jobId, StatusEnum currentStatus, StatusEnum newStatus) {
+    this(jobId, currentStatus, newStatus, null);
+  }
+
   public static JobStatusUpdate createFromPubSubMessage(PubSubMessage message) {
     Map<String, String> attributes = message.attributes();
     UUID jobId = UUID.fromString(attributes.get("import_id"));
--- a/service/src/main/java/org/databiosphere/workspacedataservice/pubsub/JobStatusUpdate.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/pubsub/JobStatusUpdate.java
@@ -22,7 +22,7 @@ public record JobStatusUpdate(
       case "Upserting" -> StatusEnum.RUNNING;
       case "Done" -> StatusEnum.SUCCEEDED;
       case "Error" -> StatusEnum.ERROR;
-      default -> throw new RuntimeException(
+      default -> throw new IllegalArgumentException(
           "Unknown Rawls import status: %s".formatted(rawlsStatus));
     };
   }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/pubsub/PubSubMessage.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/pubsub/PubSubMessage.java
@@ -1,0 +1,10 @@
+package org.databiosphere.workspacedataservice.pubsub;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Map;
+
+public record PubSubMessage(
+    @JsonProperty("attributes") Map<String, String> attributes,
+    @JsonProperty("data") String data,
+    @JsonProperty("messageId") String messageId,
+    @JsonProperty("publishTime") String publishTime) {}
--- a/service/src/main/java/org/databiosphere/workspacedataservice/pubsub/PubSubMessage.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/pubsub/PubSubMessage.java
@@ -1,10 +1,6 @@
 package org.databiosphere.workspacedataservice.pubsub;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
 import java.util.Map;
 
 public record PubSubMessage(
-    @JsonProperty("attributes") Map<String, String> attributes,
-    @JsonProperty("data") String data,
-    @JsonProperty("messageId") String messageId,
-    @JsonProperty("publishTime") String publishTime) {}
+    Map<String, String> attributes, String data, String messageId, String publishTime) {}
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
@@ -1,18 +1,28 @@
 package org.databiosphere.workspacedataservice.service;
 
+import static org.apache.commons.lang3.ObjectUtils.firstNonNull;
+
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.UUID;
 import org.databiosphere.workspacedataservice.dao.JobDao;
 import org.databiosphere.workspacedataservice.generated.GenericJobServerModel;
+import org.databiosphere.workspacedataservice.pubsub.JobStatusUpdate;
 import org.databiosphere.workspacedataservice.service.model.exception.AuthenticationMaskableException;
 import org.databiosphere.workspacedataservice.service.model.exception.MissingObjectException;
+import org.databiosphere.workspacedataservice.service.model.exception.ValidationException;
 import org.databiosphere.workspacedataservice.shared.model.CollectionId;
 import org.springframework.dao.EmptyResultDataAccessException;
 import org.springframework.stereotype.Service;
 
 @Service
 public class JobService {
+  private static final Set<GenericJobServerModel.StatusEnum> terminalJobStatuses =
+      Set.of(
+          GenericJobServerModel.StatusEnum.SUCCEEDED,
+          GenericJobServerModel.StatusEnum.ERROR,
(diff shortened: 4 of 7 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/controller/PubSubControllerMockMvcTest.java
```diff
@@ -0,0 +1,117 @@
...
+class PubSubControllerMockMvcTest extends MockMvcTestBase {
+  @MockBean private JobDao jobDao;
+  @MockBean private CollectionDao collectionDao;
+
+  // Since we're running with both control-plane and data-plane profiles simultaneously, Spring
+  // does not know what to do with beans which are satisfied by two different mutually exclusive
+  // implementations. In this @TestConfiguration, we explicitly specify @Primary beans for any
+  // conflicts.
+  @TestConfiguration
+  static class SpecifyConflictingBeans {
```
nawatts, line 54: Copied from AnnotatedApisTest.

service/src/main/java/org/databiosphere/workspacedataservice/controller/PubSubController.java
```diff
@@ -0,0 +1,48 @@
...
+    PubSubMessage message = request.message();
+    LOGGER.info(
+        "Received PubSub message: {}, published {}", message.messageId(), message.publishTime());
+    try {
+      JobStatusUpdate update = JobStatusUpdate.createFromPubSubMessage(request.message());
+      LOGGER.info(
+          "Received status update for job {}: {} -> {}",
+          update.jobId(),
+          update.currentStatus(),
+          update.currentStatus());
```
davidangb, line 38: should this be `update.newStatus()`?

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/JobStatusUpdate.java
```diff
@@ -0,0 +1,29 @@
...
+  private static StatusEnum rawlsStatusToJobStatus(String rawlsStatus) {
+    return switch (rawlsStatus) {
+      case "ReadyForUpsert" -> StatusEnum.RUNNING;
+      case "Upserting" -> StatusEnum.RUNNING;
+      case "Done" -> StatusEnum.SUCCEEDED;
+      case "Error" -> StatusEnum.ERROR;
+      default -> throw new IllegalArgumentException(
+          "Unknown Rawls import status: %s".formatted(rawlsStatus));
+    };
+  }
```
davidangb, line 28: thinking out loud: since Rawls knows how to differentiate between "ReadyForUpsert" and "Upserting", I'd love to someday save that info to WDS's job - we could possibly use the `JobResult` for this. No changes requested for this PR, just babbling.

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/PubSubMessage.java
```diff
@@ -0,0 +1,10 @@
+package org.databiosphere.workspacedataservice.pubsub;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Map;
+
+public record PubSubMessage(
+    @JsonProperty("attributes") Map<String, String> attributes,
+    @JsonProperty("data") String data,
+    @JsonProperty("messageId") String messageId,
+    @JsonProperty("publishTime") String publishTime) {}
```
davidangb, line 10: are the `@JsonProperty` annotations necessary here? I thought Jackson knew how to find these without the annotation, as long as the json property and the java class property have the same name. That said, the annotations don't _hurt_ anything either.

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
```diff
@@ -1,18 +1,28 @@
...
 import org.springframework.dao.EmptyResultDataAccessException;
 import org.springframework.stereotype.Service;
 
 @Service
 public class JobService {
+  private static final Set<GenericJobServerModel.StatusEnum> terminalJobStatuses =
+      Set.of(
+          GenericJobServerModel.StatusEnum.SUCCEEDED,
+          GenericJobServerModel.StatusEnum.ERROR,
+          GenericJobServerModel.StatusEnum.CANCELLED);
```
davidangb, line 25: predicting what Josh will say: add `import static org.databiosphere.workspacedataservice.generated.GenericJobServerModel.StatusEnum;` so you can remove the `GenericJobServerModel.` from each of these

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
```diff
@@ -44,4 +54,39 @@ public List<GenericJobServerModel> getJobsForCollection(
...
+      }
+
+      // Once a job has reached a terminal state, do not update its status again.
+      if (terminalJobStatuses.contains(currentStatus)) {
+        throw new ValidationException(
+            "Unable to update terminal status for job %s".formatted(jobId));
+      }
+
+      if (newStatus.equals(GenericJobServerModel.StatusEnum.ERROR)) {
+        jobDao.fail(jobId, firstNonNull(update.errorMessage(), "Unknown error"));
```
davidangb, line 83: very minor nit: elsewhere in the codebase we're using `Objects.requireNonNullElse` (which is built into Java) instead of `ObjectUtils.firstNonNull` from Apache

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
```diff
@@ -44,4 +54,39 @@ public List<GenericJobServerModel> getJobsForCollection(
...
+      } else {
+        jobDao.updateStatus(jobId, newStatus);
+      }
```
davidangb, line 86: for your consideration - if possible, use the higher-level `.fail()`, `.succeeded()`, `.running()` methods of `JobDao` instead of the low-level `.updateStatus()`. So maybe you could do something like:
```java
      switch (newStatus) {
        case ERROR -> jobDao.fail(jobId, firstNonNull(update.errorMessage(), "Unknown error"));
        case SUCCEEDED -> jobDao.succeeded(jobId);
        case RUNNING -> jobDao.running(jobId); // do we actually need to handle this case?
        default ->
            throw new ValidationException(
                "Unexpected status from Rawls for job %s".formatted(jobId));
      }
```
but this is really just style, the code in this PR is fine; I wouldn't block the PR on this suggestion

davidangb (commented):
some minor comments inline, nothing big, looking good!

service/src/main/java/org/databiosphere/workspacedataservice/controller/PubSubController.java
nawatts: Fixed in https://github.com/DataBiosphere/terra-workspace-data-service/pull/662/commits/a496a9b04ee46c1b2e0c3ea2871a28221171e2ef.

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/PubSubMessage.java
nawatts: Removed in https://github.com/DataBiosphere/terra-workspace-data-service/pull/662/commits/c25bd90eb28f362852fc365f9abc62ad6118ffbc.

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/662/commits/e4a4c308b1bd8631718e2f36a9c3d64042b5d803.

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
nawatts: Today I learned. Switched in https://github.com/DataBiosphere/terra-workspace-data-service/pull/662/commits/f51b7c7cb9718f4d9da6a733f99b8c63d937db30.

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/662/commits/223701a7e7acf4d6b054369c7af67dfe6850c932.

We shouldn't need to handle the running case since it should either be caught by the no-op for unchanged status or error for updating an already terminal job guards above.

davidangb (approved):
thank you! I haven't tested at all (and I realize testing is a pain without a real pubsub subscription and rawls); trusting you to validate after merging

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/JobStatusUpdate.java
```diff
@@ -0,0 +1,29 @@
+package org.databiosphere.workspacedataservice.pubsub;
+
+import java.util.Map;
+import java.util.UUID;
+import javax.annotation.Nullable;
+import org.databiosphere.workspacedataservice.generated.GenericJobServerModel.StatusEnum;
+
+public record JobStatusUpdate(
+    UUID jobId, StatusEnum currentStatus, StatusEnum newStatus, @Nullable String errorMessage) {
+  public static JobStatusUpdate createFromPubSubMessage(PubSubMessage message) {
```
jladieu, line 10: nit: not a big deal, but since you only use `attributes()` on the provided message, consider just taking a `Map<String, String>` parameter.  This might be easier to construct in tests, for example, rather than an entire `PubSubMessage` which I could imagine might have extra baggage.

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
```diff
@@ -44,4 +52,41 @@ public List<GenericJobServerModel> getJobsForCollection(
...
+      UUID jobId = update.jobId();
+      GenericJobServerModel job = getJob(jobId);
+      StatusEnum currentStatus = job.getStatus();
+      StatusEnum newStatus = update.newStatus();
+
+      // Ignore messages that don't change the job's status.
+      // Rawls and import service have more granular statuses than CWDS, so the initial update
+      // will be from "ReadyToUpsert" to "Upserting", which both translate to "RUNNING" in CWDS.
+      if (currentStatus.equals(newStatus)) {
+        return;
```
jladieu, line 71: Consider adding a log message when a message comes in that is dropped for this reason?

service/src/test/java/org/databiosphere/workspacedataservice/controller/PubSubControllerMockMvcTest.java
jladieu: Are you intentionally running with both `ControlPlane` and `DataPlane`, or just incidentally doing so because of extending `MockMvcTestBase`?  If you can run with just `ControlPlane` then you wouldn't have to do this extra mock specification.

If you want, you can pass `inheritProfiles=false` to `@ActiveProfile` to bypass the `data-plane` profile from the superclass.

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
```diff
@@ -44,4 +52,41 @@ public List<GenericJobServerModel> getJobsForCollection(
...
+      switch (newStatus) {
+        case SUCCEEDED -> jobDao.succeeded(jobId);
+        case ERROR -> jobDao.fail(
+            jobId, requireNonNullElse(update.errorMessage(), "Unknown error"));
+        default -> throw new ValidationException(
+            "Unexpected status from Rawls for job %s: %s".formatted(jobId, newStatus));
+      }
+    } catch (MissingObjectException e) {
+      // Via PubSub, CWDS will receive status updates for both CWDS and import service jobs.
+      // If the job is not found in the database (because it's an import service job), ignore it.
```
ashanhol, line 89: Is there a case where we could end up with a missingobjectexception for cwds jobs? maybe a logging message would be good just to cover us in case there's an edge case

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/JobStatusUpdate.java
jladieu, line 10: Consider adding a constructor overload that only takes 3 args and just chains to the default constructor with `null` for the 4th arg.  This would let you avoid having to explicitly pass in `null` in the tests where this is used, but also helps call attention to the fact that the 4th arg is optional in call sites.

jladieu (approved):
Approved with some optional suggestions.  Great job!

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
```diff
@@ -1,18 +1,26 @@
...
 import org.databiosphere.workspacedataservice.service.model.exception.AuthenticationMaskableException;
 import org.databiosphere.workspacedataservice.service.model.exception.MissingObjectException;
+import org.databiosphere.workspacedataservice.service.model.exception.ValidationException;
 import org.databiosphere.workspacedataservice.shared.model.CollectionId;
 import org.springframework.dao.EmptyResultDataAccessException;
 import org.springframework.stereotype.Service;
 
 @Service
 public class JobService {
+  private static final Set<StatusEnum> terminalJobStatuses =
```
ashanhol, line 22: I wonder if there's a better place for this? I remember during the discussion of the listjobs api we discussed having an "active" and "inactive" status collection

service/src/main/java/org/databiosphere/workspacedataservice/controller/PubSubController.java
```diff
@@ -0,0 +1,48 @@
...
+@RestController
+public class PubSubController {
+  private static final Logger LOGGER = LoggerFactory.getLogger(PubSubController.class);
+  private final JobService jobService;
+
+  public PubSubController(JobService jobService) {
+    this.jobService = jobService;
+  }
+
+  @PostMapping("/pubsub/import-status")
```
ashanhol, line 27: is this an api that's exposed automatically wtih pubsub? or was there a previous PR that defined this in our openapi spec?

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/JobStatusUpdate.java
nawatts: I did think about that, but creating from a PubSub message seemed a more intuitive concept. This way, the details of how the PubSub message is formatted doesn't affect the API.

service/src/main/java/org/databiosphere/workspacedataservice/controller/PubSubController.java
```diff
@@ -0,0 +1,48 @@
...
+      JobStatusUpdate update = JobStatusUpdate.createFromPubSubMessage(request.message());
+      LOGGER.info(
+          "Received status update for job {}: {} -> {}",
+          update.jobId(),
+          update.currentStatus(),
+          update.newStatus());
+      jobService.processStatusUpdate(update);
+      return new ResponseEntity<>(HttpStatus.OK);
+    } catch (ValidationException e) {
+      // Return a successful status for invalid updates to prevent PubSub from retrying the request.
```
ashanhol, line 42: what's an example of an invalid update?

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
nawatts: I don't _think_ so, since we're expecting to only get these requests from Rawls, not user requests.

service/src/test/java/org/databiosphere/workspacedataservice/controller/PubSubControllerMockMvcTest.java
nawatts: Ooh, so that's why that's happening. I didn't follow the breadcrumbs all the way to TestBase.

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
nawatts: I wanted to put it in the StatusEnum, but that class is generated.

service/src/main/java/org/databiosphere/workspacedataservice/controller/PubSubController.java
nawatts: An out of order PubSub message that would attempt to set a finished job back to running.

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
nawatts: I'll claim this is covered by the log in PubSubController. I'm inclined _not_ to log here, since this is expected to happen once per import when Rawls sends an update from "ReadyForUpsert" to "Upserting", both of which CWDS sees as "Running".

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
```diff
@@ -44,4 +52,41 @@ public List<GenericJobServerModel> getJobsForCollection(
...
+
+      switch (newStatus) {
+        case SUCCEEDED -> jobDao.succeeded(jobId);
+        case ERROR -> jobDao.fail(
+            jobId, requireNonNullElse(update.errorMessage(), "Unknown error"));
+        default -> throw new ValidationException(
+            "Unexpected status from Rawls for job %s: %s".formatted(jobId, newStatus));
+      }
+    } catch (MissingObjectException e) {
+      // Via PubSub, CWDS will receive status updates for both CWDS and import service jobs.
```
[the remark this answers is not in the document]
nawatts: I'm also inclined to _not_ log here, since it's an expected case as long as import service is around.

service/src/test/java/org/databiosphere/workspacedataservice/controller/PubSubControllerMockMvcTest.java
nawatts: Fixed in https://github.com/DataBiosphere/terra-workspace-data-service/pull/662/commits/8b3b6fd3f9378fd8acacca4cd1c115266427c644. Thanks for the `inheritProfiles=false` incantation.

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/JobStatusUpdate.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/662/commits/166f36561728861ff477ac7a0375dd174cb3e2ed.

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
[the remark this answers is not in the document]
nawatts: Added in https://github.com/DataBiosphere/terra-workspace-data-service/pull/662/commits/ca45426906bf8ce5e7f4dc98f993850bf4bcdea8.

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
nawatts: Added a log in https://github.com/DataBiosphere/terra-workspace-data-service/pull/662/commits/ca45426906bf8ce5e7f4dc98f993850bf4bcdea8.

PR #663, opened 2024-03-20
AJ-1559: Adjust referenced types for Rawls

jladieu:
Reminder:

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsModel.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsModel.java
@@ -1,11 +1,14 @@
 package org.databiosphere.workspacedataservice.recordsink;
 
+import static java.util.Objects.requireNonNull;
 import static org.databiosphere.workspacedataservice.recordsink.RawlsModel.Op.ADD_LIST_MEMBER;
 import static org.databiosphere.workspacedataservice.recordsink.RawlsModel.Op.ADD_UPDATE_ATTRIBUTE;
 import static org.databiosphere.workspacedataservice.recordsink.RawlsModel.Op.CREATE_ATTRIBUTE_ENTITY_REFERENCE_LIST;
 import static org.databiosphere.workspacedataservice.recordsink.RawlsModel.Op.CREATE_ATTRIBUTE_VALUE_LIST;
 import static org.databiosphere.workspacedataservice.recordsink.RawlsModel.Op.REMOVE_ATTRIBUTE;
 import static org.databiosphere.workspacedataservice.recordsink.RawlsModel.Op.REMOVE_LIST_MEMBER;
+import static org.databiosphere.workspacedataservice.service.RelationUtils.getRelationValue;
+import static org.databiosphere.workspacedataservice.service.RelationUtils.getTypeValue;
 
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import com.fasterxml.jackson.annotation.JsonSubTypes;
@@ -13,7 +16,18 @@ import com.fasterxml.jackson.annotation.JsonSubTypes.Type;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.annotation.JsonTypeInfo.Id;
 import com.fasterxml.jackson.annotation.JsonValue;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import com.fasterxml.jackson.databind.util.TokenBuffer;
+import java.io.IOException;
 import java.util.List;
+import org.databiosphere.workspacedataservice.shared.model.RecordType;
+import org.springframework.lang.Nullable;
 
 /**
  * Model class to implement a JSON payload compatible with Rawls <a
@@ -83,7 +97,9 @@ public class RawlsModel {
   }
 
   // Concrete subclasses of AttributeOperation
-  public record AddUpdateAttribute(String attributeName, Object addUpdateAttribute)
+  public record AddUpdateAttribute(
+      String attributeName,
+      @JsonDeserialize(using = RawlsValueDeserializer.class) Object addUpdateAttribute)
       implements AttributeOperation {
 
     @Override
@@ -92,7 +108,9 @@ public class RawlsModel {
     }
   }
 
-  public record AddListMember(String attributeListName, Object newMember)
+  public record AddListMember(
+      String attributeListName,
+      @JsonDeserialize(using = RawlsValueDeserializer.class) Object newMember)
       implements AttributeOperation {
 
     @Override
@@ -109,7 +127,9 @@ public class RawlsModel {
     }
   }
 
-  public record RemoveListMember(String attributeListName, Object removeMember)
+  public record RemoveListMember(
+      String attributeListName,
+      @JsonDeserialize(using = RawlsValueDeserializer.class) Object removeMember)
       implements AttributeOperation {
     @Override
     public Op op() {
@@ -129,6 +149,67 @@ public class RawlsModel {
     @Override
     public Op op() {
       return CREATE_ATTRIBUTE_ENTITY_REFERENCE_LIST;
+    }
+  }
+
+  /**
+   * A reference to a record in the workspace data service, used to serialize a reference to Rawls
+   * upsert JSON.
+   */
+  public record RecordReference(RecordType entityType, String entityName) {
+
+    /** Parse a wds reference string into a {@link RecordReference}. */
+    public static RecordReference fromReferenceString(String referenceString) {
+      return new RecordReference(getTypeValue(referenceString), getRelationValue(referenceString));
+    }
+  }
+
+  static class RawlsValueDeserializer extends StdDeserializer<Object> {
+    @Nullable private JsonDeserializer<Object> defaultDeserializer;
+
+    public RawlsValueDeserializer() {
+      super(Object.class);
+    }
+
+    @Override
+    public Object deserialize(JsonParser parser, DeserializationContext context)
+        throws IOException {
+      // Temporary tree model to inspect the JSON
+      JsonNode node = parser.getCodec().readTree(parser);
+      if (isReferenceObject(node)) {
+        return toRecordReference(node);
+      }
+      // Delegate to default deserializer
+      return defaultDeserialization(parser, context, node);
+    }
+
+    private boolean isReferenceObject(JsonNode node) {
+      return node.isObject() && node.has("entityType") && node.has("entityName");
+    }
+
+    private RecordReference toRecordReference(JsonNode node) {
+      return new RecordReference(
+          RecordType.valueOf(node.get("entityType").asText()), node.get("entityName").asText());
+    }
+
+    private Object defaultDeserialization(
+        JsonParser parser, DeserializationContext context, JsonNode node) throws IOException {
+      if (defaultDeserializer == null) {
+        defaultDeserializer = getDefaultDeserializer(context);
+      }
+      try (TokenBuffer tokenBuffer =
+          new TokenBuffer(parser.getCodec(), /* hasNativeIds= */ false)) {
+        tokenBuffer.writeTree(node);
+        JsonParser parserForDefaultDeserialization = tokenBuffer.asParser();
+        parserForDefaultDeserialization.nextToken(); // Prepare the parser to read the first token
+
+        return defaultDeserializer.deserialize(parserForDefaultDeserialization, context);
+      }
+    }
+
+    private static JsonDeserializer<Object> getDefaultDeserializer(DeserializationContext context)
+        throws JsonMappingException {
+      return requireNonNull(context.findRootValueDeserializer(context.constructType(Object.class)));
     }
   }
 }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
@@ -22,7 +22,9 @@ import org.databiosphere.workspacedataservice.recordsink.RawlsModel.AddUpdateAtt
 import org.databiosphere.workspacedataservice.recordsink.RawlsModel.AttributeOperation;
 import org.databiosphere.workspacedataservice.recordsink.RawlsModel.CreateAttributeValueList;
 import org.databiosphere.workspacedataservice.recordsink.RawlsModel.Entity;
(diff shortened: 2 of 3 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -128,10 +130,19 @@ private Stream<? extends AttributeOperation> toOperations(String name, Object at
     if (attributeValue instanceof List<?> values) {
       return Stream.concat(
           Stream.of(new RemoveAttribute(name), new CreateAttributeValueList(name)),
-          values.stream().map(value -> new AddListMember(name, value)));
+          values.stream()
+              .map(this::maybeCoerceRelation)
```
jladieu, line 134: This is the key functional change...right before going out the door to JSON, we take the terra-`wds:/<relation>/<relation-id>` string, parse it, and turn it into a `RecordReference` which serializes to the desired Rawls format.

I had an alternative idea here that I explored first, which was to attempt to store a `RecordReference` in `RecordAttributes`, and only turn it into a WDS string (or JSON object for Rawls) at the last minute, but that was a far bigger and extensive yakshave and refactoring.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsModel.java
```diff
@@ -131,4 +151,65 @@ public Op op() {
...
+   */
+  public record RecordReference(RecordType entityType, String entityName) {
+
+    /** Parse a wds reference string into a {@link RecordReference}. */
+    public static RecordReference fromReferenceString(String referenceString) {
+      return new RecordReference(getTypeValue(referenceString), getRelationValue(referenceString));
+    }
+  }
+
+  static class RawlsValueDeserializer extends StdDeserializer<Object> {
```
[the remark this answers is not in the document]
jladieu: I forgot David had suggested importing the rawls model classes might help cut down on some serialization stuff, I'm going to see if importing them can help me punt all this serialization junk.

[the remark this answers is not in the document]
jladieu: I took a spike trying with Scala (happy to push a branch to share) and wasn't super happy with the hacking I started having to do to get Scala interop working.

An alternative to all this deserialization complexity would be to eliminate the 'deserialization' concern from the test cases, as it's not something we actually need in prod, and replace it with just using a plain old Java Map.   I'm inclined to try that out...it'd let us replace the complexity here with simple brute force Map traversal in the test case code.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
davidangb: I don't like that in `PfbRecordConverter` and `ParquetRecordConverter` we identify relations, with (relatively) strongly typed type and id, and convert them to terra-wds:/... Strings, then over here in RawlsRecordSink we parse those strings and turn them back into something more strongly typed.

Detour: I am really wishing our `RecordAttributes` was strongly typed, it would make all of this so much easier. But it's not, and changing that is far out of scope of this PR.

BUT … what if we more strongly typed relations coming out of PFBs and Parquets? Make a record class to represent a WDS relation, e.g.
```java
public record RelationAttribute(RecordType targetType, String targetId) {

  @Override
  public String toString() {
    return RelationUtils.createRelationString(targetType, targetId);
  }
}
```
and then in `PfbRecordConverter` and `ParquetRecordConverter` where we currently call `RelationUtils.createRelationString`, create a `RelationAttribute` instead? I tried this pretty quickly, and I think it only has minor compatibility issues; you'll need to update `RelationUtils.getTypeValue` and `RelationUtils.getRelationValue` to call `toString()` before trying to parse their arguments.

I have this mostly working locally, if you want a code contribution!

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsModel.java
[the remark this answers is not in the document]
davidangb: Bah, I was hoping the rawls-model library would just work, but I'm not surprised it doesn't. I like the idea of just using a plain-old Java map to write the entity references in the Rawls json - that should really cut down on complexity here.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsModel.java
```diff
@@ -131,4 +151,65 @@ public Op op() {
       return CREATE_ATTRIBUTE_ENTITY_REFERENCE_LIST;
     }
   }
+
+  /**
+   * A reference to a record in the workspace data service, used to serialize a reference to Rawls
+   * upsert JSON.
+   */
+  public record RecordReference(RecordType entityType, String entityName) {
```
davidangb, line 159: nit: if you keep this (and don't replace it with a plain ol Map), rename it to `EntityReference`? A Rawls "entity" is the same as a WDS "record", and naming this `EntityReference` would make it clear it is intended for use only in Rawls.

jladieu: Hah!   I arrived at this same exact conclusion independently...this change will be reflected in the next commit I push up to this PR.

jladieu:
>BUT … what if we more strongly typed relations coming out of PFBs and Parquets? Make a record class to represent a WDS relation, e.g.

I am *all* for this.  I tried this out myself and couldn't pull it off without detonating a big codechange that blew up a bunch of other stuff.  Can you put up a PR onto this PR so I can see what you've got before commiting to it?

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
jladieu: Merged your downstream.  Thanks for the PR!

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsModel.java
jladieu: Fixed

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsModel.java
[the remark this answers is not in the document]
jladieu: Map approach got bad fast too, ended up circling back to the custom serializer with 1cb9b06817257f4dadc5be27ad6c9d230a2a294a

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsModel.java
```diff
@@ -131,4 +158,37 @@ public Op op() {
...
+      return switch (currentToken) {
+        case VALUE_STRING -> AttributeValue.of(parser.getValueAsString());
+        case VALUE_NUMBER_INT -> AttributeValue.of(parser.getIntValue());
+        case VALUE_NUMBER_FLOAT -> AttributeValue.of(parser.getDoubleValue());
+        case VALUE_TRUE, VALUE_FALSE -> AttributeValue.of(parser.getBooleanValue());
+        case VALUE_NULL -> null;
+        case START_OBJECT -> {
+          EntityReference entityReference = parser.readValueAs(EntityReference.class);
+          yield AttributeValue.of(entityReference);
+        }
```
davidangb, line 189: I think this would break if we encountered a Rawls [AttributeValueRawJson](https://github.com/broadinstitute/rawls/blob/bf26d6266ce3980b0f9c255445e32c4d84a450ea/model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala#L1038-L1040). But I'm ok dealing with that when it comes up, _if_ it ever comes up.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsModel.java
```diff
@@ -126,9 +187,59 @@ public Op op() {
 
   public record CreateAttributeEntityReferenceList(String attributeName)
       implements AttributeOperation {
+
+    @Override
```
jladieu, line 191: I added `attributeName` to the interface to improve test ergonomics.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbRecordConverter.java
```diff
@@ -56,8 +56,7 @@ protected final Record convertRelations(GenericRecord genericRecord) {
           String relationId = relation.get(RELATIONS_ID).toString();
           // Give the relation column the name of the record type it's linked to
           attributes.putAttribute(
-              relationType,
-              RelationUtils.createRelationString(RecordType.valueOf(relationType), relationId));
+              relationType, new RelationAttribute(RecordType.valueOf(relationType), relationId));
```
jladieu, line 59: For an explanation of this change, please take a look at [this comment](https://github.com/DataBiosphere/terra-workspace-data-service/pull/663#discussion_r1533989007)

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsModel.java
```diff
@@ -80,37 +94,67 @@ public sealed interface AttributeOperation
           CreateAttributeValueList,
           CreateAttributeEntityReferenceList {
     Op op();
+
+    @VisibleForTesting
+    String attributeName();
```
jladieu, line 99: This was added to the interface for testing ergnomics.  Of note: most of the subclasses implement this method due to having a record attribute called `attributeName` already, however a couple of the subclasses needed to explicitly provide this due to having a different name to satisfy compatibility with Rawls JSON keys.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsModel.java
jladieu: I think this comment might predate my latest commit which I _think_ at least does something intentional for non-entity objects.  PTAL [new tests for JSON](https://github.com/DataBiosphere/terra-workspace-data-service/pull/663/files#diff-bb16cae006f5b951c911abbd5258f98b4f7643f0cb6450e7f5d5acb4066933b9R184-R228) to see if I've got the expected behavior correct.

service/src/test/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkTest.java
```diff
@@ -59,6 +65,9 @@ class RawlsRecordSinkTest extends TestBase {
   @Autowired
   private GcsStorage storage;
 
+  /** ArgumentCaptor for the message passed to {@link PubSub#publishSync(Map)}. */
```
jladieu, line 68: This is unrelated to the changes I was making, but addresses a warning of using a raw parameterized type later on where we use the argument captor.

service/src/test/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkTest.java
```diff
@@ -72,13 +81,151 @@ void translatesEntityFields() {
...
+    var entity = assertSingle(entities);
+    assertThat(entity.operations())
+        .containsExactlyElementsOf(
+            expectedArrayCreationOperations(
+                new CreateAttributeValueList("booleans"), List.of(true, false, true, true)));
+  }
+
   @Test
-  void translatesAddUpdateAttribute() {
+  void translatesStringAttribute() {
```
jladieu, line 114: Renamed method to follow the pattern of 

`translates<Thing>Attribute()`
`translates<Thing>ArrayAttribute()`

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsModel.java
```diff
@@ -80,37 +94,67 @@ public sealed interface AttributeOperation
...
+      // Here, you could add additional type checks or transformations as needed
+      return new AttributeValue(value);
+    }
+
+    @Override
+    @JsonValue
+    @SuppressWarnings("java:S6207") // allow redundant #value(), which is here for annotation
+    public Object value() {
+      return value;
+    }
```
davidangb, line 115: I think this would work too:
```java
    @JsonValue
    public Object getValue() {
      return value;
    }
```

service/src/test/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkTest.java
```diff
@@ -72,13 +81,151 @@ void translatesEntityFields() {
...
+                new CreateAttributeValueList("strings"), List.of("value1", "value2")));
+  }
+
+  @Test
+  void translatesIntegerAttribute() {
+    var entities =
+        doUpsert(makeRecord(/* type= */ "widget", /* id= */ "id", Map.of("someKey", 42)));
+
+    var operation = assertSingleOperation(AddUpdateAttribute.class, entities);
+    assertThat(operation.addUpdateAttribute().value()).isEqualTo(42);
```
davidangb, line 144: for even more testing thoroughness (here and in some other tests):
```java
assertThat(operation.attributeName()).isEqualTo("someKey");
```
but I wouldn't hold the PR for that

davidangb (approved):
two very minor comments inline, optional; thank you for this!

service/src/test/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkTest.java
```diff
@@ -72,13 +81,151 @@ void translatesEntityFields() {
     assertThat(entity.entityType()).isEqualTo("widget");
   }
 
+  @AfterEach
+  void teardown() {
+    storage.getBlobsInBucket().forEach(blob -> storage.deleteBlob(blob.getName()));
```
nawatts, line 86: Does this use a real bucket? If so, could this result in conflicts if multiple instances of these tests are run concurrently?

nawatts (approved):

service/src/test/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkTest.java
jladieu: Added this in 5d87ddd711549339660d24b3c66d6387614b509a and also made the key names more specific to the test case.

service/src/test/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkTest.java
jladieu: No this is using 

```
@Qualifier("mockGcsStorage")
@Autowired
private GcsStorage storage;
```

Which configures storage to use `LocalStorageHelper` which is provided by the SDK to provide a stateful simulated fake of GCS.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsModel.java
jladieu: Fixed with 84c13496bb893a4b2b1f9ab7bb6d6dbc0dd2eb15, I slightly adjusted to chain to `value()` but otherwise was able to simplify the annotation and avoid the warning.

davidangb (approved):
:shipit:

PR #664, opened 2024-03-20
AJ-1680: rawls client retries

davidangb:
AJ-1680: retries should work correctly for the *SnapshotSupport classes.

In this PR:
* update `RestClientRetry` to support retries for the `RawlsClient`, which uses `RestTemplate` instead of a generated API client. Before this change, retries did not work at all for `RawlsClient`.
* move retry logic out of `RawlsSnapshotSupport` and down into `RawlsClient`; this is consistent with where we have retries in our other DAOs.
* remove retry logic from `WsmSnapshotSupport`, since we already have retries in `WorkspaceManagerDao`. Before this change, we would retry the retries.


service/src/main/java/org/databiosphere/workspacedataservice/rawls/RawlsClient.java
```diff
@@ -1,77 +1,83 @@
...
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
 import org.springframework.http.ResponseEntity;
-import org.springframework.stereotype.Component;
-import org.springframework.web.client.RestClientResponseException;
 import org.springframework.web.client.RestTemplate;
 import org.springframework.web.util.UriComponentsBuilder;
 
-@Component
-@ControlPlane
```
davidangb, line 23: this was declared as a `@Component` even though we also had a `@Bean`-creator for it in `RawlsClientConfig`. I've removed annotations from here and left it all in `RawlsClientConfig`.

service/src/main/java/org/databiosphere/workspacedataservice/retry/RestClientRetry.java
```diff
@@ -158,6 +159,11 @@ private int extractResponseCode(Exception e) {
     if (e instanceof org.databiosphere.workspacedata.client.ApiException wdsException) {
       return wdsException.getCode();
     }
+    // Rawls, or anything using RestTemplate
+    if (e instanceof HttpStatusCodeException httpStatusCodeException) {
+      return httpStatusCodeException.getStatusCode().value();
+    }
+
```
davidangb, line 166: this change - handling `HttpStatusCodeException` - is what allows `RestClientRetry` to understand the errors from RawlsClient and decide when to retry.

service/src/test/java/org/databiosphere/workspacedataservice/rawls/RawlsClientRetryTest.java
```diff
@@ -0,0 +1,169 @@
...
+  void enumerateDoesRetry() {
+    // ARRANGE
+    WorkspaceId workspaceId = WorkspaceId.of(UUID.randomUUID());
+
+    // define the successful REST response payload
+    ResponseEntity<SnapshotListResponse> successResponse =
+        ResponseEntity.of(Optional.of(new SnapshotListResponse(List.of())));
+
+    // argument matcher for the rest template target url
+    var urlMatcher = new UriMatcher("/api/workspaces/%s/snapshots/v2".formatted(workspaceId));
```
jladieu, line 82: nit: consider renaming variable to `uriMatcher` to match type

service/src/test/java/org/databiosphere/workspacedataservice/rawls/RawlsClientRetryTest.java
```diff
@@ -0,0 +1,169 @@
...
+    verify(mockRestTemplate, times(3))
+        .exchange(
+            argThat(urlMatcher),
+            eq(HttpMethod.POST),
+            argThat(payloadMatcher),
+            eq(DataRepoSnapshotResource.class));
+  }
+
+  // custom ArgumentMatcher - does a given URI contain the supplied substring?
+  static class UriMatcher implements ArgumentMatcher<URI> {
```
jladieu, line 140: Fancy!

jladieu (approved):
Nice!

nawatts (approved):

service/src/test/java/org/databiosphere/workspacedataservice/rawls/RawlsClientRetryTest.java
```diff
@@ -0,0 +1,169 @@
...
+import org.springframework.http.HttpStatusCode;
+import org.springframework.http.ResponseEntity;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.TestPropertySource;
+import org.springframework.web.client.HttpServerErrorException;
+import org.springframework.web.client.RestTemplate;
+
+@DirtiesContext
+@ActiveProfiles(value = "control-plane", inheritProfiles = false)
```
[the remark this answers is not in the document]
davidangb: heh, this is very true - I left it in because then it will do the right thing when the next engineer comes along and copy/pastes it :)

PR #667, opened 2024-03-21
suggestion

davidangb:
https://github.com/DataBiosphere/terra-workspace-data-service/pull/663#issuecomment-2013021116

> Can you put up a PR onto this PR so I can see what you've got before commiting to it?

Here you go! This is what I have working - at least, what is passing all the tests.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsModel.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsModel.java
@@ -26,6 +26,7 @@ import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import java.io.IOException;
 import java.util.List;
 import org.databiosphere.workspacedataservice.shared.model.RecordType;
+import org.databiosphere.workspacedataservice.shared.model.RelationAttribute;
 import org.springframework.lang.Nullable;
 
 /**
@@ -64,7 +65,7 @@ public class RawlsModel {
 
     @Override
     public String toString() {
-      return String.valueOf(value);
+      return value;
     }
   }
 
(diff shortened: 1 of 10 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsModel.java
```diff
@@ -170,6 +171,10 @@ public record EntityReference(RecordType entityType, String entityName) {
     public static EntityReference fromReferenceString(String referenceString) {
       return new EntityReference(getTypeValue(referenceString), getRelationValue(referenceString));
     }
```
davidangb, line 173: I think `fromReferenceString` is unused as of this PR, you could remove it

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -140,8 +140,8 @@ private Stream<? extends AttributeOperation> toOperations(String name, Object at
   }
 
   private AttributeValue maybeCoerceRelation(Object originalValue) {
-    if (RelationUtils.isRelationValue(originalValue)) {
-      return AttributeValue.of(EntityReference.fromReferenceString(originalValue.toString()));
+    if (originalValue instanceof RelationAttribute relationAttribute) {
+      return AttributeValue.of(EntityReference.fromRelationAttribute(relationAttribute));
```
davidangb, line 144: you could leave the `if` clause here alone, since `RelationUtils.isRelationValue(originalValue)` performs the same `instanceof` check

service/src/main/java/org/databiosphere/workspacedataservice/service/RelationUtils.java
```diff
@@ -58,7 +59,7 @@ public static String getRelationValue(Object obj) {
    * @return true if attribute begins with the REFERENCE_IDENTIFIER
    */
   public static boolean isRelationValue(Object obj) {
-    return obj != null && obj.toString().startsWith(RELATION_IDENTIFIER);
+    return obj instanceof RelationAttribute || obj.toString().startsWith(RELATION_IDENTIFIER);
```
davidangb, line 62: * obj should never be null
* the `instanceof` check here is not strictly necessary, but it's a nice short-circuit. Without it, we'd call `RelationAttribute.toString()` and then `startsWith` against the string, which is more work than we need

PR #668, opened 2024-03-21
[AJ-1687] Fix expected format for PubSub message

nawatts:

One more fix for PubSub 🤞 

I adapted this code from import service, which uses snake case keys for pulling information out of the PubSub message attributes.
https://github.com/broadinstitute/import-service/blob/a818658d0850818804235c5dc7f239c0de95c8cf/app/status.py#L59-L65

But it turns out, the PubSub message attributes have camel case keys. I have no idea how/where it's getting translated in import service, but this should let CWDS handle status update messages.


nawatts:
To be added in with a larger PR for pull subscriptions.

PR #670, opened 2024-03-21
[AJ-1686] Enable GCP auto configuration only in the control plane

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1686

When WDS starts up, the Spring GCP framework looks for application default credentials, which it can't find because it's not running in GCP. This results in log noise from exceptions thrown at startup.

This disables auto configuration by default and enables it only on the control plane where CWDS is running in GCP.

https://googlecloudplatform.github.io/spring-cloud-gcp/5.1.0/reference/html/index.html#configuration


nawatts:
Ugh, pushed the wrong branch

PR #671, opened 2024-03-22
[AJ-1687] Decouple CWDS project and Rawls notification topic

nawatts:

More tidying of PubSub config as part of AJ-1687...

This tweaks the configuration for the PubSub topic for CWDS to notify Rawls of an import.

Currently, ImportPubSub constructs a fully qualified topic name from CWDS's GCP project and the configured short topic name. However, this is unnecessarily complex.

It assumes that CWDS and the topic are in the same project. Because they are, we can use the short topic name instead of the fully qualified name. And if they were not, this change would allow us to configure CWDS using a fully qualified topic name.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/pubsub/ImportPubSub.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/pubsub/ImportPubSub.java
@@ -7,12 +7,11 @@ import java.util.concurrent.CompletableFuture;
 public class ImportPubSub implements PubSub {
 
   private final PubSubTemplate pubSubTemplate;
-  private final String fullTopicName;
+  private final String topic;
 
-  public ImportPubSub(PubSubTemplate pubSubTemplate, String topic, String project) {
+  public ImportPubSub(PubSubTemplate pubSubTemplate, String topic) {
     this.pubSubTemplate = pubSubTemplate;
-    /// projects/[project_name]/topics/[topic_name]
-    this.fullTopicName = "projects/" + project + "/topics/" + topic;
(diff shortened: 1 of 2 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/pubsub/ImportPubSub.java
```diff
@@ -23,8 +22,7 @@ public ImportPubSub(PubSubTemplate pubSubTemplate, String topic, String project)
   public String publishSync(Map<String, String> message) {
     // PubSubTemplate.publish returns a future
     // Rawls expects the actual data to be in the headers rather than the payload
-    CompletableFuture<String> publishFuture =
-        this.pubSubTemplate.publish(fullTopicName, "b''", message);
+    CompletableFuture<String> publishFuture = pubSubTemplate.publish(topic, "b''", message);
```
davidangb, line 25: not your code, but this should really be `""` instead of `"b''"` - not that it matters since we don't read the message payload

davidangb (approved):

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/ImportPubSub.java
nawatts: With formatting changes, my name's on the blame now. Updated in https://github.com/DataBiosphere/terra-workspace-data-service/pull/671/commits/4d2fb32ed3ca45b1d40b9c647fa7175f338ab43f.

jladieu (approved):

PR #673, opened 2024-03-22
[AJ-1687] Update PubSub configuration for tests

nawatts:

Prefactoring for AJ-1687...

We currently use a stub `PubSub` in tests. To do this, several tests configure the `spring.cloud.gcp.pubsub.enabled` property to `false` and PubSub config swaps out the implementation of `PubSub` based on that property.

Since this is only done for tests, this moves the stub PubSub and the Configuration that creates it into the test package and renames it from `Noop` to `Mock`, which aligns with other similar things like `MockSamResourcesApi`, etc.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/pubsub/PubSubConfig.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/pubsub/PubSubConfig.java
@@ -12,24 +12,11 @@ public class PubSubConfig {
 
   // when Pub/Sub autoconfiguration is enabled, use the real ImportPubSub bean
   @Bean
-  @ConditionalOnProperty(
-      name = "spring.cloud.gcp.pubsub.enabled",
-      havingValue = "true",
-      matchIfMissing = true)
-  PubSub applicationDefaultCredentialsPubSub(
+  @ConditionalOnProperty(name = "spring.cloud.gcp.pubsub.enabled", havingValue = "true")
+  PubSub getPubSub(
       GcpProjectIdProvider projectIdProvider,
       PubSubTemplate pubSubTemplate,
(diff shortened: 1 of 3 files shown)


calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/PubSubConfig.java
```diff
@@ -12,24 +12,11 @@ public class PubSubConfig {
 
   // when Pub/Sub autoconfiguration is enabled, use the real ImportPubSub bean
   @Bean
-  @ConditionalOnProperty(
-      name = "spring.cloud.gcp.pubsub.enabled",
-      havingValue = "true",
-      matchIfMissing = true)
-  PubSub applicationDefaultCredentialsPubSub(
+  @ConditionalOnProperty(name = "spring.cloud.gcp.pubsub.enabled", havingValue = "true")
+  PubSub getPubSub(
```
davidangb, line 16: we had the `matchIfMissing` set here because it matches the behavior of the pubsub starter library: if we don't set this property in any of our properties files, it defaults to true. I'm fine with this PR as-is, just want to make sure the reasoning was clear, in case it changes your mind

davidangb (approved):

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/PubSubConfig.java
nawatts: 👍 I was wondering what that was here. My thinking has been that the data plane configuration is the default for WDS and so this should be off unless it's explicitly enabled when running CWDS in the control plane.

PR #674, opened 2024-03-22
[AJ-1717] Simplify GCP project configuration

nawatts:
We get the GCP project in which CWDS is running through Spring Cloud GCP's `ProjectIdProvider`. It has several sources it can get the project ID from: https://cloud.spring.io/spring-cloud-static/spring-cloud-gcp/current/reference/html/#project-id.

Currently, we short circuit that by setting `spring.cloud.gcp.project-id` based on the `SERVICE_GOOGLE_PROJECT` environment variable.

Alternatively, we can lean into Spring Cloud GCP's project discovery and let it find the project from the environment in live environments. This would allow us to remove the `SERVICE_GOOGLE_PROJECT` environment variable.

When we do want to set the project via an environment variable, such as in local development, Spring Cloud GCP automatically looks at `GOOGLE_CLOUD_PROJECT`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,19 +65,28 @@ manually. The variables that need to be set are described below. You can also ad
 `~/.zshrc` or similar shell profile. However, in order for the app to run correctly, you still
 have to at least specify data-plane or control-plane.
 
-#### Control Plane Setup
-If you are running locally with the `control-plane` profile, you'll need some additional setup
-in order for pubsub to start up correctly.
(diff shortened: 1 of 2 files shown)


davidangb (approved):

davidangb:
ooh - could you change README.md for this too?

nawatts:
I did update one place in README.md with the new environment variable. Is there something I missed?

jladieu (approved):

PR #676, opened 2024-03-22
AJ-1716: TDR snapshot import to cWDS fails

davidangb:
[AJ-1716](https://broadworkbench.atlassian.net/browse/AJ-1716) Snapshot imports fail in the control plane with no logged error

The root cause is that the TDR import job set `ImportDetails.jobId` to `null`, but `RawlsRecordSink` required the jobId to be non-null.

Along the way to fixing that, I also:
* added logging into the `JobDao.fail()` methods to ensure we log job errors
* changed how we retrieve the end user's email address, so we never request it in the data plane and don't request it until we absolutely need it in the control plane; added unit tests for this
* tweaked constructors of `ImportDetails` to make it less easy to insert bad data
* cleaned up some IntelliJ warnings

[AJ-1716]: https://broadworkbench.atlassian.net/browse/AJ-1716?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
@@ -12,6 +12,7 @@ import java.time.ZoneOffset;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
+import javax.annotation.Nullable;
 import org.databiosphere.workspacedataservice.generated.GenericJobServerModel;
 import org.databiosphere.workspacedataservice.shared.model.CollectionId;
 import org.databiosphere.workspacedataservice.shared.model.job.Job;
@@ -129,6 +130,7 @@ public class PostgresJobDao implements JobDao {
    */
   @Override
   public GenericJobServerModel fail(UUID jobId, String errorMessage) {
+    logger.error("Job {} failed: {}", jobId, errorMessage);
     return update(jobId, StatusEnum.ERROR, errorMessage, null);
   }
 
@@ -154,11 +156,15 @@ public class PostgresJobDao implements JobDao {
    */
   @Override
   public GenericJobServerModel fail(UUID jobId, String errorMessage, Exception e) {
+    logger.error("Job {} failed: {}", jobId, errorMessage, e);
     return update(jobId, StatusEnum.ERROR, errorMessage, e.getStackTrace());
   }
 
   private GenericJobServerModel update(
-      UUID jobId, StatusEnum status, String errorMessage, StackTraceElement[] stackTrace) {
+      UUID jobId,
+      StatusEnum status,
+      @Nullable String errorMessage,
+      @Nullable StackTraceElement[] stackTrace) {
 
     // start our sql statement and map of params
     StringBuilder sb = new StringBuilder("update sys_wds.job set status = :status");
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
@@ -12,7 +12,6 @@ import java.time.ZoneOffset;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
-import javax.annotation.Nullable;
 import org.databiosphere.workspacedataservice.generated.GenericJobServerModel;
 import org.databiosphere.workspacedataservice.shared.model.CollectionId;
 import org.databiosphere.workspacedataservice.shared.model.job.Job;
@@ -23,6 +22,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.jdbc.core.RowMapper;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.lang.Nullable;
 import org.springframework.stereotype.Repository;
 
 /** Read/write jobs via the sys_wds.job Postgres table */
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/FileDownloadHelper.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/FileDownloadHelper.java
@@ -58,7 +58,8 @@ public class FileDownloadHelper {
       // TODO: this will throw DirectoryNotEmptyException; we need to delete individual files first
       Files.delete(tempFileDir);
     } catch (IOException e) {
-      logger.error("Error deleting temporary files: {} {}", e.getClass().getName(), e.getMessage());
+      logger.error(
+          "Error deleting temporary files: {} {}", e.getClass().getName(), e.getMessage(), e);
     }
   }
 
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/FileDownloadHelper.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/FileDownloadHelper.java
@@ -55,9 +55,10 @@ public class FileDownloadHelper {
 
   public void deleteFileDirectory() {
     try {
+      // TODO: this will throw DirectoryNotEmptyException; we need to delete individual files first
       Files.delete(tempFileDir);
     } catch (IOException e) {
-      logger.error("Error deleting temporary files: {}", e.getMessage());
+      logger.error("Error deleting temporary files: {} {}", e.getClass().getName(), e.getMessage());
     }
   }
 
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
@@ -1,31 +1,32 @@
 package org.databiosphere.workspacedataservice.dataimport;
 
-import jakarta.annotation.Nullable;
 import java.util.UUID;
 import org.databiosphere.workspacedataservice.recordsink.RawlsAttributePrefixer.PrefixStrategy;
 
 public record ImportDetails(
-    @Nullable UUID jobId,
-    @Nullable String userEmail,
-    UUID collectionId,
-    PrefixStrategy prefixStrategy) {
+    UUID jobId, String userEmail, UUID collectionId, PrefixStrategy prefixStrategy) {
+
+  private static final String DEFAULT_EMAIL = "unknown";
+  private static final UUID DEFAULT_JOB_ID =
+      UUID.fromString("00000000-0000-0000-0000-000000000000");
 
   /**
-   * Convenience constructor that sets jobId and userEmail to null
+   * Convenience constructor that sets jobId and userEmail to defaults
    *
    * @param collectionId target collection
    * @param prefixStrategy strategy for potentially renaming attributes
    */
   public ImportDetails(UUID collectionId, PrefixStrategy prefixStrategy) {
-    this(null, null, collectionId, prefixStrategy);
+    this(DEFAULT_JOB_ID, DEFAULT_EMAIL, collectionId, prefixStrategy);
   }
 
   /**
-   * Convenience constructor that sets jobId and userEmail to null, and sets prefixStrategy to NONE
+   * Convenience constructor that sets jobId and userEmail to defaults, and sets prefixStrategy to
+   * NONE
    *
    * @param collectionId target collection
    */
   public ImportDetails(UUID collectionId) {
-    this(null, null, collectionId, PrefixStrategy.NONE);
+    this(collectionId, PrefixStrategy.NONE);
   }
 }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
@@ -1,12 +1,16 @@
 package org.databiosphere.workspacedataservice.dataimport;
 
 import java.util.UUID;
+import java.util.function.Supplier;
 import org.databiosphere.workspacedataservice.recordsink.RawlsAttributePrefixer.PrefixStrategy;
 
 public record ImportDetails(
-    UUID jobId, String userEmail, UUID collectionId, PrefixStrategy prefixStrategy) {
+    UUID jobId,
+    Supplier<String> userEmailSupplier,
+    UUID collectionId,
+    PrefixStrategy prefixStrategy) {
 
-  private static final String DEFAULT_EMAIL = "unknown";
+  private static final Supplier<String> DEFAULT_EMAIL = () -> "unknown";
   private static final UUID DEFAULT_JOB_ID =
       UUID.fromString("00000000-0000-0000-0000-000000000000");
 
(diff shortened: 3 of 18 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/FileDownloadHelper.java
```diff
@@ -55,9 +55,10 @@ public void downloadFileFromURL(String tableName, URL pathToRemoteFile) {
 
   public void deleteFileDirectory() {
     try {
+      // TODO: this will throw DirectoryNotEmptyException; we need to delete individual files first
       Files.delete(tempFileDir);
```
davidangb, line 59: I discovered this TODO while working on this PR, but I would like to save a fix for a follow-on PR so we don't blow this one up.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/FileDownloadHelper.java
```diff
@@ -55,9 +55,10 @@ public void downloadFileFromURL(String tableName, URL pathToRemoteFile) {
...
     } catch (IOException e) {
-      logger.error("Error deleting temporary files: {}", e.getMessage());
+      logger.error("Error deleting temporary files: {} {}", e.getClass().getName(), e.getMessage());
     }
   }
```
davidangb, line 63: I discovered this TODO while working on this PR, but I would like to save a fix for a follow-on PR so we don't blow this one up.

The logging change is what I needed to figure out it was throwing DirectoryNotEmptyException.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
```diff
@@ -1,31 +1,26 @@
...
+   * Convenience constructor that sets jobId and userEmail to defaults, and sets prefixStrategy to
+   * NONE. Used by TSV and JSON APIs, which are not async and thus do not have job ids
    *
    * @param collectionId target collection
    */
   public ImportDetails(UUID collectionId) {
-    this(null, null, collectionId, PrefixStrategy.NONE);
+    this(DEFAULT_JOB_ID, DEFAULT_EMAIL_SUPPLIER, collectionId, PrefixStrategy.NONE);
   }
 }
```
davidangb, line 26: Use of a convenience constructor from `TdrManifestQuartzJob` was the root cause of snapshot import failing in the control plane - it set jobId to null, but `RawlsRecordSink` requires a non-null jobid.

I've removed one of the two convenience constructors, which was only used in tests, and changed the class so everything is always nonnull. One could make an argument that using default values is ugly and `null` is more appropriate; I could be convinced to switch that back.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
```diff
@@ -1,31 +1,26 @@
...
-import jakarta.annotation.Nullable;
 import java.util.UUID;
+import java.util.function.Supplier;
 import org.databiosphere.workspacedataservice.recordsink.RawlsAttributePrefixer.PrefixStrategy;
 
 public record ImportDetails(
-    @Nullable UUID jobId,
-    @Nullable String userEmail,
+    UUID jobId,
+    Supplier<String> userEmailSupplier,
```
davidangb, line 9: I changed `String userEmail` to `Supplier<String> userEmailSupplier` here. We only need the user's email in the control plane; in the data plane it's a totally extraneous API call to Sam. By making it a `Supplier`, which is only invoked from `RawlsRecordSink` in the control plane, I can eliminate the API call from the data plane.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -161,11 +171,14 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
...
+      throw new TdrManifestImportException(e.getMessage(), e);
+    } finally {
+      // delete temp files after everything else is completed
+      // Any failed deletions will be removed if/when pod restarts
+      fileDownloadHelper.deleteFileDirectory();
     }
-
-    // delete temp files after everything else is completed
-    // Any failed deletions will be removed if/when pod restarts
-    fileDownloadHelper.deleteFileDirectory();
```
davidangb, line 168: * add a `catch (Exception e)` clause - without this, we were not logging the cause of the import failure
* move `deleteFileDirectory` into a `finally` to make sure it always runs

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -151,7 +154,9 @@ private void publishToPubSub(ImportDetails importDetails) {
             .put("isUpsert", "true")
             .put("isCWDS", "true")
             .build();
-    pubSub.publishSync(message);
+    LOGGER.info("Publishing message to pub/sub for job {} ...", jobId);
+    String publishResult = pubSub.publishSync(message);
+    LOGGER.info("Pub/sub publishing complete for job {}: {}", jobId, publishResult);
```
davidangb, line 159: just some extra logging

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJobTest.java
```diff
@@ -196,20 +200,22 @@ void parseMalformedParquet() throws IOException {
...
+        new ImportDetails(jobId, emailSupplier, workspaceId, PrefixStrategy.TDR);
+    try (RecordSink recordSink = recordSinkFactory.buildRecordSink(importDetails)) {
+
+      // Make sure real errors on parsing parquets are not swallowed
+      assertThrows(
+          TdrManifestImportException.class,
+          () ->
+              tdrManifestQuartzJob.importTable(
+                  malformedFile, table, recordSink, ImportMode.BASE_ATTRIBUTES));
+    }
```
davidangb, line 213: fixes an IntelliJ warning that `RecordSink` was used outside of a try-with-resources

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJobTest.java
```diff
@@ -196,20 +200,22 @@ void parseMalformedParquet() throws IOException {
...
   }
 
   @Test
   @Tag(SLOW)
-  void useWorkspaceIdFromCollection() throws JobExecutionException, IOException {
+  void useWorkspaceIdFromCollection() throws IOException {
```
davidangb, line 218: fixes an IntelliJ warning about un-thrown exceptions

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
```diff
@@ -154,11 +156,15 @@ public GenericJobServerModel fail(UUID jobId, Exception e) {
    */
   @Override
   public GenericJobServerModel fail(UUID jobId, String errorMessage, Exception e) {
+    logger.error("Job {} failed: {}", jobId, errorMessage, e);
```
davidangb, line 159: these log statements may result in double-logging, in cases where the calling job also logs the error. But these log statements also catch the case we missed. So I'd rather leave them in and remove logging from calling code if we find double logs.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
```diff
@@ -12,6 +12,7 @@
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
+import javax.annotation.Nullable;
```
nawatts, line 15: Shouldn't this be the Spring annotation?

```suggestion
import org.springframework.lang.Nullable;
```

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/FileDownloadHelper.java
nawatts: Is it worth passing the full exception to `logger.error`?

```java
logger.error("Error deleting temporary files", e);
```

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
nawatts: > One could make an argument that using default values is ugly and null is more appropriate; I could be convinced to switch that back.

I think I would argue that.

Or... if the convenience constructor is only there tests, maybe a convenience constructor isn't the best way to go about it and we could have a test utility that creates these objects?

nawatts (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
jladieu: Love this!

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/FileDownloadHelper.java
davidangb: real fix in https://github.com/DataBiosphere/terra-workspace-data-service/pull/678

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
jladieu: I think I prefer explicit `@Nullable` (and then making downstream classes deal with this accordingly) over having defaults, which could very easily delay or hide failure modes.

Ultimately, I'd _really_ prefer that `ImportDetails` was a more cohesive set of attributes that were always all set, but I realize that's a yakshave for another day.

jladieu: Squinting more at this...now that TDR _does_ set the JobId, why not make it non-nullable and require it to always be set (and not let it be bypassed by the convenience constructor)?  I'd be okay with leaving in the DEFAULT_EMAIL_SUPPLIER as is.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
davidangb: hm, looking through our codebase we have both - and both seem to work; both delegate to the same underlying com.google.code.findbugs:jsr305 `@Nonnull(when…`. I've switched to using the springframework annotation exclusively.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
davidangb: I removed the convenience constructor that was only used by tests. The remaining convenience constructor _is_ used and is necessary for TSV import and JSON-api batchUpsert - both of those are synchronous and thus don't generate a jobId. I will look at putting nulls back in here and what impact that has.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
```diff
@@ -1,31 +1,26 @@
...
+      UUID.fromString("00000000-0000-0000-0000-000000000000");
 
   /**
-   * Convenience constructor that sets jobId and userEmail to null, and sets prefixStrategy to NONE
+   * Convenience constructor that sets jobId and userEmail to defaults, and sets prefixStrategy to
+   * NONE. Used by TSV and JSON APIs, which are not async and thus do not have job ids
    *
    * @param collectionId target collection
    */
   public ImportDetails(UUID collectionId) {
```
jladieu, line 23: Might be a good idea to provide a custom `toString()` method now and redact/omit the details of the userEmail supplier.  Since we know it carries an auth token around with it, my concern is that we'd want to avoid it getting accidentally printed out to the logs.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobControlPlaneE2ETest.java
```diff
@@ -232,6 +236,18 @@ void pfbToRawlsWithMultipleBatches() throws JobExecutionException, IOException {
...
+  @Test
+  @Tag(SLOW)
+  void pubSubRequestsUserEmail() throws JobExecutionException, IOException {
+    // Arrange / Act
+    UUID jobId = testSupport.executePfbImportQuartzJob(collectionId, minimalDataPfb);
+
+    // importing a PFB requires publishing to pubsub. As part of generating the pubsub message,
+    // we query Sam for the user's email. Verify we made that call correctly.
+    // the "expectedToken" is the value from PfbTestUtils.stubJobContext().
+    verify(samDao, times(1)).getUserEmail(BearerToken.of("expectedToken"));
```
jladieu, line 248: When I first scanned this, I was wondering where `"expectedToken"` was coming from, and only then did I read the comment which helped explain why.

Consider exposing `expectedToken` as a constant and then refer to that directly.

jladieu (commented):
I have a few suggestions inline.  The only one I'd consider urgent is the one about overriding `toString()`; I think the supplier approach might potentially provide an avenue to leak the auth token to the logs unless we're careful.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
jladieu: Ahh interesting!  I'd be in favor of switching over to the core java version globally if that's of interest.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
davidangb: great catch, added

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/FileDownloadHelper.java
davidangb: added the full exception to the logging statement

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
davidangb: ok, I've reverted these nullability changes and we're back to allowing jobId and userEmailSupplier to be null. I've added fail-fast validation into `RawlsRecordSinkFactory` to check that these values are non-null before even creating a `RecordSink`. How does this look?

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobControlPlaneE2ETest.java
davidangb: moved this to a constant.

jladieu (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
jladieu: The fact that you ran into this was bugging me so I tried to figure out how to `un-@Nullable` `ImportDetails` in a PR off this PR: https://github.com/DataBiosphere/terra-workspace-data-service/pull/681

PR #677, opened 2024-03-22
[AJ-1641] Support permissions syncing in cWDS in GCP only

ashanhol:
Issue: https://broadworkbench.atlassian.net/browse/AJ-1641

This PR enables permission syncing on TDR Import Jobs. It adds a call to Sam to add "reader", "writer", "owner", "project-owner" reader roles.  
Accessing this requires two checks- the first is that the syncpermissions property is set (set to true in the application-control-plane.yml and false in the data plane one), and the call has `tdrSyncPermissions:true` set in `options` in the call. 

Permission syncing working: 
image

-------------
Reminder:


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -390,4 +403,47 @@ protected void linkSnapshots(Set<UUID> snapshotIds, WorkspaceId workspaceId) {
...
+  private void syncPermissions(JobDataMap jobDataMap, WorkspaceId workspaceId, UUID snapshotId) {
+    String authToken;
+    try {
+      // only proceed if we have permission syncing enabled
+      if (!jobDataMap.getBoolean(ARG_PERMISSION_OPTION)) {
+        return;
+      }
+      authToken = getJobDataString(jobDataMap, ARG_TOKEN);
+    } catch (Exception e) {
+      // don't throw an exception if we can't get values, just return
```
ashanhol, line 423: Current behavior if we mess up getting the bearer token or the permission option is to not throw an exception so the job will complete. Not sure if I should?

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
```diff
@@ -69,4 +74,25 @@ public String getPetToken() {
...
     return restClientRetry.withRetryAndErrorHandling(samFunction, "Sam.getPetToken");
   }
+
+  public void addMemberPolicy(
+      String resourceTypeName,
+      WorkspaceId workspaceId,
+      UUID snapshotId,
+      String readerRole,
+      BearerToken token) {
+    restClientRetry.withRetryAndErrorHandling(
```
ashanhol, line 84: Followed same call done in import service here: https://github.com/broadinstitute/import-service/blob/a818658d0850818804235c5dc7f239c0de95c8cf/app/translators/sync_permissions.py#L24

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -166,6 +174,11 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
     // delete temp files after everything else is completed
     // Any failed deletions will be removed if/when pod restarts
     fileDownloadHelper.deleteFileDirectory();
+
+    // sync permissions if running in control-plane
+    if (dataImportProperties.isPermissionSync()) {
+      syncPermissions(jobDataMap, workspaceId, snapshotId);
```
ashanhol, line 180: two checks to enable sync: 
1. are we in control plane (hence the property check, as it'll only be set to true in application-control-plane.yml)
2. is the option passed via context args

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
```diff
@@ -101,7 +101,7 @@ class TestableQuartzJob extends QuartzJob {
     private boolean shouldThrowError = false;
 
     public TestableQuartzJob(String expectedToken, ObservationRegistry registry) {
-      super(registry, dataImportProperties);
+      super(registry, QuartzJobTest.this.dataImportProperties);
```
ashanhol, line 104: compiler required this change because I changed dataimportproperties from private to protected and am accessing it in a child class. error says it was accessed too soon

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
jladieu: In the absence of existing behavior to copy, I would bias toward crashing hard on this sort of weirdness.  Since there seems to be prior art here, I'd look to a historian :eyes: @davidangb for opinions :)

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
davidangb: I find it awkward that these two checks are separated. Could you elevate the context-arg check up to here, and remove it from `syncPermissions`?
```java
    if (dataImportProperties.isPermissionSync() && jobDataMap.getBoolean(ARG_PERMISSION_OPTION)) {
      syncPermissions(jobDataMap, workspaceId, snapshotId);
    }
```
if you do this - coupled with the fact that code on `main` already retrieves the auth token from the context map - you can avoid passing the whole job context map into `syncPermissions` and just pass the auth token instead.

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
```diff
@@ -69,4 +74,25 @@ public String getPetToken() {
...
+        () ->
+            samClientFactory
+                .getResourcesApi()
+                .addMemberPolicyV2(
```
jladieu, line 88: I'll be curious what your experimentation with this API call shows once you try this out.

One thing I'm thinking is: I _suspect_ that you don't want/need to wire the `bearerToken` through here and can just omit it in the call to `addMemberPolicyV2` and from `addMemberPolicy`.  `SamClientFactory` will consult the request or job context to determine the bearer token of the current user.

I looked at the source for `addMemberPolicyV2` and I think that last arg is intended to be for the optional request body and not for the token:

```
    /**
     * Add a member policy to a parent policy
     * 
     * @param resourceTypeName Type of resource (required)
     * @param resourceId Id of resource (required)
     * @param policyName Name of the policy (required)
     * @param memberResourceTypeName Type of resource for the member policy (required)
     * @param memberResourceId Resource id for the member policy (required)
     * @param memberPolicyName Member policy name (required)
     * @param body  (optional)
     * @throws ApiException If fail to call the API, e.g. server error or cannot deserialize the response body
     * @http.response.details
     <table summary="Response Details" border="1">
        <tr><td> Status Code </td><td> Description </td><td> Response Headers </td></tr>
        <tr><td> 204 </td><td> Successfully added a member policy to the parent policy </td><td>  -  </td></tr>
        <tr><td> 403 </td><td> You do not have permission to alter this policy </td><td>  -  </td></tr>
        <tr><td> 404 </td><td> Resource type does not exist, you are not a member of any policy on the resource, or member policy was not found </td><td>  -  </td></tr>
        <tr><td> 500 </td><td> Internal Server Error </td><td>  -  </td></tr>
     </table>
     */
    public void addMemberPolicyV2(String resourceTypeName, String resourceId, String policyName, String memberResourceTypeName, String memberResourceId, String memberPolicyName, Object body) throws ApiException {
        addMemberPolicyV2WithHttpInfo(resourceTypeName, resourceId, policyName, memberResourceTypeName, memberResourceId, memberPolicyName, body);
    }
```

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
```diff
@@ -95,6 +96,12 @@ public GenericJobServerModel createImport(
       arguments.put(ARG_TOKEN, petToken);
       arguments.put(ARG_URL, importRequest.getUrl().toString());
       arguments.put(ARG_COLLECTION, collectionId.toString());
+      // try to retrieve the tdrSyncPermissions option if available
+      arguments.put(
+          ARG_PERMISSION_OPTION,
+          importRequest.getOptions().containsKey("tdrSyncPermissions")
+              ? (boolean) importRequest.getOptions().get("tdrSyncPermissions")
```
jladieu, line 103:
```suggestion
              ? Boolean.parseBoolean(importRequest.getOptions().get("tdrSyncPermissions"))
```

I suspect you need to do more than cast here

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
```diff
@@ -69,4 +74,25 @@ public String getPetToken() {
...
+                    resourceTypeName,
+                    snapshotId.toString(),
+                    READER_POLICY_NAME,
+                    RESOURCE_NAME_WORKSPACE,
+                    workspaceId.toString(),
+                    readerRole,
+                    token),
+        "Sam.addMemberPolicy");
+  }
```
davidangb, line 97: nit: could you name this something a little less generic? This doesn't just add a generic member policy to some resource - it adds a _workspace's_ policy as _reader_ on some resource. So there's some built-in business logic.

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
ashanhol: Got it, I was copying the python call which included the bearer token as the last param. I didn't realize the http call could automatically get it from context

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJobTest.java
```diff
@@ -235,4 +245,40 @@ void useWorkspaceIdFromCollection() throws JobExecutionException, IOException {
...
+    JobExecutionContext mockContext =
+        stubJobContext(
+            jobId, v2fManifestResource, collectionId.id(), /* shouldPermissionSync= */ true);
+
+    // ACT
+    // execute the job
+    tdrManifestQuartzJob.executeInternal(jobId, mockContext);
+
+    // ASSERT
+    verify(samDao, times(4)).addMemberPolicy(anyString(), any(), any(), anyString(), any());
```
jladieu, line 282:
```suggestion
    ArgumentCaptor<String> roleCaptor = ArgumentCaptor.forClass(String.class);
    verify(samDao, times(4))
        .addMemberPolicy(anyString(), any(), any(), roleCaptor.capture(), any());
    assertThat(roleCaptor.getAllValues())
        .containsExactlyInAnyOrder("reader", "writer", "owner", "project-owner");
```

I'd suggest using an `ArgumentCaptor` here to check that the right roles were passed through.

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
```diff
@@ -95,6 +96,12 @@ public GenericJobServerModel createImport(
...
+              : false);
```
davidangb, line 104: the cast to `boolean` here is unsafe. What about something like this?
```java
      // we expect the options map to contain a boolean for tdrSyncPermissions, but people could easily call the
      // API with a String value of `"true"` instead of the bare boolean `true`, so first grab the value as a String:
      String tdrSyncPermissions =
          importRequest.getOptions().getOrDefault("tdrSyncPermissions", "false").toString();
      // then, parse that String into a boolean:
      arguments.put(ARG_PERMISSION_OPTION, Boolean.parseBoolean(tdrSyncPermissions));
```

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
ashanhol: I believe you, but for my learnings sake, why? The map is <string, object>, so if you have an object that's just `true`, wouldn't that (boolean) cast be sufficient?

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
jladieu: This looks fishy...I actually don't even know how to reason about how this works...

To that end, I recommend just adding a `DataImportProperties` arg to `TestableQuartzJob` constructor and wiring that through from `QuartzJobTest`

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJobTest.java
davidangb, line 282: you know all of the values that `addMemberPolicy` should be called with here. Could you split this out into 4 separate `verify` checks, each of which checks `times(1)` and specifies exactly the values you expect it to be called with?

The weirdest value to figure out here is the token, which is specified in `stubJobContext`. The latest code on `main` exposes that value as a constant to make it easier to grab and use in cases like this.

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
```diff
@@ -36,7 +36,7 @@
 public abstract class QuartzJob implements Job {
 
   private final ObservationRegistry observationRegistry;
-  private final DataImportProperties dataImportProperties;
+  protected final DataImportProperties dataImportProperties;
```
davidangb, line 39: Is this necessary? Could you leave this as private, but add `DataImportProperties dataImportProperties` as a private member var on `TdrManifestQuartzJob` so it can be used there too?

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbTestUtils.java
```diff
@@ -160,7 +162,9 @@ public static JobExecutionContext stubJobContext(UUID jobId, Resource resource,
                     ARG_URL,
                     resource.getURL().toString(),
                     ARG_COLLECTION,
-                    collectionId.toString())));
+                    collectionId.toString(),
+                    ARG_PERMISSION_OPTION,
+                    shouldPermissionSync)));
```
davidangb, line 167: if you added a convenience method:
```java
  public static JobExecutionContext stubJobContext(
      UUID jobId, Resource resource, UUID collectionId) throws IOException {
    return stubJobContext(jobId, resource, collectionId, false);
  }
```
you could avoid having to add `, /* shouldPermissionSync= */ false` on almost all invocations. Nice side effect: the argument isn't even visible in PFB tests if they use the convenience method.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobTest.java
```diff
@@ -216,7 +219,9 @@ void snapshotIdsAreParsed() throws JobExecutionException, IOException {
     UUID jobId = UUID.randomUUID();
     CollectionId collectionId = CollectionId.of(UUID.randomUUID());
     WorkspaceId workspaceId = WorkspaceId.of(UUID.randomUUID());
-    JobExecutionContext mockContext = stubJobContext(jobId, testAvroResource, collectionId.id());
+    JobExecutionContext mockContext =
+        stubJobContext(
+            jobId, testAvroResource, collectionId.id(), /* shouldPermissionSync= */ false);
```
davidangb, line 224: bleh, this is a weird side effect of TDR tests also using `PfbTestUtils.stubJobContext` - PFB tests have to specify a value they never use. See my comment on `PfbTestUtils` for more …

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -390,4 +403,47 @@ protected void linkSnapshots(Set<UUID> snapshotIds, WorkspaceId workspaceId) {
...
+      }
+      authToken = getJobDataString(jobDataMap, ARG_TOKEN);
+    } catch (Exception e) {
+      // don't throw an exception if we can't get values, just return
+      logger.error(
+          permissionFailureErrorMessage(
+              snapshotId, workspaceId, "failed to retrieve arguments from job"));
+      return;
+    }
+    String[] readerRoles = {"reader", "writer", "owner", "project-owner"};
```
davidangb, line 429: Could you move this to a class-level static constant, so we're not recreating it on every invocation of this method? Alternately, it could move to a constant somewhere else … `SamAuthorizationDao` maybe?

davidangb (commented):
See comments inline!

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
ashanhol: done

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbTestUtils.java
ashanhol: that's so much better 👍

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
ashanhol: yeah that works and gets rid of that weird issue in that one test

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
ashanhol: yes thank you

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
ashanhol: Sam.addWorkspaceReaderPolicyToResource?

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
ashanhol: verified that yeah, getResourcesApi does do the token thing

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJobTest.java
ashanhol: ccing @davidangb on this thread since he left a comment below- Id think the roles were the only significant param to check, right? This is a much cleaner way to do it so we don't have to call times(1) and check each param

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
ashanhol: fixed with a suggestion david made

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
ashanhol: ah yeah that's way better

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJobTest.java
```diff
@@ -241,4 +252,42 @@ void useWorkspaceIdFromCollection() throws IOException {
     // ASSERT
     verify(wsmDao).enumerateDataRepoSnapshotReferences(eq(workspaceId.id()), anyInt(), anyInt());
   }
+
+  @Test
+  @Tag(SLOW)
+  /* note: this functionality is control plane only.*/
```
calypsomatic, line 258: If the test extends `TestBase`, isn't this test for `data-plane` then?  Do we need a new test class that tests Tdr for control-plane?

ashanhol: I _could_ but I didn't want to if I didn't have to. The control plane part of it can be mocked with the `dataImportProperties.isPermissionSync()`.

service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
```diff
@@ -63,6 +64,20 @@ public void setSucceedOnCompletion(boolean succeedOnCompletion) {
     this.succeedOnCompletion = succeedOnCompletion;
   }
 
+  /**
+   * Permissions syncing on TDR import jobs is only enabled on the control plane.
+   *
+   * @see org.databiosphere.workspacedataservice.dataimport.tdr.TdrManifestQuartzJob
+   * @return the configured value
+   */
+  public boolean isPermissionSync() {
```
nawatts, line 73: This reads a little awkwardly. Maybe something like:

```suggestion
  public boolean isTdrPermissionSyncingEnabled() {
```

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -71,6 +74,8 @@ public class TdrManifestQuartzJob extends QuartzJob {
   private final RecordSourceFactory recordSourceFactory;
   private final SnapshotSupportFactory snapshotSupportFactory;
   private final SamDao samDao;
+  private final DataImportProperties dataImportProperties;
+  private static final String RESOURCE_TYPE_DATASNAPSHOT = "datasnapshot";
```
nawatts, line 78: Since it's Sam-related, is there a more appropriate place for this constant than the TDR import job?

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
```diff
@@ -69,4 +74,21 @@ public String getPetToken() {
...
   }
+
+  public void addMemberPolicy(
+      String resourceTypeName, WorkspaceId workspaceId, UUID snapshotId, String readerRole) {
+    restClientRetry.withRetryAndErrorHandling(
+        () ->
+            samClientFactory
+                .getResourcesApi()
+                .addMemberPolicyV2(
+                    resourceTypeName,
```
nawatts, line 85: Is this ever called with anything other than the snapshot resource type? Do we need that as a parameter?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
nawatts: Also, another naming suggestion
```suggestion
  private static final String TDR_SNAPSHOT_SAM_RESOURCE_TYPE = "datasnapshot";
```

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -398,4 +408,29 @@ protected void linkSnapshots(Set<UUID> snapshotIds, WorkspaceId workspaceId) {
...
+      try {
+        samDao.addMemberPolicy(RESOURCE_TYPE_DATASNAPSHOT, workspaceId, snapshotId, role);
+      } catch (RestException e) {
+        throw new TdrManifestImportException(
+            "Failed to sync permissions for TDR snapshot "
+                + snapshotId
+                + " into workspace "
+                + workspaceId.toString()
+                + "RestException on role "
+                + role,
```
nawatts, line 431: What's the convention on concatenation vs `"<format string>".formatted(...)` in Java?

service/src/main/resources/application-control-plane.yml
```diff
@@ -11,14 +11,14 @@ twds:
     batch-write-record-sink: "rawls"
     succeed-on-completion: false
     rawls-bucket-name: ${SERVICE_GOOGLE_BUCKET}
+    permission-sync: true
```
nawatts, line 14: Since this is specific to one type of import (TDR snapshots) and allows a user to sync permissions, suggest a more specific name:

```suggestion
    enable-tdr-permission-sync: true
```

service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
```diff
@@ -22,6 +22,7 @@ public class DataImportProperties {
   private RecordSinkMode batchWriteRecordSink;
   private String rawlsBucketName;
   private boolean succeedOnCompletion;
+  private boolean permissionSync;
```
nawatts, line 25: Initializing to false would allow omitting this from the data plane profile.

```suggestion
  private boolean permissionSync = false;
```

service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
ashanhol: I agree, but how married are we on the convention of the get function being is<PropertyName>? That's how this function is named

nawatts: In my opinion, we're not very attached to that convention. TenancyProperties contains some boolean properties with getters named `getProperty`.

And I also proposed renaming the property in the profile YAML 😁.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
davidangb: `SamAuthorizationDao` already has `String RESOURCE_NAME_WORKSPACE = "workspace";`, I'd suggest moving this over there and have both resource names in the same place

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -167,6 +173,10 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
                               .record()
                               .withRecordType(entry.getKey())
                               .ofQuantity(entry.getValue())));
+      // sync permissions if running in control-plane
```
davidangb, line 176: nit: comment is now slightly out of date

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -398,4 +408,29 @@ protected void linkSnapshots(Set<UUID> snapshotIds, WorkspaceId workspaceId) {
...
     snapshotSupport.linkSnapshots(snapshotIds);
   }
+
+  /**
+   * Sync permissions with SAM if option is provided.
+   *
+   * @param jobDataMap map to extract the argument options if exists
+   * @param workspaceId current workspace
+   * @param snapshotId id of the TDR snapshot
+   */
```
davidangb, line 418: javadoc is now incorrect

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
davidangb: My $.02 is to do whatever is more readable, _unless_ it's in a high-performance area in which case we should prefer a `StringBuilder` (or `StringBuffer` if we need thread safety). This isn't a high-performance area so I don't have an opinion on what happens here.

Good article: https://www.baeldung.com/java-string-concatenation-methods

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
davidangb: this is adding more than the workspace reader policy - it's adding the workspace reader policy, writer policy, owner policy, and project-owner policy. How about "Sam.addWorkspacePoliciesAsReader"? Or, if you hardcode the snapshot resource as Nick commented, it could be "Sam.addWorkspacePoliciesAsSnapshotReader"

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamAuthorizationDao.java
```diff
@@ -10,6 +10,12 @@ public interface SamAuthorizationDao {
   /** Sam action name for read permission */
   String ACTION_READ = "read";
 
+  /** Sam policy name for read permission */
+  String READER_POLICY_NAME = "reader";
+
+  /** All roles to add reader policy */
+  String[] READER_ROLES = {"reader", "writer", "owner", "project-owner"};
```
davidangb, line 17: another rename suggestion: `WORKSPACE_ROLES`?

High-level, you are taking all available _workspace_ roles (reader, writer, owner, project-owner) and adding those to a single _snapshot_ role (reader). This variable represents the list of workspace roles.

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
```diff
@@ -32,6 +32,7 @@
 
 @Service
 public class ImportService {
+  public static final String ARG_PERMISSION_OPTION = "permission_option";
```
davidangb, line 35: nit: we use camelCase everywhere else, not snake case; change to `permissionOption`? Or, to mirror the inbound option, change to `tdrSyncPermissions`?

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
ashanhol: It isn't called with anything else _yet_. I'm a fan of keeping this as dynamic as possible in case there's ever a case where we need to support other scenarios

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJobTest.java
```diff
@@ -237,7 +246,9 @@ void useWorkspaceIdFromCollection() throws IOException {
     // set up the job
     TdrManifestQuartzJob tdrManifestQuartzJob =
         testSupport.buildTdrManifestQuartzJob(workspaceId.id());
-    JobExecutionContext mockContext = stubJobContext(jobId, v2fManifestResource, collectionId.id());
+    JobExecutionContext mockContext =
+        stubJobContext(
+            jobId, v2fManifestResource, collectionId.id(), /*shouldPermissionSync*/ false);
```
davidangb, line 251: you can remove the `/*shouldPermissionSync*/ false` here

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
ashanhol: oh man i didn't even realize there was a "".formatted option, that's a lot nicer looking than String.format

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJobTest.java
```diff
@@ -246,4 +257,42 @@ void useWorkspaceIdFromCollection() throws IOException {
...
+            jobId, v2fManifestResource, collectionId.id(), /* shouldPermissionSync= */ true);
+
+    // ACT
+    // execute the job
+    tdrManifestQuartzJob.executeInternal(jobId, mockContext);
+
+    // ASSERT
+    ArgumentCaptor<String> roleCaptor = ArgumentCaptor.forClass(String.class);
+    verify(samDao, times(4)).addMemberPolicy(anyString(), any(), any(), roleCaptor.capture());
+    assertThat(roleCaptor.getAllValues()).containsExactlyInAnyOrder(READER_ROLES);
```
davidangb, line 296: nice verification of the workspace roles!

I still think it's worth verifying all the other arguments to `addMemberPolicy` - the resourceTypeName, workspaceId, and snapshotId.

davidangb (commented):
a bunch of naming suggestions, a test suggestion, some doc suggestions; this is very close

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
nawatts: Currently, it's kind of in an odd intermediate state between generic and specific. It accepts a resource type, but the name of the `snapshotId` parameter suggests it should be a snapshot. And the other pieces are hard coded: `READER_POLICY_NAME`, `RESOURCE_NAME_WORKSPACE`.

I don't have a strong opinion on _which_ way to go, but I think it would be preferable to commit to _one_ of those two directions.

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
ashanhol: hardcoded it and renamed

service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
ashanhol: aight, renamed

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
ashanhol: Renamed to TDR_SNAPSHOT_RESOURCE_TYPE

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
ashanhol: went with the hardcoded way and left a comment when we call addMemberPolicyV2

service/src/main/resources/application-control-plane.yml
ashanhol: renamed

service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
ashanhol: ha thanks that's great, changed

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
ashanhol: updated to be more specific

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
ashanhol: ah fixed it but didn't commit it, fixed now

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamAuthorizationDao.java
ashanhol: :( i dislike trying to name all these things

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
ashanhol: fair, renamed to use tdrSyncPermissions. I was going based off of what I thought the other args were but they were all one word so i matched what the const was

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJobTest.java
ashanhol: thanks, missed one -_-

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJobTest.java
ashanhol: aight, added checks for snapshotid and workspaceid.

nawatts (approved):

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamAuthorizationDao.java
```diff
@@ -10,6 +10,15 @@ public interface SamAuthorizationDao {
   /** Sam action name for read permission */
   String ACTION_READ = "read";
 
+  /** Sam policy name for read permission */
```
jladieu, line 13: Since these constants are only used as internal details of `HttpSamDao`, consider moving them to become 
`private static final String` within that class.  For the constants that need broader access for testing purposes only, make them package-protected (or public if needed) and mark them as `@VisibleForTesting`

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
```diff
@@ -95,6 +96,10 @@ public GenericJobServerModel createImport(
       arguments.put(ARG_TOKEN, petToken);
       arguments.put(ARG_URL, importRequest.getUrl().toString());
       arguments.put(ARG_COLLECTION, collectionId.toString());
+      // try to retrieve the tdrSyncPermissions option if available (first as string, then bool)
+      String tdrSyncPermissions =
+          importRequest.getOptions().getOrDefault("tdrSyncPermissions", "false").toString();
```
jladieu, line 101: nit: use `ARG_TDR_SYNC_PERMISSION` for the key value here too?

jladieu (approved):

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamAuthorizationDao.java
```diff
@@ -10,6 +10,15 @@ public interface SamAuthorizationDao {
...
+  String READER_POLICY_NAME = "reader";
+
+  /* Resource type to specify when adding Sam member policies */
+  String TDR_SNAPSHOT_RESOURCE_TYPE = "datasnapshot";
```
davidangb, line 17: naming is hard. This class now has:
`RESOURCE_NAME_WORKSPACE`
`TDR_SNAPSHOT_RESOURCE_TYPE`

could you align those? Either
`RESOURCE_NAME_WORKSPACE`
`RESOURCE_NAME_TDR_SNAPSHOT`

or 
`WORKSPACE_RESOURCE_TYPE`
`TDR_SNAPSHOT_RESOURCE_TYPE`

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamAuthorizationDao.java
ashanhol: that interface requires it all to be `public`, but i can make all of these variables static and final

service/src/main/java/org/databiosphere/workspacedataservice/sam/HttpSamDao.java
ashanhol: renamed method

service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
```diff
@@ -63,6 +64,20 @@ public void setSucceedOnCompletion(boolean succeedOnCompletion) {
...
+   * Permissions syncing on TDR import jobs is only enabled on the control plane.
+   *
+   * @see org.databiosphere.workspacedataservice.dataimport.tdr.TdrManifestQuartzJob
+   * @return the configured value
+   */
+  public boolean isTdrPermissionSyncingEnabled() {
+    return enableTdrPermissionSync;
+  }
+
+  public void setTdrPermissionSync(boolean enableTdrPermissionSync) {
```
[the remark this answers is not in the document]
ashanhol: asdkl;jfhasdl;fkjsda;l ofc it does

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -96,6 +100,7 @@ public TdrManifestQuartzJob(
     this.mapper = mapper;
     this.snapshotSupportFactory = snapshotSupportFactory;
     this.samDao = samDao;
+    this.dataImportProperties = dataImportProperties;
```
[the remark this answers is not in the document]
ashanhol: thats a great point, made that change

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -398,4 +408,24 @@ protected void linkSnapshots(Set<UUID> snapshotIds, WorkspaceId workspaceId) {
     // TODO AJ-1673: don't use the env-var workspaceId here
     snapshotSupport.linkSnapshots(snapshotIds);
   }
+
+  /**
+   * Sync permissions with SAM if option is provided.
```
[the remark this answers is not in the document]
ashanhol: made more sense when i included the option check in the function

service/src/main/java/org/databiosphere/workspacedataservice/sam/SamAuthorizationDao.java
ashanhol: renamed to be resource_name

davidangb (approved):
thank you!!1!1

PR #680, opened 2024-03-25
[AJ-1687][AJ-1715] Switch to pull subscription for import status updates from Rawls

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1715

Currently, CWDS uses a push subscription to receive import status updates from Rawls. This works ok in live environments, where the push subscription is Terraformed. But in BEEs, there's no way to Terraform the subscription.

To work around that, this switches to using a pull subscription. When CWDS starts up, it creates the topic and subscription if they don't already exist.


service/src/main/java/org/databiosphere/workspacedataservice/pubsub/JobStatusUpdate.java
```diff
@@ -27,6 +28,9 @@ public static JobStatusUpdate createFromPubSubMessage(PubSubMessage message) {
   }
 
   private static StatusEnum rawlsStatusToJobStatus(String rawlsStatus) {
```
jladieu, line 30: Since this method is expecting and accounting for `rawlsStatus` to be null, it'd be good to annotate the arg as `@Nullable` which signals to calling code that it's okay to send a `null` in, and to the code _in_ the method that it needs to be `null` cautious.

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/JobStatusUpdate.java
```diff
@@ -12,13 +13,13 @@ public JobStatusUpdate(UUID jobId, StatusEnum currentStatus, StatusEnum newStatu
...
+  public static JobStatusUpdate createFromPubSubMessage(PubsubMessage message) {
     try {
-      Map<String, String> attributes = message.attributes();
-      UUID jobId = UUID.fromString(attributes.get("import_id"));
-      StatusEnum newStatus = rawlsStatusToJobStatus(attributes.get("new_status"));
-      StatusEnum currentStatus = rawlsStatusToJobStatus(attributes.get("current_status"));
-      String errorMessage = attributes.get("error_message");
+      Map<String, String> attributes = message.getAttributesMap();
+      UUID jobId = UUID.fromString(attributes.get("importId"));
+      StatusEnum newStatus = rawlsStatusToJobStatus(attributes.get("newStatus"));
```
jladieu, line 20: Is there a specific reason to camelize these keys?  I'm asking because I think this causes the JSON format to drift further from that of the import-service.  If we wanted to build code or just do a visual diff between messages it'll be harder to do so the more divergent the messages are.

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/PubSubClient.java
```diff
@@ -0,0 +1,46 @@
...
+
+import com.google.cloud.spring.pubsub.PubSubAdmin;
+import com.google.protobuf.Duration;
+import com.google.pubsub.v1.ExpirationPolicy;
+import com.google.pubsub.v1.Subscription;
+import com.google.pubsub.v1.Topic;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PubSubClient {
```
jladieu, line 11: nit: not needed for this PR, but since this class is currently only used by `getInboundChannelAdapter`, you could also just inline this logic entirely into `PubSubConfig` as a private static method and not bother exposing `PubSubClient` as a class or a bean.

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
```diff
@@ -59,12 +55,12 @@ public List<GenericJobServerModel> getJobsForCollection(
...
   /**
    * Process a job status update from Rawls received via PubSub. This method is only used in the
-   * control plane (it's exposed through the control plane only PubSubController).
+   * control plane.
    */
   public void processStatusUpdate(JobStatusUpdate update) {
     UUID jobId = update.jobId();
     try {
-      GenericJobServerModel job = getJob(jobId);
+      GenericJobServerModel job = jobDao.getJob(jobId);
```
jladieu, line 63: hmm why call the `jobDao` directly here instead of using the `getJob` method?

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/JobStatusUpdate.java
davidangb: +1, this looks inconsistent with Import Service: https://github.com/broadinstitute/import-service/blob/a818658d0850818804235c5dc7f239c0de95c8cf/app/status.py#L59-L60

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/PubSubConfig.java
```diff
@@ -1,19 +1,77 @@
...
+      BasicAcknowledgeablePubsubMessage originalMessage =
+          springMessage
+              .getHeaders()
+              .get(GcpPubSubHeaders.ORIGINAL_MESSAGE, BasicAcknowledgeablePubsubMessage.class);
+
+      pubSubService.processPubSubMessage(originalMessage.getPubsubMessage());
+
+      originalMessage.ack();
+    };
+  }
```
davidangb, line 76: do we need `@EnableIntegration` on our `WorkspaceDataServiceApplication` to turn any of this on?

davidangb (commented):
I'm really glad the spring integration stuff worked out so well (or at least appears to)! It definitely simplifies our codebase.

I have some comments inline. Also some broader questions for followup at some point:
* should we also change our *outbound* pubsub messages to use a channel adapter?
* is there any contention with our multiple k8s replicas both pulling from the subscription?
* are there any [config properties](https://googlecloudplatform.github.io/spring-cloud-gcp/5.1.0/reference/html/appendix.html) for the pubsub channel adapter that make sense to tweak?
* do you think the [pubsub emulator](https://cloud.google.com/pubsub/docs/emulator) could be useful for integration tests?
* can we get observability metrics for any of this? It looks like by default we _can_ easily turn these on: https://docs.spring.io/spring-integration/reference/metrics.html#micrometer-observation

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/PubSubClient.java
```diff
@@ -0,0 +1,46 @@
...
+          admin.createSubscription(
+              Subscription.newBuilder()
+                  .setName(subscriptionName)
+                  .setTopic(topic.getName())
+                  .setExpirationPolicy(ExpirationPolicy.getDefaultInstance())
+                  .setMessageRetentionDuration(Duration.newBuilder().setSeconds(86_400)));
+    }
+
+    return subscription;
+  }
```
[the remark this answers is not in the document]
nawatts: That's fine with me for the topic, but it does seem like unnecessary complexity. The subscription, however, are not currently Terraformed. As is, in all environments, CWDS will need to create the subscription when it's first deployed to the environment.

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/PubSubConfig.java
```diff
@@ -1,19 +1,77 @@
...
+
+  @Bean
+  PubSubClient getPubSubClient(PubSubAdmin pubSubAdmin) {
+    return new PubSubClient(pubSubAdmin);
+  }
+
+  @Bean
+  public MessageChannel getImportStatusUpdateMessageChannel() {
+    return new DirectChannel();
+  }
```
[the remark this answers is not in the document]
nawatts: 👍 IntelliJ complained if there was no Bean matching the name in Qualifier / ServiceActivator, so I think we may already have that protection.

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/JobStatusUpdate.java
nawatts: Yeah, I was also confused here because I copied the implementation from import service and then got errors when the incoming messages weren't in that format. As far as I can tell though, it looks like Rawls is sending camel case keys.

https://github.com/broadinstitute/rawls/blob/da8c749603ba1d53a9956fd09d076f40b751072c/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala#L94-L98

https://github.com/broadinstitute/rawls/blob/da8c749603ba1d53a9956fd09d076f40b751072c/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala#L424-L444

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/PubSubClient.java
nawatts: I had it that way at first, but `PubSubConfig` was growing large. I'm thinking to absorb ImportPubSub (which currently only has a `publishSync` method), into this class.

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/JobStatusUpdate.java
nawatts: Aha! Found where import service converts from camel case to snake case, https://github.com/broadinstitute/import-service/blob/a818658d0850818804235c5dc7f239c0de95c8cf/app/server/routes.py#L109-L110

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/JobStatusUpdate.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/680/commits/bc826c05efc0396ca61ba6b8baf8d8147dd12ced.

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/PubSubConfig.java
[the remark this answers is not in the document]
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/680/commits/d223fcfe04191aa6f2c033c67a218ab56b4cd100.

service/src/main/java/org/databiosphere/workspacedataservice/service/PubSubService.java
```diff
@@ -0,0 +1,49 @@
...
+    this.jobService = jobService;
+  }
+
+  public void processPubSubMessage(PubsubMessage message) {
+    LOGGER.info(
+        "Received PubSub message: {} {}",
+        message.getMessageId(),
+        message.getAttributesMap().entrySet().stream()
+            .map(e -> "%s: %s".formatted(e.getKey(), e.getValue()))
+            .collect(Collectors.joining(", ")));
```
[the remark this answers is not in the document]
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/680/commits/a126c5587f53f8b1c2f314556eac3c5c7b0c2700.

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/PubSubConfig.java
```diff
@@ -1,19 +1,77 @@
...
+
+  @Bean
+  public PubSubInboundChannelAdapter getInboundChannelAdapter(
+      PubSubClient pubSubClient,
+      PubSubTemplate pubSubTemplate,
+      @Qualifier("getImportStatusUpdateMessageChannel") MessageChannel messageChannel,
+      @Value("${twds.data-import.status-updates.topic}") String topicName,
+      @Value("${twds.data-import.status-updates.subscription}") String subscriptionName) {
```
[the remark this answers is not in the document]
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/680/commits/1e26627a8af997288b4c4a9cc936e6663edd3ede.

service/src/main/java/org/databiosphere/workspacedataservice/pubsub/PubSubClient.java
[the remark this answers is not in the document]
davidangb: I'm fine skipping the config property. If we find we somehow have problems with this in live environments, we can always revisit at that point.

davidangb (approved):
Thank you for all this!

It looks like we get at least some Prometheus metrics for this out of the box; I saw them when running locally. It also looks like we get a new "pubSub" key in our /status API.

I'd still love to hear your opinions on some of the other post-PR questions, like whether we should rewrite our outbound messages as a channel adapter.

nawatts:
> should we also change our outbound pubsub messages to use a channel adapter?

This makes sense to me if for no other reason than consistency. I think this might also end up removing the ImportPubSub class and its `publishSync` method.

> is there any contention with our multiple k8s replicas both pulling from the subscription?

https://cloud.google.com/pubsub/docs/reference/rpc/google.pubsub.v1#streamingpullrequest references a `clientId`, but I haven't yet found what API (streaming or not) Spring Cloud GCP uses or a way to set this.

> are there any [config properties](https://googlecloudplatform.github.io/spring-cloud-gcp/5.1.0/reference/html/appendix.html) for the pubsub channel adapter that make sense to tweak?

I didn't look at them in detail, but given that our message handling is pretty simple, I think we could probably try out the defaults.

> do you think the [pubsub emulator](https://cloud.google.com/pubsub/docs/emulator) could be useful for integration tests?

Maybe. The current PubSub publishing doesn't have any integration tests either. Something, something, test containers?

> can we get observability metrics for any of this? It looks like by default we can easily turn these on: https://docs.spring.io/spring-integration/reference/metrics.html#micrometer-observation

👍 

> should we - in cWDS - add a pubsub connection to our /status health? https://cloud.spring.io/spring-cloud-static/spring-cloud-gcp/current/reference/html/#cloud-pubsub-health-indicator

It looks like Spring Cloud GCP automatically plugs something into Actuator: https://cloud.spring.io/spring-cloud-static/spring-cloud-gcp/current/reference/html/#cloud-pubsub-health-indicator

davidangb:
oh also it looks like there are a few Sonar findings :(

nawatts:
> oh also it looks like there are a few Sonar findings :(

Resolved the one about the test visibility. The log formatting one seems not worth it. And the nested try is necessary with the current logging.

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
nawatts: Added a comment about this in https://github.com/DataBiosphere/terra-workspace-data-service/pull/680/commits/5edbed13ce175883260c9bc0a3b0e0625741eee6.

jladieu (approved):

PR #681, opened 2024-03-25
AJ-1716: Attempt to disentangle `ImportDetails`

jladieu:
Followup refactoring to https://github.com/DataBiosphere/terra-workspace-data-service/pull/676 for ticket: [AJ-1716](https://broadworkbench.atlassian.net/browse/AJ-1716)

Two of the four attributes of `ImportDetails` were `@Nullable` to support its use being threaded through `WdsRecordSink`.  However, `WdsRecordSink` only needs the `CollectionId`, so support that explicitly with a dedicated factory method.

[AJ-1716]: https://broadworkbench.atlassian.net/browse/AJ-1716?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
```diff
@@ -3,24 +3,14 @@
 import java.util.UUID;
 import java.util.function.Supplier;
 import org.databiosphere.workspacedataservice.recordsink.RawlsAttributePrefixer.PrefixStrategy;
-import org.springframework.lang.Nullable;
+import org.databiosphere.workspacedataservice.shared.model.CollectionId;
 
 public record ImportDetails(
-    @Nullable UUID jobId,
-    @Nullable Supplier<String> userEmailSupplier,
-    UUID collectionId,
```
jladieu, line 11: Sorry, I firmed up the type of `CollectionId` while I was doing this and kind of regret the amount of diff it created.  I can pull this back to make this PR more focused for review or merge if preferred.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -170,7 +171,9 @@ static String getBlobName(UUID jobId) {
 
   private void publishToPubSub(ImportDetails importDetails) {
     UUID jobId = requireNonNull(importDetails.jobId());
-    UUID workspaceId = importDetails.collectionId();
+    // TODO: do a lookup of CollectionId to get WorkspaceId rather than assuming it's the same,
+    //   or pass the WorkspaceId from where ImportDetails is created
+    WorkspaceId workspaceId = WorkspaceId.of(importDetails.collectionId().id());
```
davidangb, line 176: You could conceivably do this translation in `RawlsRecordSinkFactory.buildRecordSink()` too

davidangb (approved):
I like it, thank you!

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
jladieu: Good call!  

...I started to pull this comment up there, and then looked a little further and learned that the two places that need `ImportDetails` are already doing a lookup of `WorkspaceId` based on `CollectionId` within a couple lines of code, so we already have it in hand.

However, while threading that down through, I also realized that WDS wants a `CollectionId`, but Rawls wants a `WorkspaceId`; we were relying on a mixed usage of one field to act in both roles, depending on the downstream sink.  Rather than have this mixed usage, I just set _both_ since we already have both in hand.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -115,13 +115,11 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
     String authToken = getJobDataString(jobDataMap, ARG_TOKEN);
     Supplier<String> userEmailSupplier = () -> samDao.getUserEmail(BearerToken.of(authToken));
 
-    // TDR import is interested in the collectionId (not the workspaceId)
```
jladieu, line 118: I think this comment wasn't exactly true...I think it was that _`WDS`_-based record sink was interested in `CollectionId` rather than `WorkspaceId`.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/WdsRecordSinkFactory.java
```diff
@@ -23,9 +23,6 @@ public WdsRecordSinkFactory(
     this.dataTypeInferer = dataTypeInferer;
   }
 
-  // TODO(AJ-1589): make prefix assignment dynamic. However, of note: the prefix is currently
```
jladieu, line 26: Just saw this leftover comment which has already been addressed by the Prefixer work.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetails.java
jladieu: This turned into adding both `WorkspaceId` and `CollectionId` to the `ImportDetails`.  Thankfully, in both production code paths where this gets created, they were both in hand.

davidangb (approved):
nice, thank you!

PR #682, opened 2024-03-26
AJ-1587: Clarify job metrics with unit test.

jladieu:
[AJ-1587](https://broadworkbench.atlassian.net/browse/AJ-1587): While trying to understand the job counts dashboard, I was confused by what metrics I could be able to count on, and wanted a way to document them.  I thought a unit test to capture expectations of what appears in the Prometheus scrape for this would be a helpful start.

[AJ-1587]: https://broadworkbench.atlassian.net/browse/AJ-1587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobControlPlaneE2ETest.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobControlPlaneE2ETest.java
@@ -1,5 +1,6 @@
 package org.databiosphere.workspacedataservice.dataimport.pfb;
 
+import static java.util.Arrays.stream;
 import static java.util.stream.Collectors.counting;
 import static java.util.stream.Collectors.groupingBy;
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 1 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
```diff
@@ -63,7 +63,6 @@ public void execute(JobExecutionContext context) throws org.quartz.JobExecutionE
     try {
       // mark this job as running
       getJobDao().running(jobId);
-      observation.event(Observation.Event.of("job.running"));
```
jladieu, line 66: This was producing a metric named: `wds_job_execute_job_running_total` which was redundant with `wds_job_execute_seconds_count`; given the somewhat confusing set of metrics already made present by `Observation` automatically, I thought it'd be preferable to eliminate the duplication and slim down the list of metrics we have to think about.

davidangb (approved):
:+1:, once tests actually pass

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobControlPlaneE2ETest.java
```diff
@@ -253,6 +305,79 @@ void pubSubRequestsUserEmail() throws JobExecutionException, IOException {
...
+        metrics, "wds_job_execute_job_running_total", "1.0", Map.of("jobType", "PfbQuartzJob"));
+
+    // (counter) of job.execute events
+    assertMetric(metrics, "wds_job_execute_seconds_count", "1.0", successTags);
+
+    // (summary) sum of all job durations, and when divided by count, can get the average
+    assertMetric(metrics, "wds_job_execute_seconds_sum", ".*", successTags);
+
+    // (gauge) maximum duration tracked
+    assertMetric(metrics, "wds_job_execute_seconds_max", ".*", successTags);
```
ashanhol, line 341: A thought- if you think we'll be asserting counter/summary/gauge for other things in wds/cwds in the future (not just pfbs) I wonder if theres a way to abstract this so we can make it easier to check test observations for other things. 

but if you dont feel like thinking about that, maybe use a variable so you dont have to use the string literals for both success/fail tests.

ashanhol (approved):

jladieu:
I think the test failure is `can-i-deploy` which is expected; clicking through there, it's ok to deploy to dev, but staging & prod are still waiting on the updated version of sam.  Thus, I think it's okay to merge this as it's only destined for dev out the gate anyway.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobControlPlaneE2ETest.java
jladieu: I'm so sorry, I scrolled right past this comment and merged without seeing it.

I think for the sake of this test, having the duplicated string literals is fine since it's only two occurrences in relatively the same location.  Owing to the goal of having these tests provide a form of documentation, I'd rather have the expected prometheus strings jump off the page and be clearly visible over requiring you to have to click through to the constant definition to look them up.  That said, if the duplication starts to proliferate and spread over multiple files, I'd see that as a signal to pull back and de-dupe somehow.  For now, I'd like to leave as is though.

PR #689, opened 2024-03-28
AJ-1573: cwds smoke test framework

davidangb:
This adds smoke tests for cWDS.

Based on discussion in the [proposal](https://docs.google.com/document/d/1xyLa62UZ_uzRvT67VWBuL3z2NZ1txYns_O0EqQmEGxo/edit#heading=h.xseqj4pczzeh), cWDS smoke tests are copied from [Sam's smoke tests](https://github.com/broadinstitute/sam/tree/develop/smoke_test) and therefore are written in Python.

Tactically, I copied the `smoke_test` directory from the Sam repo to here, renamed "sam" to "cwds" in a lot of places, and modified the individual tests to pass.

This is a first spike, with the intent of getting peer review and threading tests through. Remaining work for follow-on PRs/activities:
- [ ] modify the authenticated tests to do something useful. Right now they pass, but they aren't validating anything that makes sense. I'd like to commit them as-is to allow others the opportunity to collab on these!
- [ ] create a `cwds-smoke-tests.yaml` GHA workflow in terra-github-workflows, modeled after https://github.com/broadinstitute/terra-github-workflows/blob/main/.github/workflows/sam-smoke-tests.yaml
- [ ] create Beehive deploy hooks to call the new `cwds-smoke-tests.yaml` GHA workflow


jladieu (approved):
Awesome!

smoke_test/README.md
```diff
@@ -0,0 +1,74 @@
...
+```pip install -r requirements.txt```
+
+## Run
+
+The smoke tests have 2 different modes that they can run in: authenticated or unauthenticated. The
+mode will be automatically selected based on the arguments you pass to `smoke_test.py`.
+
+To run the _unauthenticated_ smoke tests:
+
+```python smoke_test.py {CWDS_HOST}```
```
calypsomatic, line 37: Just clarifying - these two lines are the format followed by an example?  Maybe a word or two explaining that for people like me who get confused easily?

smoke_test/README.md
```diff
@@ -0,0 +1,74 @@
...
+## Requirements
+
+Python 3.10.3 or higher
+
+## Setup
+
+Recommended, but optional, create and active a Python virtual env
+
+```
+python - m venv venv
```
calypsomatic, line 22: There's an extra space between "-" and "m" that makes this not work if you copy paste

smoke_test/README.md
```diff
@@ -0,0 +1,74 @@
...
+source venv / bin / activate
+```
+
+You will need to install required pip libraries:
+
+```pip install -r requirements.txt```
```
calypsomatic, line 28: You need to be in the `smoke_test` directory for this to work, I was able to figure that out so maybe it doesn't need to be written but maybe you could anyway!

smoke_test/smoke_test.py
```diff
@@ -0,0 +1,103 @@
...
+    args = parser.parse_args()
+
+    # Need to pop off sys.argv values to avoid messing with args passed to unittest.main()
+    for _ in range(len(sys.argv[1:])):
+      sys.argv.pop()
+
+    main(args)
+    sys.exit(0)
+
+  except Exception as e:
```
nawatts, line 101: Do we need the try/catch here? Isn't print the exception and exit with an error code the default behavior?

smoke_test/smoke_test.py
```diff
@@ -0,0 +1,103 @@
...
+      nargs='?',
+      default=None,
+      help="Optional. If present, will test additional authenticated endpoints using the specified token"
+    )
+
+    args = parser.parse_args()
+
+    # Need to pop off sys.argv values to avoid messing with args passed to unittest.main()
+    for _ in range(len(sys.argv[1:])):
+      sys.argv.pop()
```
nawatts, line 96: If I'm understanding this correctly...

```suggestion
    del sys.argv[1:]
```

https://docs.python.org/3/tutorial/datastructures.html#the-del-statement

smoke_test/smoke_test.py
```diff
@@ -0,0 +1,103 @@
...
+def verify_user_token(user_token: str) -> bool:
+  response = requests.get(
+    f"https://www.googleapis.com/oauth2/v1/tokeninfo?access_token={user_token}")
+  assert response.status_code == 200, "User Token is no longer valid.  Please generate a new token and try again."
+
+
+if __name__ == "__main__":
+  try:
+    parser = argparse.ArgumentParser(
+      description=DESCRIPTION,
```
nawatts, line 67: If you want to get fancy, you could make the contents of `DESCRIPTION` the module docstring and still use it here with:

```suggestion
      description=__doc__,
```

smoke_test/README.md
```diff
@@ -0,0 +1,74 @@
...
+_These tests are valid for cWDS in the control plane. They have not been evaluated against WDS as an
+app in the data plane._
+
+## Requirements
+
+Python 3.10.3 or higher
+
+## Setup
+
+Recommended, but optional, create and active a Python virtual env
```
nawatts, line 19:
```suggestion
Recommended, but optional, create and activate a Python virtual env
```

nawatts (approved):
👍, though I do wish it were pytest instead of unittest

smoke_test/README.md
davidangb: updated, hope this is better

smoke_test/README.md
davidangb: weird - this, and the extra spaces in `venv/bin/activate` - were IntelliJ auto-formatting that I failed to notice! Fixed.

smoke_test/README.md
davidangb: added a note to the Setup section

smoke_test/smoke_test.py
davidangb: it also looks extraneous to me; removed

calypsomatic (approved):

smoke_test/smoke_test.py
davidangb: done, good call

PR #690, opened 2024-03-29
[AJ-1742] Use dedicated subscription for local CWDS

nawatts:
It occurred to me yesterday, we shouldn't have local CWDS' pulling messages from the same subscription as dev, since that could prevent dev from receiving messages for imports in the dev environment and updating its database accordingly.

It's probably also worthwhile to remove the default configuration for the subscription name to help avoid this, but that will require a change to terra-helmfile.


davidangb (approved):

nawatts:
> It's probably also worthwhile to remove the default configuration for the subscription name to help avoid this, but that will require a change to terra-helmfile.

Went ahead with this.

https://github.com/broadinstitute/terra-helmfile/pull/5343

davidangb (approved):

PR #692, opened 2024-03-29
AJ-1573: smoke tests add workspace_id CLI argument; use it in the job-listing test

davidangb:
* accept a workspace id on the command-line when running smoke tests
* use the workspace id in the job-listing test, so that test does something useful

In support of this, I have created a "cwds-smoketest-do-not-delete" workspace in the dev env to use for smoke tests. Before this rolls out to staging and prod I will create workspaces there too.

### DIFF
--- a/smoke_test/smoke_test.py
+++ b/smoke_test/smoke_test.py
@@ -27,13 +27,14 @@ def gather_tests(is_authenticated: bool = False) -> TestSuite:
   suite.addTests(version_tests)
 
   if is_authenticated:
+    print("user_token and workspace_id both provided.  Running additional authenticated tests.")
     user_info_tests = unittest.defaultTestLoader.loadTestsFromTestCase(CwdsJobListingTests)
     resource_types_tests = unittest.defaultTestLoader.loadTestsFromTestCase(CwdsJobStatusTests)
 
     suite.addTests(user_info_tests)
     suite.addTests(resource_types_tests)
   else:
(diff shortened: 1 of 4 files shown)


jladieu (approved):

smoke_test/smoke_test.py
```diff
@@ -27,13 +27,14 @@ def gather_tests(is_authenticated: bool = False) -> TestSuite:
...
   if is_authenticated:
+    print("user_token and workspace_id both provided.  Running additional authenticated tests.")
     user_info_tests = unittest.defaultTestLoader.loadTestsFromTestCase(CwdsJobListingTests)
     resource_types_tests = unittest.defaultTestLoader.loadTestsFromTestCase(CwdsJobStatusTests)
 
     suite.addTests(user_info_tests)
     suite.addTests(resource_types_tests)
   else:
-    print("No User Token provided.  Skipping authenticated tests.")
+    print("user_token and/or workspace_id not provided.  Skipping authenticated tests.")
```
jladieu, line 37:
```suggestion
    print("neither user_token nor workspace_id provided.  Skipping authenticated tests.")
```

davidangb: this isn't actually the case - since both are required, if the caller supplies either but not both, it will still skip the authenticated tests

nawatts (approved):
It looks like we're accumulating options. It might be nice to switch to named options (`--workspace-id=...`) instead of positional arguments.

PR #697, opened 2024-04-01
AJ-1726 Add measurement to `processStatusUpdate`

jladieu:
For [AJ-1726](https://broadworkbench.atlassian.net/browse/AJ-1726):  Add measurement to `processStatusUpdate`.

Commits have been intentionally broken apart for easier review if needed.

Introduces two new measures:
- `wds.job.update` is the time it takes to update a job in the database
- `wds.job.elapsed` is the time elapsed since the job was created

Also changes the `jobType` label values of the existing `wds.job.execute` measurement to use the value from `JobKey` so that all the `wds.job` metrics are keyed the same way.

[AJ-1726]: https://broadworkbench.atlassian.net/browse/AJ-1726?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/QuartzSchedulerDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/QuartzSchedulerDao.java
@@ -3,7 +3,6 @@ package org.databiosphere.workspacedataservice.dao;
 import static org.quartz.TriggerBuilder.newTrigger;
 
 import org.databiosphere.workspacedataservice.shared.model.Schedulable;
-import org.quartz.JobBuilder;
 import org.quartz.JobDetail;
 import org.quartz.JobKey;
 import org.quartz.Scheduler;
@@ -23,17 +22,8 @@ public class QuartzSchedulerDao implements SchedulerDao {
   }
 
   public void schedule(Schedulable schedulable) {
-    JobKey jobKey = new JobKey(schedulable.getId(), schedulable.getGroup());
-
-    JobDetail jobDetail =
-        JobBuilder.newJob()
-            .ofType(schedulable.getImplementation())
-            .withIdentity(jobKey)
-            .setJobData(schedulable.getArgumentsAsJobDataMap())
-            .storeDurably(false) // delete from the quartz table after the job finishes
-            .withDescription(schedulable.getDescription())
-            .build();
-
+    JobDetail jobDetail = schedulable.getJobDetail();
+    JobKey jobKey = jobDetail.getKey();
     // tell Quartz to run the job: run only once, start immediately
     Trigger trigger = newTrigger().forJob(jobKey).startNow().build();
     try {
--- a/service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
@@ -14,6 +14,7 @@ import org.databiosphere.workspacedataservice.service.MDCServletRequestListener;
 import org.quartz.Job;
 import org.quartz.JobDataMap;
 import org.quartz.JobExecutionContext;
+import org.quartz.JobKey;
 import org.slf4j.MDC;
 
 /**
@@ -50,7 +51,8 @@ public abstract class QuartzJob implements Job {
   @Override
   public void execute(JobExecutionContext context) throws org.quartz.JobExecutionException {
     // retrieve jobId
-    UUID jobId = UUID.fromString(context.getJobDetail().getKey().getName());
+    JobKey jobKey = context.getJobDetail().getKey();
+    UUID jobId = UUID.fromString(jobKey.getName());
 
     // (try to) set the MDC request id based on the originating thread
     propagateMdc(context);
@@ -58,7 +60,7 @@ public abstract class QuartzJob implements Job {
     Observation observation =
         Observation.start("wds.job.execute", observationRegistry)
             .contextualName("job-execution")
-            .lowCardinalityKeyValue("jobType", getClass().getSimpleName())
+            .lowCardinalityKeyValue("jobType", jobKey.getGroup())
             .highCardinalityKeyValue("jobId", jobId.toString());
     try {
       // mark this job as running
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
@@ -4,6 +4,7 @@ import static org.databiosphere.workspacedataservice.shared.model.Schedulable.AR
 import static org.databiosphere.workspacedataservice.shared.model.Schedulable.ARG_TOKEN;
 import static org.databiosphere.workspacedataservice.shared.model.Schedulable.ARG_URL;
 
+import com.google.common.annotations.VisibleForTesting;
 import java.io.Serializable;
 import java.util.HashMap;
 import java.util.Map;
@@ -137,14 +138,15 @@ public class ImportService {
     }
   }
 
-  protected Schedulable createSchedulable(
+  @VisibleForTesting
+  public static Schedulable createSchedulable(
       ImportRequestServerModel.TypeEnum importType,
       UUID jobId,
       Map<String, Serializable> arguments) {
     return switch (importType) {
       case TDRMANIFEST -> new TdrManifestSchedulable(
           jobId.toString(), "TDR manifest import", arguments);
-      case PFB -> new PfbSchedulable(jobId.toString(), "TODO: PFB import", arguments);
+      case PFB -> new PfbSchedulable(jobId.toString(), "PFB import", arguments);
     };
   }
 }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
@@ -1,8 +1,16 @@
 package org.databiosphere.workspacedataservice.service;
 
+import static java.time.OffsetDateTime.now;
 import static java.util.Objects.requireNonNullElse;
 
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.observation.Observation;
+import io.micrometer.observation.ObservationRegistry;
+import java.time.Duration;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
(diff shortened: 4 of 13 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
```diff
@@ -209,7 +209,7 @@ void observationLogsFailure() throws org.quartz.JobExecutionException {
         .hasObservationWithNameEqualTo("wds.job.execute")
         .that()
         .hasHighCardinalityKeyValue("jobId", jobUuid)
-        .hasLowCardinalityKeyValue("jobType", "TestableQuartzJob")
+        .hasLowCardinalityKeyValue("jobType", "jobGroup")
```
nawatts, line 212: Super nit: I think a different name would make it more obvious that this is a placeholder value for tests.

```suggestion
        .hasLowCardinalityKeyValue("jobType", "testJobType")
```

nawatts (changes requested):
I think `JobDao` would be a preferable place for these metrics than `JobService.processStatusUpdate`. The issue with `JobService.processStatusUpdate` is that it _only_ handles status updates sent from Rawls via PubSub. So it only applies to a subset of status updates on the control plane.

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceTest.java
```diff
@@ -0,0 +1,280 @@
...
+import org.springframework.test.context.ActiveProfiles;
+
+@ActiveProfiles(
+    profiles = {
+      "data-plane", // need _a_ profile even if this test only contains tests for common behavior
+    })
+@DirtiesContext
+@SpringBootTest(properties = {"twds.instance.workspace-id=f01dab1e-0000-1111-2222-000011112222"})
+@WithTestObservationRegistry
+class JobServiceTest extends JobServiceTestBase {
```
jladieu, line 47: The `processStatusUpdate` tests were moved into this test file so all those tests existed previously.  New tests start at `processJobStatusUpdateMeasuresDurationSinceCreationOnSuccess` and continue to the end of the file.

As noted in one of the refactoring commits, the stubbing logic for these tests was adjusted slightly to make it easier to add more customization to the stubbed job.

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
```diff
@@ -50,15 +51,16 @@ protected QuartzJob(
...
+    UUID jobId = UUID.fromString(jobKey.getName());
 
     // (try to) set the MDC request id based on the originating thread
     propagateMdc(context);
 
     Observation observation =
         Observation.start("wds.job.execute", observationRegistry)
             .contextualName("job-execution")
-            .lowCardinalityKeyValue("jobType", getClass().getSimpleName())
+            .lowCardinalityKeyValue("jobType", jobKey.getGroup())
```
jladieu, line 63: This is a key change which caused a new set of labels to get used for `jobType`.  I'm doing this here because later on at the end of the job processing (during `processStatusUpdate`) we no longer have the class name in hand to use to label the metrics.  I think it's important to have the metrics labeled consistently, so wanted to use something that was accessible in both places.  The `jobInput` stores this information and it's available later on.

service/src/main/java/org/databiosphere/workspacedataservice/dao/QuartzSchedulerDao.java
```diff
@@ -23,17 +22,8 @@ public QuartzSchedulerDao(Scheduler scheduler) {
   }
 
   public void schedule(Schedulable schedulable) {
-    JobKey jobKey = new JobKey(schedulable.getId(), schedulable.getGroup());
-
-    JobDetail jobDetail =
```
jladieu, line 28: Pushing this logic into `Schedulable` makes it easier to reuse in tests and reduce the drift of test code from production code.

jladieu:
>I think JobDao would be a preferable place for these metrics than JobService.processStatusUpdate. The issue with JobService.processStatusUpdate is that it only handles status updates sent from Rawls via PubSub. So it only applies to a subset of status updates on the control plane.

Can you say more?  Are there other mechanisms where we update status in control plane?  I do ideally want to catch them all, but I also think pushing this down a layer into `JobDao` may bring with it some tradeoffs I'd want to think about more.

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
ashanhol: Where does a job key's group get set? Will this be a unique value or just a generic thing for all jobs?

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
jladieu: Fixed with 9f7851bc71801c8d27803e5ffd418fb5f1c0e785

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
```diff
@@ -137,14 +138,15 @@ private Optional<String> safeGetMdcId(UUID jobId) {
...
+  @VisibleForTesting
+  public static Schedulable createSchedulable(
       ImportRequestServerModel.TypeEnum importType,
       UUID jobId,
       Map<String, Serializable> arguments) {
     return switch (importType) {
       case TDRMANIFEST -> new TdrManifestSchedulable(
           jobId.toString(), "TDR manifest import", arguments);
-      case PFB -> new PfbSchedulable(jobId.toString(), "TODO: PFB import", arguments);
+      case PFB -> new PfbSchedulable(jobId.toString(), "PFB import", arguments);
```
ashanhol, line 149: are pfb imports ready to go? not sure if this was something that should have been removed and wasn't, it's kind of weird to take out a todo for this on a metrics PR otherwise

nawatts:
The status updates from Rawls would only be transitions from Running to either Succeeded or Errored. There are several other states.
https://github.com/DataBiosphere/terra-workspace-data-service/blob/98e2b157631ade98ae40805b4cf4ad59525e171e/service/src/main/java/org/databiosphere/workspacedataservice/generated/GenericJobServerModel.java#L76-L88

But more than that, what about the data plane?

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/Schedulable.java
```diff
@@ -36,29 +40,38 @@ public Schedulable(
...
   }
 
-  public JobDataMap getArgumentsAsJobDataMap() {
-    JobDataMap jobDataMap = new JobDataMap();
-    jobDataMap.putAll(arguments);
-    return jobDataMap;
+  public JobDetail getJobDetail() {
+    return JobBuilder.newJob()
+        .ofType(getImplementation())
+        .withIdentity(new JobKey(id, group))
```
jladieu, line 71: @ashanhol this is where the `group` gets set into the JobKey which gets used for the `jobType` metric.  The group is passed in as the first arg of creating the `Schedulable` in the subclasses: [PfbSchedulable](https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbSchedulable.java#L12) and [TdrManifestSchedulable](https://github.com/DataBiosphere/terra-workspace-data-service/blob/98e2b157631ade98ae40805b4cf4ad59525e171e/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestSchedulable.java#L13)

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
jladieu: It's a couple steps of indirection, I added a comment over where the group gets set: https://github.com/DataBiosphere/terra-workspace-data-service/pull/697/files#r1546779365 and that breadcrumbs to where it _actually_ gets set to either `TDRMANIFEST` or `PFB`

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
```diff
@@ -87,8 +115,47 @@ public void processStatusUpdate(JobStatusUpdate update) {
...
             "Unexpected status from Rawls for job %s: %s".formatted(jobId, newStatus));
       }
-    } catch (EmptyResultDataAccessException e) {
-      throw new MissingObjectException("Job");
+    } catch (Exception e) {
+      observation.error(e);
+      if (e instanceof EmptyResultDataAccessException) {
+        throw new MissingObjectException("Job");
+      }
+      throw e;
```
ashanhol, line 123: what other exceptions would we expect? could you leave a comment if there's a new scenario to anticipate?

davidangb:
Consider for instance an import job with some invalid input, which fails early. That job will move to failed state via `getJobDao().fail()`, here:

https://github.com/DataBiosphere/terra-workspace-data-service/blob/98e2b157631ade98ae40805b4cf4ad59525e171e/service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java#L87

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
jladieu: I assumed its removal was just something we missed.

Looking at git history, I think this TODO was put in a long time ago (sixish months) when @davidangb was first introducing the quartz job stuff.  It strongly predates the PFB work which is mostly wrapped at this point.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbTestUtils.java
```diff
@@ -160,21 +162,20 @@ public static JobExecutionContext stubJobContext(
...
-                    resource.getURL().toString(),
-                    ARG_COLLECTION,
-                    collectionId.toString(),
-                    ARG_TDR_SYNC_PERMISSION,
-                    shouldPermissionSync)));
-
-    JobDetailImpl jobDetail = new JobDetailImpl();
-    jobDetail.setKey(new JobKey(jobId.toString(), "bar"));
+
+    var schedulable =
```
ashanhol, line 166: do we use `var`s? Everything's been strongly typed that I've seen so I was under the impression this teams coding standards shied away from vars (but I could have just assumed wrong)

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/Schedulable.java
ashanhol: Thank you!

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceTest.java
```diff
@@ -0,0 +1,280 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.dao.EmptyResultDataAccessException;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.ActiveProfiles;
+
+@ActiveProfiles(
+    profiles = {
+      "data-plane", // need _a_ profile even if this test only contains tests for common behavior
```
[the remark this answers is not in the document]
jladieu: Fixed in 643d2f44f4d5eccab0eaf5840a7439afc2d5bb5c

service/src/main/java/org/databiosphere/workspacedataservice/service/JobService.java
jladieu: Added an explanatory comment in f7b4fcae8a09f4d167416cdd8d9d9d6e74361262

I wanted to make sure that any exception thrown out of the wrapped logic would correctly mark the observation with the error, including the exceptions thrown explicitly in the try block.  I didn't have any other categories of exception in mind besides those though.

jladieu:
Switching this back to draft while I move the observation logic down to the Dao.  I might end up with another PR to replace this one instead.

jladieu:
Replaced with https://github.com/DataBiosphere/terra-workspace-data-service/pull/699

PR #698, opened 2024-04-02
AJ-1773: more helpful error if user is read-only for imports

davidangb:
When performing a data import … 

If the user has _neither_ read nor write permission to the target workspace, we return a 404 that "Workspace does not exist or you do not have permission to see it". We choose 404 here to prevent any info leaks; we don't want to give malicious users the ability to test if a workspace exists by comparing 403 responses vs 404 responses. We return 404 for both workspace-doesn't-exist and user-doesn't-have-access.

BUT, if the user _does_ have read permission to the workspace, just not write permission, the 404 response is misleading. In this case, there's no information leakage: the user can already see that the workspace exists so we should return a more helpful response.

In this PR:
* if the user does not have write permission to the target workspace when creating a data import, make an additional request to Sam to ask if the user has read permission.
    * if the user has read permission, return a 403
    * if the user does not have read permission, return 404


nawatts (approved):

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceDataPlaneTest.java
```diff
@@ -73,15 +78,71 @@ void userHasAccess() {
     assertDoesNotThrow(() -> importService.createImport(collectionUuid, importRequest));
   }
 
+  /* collection exists, workspace matches env var, user has write permission */
```
calypsomatic, line 81: Nit: comment says "user has write permission" but the test says otherwise

calypsomatic (approved):

PR #699, opened 2024-04-02
AJ-1726 Add measurement to job updates

jladieu:
For [AJ-1726](https://broadworkbench.atlassian.net/browse/AJ-1726): 
Add measurement to job updates.

Commits have been intentionally broken apart for easier review if needed,
but the summary of functional changes is as follows:
* Switch `jobType` label to actually be the `JobType`, which is based
on the `JobType` enum rather than the subclass name.  For now, this
value is only `DATA_IMPORT`.
* Add a new label called `importType` which is populated based on the
`TypeEnum` import type, which will be either `PFB` or `TDRMANIFEST`.
* `wds.job.update` is the time it takes to do update a job in the database
* `wds.job.elapsed` is the time elapsed since the job was created

Note: these changes to metrics definition are intentionally done in one
PR to avoid multiple churns in the metrics.


service/src/main/java/org/databiosphere/workspacedataservice/shared/model/job/JobInput.java
```diff
@@ -18,4 +32,28 @@ class Empty implements JobInput {}
...
+   * This deserializer allows expressive testing of serialized JSON, no parts of the current
+   * production runtime require the Rawls JSON to be deserialized.
+   */
+  class Deserializer extends JsonDeserializer<JobInput> {
+    private static final Logger logger = LoggerFactory.getLogger(Deserializer.class);
+
+    @Override
+    public JobInput deserialize(JsonParser parser, DeserializationContext context)
+        throws IOException {
+      if (parser.currentToken() == JsonToken.START_OBJECT) {
```
jladieu, line 46: I needed to bite off some of AJ-1402 as part of figuring out the importType for metrics.

jladieu:
>[4 New issues](https://sonarcloud.io/project/issues?id=DataBiosphere_workspace-data-service&pullRequest=699&resolved=false&inNewCodePeriod=true)

I took care of 2 of these.

The two remaining are
- a pre-existing TODO I don't plan to address but seems relevant still
- a recommendation to use a constant but which I'd rather not address because using a constant would couple together two orthogonal terms; I don't want to silence it but also don't want to deal with it with this change

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
```diff
@@ -165,41 +185,50 @@ private GenericJobServerModel update(
...
-
-    // return the updated job
-    return getJob(jobId);
+    Observation observation =
+        Observation.createNotStarted("wds.job.update", observations)
+            .highCardinalityKeyValue("jobId", jobId.toString())
+            .lowCardinalityKeyValue("newStatus", status.toString());
+    return requireNonNull(
+        observation.observe(
+            () -> {
```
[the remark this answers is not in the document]
nawatts: Agreed. I think the most interesting thing is how long it takes the job to reach each status since it was created.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
```diff
@@ -269,9 +304,48 @@ public GenericJobServerModel mapRow(ResultSet rs, int rowNum) throws SQLExceptio
...
   }
+
+  private GenericJobServerModel measure(GenericJobServerModel job, Observation observation) {
+    applyJobSpecificTags(job, observation);
+
+    metrics
+        .timer("wds.job.elapsed", getTags(observation))
+        .record(
+            Duration.between(
+                job.getCreated(), Instant.ofEpochMilli(clock.wallTime()).atOffset(ZoneOffset.UTC)));
```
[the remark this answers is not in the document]
nawatts: Agreed that high precision doesn't seem necessary here. For imports into GCP workspaces, this will already be somewhat delayed (at least for some definition of the elapsed time) due to Rawls having to tell cWDS to update the job status via PubSub.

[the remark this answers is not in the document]
jladieu: >The Javadoc for clock.wallTime() says "Should not be used to determine durations". I'm not too worried about super-precise accuracy here, but should we use that method here? Could this be simplified to just Instant.now().atOffset(ZoneOffset.UTC)? Edit: I just noticed you mock Clock in tests; does removing use of Clock here help or hurt?

This clumsy date logic is the result of a bunch of struggles with trying to simulate "5 minutes in the future" which ended up pulling in the `Clock`.  I think the actual logic could be much, much simpler if I didn't want to actually test it.  I'll take another pass while addressing other feedback to see if I can at least improve it a bit.

I think I'd like to steer clear from option 2 though, simply because (_I think_) it'd make it nearly impossible to test.  While struggling with this, I actually wanted to go the opposite direction and pull the `now()` computation _up_ from postgres to introduce testability, but figured that'd also be way out of scope of this change and not necessarily a popular change.

service/src/test/java/org/databiosphere/workspacedataservice/controller/JobControllerTest.java
```diff
@@ -56,10 +62,21 @@ class JobControllerTest extends TestBase {
   @BeforeAll
   void beforeAll() {
     // push jobs into the db, one of different collectionId
-    Job<JobInput, JobResult> testJob1 = newJob(collectionId, JobType.DATA_IMPORT, JobInput.empty());
-    Job<JobInput, JobResult> testJob2 = newJob(collectionId, JobType.DATA_IMPORT, JobInput.empty());
+    Job<JobInput, JobResult> testJob1 =
```
jladieu, line 65: I added `JobInput` here as it is now deserialized during `jobDao.getDao` -- without a `JobInput` the deserialization logic violates the `@NonNullable` contract of `ImportJobInput` and throws an NPE.

I could either mark `ImportJobInput` as `@Nullable` then program defensively against that with null checks, or update all the tests that put these in the database to put realistic / non-nully values in for `DATA_IMPORT` jobs.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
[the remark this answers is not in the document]
jladieu: Got rid of the observation in https://github.com/DataBiosphere/terra-workspace-data-service/pull/699/commits/8483611044f3b1b629c8483688db94cea1389a98

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/job/JobInput.java
```diff
@@ -18,4 +30,26 @@ class Empty implements JobInput {}
   static Empty empty() {
     return new Empty();
   }
+
+  /**
+   * This deserializer allows expressive testing of serialized JSON, no parts of the current
+   * production runtime require the Rawls JSON to be deserialized.
+   */
```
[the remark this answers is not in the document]
jladieu: Removed as part of removing the deserializer in https://github.com/DataBiosphere/terra-workspace-data-service/pull/699/commits/20a126da4858eb724f0e6a070cc8277d5eb08951

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
```diff
@@ -269,9 +304,48 @@ public GenericJobServerModel mapRow(ResultSet rs, int rowNum) throws SQLExceptio
...
       job.errorMessage(rs.getString("error"));
       job.instanceId(rs.getObject("collection_id", UUID.class));
+      try {
+        job.input(mapper.readValue(rs.getString("input"), JobInput.class));
+      } catch (JsonProcessingException e) {
+        logger.error(
+            "Error deserializing input for job {}: {}; input will be empty.",
+            jobId,
+            e.getMessage());
+      }
```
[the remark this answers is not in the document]
jladieu: Love this, fixed in: https://github.com/DataBiosphere/terra-workspace-data-service/pull/699/commits/20a126da4858eb724f0e6a070cc8277d5eb08951

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
```diff
@@ -235,43 +249,87 @@ public List<GenericJobServerModel> getJobsForCollection(
...
-      return job;
+  private static Tags getTags(GenericJobServerModel job, StatusEnum status) {
+    Tags tags = Tags.of("jobType", job.getJobType().toString());
+    tags = tags.and("newStatus", status.toString());
+
+    Object jobInput = job.getInput();
+    if (jobInput != null
+        && job.getJobType() == DATA_IMPORT
+        && jobInput instanceof ImportJobInput inputJobInput
+    //    && inputJobInput.importType() != null
```
nawatts, line 326: Is this commented out code needed or should it be deleted?

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
```diff
@@ -235,43 +249,87 @@ public List<GenericJobServerModel> getJobsForCollection(
...
+            Duration.between(
+                job.getCreated(), Instant.ofEpochMilli(clock.wallTime()).atOffset(ZoneOffset.UTC)));
+    return job;
+  }
 
-      // TODO: AJ-1011 also return stacktrace, input, result.
-      return job;
+  private static Tags getTags(GenericJobServerModel job, StatusEnum status) {
+    Tags tags = Tags.of("jobType", job.getJobType().toString());
+    tags = tags.and("newStatus", status.toString());
```
nawatts, line 320: Can these two lines be combined?

```suggestion
    Tags tags = Tags.of("jobType", job.getJobType().toString(), "newStatus", status.toString());
```

service/src/main/java/org/databiosphere/workspacedataservice/config/ClockConfig.java
```diff
@@ -0,0 +1,18 @@
...
+
+@Configuration
+public class ClockConfig {
+
+  /**
+   * Expose {@link InstantSource#system())} as a bean to provide an override seam for tests that
+   * need to verify clock behavior.
+   */
+  @Bean
+  InstantSource instanceSource() {
```
nawatts, line 15: Was this intended to be `instantSource`?

```suggestion
  InstantSource instantSource() {
```

nawatts (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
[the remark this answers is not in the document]
jladieu: Fixed with bf5b86931245edeb765ab65d8672ee8d6690b314

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
jladieu: Ahh yes! This is the null protection that I wanted to avoid needing to implement by having `ImportJobInput` actually `@NonNullable`, you can see where I would have had to add it.  Removed by 5da625cb436f6bfda03959c516f3d0e5ea120a80

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
jladieu: Yes they can be, but I actively try to avoid `...varargs` invocations where you can lose track of key/value pairs.  I prefer to be explicit with the 2-arg constructors where it's clear which thing is key vs. value

This is why I always use `ImmutableMap.Builder` for building up many-valued `Map`s instead of `Map.of`

service/src/main/java/org/databiosphere/workspacedataservice/config/ClockConfig.java
jladieu: Yes, thanks for spotting!  Fixed in 87bdb44a5691af001b2470c0494d77b26a868848

service/src/main/java/org/databiosphere/workspacedataservice/config/ClockConfig.java
```diff
@@ -0,0 +1,18 @@
...
+
+@Configuration
+public class ClockConfig {
+
+  /**
+   * Expose {@link InstantSource#system())} as a bean to provide an override seam for tests that
+   * need to verify clock behavior.
+   */
+  @Bean
+  InstantSource instantSource() {
```
ashanhol, line 15: can you share the resource you used to determine InstanceSource was the right thing to use for a mock clock?

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
```diff
@@ -209,7 +218,8 @@ void observationLogsFailure() throws org.quartz.JobExecutionException {
         .hasObservationWithNameEqualTo("wds.job.execute")
         .that()
         .hasHighCardinalityKeyValue("jobId", jobUuid)
-        .hasLowCardinalityKeyValue("jobType", "TestableQuartzJob")
+        .hasLowCardinalityKeyValue("extraKey1", "extraValue1")
+        .hasLowCardinalityKeyValue("extraKey2", "extraValue2")
```
ashanhol, line 222: What is the value of adding and asserting the extrakeys/extravalues?

ashanhol (approved):

service/src/main/java/org/databiosphere/workspacedataservice/config/ClockConfig.java
jladieu: I landed on it organically.  I was fumbling around with the `MockClock` from micrometer and decided it wasn't going to work.  So then I was looking for the native java.util.Clock and looked at its source in IDE, and saw that it implemented `InstantSource` interface, which I clicked through and read about and it seemed to fit the bill.

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
jladieu: I wanted to verify the behavior that subclasses extending have the ability to provide their own tags, but I didn't want to imply that they were always going to be a `jobType` or `importType` which is what the production classes do.

davidangb (approved):
lgtm, giving this a 👍 . I took a quick look at our OpenAPI spec and I don't think it needs any updates - it just says we generically return an object for `input` when creating, retrieving, or listing jobs. Do you think we need any changes to the OpenAPI spec?

jladieu:
>lgtm, giving this a 👍 . I took a quick look at our OpenAPI spec and I don't think it needs any updates - it just says we generically return an object for input when creating, retrieving, or listing jobs. Do you think we need any changes to the OpenAPI spec?

Hmm, good point!  I added 62f41c04cc825a1be5cdb2545041aee61656f716 to add some test coverage to verify this assertion and I think this corroborates your assertion that it's just an object.

davidangb (approved):
ty!

jladieu:
Just to keep it visible, I pushed one more amendment to address Sonar warnings on intentionality in the unit test, doesn't change test functionality or behavior so I'm going to merge.

jladieu:
Did some git tidying to smoosh the tiny commits into the last one since they all belong to the same functional change.

PR #701, opened 2024-04-04
AJ-1727 Measure outbound requests in `RestClientRetry`

jladieu:
[AJ-1727](https://broadworkbench.atlassian.net/browse/AJ-1727): Add `wds.outbound` observation

This will measure count & latency for outbound requests, using the `loggerHint` as a label.

[AJ-1727]: https://broadworkbench.atlassian.net/browse/AJ-1727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/retry/RestClientRetry.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/retry/RestClientRetry.java
@@ -1,5 +1,7 @@
 package org.databiosphere.workspacedataservice.retry;
 
+import io.micrometer.observation.Observation;
+import io.micrometer.observation.ObservationRegistry;
 import java.util.Objects;
 import org.databiosphere.workspacedataservice.service.model.exception.AuthenticationException;
 import org.databiosphere.workspacedataservice.service.model.exception.AuthorizationException;
@@ -18,6 +20,11 @@ import org.springframework.web.client.HttpStatusCodeException;
 @Component
 public class RestClientRetry {
   private static final Logger LOGGER = LoggerFactory.getLogger(RestClientRetry.class);
+  private final ObservationRegistry observations;
+
+  public RestClientRetry(ObservationRegistry observations) {
+    this.observations = observations;
+  }
 
   /**
    * Perform a REST client request with logging and exception handling, and return the result of
@@ -41,6 +48,8 @@ public class RestClientRetry {
       listeners = {"retryLoggingListener"})
   public <T> T withRetryAndErrorHandling(RestCall<T> restCall, String loggerHint)
       throws RestException, AuthenticationException, AuthorizationException {
+    Observation observation =
+        Observation.start("wds.outbound", observations).lowCardinalityKeyValue("hint", loggerHint);
     try {
       LOGGER.debug("Sending {} request to REST target ...", loggerHint);
       T functionResult = restCall.run();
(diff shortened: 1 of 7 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/leonardo/LeonardoDaoTest.java
```diff
@@ -29,7 +28,7 @@
 import org.springframework.test.context.TestPropertySource;
 
 @DirtiesContext
-@SpringBootTest(classes = {LeonardoConfig.class, RestClientRetry.class})
```
jladieu, line 32: This test started failing due to missing the dependencies of `ObservationRegistry`; rather than try maintain the list of those dependencies I just removed the customization.

service/src/test/java/org/databiosphere/workspacedataservice/retry/RestClientRetryTest.java
```diff
@@ -27,15 +32,16 @@
 /** Tests for @see RestClientRetry */
 @DirtiesContext
 @SpringBootTest(
-    classes = {RestClientRetry.class, RetryLoggingListener.class},
```
jladieu, line 30: Same as before: this test started failing due to missing the dependencies of `ObservationRegistry`; rather than try maintain the list of those dependencies I just removed the customization.

service/src/test/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDaoTest.java
```diff
@@ -38,7 +37,7 @@
 import org.springframework.test.context.ActiveProfiles;
 
 @ActiveProfiles(profiles = "mock-sam")
-@SpringBootTest(classes = {WorkspaceManagerConfig.class, RestClientRetry.class})
```
jladieu, line 41: Same as before: this test started failing due to missing the dependencies of `ObservationRegistry`; rather than try maintain the list of those dependencies I just removed the customization.

nawatts (approved):

jladieu:
I pushed another commit to incorporate a fix for a failing test and reset your reviews.  If you don't mind, another 👀 would be appreciated.

service/src/main/java/org/databiosphere/workspacedataservice/retry/RestClientRetry.java
```diff
@@ -81,6 +94,8 @@ public <T> T withRetryAndErrorHandling(RestCall<T> restCall, String loggerHint)
               resolvedStatus,
               "Error from " + loggerHint + " REST target: " + e.getMessage()); // not retryable
       }
+    } finally {
+      observation.stop();
     }
```
davidangb, line 99: how would you feel about adding a metric here to measure number of retries? Something like:
```java
    } finally {
      Optional.ofNullable(RetrySynchronizationManager.getContext())
          .ifPresent(
              context -> {
                int retryCount = context.getRetryCount();
                observation.lowCardinalityKeyValue("retryCount", Integer.toString(retryCount));
              });
      observation.stop();
    }
```
note: I haven't actually tried this, I don't know if it will work

davidangb (approved):
code lgtm, one suggestion inline.

Since we are already recording rest-call metrics for cWDS->Rawls, does this mean we are double-counting them? Should we disable those metrics, or keep them both (at least for a while) for comparison?

nawatts (approved):

service/src/main/java/org/databiosphere/workspacedataservice/retry/RestClientRetry.java
jladieu: I thought about this but figured I should try to control the scope of this PR.  Either way it looks like there's a ticket for this: https://broadworkbench.atlassian.net/browse/AJ-1515

Given you did the thinking of how to get the retryCount out of the context, and now there's an observation that we could easily wire metrics into, I bet we could knock the point cost down on that ticket when we pull it in.  I'll link back to this comment to give whomever pulls that a speed boost.

jladieu:
>Since we are already recording rest-call metrics for cWDS->Rawls, does this mean we are double-counting them? Should we disable those metrics, or keep them both (at least for a while) for comparison?

Ahh good point, I'd say let's keep them enabled for now.  It might be interesting to see if/how they differ as it might point out some gaps in our custom measurement.  I think eventually we'd want to only have one enabled.

PR #702, opened 2024-04-04
Aj 1697 Demonstrate newer pact DSL

jladieu:
This is just to show how to use the newer Pact DSL to get an equivalent pact file.

Verified equivalence by diffing contents of service/build/pacts-wds-rawls.json before and after.


calypsomatic:
The [docs](https://docs.pact.io/implementation_guides/jvm/consumer#why-a-new-dsl-implementation) say that this way avoids some errors compared to the old way, which sounds good, and it's supposedly less cumbersome although I'm not sure I'm all that convinced of that.  I'm happy to switch to this style to avoid gotchas, but I was just curious if you had any other particular reasons why you prefer it.

calypsomatic (approved):

PR #705, opened 2024-04-10
AJ-1571: Add Orchestration to smoke test suite

jladieu:
For [AJ-1571](https://broadworkbench.atlassian.net/browse/AJ-1571): Add an integration test to the smoke_test suite.  This test will hit an orchestration backend which interacts with cwds.

Following approach described in [doc](https://docs.google.com/document/d/1xyLa62UZ_uzRvT67VWBuL3z2NZ1txYns_O0EqQmEGxo/edit#bookmark=id.k5adiljik6a7)

Paired with: https://github.com/broadinstitute/terra-github-workflows/pull/185

[AJ-1571]: https://broadworkbench.atlassian.net/browse/AJ-1571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


smoke_test/smoke_test.py
```diff
@@ -95,6 +105,21 @@ def verify_user_token(user_token: str) -> bool:
     default=None,
     help="Optional; auth token for authenticated tests. If this and workspace_id are present, enables additional tests."
   )
+  parser.add_argument(
```
jladieu, line 108: I prefer named args, following @nawatts's suggestion [here](https://github.com/DataBiosphere/terra-workspace-data-service/pull/692#pullrequestreview-1969261142) -- I could also adjust the other existing args at the same time in this PR but wasn't sure if anything automated will start breaking as a result of this and wanted to get buy-in first.

smoke_test/README.md
```diff
@@ -82,6 +99,19 @@ and this is the default if no protocol is specified:
 Optional - A `gcloud` access token. If present, `smoke_test.py` will execute all unauthenticated
 tests as well as all authenticated tests using the access token provided in this argument.
 
+### --orchestration-host
+
```
calypsomatic, line 103: These three are optional, but if any one of them is missing the orchestration/integration tests won't run?  Can that be explicitly stated in here somewhere?

smoke_test/smoke_test.py
davidangb: There may be problems in the github actions:
* https://github.com/broadinstitute/terra-github-workflows/blob/main/.github/workflows/cwds-smoke-tests.yaml
* https://github.com/broadinstitute/terra-github-workflows/blob/main/.github/actions/run-smoke-tests/action.yaml

The latter of these is shared with Sam, so it requires some care to change. We could of course stop sharing with Sam, if necessary.

smoke_test/README.md
```diff
@@ -82,6 +99,19 @@ and this is the default if no protocol is specified:
...
+orchestration backend.
+
+### --workspace-namespace
+
+Optional - The namespace of the workspace to be used in the orchestration integration tests.
+
+### --workspace-name
+
+Optional - The name of the workspace to be used in the orchestration integration tests.
+
```
davidangb, line 114: is the expectation that the workspace namespace/name will match the workspace id?

smoke_test/tests/authenticated/orchestration_tests.py
```diff
@@ -0,0 +1,67 @@
...
+  @cache
+  def call_orchestration(path: str) -> Response:
+    """Function is memoized so that we only make the call once"""
+    access_token = OrchestrationTests.ACCESS_TOKEN
+    oidc_headers = {
+      "Authorization": f"Bearer {access_token}",
+      "OIDC_access_token": access_token,
+      "OIDC_CLAIM_user_id": "cwds_orchestration_test_id",
+      "OIDC_CLAIM_expires_in": "10000",
+      "OIDC_CLAIM_email": "cwds.orchestration.test@example.com"
```
davidangb, line 36: I'm confused why you need to specify the `OIDC_*` headers here. These should be automatically created by the Apache proxy when a valid `Authorization` bearer token is present. Are we somehow bypassing Orch's Apache proxy?

smoke_test/tests/orchestration_smoke_test_case.py
```diff
@@ -0,0 +1,29 @@
...
+    else:
+      return urljoin(f"https://{OrchestrationSmokeTestCase.ORCHESTRATION_HOST}", path)
+
+  @staticmethod
+  @cache
+  def call_orchestration(path: str) -> Response:
+    """Function is memoized so that we only make the call once"""
+    user_token = OrchestrationSmokeTestCase.USER_TOKEN
+    headers = {"Authorization": f"Bearer {user_token}"} if user_token else {}
+    return requests.get(OrchestrationSmokeTestCase.build_orchestration_url(path), headers=headers)
```
davidangb, line 29: looks like these two methods are duplicated between this file and `orchestration_tests.py`

smoke_test/README.md
nawatts: And if so, could the tests automatically get the namespace and name from Rawls?

smoke_test/tests/authenticated/orchestration_tests.py
```diff
@@ -0,0 +1,67 @@
...
+    response = OrchestrationTests.call_orchestration(self.job_path(uuid4()))
+    self.assertEqual(response.status_code, 404,
+                     f"Job Status HTTP Status is not 404: {response.text}")
+
+  def test_job_listing_status_is_200_with_array(self):
+    """Call the job-listing url; it should return 200 and the response should be an array"""
+    response = OrchestrationTests.call_orchestration(self.job_listing_path())
+    self.assertEqual(response.status_code, 200,
+                     f"Job Listing HTTP Status is not 200: {response.text}")
+    job_list = json.loads(response.text)
```
nawatts, line 62: Nit: Requests has a convenience method for this

```suggestion
    job_list = response.json()
```

https://requests.readthedocs.io/en/latest/api/#requests.Response.json

nawatts (approved):

smoke_test/tests/authenticated/orchestration_tests.py
jladieu: Removed.

smoke_test/tests/authenticated/orchestration_tests.py
jladieu: I'm not sure what was going on here but I've managed to get the tests to pass by removing all the OIDC stuff.  I think I must have introduced an error by unintentionally having duplicate copies of the code checked in at the same time (and possibly interacting).

smoke_test/tests/orchestration_smoke_test_case.py
jladieu: Forgot to remove this...I bet this duplication is what was causing me to think I needed OIDC somehow.

jladieu:
>And if so, could the tests automatically get the namespace and name from Rawls?

I mentioned this as an optional bonus in the doc, although I was thinking the other way around:
```
(optional extension: make workspaceNamespace / name always required, independent from orchestration, then use these to go out to Rawls and look up workspaceId, thereby not requiring it as an arg for any of the tests), this introduces complexity but might also add an extra verification of a Rawls integration point used by cWDS
```

I like your idea better...if we can just take one arg (the workspaceId) and look up the other info.

smoke_test/README.md
jladieu: > is the expectation that the workspace namespace/name will match the workspace id?

Updated README to capture this assumption

smoke_test/smoke_test.py
jladieu: I'll leave the existing args as is to maintain compatibility with Sam.  The additional args will be wired as such in https://github.com/broadinstitute/terra-github-workflows/pull/185 and should hopefully not interact with Sam at all.

smoke_test/README.md
```diff
@@ -59,8 +59,25 @@ syntax: python smoke_test.py {CWDS_HOST} {WORKSPACE_ID} $(gcloud auth print-acce
 
 example to run tests against dev:
 
-```python smoke_test.py cwds.dsde-dev.broadinstitute.org 123e4567-e89b-12d3-a456-426614174000 $(gcloud auth print-access-token)```
-_(replace the example workspace uuid with a real workspace's uuid)_
+```python smoke_test.py cwds.dsde-dev.broadinstitute.org 1df7c7c1-ce85-4dc3-8539-4766896821bb $(gcloud auth print-access-token)```
```
yuliadub, line 62: this right now references the https://bvdp-saturn-dev.appspot.com/#workspaces/general-dev-billing-account/cwds-smoketest-do-not-delete, is it okay to run smoke tests against this workspace for local tests/etc or should each person create their own?

smoke_test/tests/authenticated/orchestration_tests.py
```diff
@@ -0,0 +1,59 @@
...
+    assert OrchestrationTests.ORCHESTRATION_HOST, "ERROR - OrchestrationTests.ORCHESTRATION_HOST not properly set"
+    if re.match(r"^\s*https?://", OrchestrationTests.ORCHESTRATION_HOST):
+      return urljoin(OrchestrationTests.ORCHESTRATION_HOST, path)
+    else:
+      return urljoin(f"https://{OrchestrationTests.ORCHESTRATION_HOST}", path)
+
+  @staticmethod
+  @cache
+  def call_orchestration(path: str) -> Response:
+    """Function is memoized so that we only make the call once"""
```
yuliadub, line 28: learned a new word -  "memoized" 😄

smoke_test/README.md
jladieu: I picked this one because I figured it'll be around for the foreseeable future and shouldn't get deleted.

The smoke tests are completely stateless so there should be no impact to having multiple people running this command.

ashanhol: Wasn't there discussion that for some reason dev GCP workspaces end up deleting if they're a week old?

nawatts: I know of an automated job to clean up workspaces created by Terra UI e2e tests. It deletes old workspaces if the name starts with `terra-ui-test-workspace-`.
https://github.com/DataBiosphere/terra-ui/blob/dev/integration-tests/tests/delete-orphaned-workspaces.js

But I don't think there's an expiration on dev GCP workspaces in general... I can see a few on my workspaces list that are years old.

calypsomatic: It's on BEEs that GCP workspaces get murdered after a week.

smoke_test/tests/authenticated/orchestration_tests.py
yuliadub: as Adina pointed out to me, I may have known it in the past (college) but it clearly left my brain as of now 😄

davidangb (approved):

PR #709, opened 2024-04-11
AJ-1697 Bump WDS version on every merge to main

calypsomatic:
Throughout DSP, services update their version on every merge to main.  This is required for the pact broker to be able to manage versioning for publishing and verifying pacts.  Previously, WDS only bumped its version on release, which was manually done by a member of a team.  To enable contract testing and maintain consistency throughout the department, this PR starts bumping WDS with every merge to main.  When we want to manually release WDS along with its clients, this will now be a *minor* instead of a patch version bump.
Current GHA structure and the changes in this PR are outlined [here](https://github.com/DataBiosphere/terra-workspace-data-service/wiki/GHA-structure-in-WDS).  This document will be updated to reflect only the current situation after the PR merges.
Questions for reviewers:
- `tag-and-publish` is still calling the `bumper` action directly, instead of the `tag` workflow.  Is there any reason to call `tag` or to keep it as is?  
- Are there any repercussions for clients or releases that we need to deal with?  
- Have I accounted for all the inputs required in all situations where the updated workflows are called?

Reminder:


.github/workflows/tag.yml
```diff
@@ -1,90 +1,72 @@
 name: Tag
 on:
   workflow_dispatch:
-    # Inputs the workflow accepts.
+  workflow_call:
```
calypsomatic, line 4: Sorry the renaming makes this confusing.  This file wasn't changed at all, just renamed.

.github/workflows/tag-and-publish.yml
```diff
@@ -0,0 +1,90 @@
...
+        with:
+          token: ${{ secrets.BROADBOT_TOKEN }} # this allows the push to succeed later
+
+      - name: Bump the tag to a new version
+        # https://github.com/DataBiosphere/github-actions/tree/master/actions/bumper
+        uses: databiosphere/github-actions/actions/bumper@bumper-0.3.0
+        id: tag
+        env:
+          GITHUB_TOKEN: ${{ secrets.BROADBOT_TOKEN }}
+          DEFAULT_BUMP: minor
```
calypsomatic, line 33: The renaming makes this seem like a new file; in reality the only thing changed is line 33 from patch to minor.

.github/workflows/build-and-test.yml
```diff
@@ -85,3 +85,21 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Skip version bump merges
+        id: bumpcheck
+        uses: ./.github/actions/bump-skip
+        with:
+          event-name: ${{ github.event_name }}
+
+  tag:
+    if: always() && needs.bump-check.outputs.is-bump == 'no' && github.event_name == 'push' && github.ref == 'refs/heads/main'
```
nawatts, line 102: `always()` looks unnecessary here since there are other conditions.

```suggestion
    if: needs.bump-check.outputs.is-bump == 'no' && github.event_name == 'push' && github.ref == 'refs/heads/main'
```

.github/workflows/tag-and-publish.yml
```diff
@@ -0,0 +1,90 @@
...
+          RELEASE_BRANCHES: main
+          VERSION_FILE_PATH: build.gradle
+          VERSION_LINE_MATCH: "^\\s*version\\s*=\\s*'.*'"
+          VERSION_SUFFIX: SNAPSHOT
```
nawatts, line 37: Since every commit is going to get a new tag anyway, I don't think we need to bump the tag _again_ in this job.

```suggestion
```

.github/workflows/tag-and-publish.yml
nawatts, line 1: I think we'd want to publish a Docker image and report to Sherlock on every commit to main so that the latest version is always available for the dev environment and BEEs.

.github/workflows/build-and-test.yml
calypsomatic: I wondered about that!  I looked it up and the `always()` makes sure it runs (or checks to see if it should run) if previous steps fail.  I'm not sure if we want it to run if other steps fail or not, but it at least isn't as senseless as it seems at first.

.github/workflows/tag-and-publish.yml
calypsomatic: But we do want to distinguish between releases and non-releases with patch v minor, so when this workflow is called for releases we would want to bump it again to minor.

.github/workflows/tag-and-publish.yml
calypsomatic: Ah good idea.  I'll see if I can do that.

.github/workflows/build-and-test.yml
nawatts: Ah.
> You can use the following status check functions as expressions in if conditionals. A default status check of success() is applied unless you include one of these functions
https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions

.github/workflows/tag-and-publish.yml
nawatts: Oohhhh. Missed the difference in the `DEFAULT_BUMP`.

.github/workflows/tag-and-publish.yml
calypsomatic: Ok I just put it in `build-and-test`; I'm open to suggestions for whether that means we should renamed that workflow for clarity, or move it somewhere else.

.github/workflows/tag-and-publish.yml
calypsomatic: Update: also changed the name of the workflow (on line 1)

.github/actions/bump-skip/action.yml
```diff
@@ -0,0 +1,27 @@
...
+runs:
+  using: "composite"
+  steps:
+    - name: Bump test
+      id: bumptest
+      run: |
+        log=$(git log --pretty='%B')
+        echo "log=$log"
+        pattern="^bump .*"
+        IS_BUMP=no
```
yuliadub, line 21: more of a nit then anything, but it may be more straightforward to use bools like "false" here vs "no"? :)

calypsomatic: This is one of those workflows that is copy-pasted throughout the organization; I'm mildly inclined to leave it as-is so that other copy-paste jobs that rely on it are correct

yuliadub:
looking at the diagram you created (super helpful, thank you!), I am a bit confused why we public python client on PR -- doesnt seem like the GHAs are actually doing that so maybe there is an accidental extra arrow in there?

.github/actions/bump-skip/action.yml
yuliadub: ah if that is the case, I agree best to leave it as is!

.github/workflows/tag.yml
```diff
@@ -1,90 +1,72 @@
...
 jobs:
+  # On tag vs. new-tag.
+  # The new-tag is always the tag resulting from a bump to the original tag.
+  # However, the tag is by definition the value of the latest tag after running the action,
+  # which might not change if dry run is used, and remains same as the original tag.
   tag-job:
     runs-on: ubuntu-latest
     outputs:
-      new_tag: ${{ steps.tag.outputs.tag }}
+      tag: ${{ steps.tag.outputs.tag }}
```
yuliadub, line 45: maybe I am just missing it - but where do we set tag?

calypsomatic: Next step - "Bump the tag to a new version", line 53

calypsomatic:
> looking at the diagram you created (super helpful, thank you!), I am a bit confused why we public python client on PR -- doesnt seem like the GHAs are actually doing that so maybe there is an accidental extra arrow in there?
> 
> another question I have is - when we say we publish docker/bump tag "on PR", what do we mean? like on PR being created/updated or is this more specifically on when PR is merged into main? If we push docker image on each PR update that may end up being a lot of images...

@yuliadub I made the diagram based on what was written at the top of each file - the "on" part, and `release-python-client` has "on: pull_request" so I took that to mean it runs on every PR.  Maybe I'm misunderstanding it or there's something else I'm missing!  I'll investigate.
As for a lot of images... I think there just will be (maybe we need some process for cleaning up old images?  Do they expire?).  We wanted an image for every push to a PR so that you can deploy your version to a BEE for testing.

.github/workflows/tag.yml
nawatts: Call and dispatch are different events. Call allows the workflow to be run from another workflow. Dispatch allows the workflow to be manually run.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_call
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

That being said, I do think we could get rid of dispatch here since this is only called from other workflows, not run on its own.

yuliadub:
> > looking at the diagram you created (super helpful, thank you!), I am a bit confused why we public python client on PR -- doesnt seem like the GHAs are actually doing that so maybe there is an accidental extra arrow in there?
> > another question I have is - when we say we publish docker/bump tag "on PR", what do we mean? like on PR being created/updated or is this more specifically on when PR is merged into main? If we push docker image on each PR update that may end up being a lot of images...
> 
> @yuliadub I made the diagram based on what was written at the top of each file - the "on" part, and `release-python-client` has "on: pull_request" so I took that to mean it runs on every PR. Maybe I'm misunderstanding it or there's something else I'm missing! But just looking at this PR, it seems like it has been publishing the python client. Maybe the question is whether we want to do that or not? As for a lot of images... I think there just will be (maybe we need some process for cleaning up old images? Do they expire?). We wanted an image for every push to a PR so that you can deploy your version to a BEE for testing.

we run and install the client locally on each PR -- but we only publish the client when we merge into main. I can see how that can be confusing since in the release python client here is this clause for the [publish step](https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/.github/workflows/release-python-client.yml#L121). 

makes sense for images, I think clean up would be nice if we do push a new image on every PR update.

.github/workflows/tag.yml
yuliadub: ah yes, sorry I got confused and those this was actually a step definition but it is not. Thanks for clarifying!

.github/workflows/tag-and-publish.yml
yuliadub: ah I see, and that then becomes the output. makes sense!

.github/workflows/tag.yml
yuliadub: ha, how did I miss that 😄

.github/workflows/tag-and-publish.yml
calypsomatic: Ok I removed the "push: main" part of `publish-docker`, so it's still called on push to main, but FROM `build-and-test` so it has the correct tag.

pull_request_template.md
```diff
@@ -1,7 +1,11 @@
 Reminder:
 
-PRs merged into main will not automatically generate a PR in https://github.com/broadinstitute/terra-helmfile to update the WDS image deployed to kubernetes - this action will need to be triggered manually by running the following github action: https://github.com/DataBiosphere/terra-workspace-data-service/actions/workflows/tag.yml. Dont forget to provide a Jira Id when triggering the manual action, if no Jira ID is provided the action will not fully succeed. 
+#### Releasing WDS ####
+PRs merged into main will not automatically generate a PR in https://github.com/broadinstitute/terra-helmfile to update the WDS image deployed to kubernetes - this action will need to be triggered manually by running the following github action: https://github.com/DataBiosphere/terra-workspace-data-service/actions/workflows/tag-and-publish.yml. Dont forget to provide a Jira Id when triggering the manual action, if no Jira ID is provided the action will not fully succeed.
```
calypsomatic, line 4: Noticed this link will now be wrong, so I updated it and added in a reminder about the GHA diagram while I was at it.

yuliadub (approved):
thanks for tackling this! now we just need to make sure the release works with no issues next time we do one :)

nawatts (approved):

PR #710, opened 2024-04-11
AJ-1759: Fix TDR snapshot `EntityReferenceList`

jladieu:
* CreateAttributeEntityReferenceList value should be named `attributeListName` in the Rawls JSON output.
* The actual entity reference value, when deserialized from Parquet, needs to be unwrapped from the bigquery-based format to populate the entity reference ID values.


service/src/main/java/org/databiosphere/workspacedataservice/service/RecordService.java
```diff
@@ -102,17 +102,19 @@ private Map<Relation, List<RelationValue>> getAllRelationArrayValues(
       for (Map.Entry<String, Object> attribute : arrayAttributesForThisRecord) {
         // How to read relation list depends on its source, which we don't know here, so we have to
         // check
-        List<String> rels;
+        List<Object> rels;
         if (attribute.getValue() instanceof List<?>) {
-          rels = (List<String>) attribute.getValue();
+          // attribute.getValue() might be a Collection<RelationAttribute>, so we can't just cast
```
jladieu, line 107: This was the strangest behavior I was observing, which is that the attribute value was coming out as a `List<RelationAttribute>` and not a `List<String>`.  Assigning it to a `List<Object>` avoids a class cast exception but still lets us use `toString()` later on during storage to `RecordService`

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrTestSupport.java
```diff
@@ -51,8 +49,13 @@ TdrManifestQuartzJob buildTdrManifestQuartzJob(UUID workspaceId) {
       @Override
       protected URL parseUrl(String path) {
         if (path.startsWith("classpath:")) {
-          return requireNonNull(
-              getClass().getClassLoader().getResource(path.substring("classpath:".length())));
+          var url = getClass().getClassLoader().getResource(path.substring("classpath:".length()));
```
jladieu, line 52: This just surfaces a more friendly error when there are issues with the classpath resources used by E2E tests.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJobE2ETest.java
```diff
@@ -216,6 +219,43 @@ void importV2FManifest() throws IOException, JobExecutionException {
     assertEquals(expectedCounts, actualCounts);
   }
 
+  @Test
+  @Tag(SLOW)
+  void devSampleManifest() throws IOException, JobExecutionException {
```
jladieu, line 224: This new E2E test repros the issue encountered with the snapshot.

nawatts (approved):

jladieu:
@nawatts Thanks for taking a look!  I reset your review because I added some extra test coverage & sonar fixen after you took a look.

nawatts (approved):

service/src/test/resources/tdrmanifest/devsample.json
```diff
@@ -0,0 +1,181 @@
+{
+  "snapshot": {
+    "id": "26f8ca1a-06d7-4466-b3f6-f56d84cbe0b8",
```
calypsomatic, line 3: Extreme nit: the title "devsample" is pretty vague and doesn't really reflect what's in it.

calypsomatic (approved):
LGTM, did not test

service/src/test/resources/tdrmanifest/devsample.json
jladieu: I'll rename to with-entity-reference-lists

PR #713, opened 2024-04-12
AJ-1697 Add bump-check to publish_pacts

calypsomatic:
`can-i-deploy` doesn't need to run on the bump commit, since the exact same code was just verified.  It fails because it doesn't receive a version parameter, which isn't present in the bump commit, since it's not a PR event.  This uses the bump-check action to avoid re-running contract tests on commits that are only bumping the version.

Reminder:

#### Releasing WDS ####

### DIFF
--- a/.github/workflows/publish-pacts.yml
+++ b/.github/workflows/publish-pacts.yml
@@ -17,6 +17,17 @@ env:
   CAN_I_DEPLOY_RUN_NAME: 'can-i-deploy-${{ github.event.repository.name }}-${{ github.run_id }}-${{ github.run_attempt }}'
 
 jobs:
+  bump-check:
+    runs-on: ubuntu-latest
+    outputs:
+      is-bump: ${{ steps.bumpcheck.outputs.is-bump }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Skip version bump merges
+        id: bumpcheck
+        uses: ./.github/actions/bump-skip
+        with:
+          event-name: ${{ github.event_name }}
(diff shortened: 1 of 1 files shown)


nawatts (approved):

.github/workflows/publish-pacts.yml
```diff
@@ -38,6 +49,8 @@ jobs:
   # Note: All workflows from the same PR merge should have the same copy of settings.gradle on disk,
   # which should be the one from the HEAD of the main branch before the workflow starts running
   regulated-tag-job:
+    needs: [ bump-check ]
+    if: ${{ needs.bump-check.outputs.is-bump == 'no' }}
```
yuliadub, line 53: should this not have a "-"? looking at same variable on line 23

jladieu (approved):

yuliadub (approved):
just one clarifying question, otherwise LGTM

.github/workflows/publish-pacts.yml
nawatts: It's the difference between the "bump-check" _job_ ID and the "bumpcheck" _step_ ID.

calypsomatic: No, the difference in hyphen usage is there on purpose to confuse us.  Since this is a different job, it's referring to the JOB "bump-check", defined on line 20.  WITHIN the "bump-check" job, it has a STEP that's just called "bumpcheck"

yuliadub: haha wow those names are quite misleading 😄

PR #719, opened 2024-04-17
AJ-1776: use real commit SHA for tagging PR builds

davidangb:
When publishing docker images / reporting to Sherlock on PRs, use the SHA from the latest PR commit instead of the SHA or the merge commit. This makes it much easier to correlate image tags/Sherlock versions with your PR.

Check out the actions that ran on this PR, especially `report-to-sherlock`, and compare to the list of commits in this PR.

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -43,6 +43,7 @@ jobs:
           echo "github.ref: ${{ github.ref }}"
           echo "github.ref_name: ${{ github.ref_name }}"
           echo "github.sha: ${{ github.sha }}"
+          echo "github.event.pull_request.head.sha: ${{ github.event.pull_request.head.sha }}"
 
       - name: Set commit short hash
         id: setHash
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-docker.yml
```diff
@@ -37,8 +37,10 @@ jobs:
 
       - name: Set commit short hash
         id: setHash
+        # github.event.pull_request.head.sha is the latest commit for a pull request
+        # github.sha is the latest commit for a merge to main (but is the merge commit on PRs)
         run: |
-          git_short_sha=$(git rev-parse --short HEAD)
+          git_short_sha=$(echo ${{ github.event.pull_request.head.sha || github.sha }} | cut -c1-7)
           echo $git_short_sha
           echo "git_short_sha=${git_short_sha}" >> $GITHUB_OUTPUT
```
davidangb, line 45: `$(git rev-parse --short HEAD)` is the equivalent of `github.sha`, which returns the _merge_ commit for PRs.

.github/workflows/publish-docker.yml
```diff
@@ -37,8 +37,10 @@ jobs:
 
       - name: Set commit short hash
         id: setHash
+        # github.event.pull_request.head.sha is the latest commit for a pull request
+        # github.sha is the latest commit for a merge to main (but is the merge commit on PRs)
```
nawatts, line 41: Super nit: the difference here is the event that triggered the workflow (push vs pull request), not the type of commit.
https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

```suggestion
        # github.sha is the latest commit for a push to main (but is the merge commit on PRs)
```

nawatts (approved):

PR #739, opened 2024-04-23
[AJ-1786] Generate Pacts for both WDS and CWDS

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1786

Pacts are between services. Since the WDS repository now contains two services (WDS and CWDS), we want to publish Pacts for both of those.

Some Pacts are the same for WDS and CWDS. Both services have the same interactions with Sam and DataRepo. For those, we want to use one Pact consumer test to publish a Pact for both services.

To accomplish that, this replaces the hard coded `wds` consumer name with a consumer name from an environment variable. The Pact tests are run twice: once with the consumer name set to `wds` and once with it set to `cwds`. All of the resulting Pacts will be published to the Pact broker.

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -246,13 +246,29 @@ test {
     systemProperties['user.timezone'] = "UTC"
 }
 
-tasks.register("pactTests", Test) {
+tasks.register("wdsPactTests", Test) {
+    environment "CONSUMER_NAME", "wds"
     useJUnitPlatform {
         includeTags "pact-test"
     }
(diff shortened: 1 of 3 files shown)


service/build.gradle
nawatts, line 1: Not sure if there's a better way to do this to avoid duplicating configuration between two tasks.

calypsomatic (approved):

davidangb (approved):

PR #742, opened 2024-04-24
AJ-1700: Initial scaffolding for cWDS support of `RAWLSJSON`

jladieu:
Some small no-op / skeletal implementations to lay the groundwork for the `RAWLSJSON` format, which will be used to support async TSV in cWDS.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/rawlsjson/RawlsJsonSchedulable.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/rawlsjson/RawlsJsonSchedulable.java
@@ -1,0 +1,14 @@
+package org.databiosphere.workspacedataservice.dataimport.rawlsjson;
+
+import static org.databiosphere.workspacedataservice.generated.ImportRequestServerModel.TypeEnum.RAWLSJSON;
+
+import java.io.Serializable;
+import java.util.Map;
+import org.databiosphere.workspacedataservice.shared.model.Schedulable;
+
+public class RawlsJsonSchedulable extends Schedulable {
+  public RawlsJsonSchedulable(
+      String name, String description, Map<String, Serializable> arguments) {
+    super(RAWLSJSON.name(), name, RawlsJsonQuartzJob.class, description, arguments);
+  }
+}
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
@@ -68,7 +68,6 @@ import org.springframework.stereotype.Component;
 @Component
 public class TdrManifestQuartzJob extends QuartzJob {
 
-  private final JobDao jobDao;
   private final RecordSinkFactory recordSinkFactory;
   private final BatchWriteService batchWriteService;
   private final CollectionService collectionService;
@@ -93,8 +92,7 @@ public class TdrManifestQuartzJob extends QuartzJob {
       DataImportProperties dataImportProperties,
       SnapshotSupportFactory snapshotSupportFactory,
       SamDao samDao) {
-    super(observationRegistry, dataImportProperties);
-    this.jobDao = jobDao;
+    super(jobDao, observationRegistry, dataImportProperties);
     this.recordSinkFactory = recordSinkFactory;
     this.recordSourceFactory = recordSourceFactory;
     this.batchWriteService = batchWriteService;
@@ -104,11 +102,6 @@ public class TdrManifestQuartzJob extends QuartzJob {
     this.snapshotSupportFactory = snapshotSupportFactory;
     this.samDao = samDao;
     this.isTdrPermissionSyncingEnabled = dataImportProperties.isTdrPermissionSyncingEnabled();
-  }
-
-  @Override
-  protected JobDao getJobDao() {
-    return this.jobDao;
   }
 
   @Override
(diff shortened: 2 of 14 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -106,11 +104,6 @@ public TdrManifestQuartzJob(
     this.isTdrPermissionSyncingEnabled = dataImportProperties.isTdrPermissionSyncingEnabled();
   }
 
-  @Override
```
jladieu, line 109: This is just a small refactoring to change `JobDao` to be injected at construction time, rather than provided via getter in a subclass.  I think this code pre-dated when we had constructor injection when it was first introduced, and so was the lightest weight way to do this at that time, but now that there's a constructor injection point for the other instance variables, this seemed to make more sense to be handled similarly.

service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
```diff
@@ -242,8 +240,7 @@ void jobCompletion(boolean isSucceedOnCompletion) throws org.quartz.JobExecution
     // execute the TestableQuartzJob. This will move the job to SUCCESS when
     // dataImportProperties.isSucceedOnCompletion() is true; else it will leave the job
     // in RUNNING
-    new TestableQuartzJob(randomToken, observationRegistry, /* shouldThrowError= */ false)
-        .execute(mockContext);
+    getQuartzJob(randomToken).execute(mockContext);
```
jladieu, line 243: This tweak to the tests just made adding `jobDao` as a constructor arg easier to introduce.

service/src/test/java/org/databiosphere/workspacedataservice/service/ImportServiceTest.java
```diff
@@ -99,8 +100,11 @@ void setUp() throws ApiException {
 
   // does createSchedulable properly store the jobId, job group, and job data map?
   @ParameterizedTest(name = "for import type {0}")
-  @EnumSource(ImportRequestServerModel.TypeEnum.class)
-  void createSchedulableValues(ImportRequestServerModel.TypeEnum importType) {
+  @EnumSource(
+      value = TypeEnum.class,
+      names = {"RAWLSJSON"},
```
jladieu, line 105: For now, exclude `RAWLSJSON` from the expected/success/happy path tests.  It shouldn't actually be supported in `DataPlane` and so we don't want to assert that it works here alongside the other types.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/rawlsjson/RawlsJsonSchedulable.java
```diff
@@ -0,0 +1,15 @@
...
+
+import java.io.Serializable;
+import java.util.Map;
+import org.databiosphere.workspacedataservice.dataimport.pfb.PfbQuartzJob;
+import org.databiosphere.workspacedataservice.shared.model.Schedulable;
+
+public class RawlsJsonSchedulable extends Schedulable {
+  public RawlsJsonSchedulable(
+      String name, String description, Map<String, Serializable> arguments) {
+    super(RAWLSJSON.name(), name, PfbQuartzJob.class, description, arguments);
```
calypsomatic, line 13: `PfbQuartzJob.class` seems incorrect here

jladieu: Amended in a fix on the commit where I added this.  Thanks for catching!

jladieu:
Amended in a fix to remove the `public` modifier from a test as suggested by Sonar.

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
```diff
@@ -156,9 +157,10 @@ public static Schedulable createSchedulable(
       UUID jobId,
       Map<String, Serializable> arguments) {
     return switch (importType) {
+      case PFB -> new PfbSchedulable(jobId.toString(), "PFB import", arguments);
+      case RAWLSJSON -> new RawlsJsonSchedulable(jobId.toString(), "RAWLSJSON import", arguments);
```
davidangb, line 161: pretty sure that Orchestration sends "rawlsjson" instead of "RAWLSJSON" - will we need to change orchestration? Is there an option to make the enum `fromValue` case-insensitive? Or any other good way to handle this?

davidangb (approved):
:+1:, but one question about "rawlsjson" vs "RAWLSJSON"

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
jladieu: 🤔 Thanks for pointing this out.  I'll take a look and see what I can do.  I'd like to avoid having to mess with orchestration (for this reason at least) if possible.

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
jladieu: I'm going to merge as is here, under the assumption that I'll need to use the existing type translation logic as part of integrating orchestration, but if you feel there's followup to this I'll be happy to tackle whatever makes sense.

PR #743, opened 2024-04-25
AJ-1628: TSV uploads detect datatypes across batches

davidangb:
## Before this PR
WDS would read the first _batch_ of records from a TSV upload (or other input!), infer a schema based on the contents of those records, and then use that schema for the remainder of the batches. The batch size is relatively large, at 5000 records, defined [here](https://github.com/DataBiosphere/terra-workspace-data-service/blob/d8212da81efe1e1934a9f84e73ba72f7668c6f1f/service/src/main/resources/application.yml#L126).

However, this means that if a record in any batch later than the first has a conflicting datatype, the upload will fail. For instance, if the first 5000 records have integers in a given column but the 5001st record has a string, we'll hit an error trying to insert a string into a number column.

## After this PR
We now infer schema on every batch. This trades performance for correctness - it's definitely more work inside the codebase, but it eliminates errors.

This PR adds one unit test in `TsvBatchTypeDetectionTest`, deletes two tests that are no longer needed, and updates another unit test for correctness.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
@@ -67,7 +67,7 @@ public class BatchWriteService {
     BatchWriteResult result = BatchWriteResult.empty();
 
     // tracker to stash the schemas for the record types seen while processing this stream
-    Map<RecordType, Map<String, DataTypeMapping>> typesSeen = new HashMap<>();
+    Map<RecordType, Map<String, DataTypeMapping>> typeSchemas = new HashMap<>();
 
     // loop through, in batches, the records provided by the RecordSource. This loops
     // until the RecordSource returns an empty batch.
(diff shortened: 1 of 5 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -60,7 +60,7 @@ public Map<String, DataTypeMapping> inferTypes(RecordAttributes updatedAtts) {
     return result;
   }
 
-  Map<String, DataTypeMapping> inferTypes(List<Record> records) {
+  public Map<String, DataTypeMapping> inferTypes(List<Record> records) {
```
jladieu, line 63: nit: `@VisibleForTesting`?

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
```diff
@@ -92,18 +92,16 @@ private BatchWriteResult consumeWriteStream(
...
-        // record type's schema, then save that schema back to the `typesSeen` map
-        if (!isTypeAlreadySeen && opType == OperationType.UPSERT) {
+        // infer and update this record type's schema, then save that schema back to the
+        // `typeSchemas` map
+        if (opType == OperationType.UPSERT) {
           Map<String, DataTypeMapping> inferredSchema = inferer.inferTypes(records);
           Map<String, DataTypeMapping> finalSchema =
               recordSink.createOrModifyRecordType(recType, inferredSchema, records, primaryKey);
-          typesSeen.put(recType, finalSchema);
+          typeSchemas.put(recType, finalSchema);
```
jladieu, line 101: I'm not sure I'm entirely grokking this yet but I think it's because I only have a shallow understanding of how `inferTypes` works.

What if this flip-flops across multiple batches?  So...it starts as a long, then a string is encountered in a subsequent batch, then another batch has _only_ longs again?  Will it do the right thing here?

jladieu (approved):
Looks like a very graceful fix!

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
yuliadub: I am also wondering the same - would going from int to string back to int make this fail somehow?

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
davidangb: heh - I had that in there 117d64475dd7e372f4ac0982ff7a7f9a931587c6, but Sonar complained because this method is called from runtime code.

service/src/main/java/org/databiosphere/workspacedataservice/service/BatchWriteService.java
davidangb: The inferer considers the current datatype as well as the incoming data. So if the column is already a string, if the inferer detects numbers it will NOT suggest a numeric datatype; it will keep it as a string. See https://github.com/DataBiosphere/terra-workspace-data-service/blob/58039b249fa32a4b1a92ed0c6395a00657b71d09/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java#L82-L105.

This is why all columns end up as strings eventually - in almost every case as soon as any non-compliant data is encountered, we say the column is a string.

davidangb: I've updated the unit test to reflect this case!

yuliadub (approved):
thank you for clearing up our question :) LGTM

PR #744, opened 2024-04-25
AJ-1675 - Disable Web specific logging to limit log volume

yuliadub:
Bottom line: 
I think it will be best to disable DispatcherServlet logs and rely on other logging for request URLs and response codes. Below is the summary of my investigations. 

A bit of context, here is what a DispatcherServlet log looks like for any API call: 
**2024-04-25 21:38:14.914 DEBUG [VvWDLElx] [http-nio-8080-exec-1] o.s.web.servlet.DispatcherServlet        : GET "/status/liveness", parameters={}
2024-04-25 21:38:14.918 DEBUG [Vp21QnjW] [http-nio-8080-exec-9] o.s.web.servlet.DispatcherServlet        : Completed 200 OK
2024-04-25 21:37:52.146 DEBUG [4KB9GGeq] [http-nio-8080-exec-3] o.s.web.servlet.DispatcherServlet        : GET "/dee19d25-f877-4c5f-b2fe-398311b8b633/types/v0.2", parameters={}
2024-04-25 21:37:52.146 DEBUG [360PdlrB] [http-nio-8080-exec-8] o.s.web.servlet.DispatcherServlet        : GET "/capabilities/v1", parameters={}
2024-04-25 21:37:52.148 DEBUG [360PdlrB] [http-nio-8080-exec-8] o.s.web.servlet.DispatcherServlet        : Completed 200 OK
2024-04-25 21:37:52.173 DEBUG [4KB9GGeq] [http-nio-8080-exec-3] o.s.web.servlet.DispatcherServlet        : Completed 200 OK**

1. First I was trying to gauge if there is any specific filtering that can be done for what DispatcherServlet  logs - for example not logging specific paths (like status/liveness) but that doesnt seem to be an option. 
2. Next I looked into if we can only log via DispatcherServlet for requests that do not return 200s (since those are more interesting anyways, especially for liveness/readiness) but that doesnt appear to be an option. There are other logging options that you can use instead of DispatcherServlet, but that would mean writing that implementation (it is not a plug in and it works, or at least so I thought at first 😄 )
3. What I later realized is that we have activity logging that will actually log on Liveness/Readiness changes [here](https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/service/src/main/java/org/databiosphere/workspacedataservice/logging/AvailabilityListener.java#L16). To test this I triggered a pod restart and confirmed that right before the pod was shut down a state change on the Readiness side was registered as expected.  Basically this means that we do have logging for when liveness/readiness state changes independently of DispatcherServlet. 
**2024-04-25 21:30:14.537 ERROR [] [SpringApplicationShutdownHook] o.d.w.logging.AvailabilityListener       : ReadinessState: REFUSING_TRAFFIC**
5. For other APIs, like uploading a new table or loading contents of an existing tables, we have some logging already that lives in WDS Activity logger, as well as in RecordDao, so I do not think we would be loosing anything by not having the exact api path that was hit. However, if we do want to log more info - we can add that in our own logging. 
**2024-04-25 21:37:52.280 DEBUG [7Zxx2OO4] [http-nio-8080-exec-6] o.s.web.servlet.DispatcherServlet        : POST "/dee19d25-f877-4c5f-b2fe-398311b8b633/search/v0.2/sample", parameters={}
2024-04-25 21:37:52.292  INFO [7Zxx2OO4] [http-nio-8080-exec-6] o.d.w.service.RecordOrchestratorService  : queryForEntities: sample
2024-04-25 21:37:52.292  INFO [7Zxx2OO4] [http-nio-8080-exec-6] o.d.workspacedataservice.dao.RecordDao   : queryForRecords: sample**
6. Ended up settling on logbook for [logging ](https://github.com/zalando/logbook)requests/responses (basically found some road blocks that were annoying with a custom implementation in the end). This solution lets us exclude liveness/readiness events and even thought it logs quite a bit of info for other requests, it seems valuable. 
 
example of 404 request/response
image
image

example of 200 request/response
image

**after PR feedback**
image

Reminder:

#### Releasing WDS ####


service/src/main/java/org/databiosphere/workspacedataservice/logging/LogbookConfiguration.java
```diff
@@ -0,0 +1,18 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.zalando.logbook.Logbook;
+
+@Configuration
+public class LogbookConfiguration {
+
+  @Bean
+  public Logbook logbook() {
+    Logbook logbook = Logbook.builder().condition(exclude(requestTo("/status/**"))).build();
```
yuliadub, line 15: this will exclude liveness/readiness checks but not simple "status" checks

service/src/main/java/org/databiosphere/workspacedataservice/logging/LogbookConfiguration.java
```diff
@@ -0,0 +1,17 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.zalando.logbook.Logbook;
+
+@Configuration
+public class LogbookConfiguration {
+
+  @Bean
+  public Logbook logbook() {
+    return Logbook.builder().condition(exclude(requestTo("/status/**"))).build();
```
davidangb, line 15: what do you think about using
``
.sink(new ExtendedLogFormatSink(new DefaultHttpLogWriter()))
```
or
```
.sink(new CommonsLogFormatSink(new DefaultHttpLogWriter()))
```
to get the requests/responses on a single line? It's a terser format but I don't think we need all the request/response headers.

**Comment by jladieu** [comment]
Approach SGTM!

File: service/src/main/resources/logback-spring.xml [hunk 3]
```diff
@@ -67,17 +67,12 @@
...
-  </logger>
-
   <!-- log the remainder of org.springframework.web at INFO level -->
   <logger name="org.springframework.web" level="info" additivity="false">
     <appender-ref ref="CONSOLE"/>
     <appender-ref ref="Sentry"/>
   </logger>
 
+  <logger name="org.zalando.logbook" level="TRACE"/>
+
```
davidangb, line 77: do you need both this setting in the XML and the setting in application.yml? Will only one suffice?

service/src/main/java/org/databiosphere/workspacedataservice/logging/LogbookConfiguration.java
yuliadub: I lied, status/** excludes stats as well so going to add the two path on its own to avoid that

service/src/main/resources/logback-spring.xml
yuliadub: I swear when I checked first it didnt work until I added the line here also, but now that I try again it seems to not be needed in here 😄

service/src/main/java/org/databiosphere/workspacedataservice/logging/LogbookConfiguration.java
```diff
@@ -0,0 +1,22 @@
...
+import org.zalando.logbook.core.CommonsLogFormatSink;
+import org.zalando.logbook.core.DefaultHttpLogWriter;
+
+@Configuration
+public class LogbookConfiguration {
+
+  @Bean
+  public Logbook logbook() {
+    return Logbook.builder()
+        .condition(exclude(requestTo("/status/liveness"), requestTo("/status/readiness")))
```
davidangb, line 18: could we/should we also exclude `/prometheus` to avoid yet more log spam?

davidangb (approved):
:+1: and thank you! One suggestion inline to also ignore `/prometheus`.

There's a kinda subtle change with this implementation. We used to get:
1. log message that we received the request
2. any log messages that happen as part of the request
3. log message showing the response

But now we get:
1. any log messages that happen as part of the request
2. log message showing the request and response

This is fine to me and I understand why it happens - we just need to remember that when if we ever need to debug logfiles!

yuliadub:
> 👍 and thank you! One suggestion inline to also ignore `/prometheus`.
> 
> There's a kinda subtle change with this implementation. We used to get:
> 
> 1. log message that we received the request
**> 2. any log messages that happen as part of the request**
> 3. log message showing the response
> 
> But now we get:
> 
> 1. any log messages that happen as part of the request
> 2. log message showing the request and response
> 
> This is fine to me and I understand why it happens - we just need to remember that when if we ever need to debug logfiles!

I am not sure I have seen number 2 really - all the testing I have done just gave request and response. But its possible there are more specific cases that I have missed. Based on observation aside from it being one log statement vs 2(request/response), these two should be relatively similar 🤷 I do agree it'd be good to document this change (maybe WDS playbook?) so we know why we did this in case this ticket gets lost.

service/src/main/java/org/databiosphere/workspacedataservice/logging/LogbookConfiguration.java
yuliadub: would the requests just be /prometheus, not /prometheus/****?

nawatts (approved):

service/src/main/java/org/databiosphere/workspacedataservice/logging/LogbookConfiguration.java
jladieu: `/prometheus` should be fine -- that's what the default scrape should use, and non-default scrapes will be human-driven which might actually be interesting to see in the logs.

PR #749, opened 2024-04-26
[AJ-1798] Remove data-import.allowed-schemes configuration

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1798

By default, `ImportValidator` only allows HTTPS import URLs. Attempts to import from URLs with other schemes are rejected with a 400 response.

However, `ImportValidator` can be configured to accept other schemes via the `twds.data-import.allowed-schemes` property. This is used only for tests, to allow a few e2e tests to import test fixtures from file URLs and to allow a Python test to import from an HTTP URL.

However, there is some risk that a misconfiguration would allow users to import file URLs, potentially giving them access to data on the server. To remove that risk, this PR removes the `twds.data-import.allowed-schemes` property.

e2e tests now use a `MockBean` annotation to skip `ImportValidator`'s validation entirely.

To support the Python test, this adds an `ImportValidatorConfiguration` that replaces the default import validator with a `NoopImportValidator` when the `twds.data-import.disable-validation` property is set to true. However, it will only do that when the `local` profile is active. If `twds.data-import.disable-validation` is set to true when the `local` profile is not active, the app will fail to start.

### DIFF
--- a/.github/workflows/release-python-client.yml
+++ b/.github/workflows/release-python-client.yml
@@ -91,7 +91,7 @@ jobs:
           export SPRING_PROFILES_ACTIVE=data-plane,local
           ./gradlew --build-cache build -x test
           # Allow http imports from localhost to test with files from resources served by the mock API server
-          ./gradlew bootRun --args='--twds.data-import.allowed-hosts=localhost --twds.data-import.allow-http-urls=true' &
+          ./gradlew bootRun --args='--twds.data-import.allowed-hosts=localhost --twds.data-import.disable-validation=true' &
           count=20
           until $(curl --output /dev/null --silent --head --fail http://localhost:8080/status); do
             printf '.'
--- a/.github/workflows/release-python-client.yml
+++ b/.github/workflows/release-python-client.yml
@@ -91,7 +91,7 @@ jobs:
           export SPRING_PROFILES_ACTIVE=data-plane,local
           ./gradlew --build-cache build -x test
           # Allow http imports from localhost to test with files from resources served by the mock API server
-          ./gradlew bootRun --args='--twds.data-import.allowed-hosts=localhost --twds.data-import.disable-validation=true' &
+          ./gradlew bootRun --args='--twds.data-import.allowed-hosts=localhost --twds.data-import.require-validation=false' &
           count=20
           until $(curl --output /dev/null --silent --head --fail http://localhost:8080/status); do
             printf '.'
--- a/.github/workflows/release-python-client.yml
+++ b/.github/workflows/release-python-client.yml
@@ -91,7 +91,7 @@ jobs:
           export SPRING_PROFILES_ACTIVE=data-plane,local
           ./gradlew --build-cache build -x test
           # Allow http imports from localhost to test with files from resources served by the mock API server
-          ./gradlew bootRun --args='--twds.data-import.allowed-hosts=localhost --twds.data-import.allowed-schemes=http' &
+          ./gradlew bootRun --args='--twds.data-import.allowed-hosts=localhost --twds.data-import.allow-http-urls=true' &
           count=20
           until $(curl --output /dev/null --silent --head --fail http://localhost:8080/status); do
             printf '.'
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportValidator.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportValidator.java
@@ -1,35 +1,7 @@
 package org.databiosphere.workspacedataservice.dataimport;
 
-import java.net.URI;
-import java.util.Set;
-import java.util.regex.Pattern;
-import org.databiosphere.workspacedataservice.config.DataImportProperties;
 import org.databiosphere.workspacedataservice.generated.ImportRequestServerModel;
-import org.databiosphere.workspacedataservice.service.model.exception.ValidationException;
-import org.springframework.stereotype.Component;
 
-@Component
-public class ImportValidator {
-  private final boolean allowHttpUrls;
-  private final Set<Pattern> allowedHosts;
-
-  public ImportValidator(DataImportProperties dataImportProperties) {
-    this.allowHttpUrls = dataImportProperties.areHttpUrlsAllowed();
-    this.allowedHosts = dataImportProperties.getAllowedHosts();
-  }
-
-  public void validateImport(ImportRequestServerModel importRequest) {
-    URI importUrl = importRequest.getUrl();
-    String urlScheme = importUrl.getScheme();
-    if (!(urlScheme.equals("https") || (urlScheme.equals("http") && allowHttpUrls))) {
-      throw new ValidationException(
-          "Files may not be imported from %s URLs.".formatted(importUrl.getScheme()));
-    }
-
-    if (allowedHosts.stream()
-        .noneMatch(allowedHost -> allowedHost.matcher(importUrl.getHost()).matches())) {
-      throw new ValidationException(
-          "Files may not be imported from %s.".formatted(importUrl.getHost()));
-    }
-  }
+public interface ImportValidator {
+  void validateImport(ImportRequestServerModel importRequest);
 }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportValidator.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportValidator.java
@@ -10,26 +10,24 @@ import org.springframework.stereotype.Component;
 
 @Component
 public class ImportValidator {
-  private final Set<String> allowedSchemes;
+  private final boolean allowHttpUrls;
   private final Set<Pattern> allowedHosts;
 
   public ImportValidator(DataImportProperties dataImportProperties) {
-    this.allowedSchemes = dataImportProperties.getAllowedSchemes();
+    this.allowHttpUrls = dataImportProperties.areHttpUrlsAllowed();
     this.allowedHosts = dataImportProperties.getAllowedHosts();
   }
 
   public void validateImport(ImportRequestServerModel importRequest) {
     URI importUrl = importRequest.getUrl();
-    if (!allowedSchemes.contains(importUrl.getScheme())) {
+    String urlScheme = importUrl.getScheme();
+    if (!(urlScheme.equals("https") || (urlScheme.equals("http") && allowHttpUrls))) {
       throw new ValidationException(
           "Files may not be imported from %s URLs.".formatted(importUrl.getScheme()));
     }
 
-    // File URLs don't have a host to validate.
-    boolean isFileUrl = importUrl.getScheme().equals("file");
-    if (!isFileUrl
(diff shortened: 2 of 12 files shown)


.github/workflows/release-python-client.yml
```diff
@@ -91,7 +91,7 @@ jobs:
           export SPRING_PROFILES_ACTIVE=data-plane,local
           ./gradlew --build-cache build -x test
           # Allow http imports from localhost to test with files from resources served by the mock API server
-          ./gradlew bootRun --args='--twds.data-import.allowed-hosts=localhost --twds.data-import.allowed-schemes=http' &
+          ./gradlew bootRun --args='--twds.data-import.allowed-hosts=localhost --twds.data-import.allow-http-urls=true' &
```
jladieu, line 94: Is there possibly an option here to just have a more aggressive boolean like `disable-import-validation` that just replaces the `ImportValidator` with an always-allow implementation?  This could be coupled with a fail-fast startup check to only allow it to happen in the development environment?

This suggestion is in the spirit of trying to keep the _actual_ prod code path as simple as possible, and avoid introducing support for undesirable schemes, and the complexity the bring.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportValidator.java
```diff
@@ -10,26 +10,24 @@
...
-    this.allowedSchemes = dataImportProperties.getAllowedSchemes();
+    this.allowHttpUrls = dataImportProperties.areHttpUrlsAllowed();
     this.allowedHosts = dataImportProperties.getAllowedHosts();
   }
 
   public void validateImport(ImportRequestServerModel importRequest) {
     URI importUrl = importRequest.getUrl();
-    if (!allowedSchemes.contains(importUrl.getScheme())) {
+    String urlScheme = importUrl.getScheme();
+    if (!(urlScheme.equals("https") || (urlScheme.equals("http") && allowHttpUrls))) {
```
jladieu, line 24: I'd rather see us bypass the check entirely than have the check that will be used to enforce production get more complicated, so I'd advocate for allowing this passthrough by swapping in a different implementation of this class rather than introducing complexity to this one.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJobE2ETest.java
```diff
@@ -57,18 +58,16 @@
...
+@SpringBootTest
 @AutoConfigureMockMvc
 class TdrManifestQuartzJobE2ETest extends TestBase {
   @Autowired private RecordOrchestratorService recordOrchestratorService;
   @Autowired private ImportService importService;
   @Autowired private CollectionService collectionService;
   @Autowired private TdrTestSupport testSupport;
 
+  // Mock ImportValidator to allow importing test data from a file:// URL.
+  @MockBean ImportValidator importValidator;
```
jladieu, line 70: Oh slick!

.github/workflows/release-python-client.yml
nawatts: Ok, what do you think after https://github.com/DataBiosphere/terra-workspace-data-service/pull/749/commits/60d7e155c275f632e3a00912579465ab6d4fdafb?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportValidatorConfiguration.java
```diff
@@ -0,0 +1,26 @@
...
+      name = "twds.data-import.disable-validation",
+      havingValue = "false",
+      matchIfMissing = true)
+  ImportValidator getDefaultImportValidator(DataImportProperties dataImportProperties) {
+    return new DefaultImportValidator(dataImportProperties.getAllowedHosts());
+  }
+
+  @Bean
+  @ConditionalOnProperty(name = "twds.data-import.disable-validation", havingValue = "true")
+  @Profile("local")
```
nawatts, line 22: There's probably a better way to require the `local` profile here. As is, there's no clear error message.

nawatts: Improved in https://github.com/DataBiosphere/terra-workspace-data-service/pull/749/commits/29931da035cc708ec93f2a21832734320c0c7e2f.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/NoopImportValidator.java
```diff
@@ -0,0 +1,13 @@
+package org.databiosphere.workspacedataservice.dataimport;
+
+import org.databiosphere.workspacedataservice.generated.ImportRequestServerModel;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class NoopImportValidator implements ImportValidator {
+  private static final Logger LOGGER = LoggerFactory.getLogger(NoopImportValidator.class);
```
jladieu, line 8: Optional: after I've repeatedly made copy/paste errors with bringing the wrong `.class` over when copying the Logger line to a new class, I've taken to preferring copy-pasting a slightly different logger idiom when adding new ones:

`private final Logger logger = LoggerFactory.getLogger(getClass());`

I think there are some minor tradeoffs when doing it this way, but I like that it doesn't require a manual edit after copypasta.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobDataPlaneE2ETest.java
```diff
@@ -52,17 +53,15 @@
...
-    })
+@SpringBootTest
 class PfbQuartzJobDataPlaneE2ETest {
   @Autowired @SingleTenant WorkspaceId workspaceId;
   @Autowired RecordOrchestratorService recordOrchestratorService;
   @Autowired CollectionService collectionService;
   @Autowired PfbTestSupport testSupport;
   @MockBean WorkspaceManagerDao wsmDao;
+  // Mock ImportValidator to allow importing test data from a file:// URL.
+  @MockBean ImportValidator importValidator;
```
jladieu, line 64: Consider using the `NoopImportValidator` here?

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/rawlsjson/RawlsJsonQuartzJobControlPlaneE2ETest.java
```diff
@@ -22,13 +24,14 @@
...
       // turn off pubsub autoconfiguration for tests
       "spring.cloud.gcp.pubsub.enabled=false",
-      // Allow file imports to test with files from resources.
-      "twds.data-import.allowed-schemes=file",
       // Rawls url must be valid, else context initialization (Spring startup) will fail
       "rawlsUrl=https://localhost/",
     })
 @AutoConfigureMockMvc
 class RawlsJsonQuartzJobControlPlaneE2ETest {
+  // Mock ImportValidator to allow importing test data from a file:// URL.
```
jladieu, line 32: Consider using the `NoopImportValidator`?

(This isn't a big deal, but I guess I'm picturing only having to grep for _one_ thing when trying to figure out all our use of fake/no-op `ImportValidator`).

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJobE2ETest.java
jladieu: Same as others, not a big deal but maybe good spot to use `NoopImportValidator`

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportValidatorConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.env.Environment;
+
+@Configuration
+public class ImportValidatorConfiguration {
+  @Bean
+  @ConditionalOnProperty(
+      name = "twds.data-import.disable-validation",
```
jladieu, line 14: My brain glitched while reading this because of the double-negative boolean logic.

Would you consider calling this `require-validation` and inverting it where it's used?

jladieu (approved):
I had some optional questions/suggestions, but looks good.  Thanks for doing this!  I particularly like how you have a safeguard against disabling validation for non-local profile.

.github/workflows/release-python-client.yml
jladieu: I like!

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobDataPlaneE2ETest.java
nawatts: Would that be via configuration properties similar to before this PR? Or is there another way to tell Spring to use a specific implementation for an interface here?

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportValidatorConfiguration.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/749/commits/8596fdc372ab9612fdd772703db0f19cdfdf031a.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/NoopImportValidator.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/749/commits/45b4b32fb4b2379295f57244f5942fbcc0f7f44b.

Is it worth mentioning this at office hours or something like that? I think we're now up to 3 conventions for creating Loggers.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobDataPlaneE2ETest.java
nawatts: After reading https://www.baeldung.com/spring-beans-integration-test-override, I tried configuration properties but ran into the error where disabling validation is only allowed with the local profile enabled. I wasn't too keen on having this change the profiles enabled for tests, so I left this as-is.

nawatts:
> I particularly like how you have a safeguard against disabling validation for non-local profile.

That was your idea! 😁 
https://github.com/DataBiosphere/terra-workspace-data-service/pull/749#discussion_r1581561234

PR #750, opened 2024-04-26
AJ-1700 Implement `RAWLSJSON` support.

jladieu:
This is logic replicated from import-service to just move a given incoming rawls JSON file from its original location to the upsert location and notify Rawls.

Some prefactoring commits which eased implementation can be more easily reviewed on their own.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
@@ -18,7 +18,6 @@ public class DataImportProperties {
   private String rawlsNotificationsTopic;
   private String statusUpdatesTopic;
   private String statusUpdatesSubscription;
-  private String rawlsJsonDirectImportBucket;
 
   /** Where to write records after import, options are defined by {@link RecordSinkMode} */
   public RecordSinkMode getBatchWriteRecordSink() {
@@ -36,15 +35,6 @@ public class DataImportProperties {
 
   void setRawlsBucketName(String rawlsBucketName) {
     this.rawlsBucketName = rawlsBucketName;
-  }
-
-  /** A bucket from which Rawls JSON files can be directly imported without conversion. */
-  public String getRawlsJsonDirectImportBucket() {
-    return rawlsJsonDirectImportBucket;
-  }
-
-  void setRawlsJsonDirectImportBucket(String rawlsJsonDirectImportBucket) {
-    this.rawlsJsonDirectImportBucket = rawlsJsonDirectImportBucket;
   }
 
   /**
--- a/service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
@@ -6,6 +6,7 @@ import static java.util.Collections.emptySet;
 import java.util.Set;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import org.springframework.lang.Nullable;
 
 /** Properties that dictate how data import processes should behave. */
 public class DataImportProperties {
@@ -29,6 +30,7 @@ public class DataImportProperties {
   }
 
   /** Where to write Rawls JSON files after import. */
+  @Nullable
   public String getRawlsBucketName() {
     return rawlsBucketName;
   }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
@@ -18,6 +18,7 @@ public class DataImportProperties {
   private String rawlsNotificationsTopic;
   private String statusUpdatesTopic;
   private String statusUpdatesSubscription;
+  private String rawlsJsonDirectImportBucket;
 
   /** Where to write records after import, options are defined by {@link RecordSinkMode} */
   public RecordSinkMode getBatchWriteRecordSink() {
@@ -28,12 +29,22 @@ public class DataImportProperties {
     this.batchWriteRecordSink = RecordSinkMode.fromValue(batchWriteRecordSink);
   }
 
+  /** Where to write Rawls JSON files after import. */
   public String getRawlsBucketName() {
     return rawlsBucketName;
   }
 
   void setRawlsBucketName(String rawlsBucketName) {
     this.rawlsBucketName = rawlsBucketName;
+  }
+
+  /** A bucket from which Rawls JSON files can be directly imported without conversion. */
+  public String getRawlsJsonDirectImportBucket() {
+    return rawlsJsonDirectImportBucket;
+  }
+
+  void setRawlsJsonDirectImportBucket(String rawlsJsonDirectImportBucket) {
+    this.rawlsJsonDirectImportBucket = rawlsJsonDirectImportBucket;
   }
 
   /**
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
@@ -15,11 +15,13 @@ import org.databiosphere.workspacedataservice.service.model.exception.Validation
 import org.springframework.lang.Nullable;
 
 public class DefaultImportValidator implements ImportValidator {
+  private static final String SCHEME_HTTPS = "https";
+  private static final String SCHEME_GS = "gs";
   private static final Map<TypeEnum, Set<String>> SUPPORTED_URL_SCHEMES_BY_IMPORT_TYPE =
       Map.of(
-          TypeEnum.PFB, Set.of("https"),
-          TypeEnum.RAWLSJSON, Set.of("gs"),
-          TypeEnum.TDRMANIFEST, Set.of("https"));
+          TypeEnum.PFB, Set.of(SCHEME_HTTPS),
+          TypeEnum.RAWLSJSON, Set.of(SCHEME_GS),
+          TypeEnum.TDRMANIFEST, Set.of(SCHEME_HTTPS));
   private static final Set<Pattern> ALWAYS_ALLOWED_HOSTS =
       Set.of(
           Pattern.compile("storage\\.googleapis\\.com"),
@@ -35,12 +37,12 @@ public class DefaultImportValidator implements ImportValidator {
       Set<Pattern> allowedHttpsHosts, @Nullable String allowedRawlsBucket) {
     var allowedHostsBuilder =
         ImmutableMap.<String, Set<Pattern>>builder()
-            .put("https", Sets.union(ALWAYS_ALLOWED_HOSTS, allowedHttpsHosts));
+            .put(SCHEME_HTTPS, Sets.union(ALWAYS_ALLOWED_HOSTS, allowedHttpsHosts));
 
     if (StringUtils.isNotBlank(allowedRawlsBucket)) {
-      allowedHostsBuilder.put("gs", Set.of(Pattern.compile(allowedRawlsBucket)));
+      allowedHostsBuilder.put(SCHEME_GS, Set.of(Pattern.compile(allowedRawlsBucket)));
     } else {
-      allowedHostsBuilder.put("gs", emptySet());
+      allowedHostsBuilder.put(SCHEME_GS, emptySet());
     }
 
     this.allowedHostsByScheme = allowedHostsBuilder.build();
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
@@ -2,7 +2,9 @@ package org.databiosphere.workspacedataservice.dataimport;
 
 import static java.util.Collections.emptySet;
 
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Sets;
+import io.micrometer.common.util.StringUtils;
 import java.net.URI;
 import java.util.Map;
 import java.util.Set;
@@ -10,6 +12,7 @@ import java.util.regex.Pattern;
 import org.databiosphere.workspacedataservice.generated.ImportRequestServerModel;
 import org.databiosphere.workspacedataservice.generated.ImportRequestServerModel.TypeEnum;
 import org.databiosphere.workspacedataservice.service.model.exception.ValidationException;
+import org.springframework.lang.Nullable;
 
 public class DefaultImportValidator implements ImportValidator {
   private static final Map<TypeEnum, Set<String>> SUPPORTED_URL_SCHEMES_BY_IMPORT_TYPE =
@@ -26,19 +29,25 @@ public class DefaultImportValidator implements ImportValidator {
           Pattern.compile("s3\\.amazonaws\\.com"), // path style legacy global endpoint
           Pattern.compile(".*\\.s3\\.amazonaws\\.com") // virtual host style legacy global endpoint
           );
-  private final Set<Pattern> allowedHosts;
+  private final Map<String, Set<Pattern>> allowedHostsByScheme;
 
-  public DefaultImportValidator(Set<Pattern> allowedHosts) {
-    this.allowedHosts = Sets.union(ALWAYS_ALLOWED_HOSTS, allowedHosts);
+  public DefaultImportValidator(
+      Set<Pattern> allowedHttpsHosts, @Nullable String allowedRawlsBucket) {
+    var allowedHostsBuilder =
+        ImmutableMap.<String, Set<Pattern>>builder()
+            .put("https", Sets.union(ALWAYS_ALLOWED_HOSTS, allowedHttpsHosts));
+
+    if (StringUtils.isNotBlank(allowedRawlsBucket)) {
+      allowedHostsBuilder.put("gs", Set.of(Pattern.compile(allowedRawlsBucket)));
+    } else {
+      allowedHostsBuilder.put("gs", emptySet());
+    }
+
+    this.allowedHostsByScheme = allowedHostsBuilder.build();
   }
 
   private Set<Pattern> getAllowedHosts(ImportRequestServerModel importRequest) {
-    // Allow imports from any GCS bucket.
-    if (importRequest.getUrl().getScheme().equals("gs")) {
-      return Set.of(Pattern.compile(".*"));
-    }
-
-    return allowedHosts;
+    return allowedHostsByScheme.get(importRequest.getUrl().getScheme());
   }
 
   public void validateImport(ImportRequestServerModel importRequest) {
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetailsRetriever.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetailsRetriever.java
@@ -14,6 +14,7 @@ import org.databiosphere.workspacedataservice.shared.model.CollectionId;
 import org.databiosphere.workspacedataservice.shared.model.WorkspaceId;
 import org.springframework.stereotype.Component;
 
+/** Retrieves various common details needed for import jobs. */
 @Component
 public class ImportDetailsRetriever {
 
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetailsRetriever.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetailsRetriever.java
@@ -32,8 +32,6 @@ public class ImportDetailsRetriever {
     String authToken = jobData.getString(ARG_TOKEN);
     Supplier<String> userEmailSupplier = () -> samDao.getUserEmail(BearerToken.of(authToken));
 
-    // Import all the tables and rows inside the PFB.
-    //
     // determine the workspace for the target collection
     CollectionId collectionId = CollectionId.of(targetCollection);
     WorkspaceId workspaceId = collectionService.getWorkspaceId(collectionId);
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetailsRetriever.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetailsRetriever.java
@@ -1,0 +1,41 @@
+package org.databiosphere.workspacedataservice.dataimport;
+
+import static org.databiosphere.workspacedataservice.shared.model.Schedulable.ARG_COLLECTION;
+import static org.databiosphere.workspacedataservice.shared.model.Schedulable.ARG_TOKEN;
+
+import java.util.UUID;
+import java.util.function.Supplier;
+import org.databiosphere.workspacedataservice.jobexec.JobDataMapReader;
+import org.databiosphere.workspacedataservice.recordsink.RawlsAttributePrefixer.PrefixStrategy;
+import org.databiosphere.workspacedataservice.sam.SamDao;
+import org.databiosphere.workspacedataservice.service.CollectionService;
+import org.databiosphere.workspacedataservice.shared.model.BearerToken;
+import org.databiosphere.workspacedataservice.shared.model.CollectionId;
+import org.databiosphere.workspacedataservice.shared.model.WorkspaceId;
+import org.springframework.stereotype.Component;
+
+@Component
+public class ImportDetailsRetriever {
+
+  private final SamDao samDao;
+  private final CollectionService collectionService;
+
+  ImportDetailsRetriever(SamDao samDao, CollectionService collectionService) {
+    this.samDao = samDao;
+    this.collectionService = collectionService;
+  }
+
+  public ImportDetails fetch(UUID jobId, JobDataMapReader jobData, PrefixStrategy prefixStrategy) {
+    // Collect details needed for import
+    UUID targetCollection = jobData.getUUID(ARG_COLLECTION);
+    String authToken = jobData.getString(ARG_TOKEN);
+    Supplier<String> userEmailSupplier = () -> samDao.getUserEmail(BearerToken.of(authToken));
+
+    // Import all the tables and rows inside the PFB.
+    //
+    // determine the workspace for the target collection
+    CollectionId collectionId = CollectionId.of(targetCollection);
+    WorkspaceId workspaceId = collectionService.getWorkspaceId(collectionId);
+    return new ImportDetails(jobId, userEmailSupplier, workspaceId, collectionId, prefixStrategy);
+  }
+}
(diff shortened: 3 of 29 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetailsRetriever.java
```diff
@@ -0,0 +1,41 @@
...
+
+  private final SamDao samDao;
+  private final CollectionService collectionService;
+
+  ImportDetailsRetriever(SamDao samDao, CollectionService collectionService) {
+    this.samDao = samDao;
+    this.collectionService = collectionService;
+  }
+
+  public ImportDetails fetch(UUID jobId, JobDataMapReader jobData, PrefixStrategy prefixStrategy) {
```
jladieu, line 28: Extracting this allowed me to dry up logic that was in both PFB and TDR quartz jobs, and prevent me from duplicating it a 3rd time in the new RawlsJson job.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportValidator.java
```diff
@@ -3,19 +3,16 @@
...
 
-@Component
 public class ImportValidator {
   private final Set<String> allowedSchemes;
   private final Set<Pattern> allowedHosts;
 
-  public ImportValidator(DataImportProperties dataImportProperties) {
-    this.allowedSchemes = dataImportProperties.getAllowedSchemes();
-    this.allowedHosts = dataImportProperties.getAllowedHosts();
+  ImportValidator(Set<String> allowedSchemes, Set<Pattern> allowedHosts) {
```
jladieu, line 13: @nawatts FYI in case this PR lands ahead of your changes, this will probably trigger some conflicts

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -72,19 +57,15 @@ public class RawlsRecordSink implements RecordSink {
...
+      ObjectMapper mapper, GcsStorage storage, PubSub pubSub, ImportDetails details) {
+    Blob blob = storage.createBlob(RawlsJsonQuartzJob.rawlsJsonBlobName(details.jobId()));
     return new RawlsRecordSink(
-        new RawlsAttributePrefixer(importDetails.prefixStrategy()),
+        new RawlsAttributePrefixer(details.prefixStrategy()),
         JsonWriter.create(storage.getOutputStream(blob), mapper),
-        pubSub,
-        importDetails,
-        blob);
+        new RawlsJsonPublisher(pubSub, details, blob, /* isUpsert= */ true));
```
jladieu, line 68: Having to new up a `RawlsJsonPublisher` with all of the stuff used when `publish` gets called felt really clumsy to me, but I don't have a better way to pass `details`, `blob`, and the hardcoded `isUpsert` through.

I have a _hunch_ that some kind of `@FunctionalInterface` trick might help here, such that you could do something like this:

```
return new RawlsRecordSink(
        new RawlsAttributePrefixer(details.prefixStrategy()),
        JsonWriter.create(storage.getOutputStream(blob), mapper),
        publisher -> publisher.publish(details, blob, /* isUpsert= */ true));
```

But I'm not familiar with that enough to implement it here.  Would welcome ideas of what that might look like!

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -115,7 +96,9 @@ public void deleteBatch(RecordType recordType, List<Record> records) {
   @Override
   public void close() {
     jsonWriter.close();
-    publishToPubSub(importDetails);
+
+    // TODO: add failure detection so we don't publish on failures
```
jladieu, line 100: Would welcome ideas on how to accomplish this TODO.  All my early attempts were pretty gross.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -164,31 +147,6 @@ private AttributeValue maybeCoerceRelation(Object originalValue) {
     return AttributeValue.of(originalValue);
   }
 
-  static String getBlobName(UUID jobId) {
-    return "%s.rawlsUpsert".formatted(jobId.toString());
-  }
-
-  private void publishToPubSub(ImportDetails importDetails) {
```
jladieu, line 171: Moved to `RawlsJsonPublisher`

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSinkFactory.java
```diff
@@ -25,9 +25,6 @@ public RawlsRecordSinkFactory(ObjectMapper mapper, GcsStorage storage, PubSub pu
     this.pubSub = pubSub;
   }
 
-  // TODO(AJ-1589): make prefix assignment dynamic. However, of note: the prefix is currently
```
jladieu, line 28: Just a forgotten but addressed TODO

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
```diff
@@ -84,13 +86,11 @@ public GenericJobServerModel createImport(
       }
     }
 
-    importValidator.validateImport(importRequest);
+    importValidatorFactory.create(importRequest.getType()).validateImport(importRequest);
```
jladieu, line 89: Using a Factory allowed me to easily bypass the fact that the `ImportValidator` didn't allow `gs://`, however @nawatts separate work on AJ-1798 looks like it might solve for this differently.  We'll need to keep in mind order of merge here to sort out the conflicts.

nawatts: #751 allows `ImportValidator` to accept `gs://` URLs for Rawls JSON imports.

jladieu:
FYI I pushed up a couple amendments to address non-TODO sonar warnings.

jladieu:
Going to make some changes to zip up with @nawatts 's related changes, so putting this back into draft until I'm done with that.

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
```diff
@@ -73,12 +72,11 @@ public void execute(JobExecutionContext context) throws org.quartz.JobExecutionE
       jobDao.running(jobId);
       observation.event(Observation.Event.of("job.running"));
       // look for an auth token in the Quartz JobDataMap
-      String authToken = getJobDataString(context.getMergedJobDataMap(), ARG_TOKEN);
+      String authToken = jobData.getString(ARG_TOKEN);
```
jladieu, line 75: `getString` crashes if the token is missing so this can never come back as `null`

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/QuartzJob.java
```diff
@@ -104,69 +102,12 @@ public void execute(JobExecutionContext context) throws org.quartz.JobExecutionE
...
-      String requestId =
-          getJobDataString(context.getMergedJobDataMap(), MDCServletRequestListener.MDC_KEY);
+      String requestId = reader.getString(MDCServletRequestListener.MDC_KEY);
       MDC.put(MDCServletRequestListener.MDC_KEY, requestId);
     } catch (Exception e) {
       // noop
     }
   }
-
-  /**
```
jladieu, line 117: Moved all this into `JobDataMapReader`

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/rawlsjson/RawlsJsonQuartzJob.java
```diff
@@ -31,7 +51,27 @@ protected void annotateObservation(Observation observation) {
...
+
+  private ImportDetails getImportDetails(UUID jobId, JobDataMapReader jobData) {
+    return importDetailsRetriever.fetch(jobId, jobData, PrefixStrategy.NONE);
+  }
+
+  private void publishToRawls(UUID jobId, JobDataMapReader jobData, Blob destination) {
+    new RawlsJsonPublisher(
+            pubSub, getImportDetails(jobId, jobData), destination, jobData.getBoolean("isUpsert"))
+        .publish();
+  }
```
davidangb, line 72: the `getBoolean` here will fail if the user didn't supply a value, but our `options` is not required. How about defaulting isUpsert to be true if the user didn't specify?

service/src/main/resources/application-control-plane.yml
```diff
@@ -10,7 +10,10 @@ twds:
   data-import:
     batch-write-record-sink: "rawls"
     succeed-on-completion: false
+    # Name of the Google Cloud Storage bucket where JSON files are uploaded for import.
     rawls-bucket-name: ${SERVICE_GOOGLE_BUCKET}
+    # Name of the Google Cloud Storage bucket where JSON files can be directly imported from.
+    rawls-json-direct-import-bucket: ${RAWLS_JSON_DIRECT_IMPORT_BUCKET:}
```
davidangb, line 16: same comment as in the properties class - is this necessary?

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/rawlsjson/RawlsJsonQuartzJobTest.java
```diff
@@ -1,19 +1,135 @@
...
+    JobExecutionContext mockContext =
+        stubJobContext(jobId, getUri(incoming), collectionId.id(), isUpsert);
+
+    // act
+    testSupport.buildRawlsJsonQuartzJob().execute(mockContext);
+
+    // assert
+    assertSingleBlobWritten(rawlsJsonBlobName(jobId));
+    assertPubSubMessage(expectedPubSubMessageFor(jobId, isUpsert));
+  }
```
davidangb, line 96: if you default isUpsert to true, we could use a test here to validate that

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
davidangb: This looks like a job for `Supplier`!

```java
  private final Supplier<RawlsJsonPublisher> publisherSupplier;

…

  public static RawlsRecordSink create(
      ObjectMapper mapper, GcsStorage storage, PubSub pubSub, ImportDetails details) {
    Blob blob = storage.createBlob(RawlsJsonQuartzJob.rawlsJsonBlobName(details.jobId()));
    return new RawlsRecordSink(
        new RawlsAttributePrefixer(details.prefixStrategy()),
        JsonWriter.create(storage.getOutputStream(blob), mapper),
        // create the publisher supplier via the () -> shorthand
        () -> new RawlsJsonPublisher(pubSub, details, blob, /* isUpsert= */ true));
  }

…

  @Override
  public void close() {
    jsonWriter.close();

    // make sure to get the RawlsJsonPublisher from the supplier before calling publish()
    publisherSupplier.get().publish();
  }

Comment (davidangb) [reply service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java:68]: but that said I don't think it's a big deal to eagerly create the `RawlsJsonPublisher` - it's a pretty small class that doesn't do anything other than storing its details until you call `publish()` on it

File: service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java (hunk 4 above)
Comment (davidangb) [reply service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java:100]: I think the only realistic way to accomplish this is:
* create a member variable as a volatile boolean, or possibly a state-machine enum, initialized to a false/don't-publish value
* as a last step before publishing, once everything else is complete, have calling code call "setReadyToPublish(true)" to flip the boolean/enum
* here in `close()`, check the value of that boolean/enum and condition the pubsub on its value

**Review by davidangb — COMMENTED** [review]
some questions about the config properties, and some responses to your questions, mostly small comments!

File: service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java [hunk 13]
```diff
@@ -36,6 +38,15 @@ void setRawlsBucketName(String rawlsBucketName) {
...
 
+  /** A bucket from which Rawls JSON files can be directly imported without conversion. */
+  public String getRawlsJsonDirectImportBucket() {
+    return rawlsJsonDirectImportBucket;
+  }
+
+  void setRawlsJsonDirectImportBucket(String rawlsJsonDirectImportBucket) {
+    this.rawlsJsonDirectImportBucket = rawlsJsonDirectImportBucket;
+  }
+
```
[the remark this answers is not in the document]
jladieu: This is an artifact of my first cut at this, which directly copied the logic from import-service which restricts the import bucket to a particular configured location.  After discussing with @nawatts this felt like an unnecessary restriction since end users could just directly call the upsert endpoint anyway.  I forgot to remove this after basing it off Nick's changes.  Removed now! :)

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetailsRetriever.java
```diff
@@ -0,0 +1,42 @@
...
+    // Collect details needed for import
+    UUID targetCollection = jobData.getUUID(ARG_COLLECTION);
+    String authToken = jobData.getString(ARG_TOKEN);
+    Supplier<String> userEmailSupplier = () -> samDao.getUserEmail(BearerToken.of(authToken));
+
+    // Import all the tables and rows inside the PFB.
```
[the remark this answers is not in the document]
jladieu: removed

service/src/main/resources/application-control-plane.yml
jladieu: removed

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/rawlsjson/RawlsJsonQuartzJob.java
jladieu: Hmm...thanks for this question...it made me start looking around and uncovered some gaps/questions.

Referring to: https://github.com/DataBiosphere/terra-workspace-data-service/blob/83f33984473426e3b52dc4bb4e36bbb72c6de73a/service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java#L96-L128

1) I think (in this PR) the `isUpsert` command isn't yet forwarded at all from the `ImportRequest`, so... bug found, nice catch!!
2) While looking at fixing (1), I rediscovered the one-off logic for forwarding `tdrSyncPermissions` down into the Arguments of `Schedulable`; I think this probably isn't a super durable fix for passing `options` down and want to revisit this now that I've got another option I need to put in here, but which should be defaulted to `true`.
3) I noticed that `Job.newJob` doesn't preserve the passed `options` for storage in the Jobs table, so the only thing stored will be the JobType & URI.
4) The `stubJobContext` we use in tests doesn't do a great job of catching the bug in (1) because it loses the mapping of ImportRequest -> Schedulable

I don't want to propagate the pattern in (2) such that `ImportService` has to know all the possible options for all the possible subclasses of Job to serialize them, and shoves them into the `Schedulable` with a default value.   I'd much rather have all `options` passed through into what gets stored in the `job` record and have the jobs take care of honoring or defaulting them after deserializing.

To that end, I propose:

- Pass `options` directly through into the JSON that gets stored as `JobInput`, by adjusting `Job.newJob` to include options as well
  - if necessary, we could do an allowlist of keys to prevent spam, but ideally I don't think `ImportService` should be doing defaulting here.
- Get rid of the logic that deals with `tdrSyncPermissions` here and shift the defaulting over to the Job subclass only.
- Adjust the various test fixtures which `stubJobContext` to do a better job of requiring that the actual `ImportRequest.options` in the test are populated and better thread through the options into the Job stored in the database.
- (Optional) Stop using the Schedulable arguments as a place to store args, except maybe the `AUTH_TOKEN` which I believe we actively want to avoid storing in DB, and instead store these in the Job Inputs?

Thoughts?  Wondering what makes sense here in this list, and which parts make sense to tackle as part of unblocking RAWLSJSON?

service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
```diff
@@ -18,6 +18,7 @@ public class DataImportProperties {
   private String rawlsNotificationsTopic;
   private String statusUpdatesTopic;
   private String statusUpdatesSubscription;
+  private String rawlsJsonDirectImportBucket;
```
nawatts, line 21: Is this ever used?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/rawlsjson/RawlsJsonQuartzJob.java
```diff
@@ -31,7 +51,27 @@ protected void annotateObservation(Observation observation) {
 
   @Override
   protected void executeInternal(UUID jobId, JobExecutionContext context) {
-    logger.atInfo().log(
-        "RawlsJsonQuartzJob.executeInternal: jobId={} - not yet implemented (no op)!", jobId);
+    JobDataMapReader jobData = JobDataMapReader.fromContext(context);
+    URI sourceUri = jobData.getURI(ARG_URL);
+    Blob destination = moveBlob(sourceUri, rawlsJsonBlobName(jobId));
```
nawatts, line 56: Just curious... why do we need to move the blob?

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/JobDataMapReader.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+
+  /**
+   * Retrieve a Boolean from a JobDataMap. Throws a JobExecutionException if the value is not
+   * found/null or cannot be parsed into a Boolean.
+   */
+  public Boolean getBoolean(String key) {
+    try {
+      return jobDataMap.getBooleanValue(key);
```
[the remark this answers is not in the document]
jladieu: Sonar steers us to use primitives if possible because the boxed primitives _can_ be `@Nullable` (and trigger unexpected conditional failures) even if they shouldn't be.

service/src/main/java/org/databiosphere/workspacedataservice/config/DataImportProperties.java
jladieu: Oops, nope!  Just missed this in my cleanup after rebase.  Fixed.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/rawlsjson/RawlsJsonQuartzJob.java
jladieu: I'm just copying the behavior we're migrating.  Import-service did a `mv`: https://github.com/broadinstitute/import-service/blob/a818658d0850818804235c5dc7f239c0de95c8cf/app/translate.py#L56

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/rawlsjson/RawlsJsonQuartzJobTest.java
jladieu: I added `isUpsert` defaulting, but testing this proved to be difficult due to a combination of how the `Map` passed down as Job handles `null` and our limited `JobDataReader` which needs improvement.  This all sprouts from what I think is a limited handling of pass-through options which should be addressed holistically by https://broadworkbench.atlassian.net/browse/AJ-1809.

I also added TODOs & a disabled test for that ticket.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/rawlsjson/RawlsJsonQuartzJob.java
jladieu: Added a ticket for this https://broadworkbench.atlassian.net/browse/AJ-1809

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
jladieu: I opted not to do this.

jladieu:
>I think the only realistic way to accomplish this is:

Deferred to a TODO/ticket

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/rawlsjson/RawlsJsonQuartzJob.java
davidangb: We definitely don't *have* to move the blob, and if the blob-moving is troublesome we could remove it. I think it's _ideal_ to move the blob because then by the time it gets to Rawls, it is named by its jobid, and that naming makes human debugging a lot easier. Yes we could always look up the job in cWDS's db to get the input file, but that's an annoying extra step.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/rawlsjson/RawlsJsonQuartzJobTest.java
davidangb: You could write a unit test against `ImportService.createImport` and/or `ImportService.maybePassThroughOptions` to validate the defaulting - it doesn't have to be here in this file

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
```diff
@@ -72,19 +58,15 @@ public class RawlsRecordSink implements RecordSink {
    *
    * @param mapper used to generate the stream of JSON entities
    * @param storage where the JSON stream will be written
-   * @param pubSub to notify upon JSON stream completion
-   * @param importDetails to use for generating the message to send to PubSub
+   * @param details to use for generating the message to send to PubSub
```
calypsomatic, line 61: Did we lose a param javadoc here?

davidangb (approved):
:+1: with one comment about test coverage inline.

A repercussion of NOT validating the incoming bucket name for RAWLSJSON imports: if we don't control the incoming bucket, we can't guarantee that the file-moving will work correctly, or even the file-reading. If a user supplies a bucket for which we don't have write access, the "delete original" step of the file-move will fail. If a user supplies a bucket for which we don't have read access, everything will fail. Part of the reasoning behind validating the incoming bucket is to control this - since we own that bucket and have it fairly locked down, we can make some assumptions in our code. I think it's worth putting that validation back in but am open to debate on that.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
jladieu: I turned this into a TODO/ticket as well.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/rawlsjson/RawlsJsonQuartzJobTest.java
jladieu: Added https://github.com/DataBiosphere/terra-workspace-data-service/pull/750/commits/6fa7625238816db0bb99030a22fd1e6602df5f8c to introduce coverage for the defaulting.

jladieu:
>A repercussion of NOT validating the incoming bucket name for RAWLSJSON imports: if we don't control the incoming bucket, we can't guarantee that the file-moving will work correctly, or even the file-reading. If a user supplies a bucket for which we don't have write access, the "delete original" step of the file-move will fail. If a user supplies a bucket for which we don't have read access, everything will fail. Part of the reasoning behind validating the incoming bucket is to control this - since we own that bucket and have it fairly locked down, we can make some assumptions in our code. I think it's worth putting that validation back in but am open to debate on that.

I'd prefer to maintain the existing behavior during migration and do the renaming for easier operation; I'll start by assuming the bucket used for input is the same as the bucket used to relay to Rawls.

service/src/main/resources/application-control-plane.yml
```diff
@@ -10,7 +10,9 @@ twds:
   data-import:
     batch-write-record-sink: "rawls"
     succeed-on-completion: false
-    rawls-bucket-name: ${SERVICE_GOOGLE_BUCKET}
+    # Name of the Google Cloud Storage bucket where JSON files are uploaded for import and sent to
+    # Rawls for upserting.
+    rawls-bucket-name: ${SERVICE_GOOGLE_BUCKET:}
```
jladieu, line 15: I needed to add the `:` to make this optional since it's not set in `data-plane`.  Without the `:` there's a startup crash while trying to configure the `DefaultImportValidator` which now depends on this, but we can't rely on `SERVICE_GOOGLE_BUCKET` being set everywhere.

davidangb (approved):
ty!

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsRecordSink.java
jladieu: Re-added this, sorry for the churn here.  Thanks for spotting it!

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
```diff
@@ -26,19 +31,25 @@ public class DefaultImportValidator implements ImportValidator {
...
+    var allowedHostsBuilder =
+        ImmutableMap.<String, Set<Pattern>>builder()
+            .put(SCHEME_HTTPS, Sets.union(ALWAYS_ALLOWED_HOSTS, allowedHttpsHosts));
 
-  private Set<Pattern> getAllowedHosts(ImportRequestServerModel importRequest) {
-    // Allow imports from any GCS bucket.
-    if (importRequest.getUrl().getScheme().equals("gs")) {
-      return Set.of(Pattern.compile(".*"));
+    if (StringUtils.isNotBlank(allowedRawlsBucket)) {
+      allowedHostsBuilder.put(SCHEME_GS, Set.of(Pattern.compile(allowedRawlsBucket)));
```
nawatts, line 43: Technically, I think the configured bucket should be the only host allowed for the Rawls JSON _import type_, not the `gs://` _scheme_. But practically, those two are currently the same.

service/src/main/resources/application-control-plane.yml
nawatts: Instead, could it be set to an empty string in `application-data-plane.yml`? Since it's not actually optional for the control plane.

nawatts (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
jladieu: My logic here is since we currently only support `gs` for `RAWLSJSON` we could keep it like this until we need to adjust it again for some other circumstance.

jladieu:
>Instead, could it be set to an empty string in application-data-plane.yml? Since it's not actually optional for the control plane.

So if I take this approach, the error that caused me to make these optional still happens for every test that has `@ActiveProfile("control-plane")`.

The workaround is to adjust those same tests to:
```
@TestPropertySource(
    properties = {
      // SERVICE_GOOGLE_BUCKET isn't available in unit tests, so we hardcode this to empty string
      "twds.data-import.rawls-bucket-name="
})
```

This change is required across 7 different tests.  Do you think it's worth it to keep the config unchanged?

nawatts:
😭 I vote for your current approach then.

PR #751, opened 2024-04-26
[AJ-1798] Support gs:// URLs for Rawls JSON imports in ImportValidator

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1798

Unlike the other import types, Rawls JSON import URLs will be `gs://` URLs. `ImportValidator` must accept those URLs.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
@@ -32,7 +32,7 @@ public class DefaultImportValidator implements ImportValidator {
     this.allowedHosts = Sets.union(ALWAYS_ALLOWED_HOSTS, allowedHosts);
   }
 
-  private Set<Pattern> getAllowedHostsForImportRequest(ImportRequestServerModel importRequest) {
+  private Set<Pattern> getAllowedHosts(ImportRequestServerModel importRequest) {
     // Allow imports from any GCS bucket.
     if (importRequest.getUrl().getScheme().equals("gs")) {
       return Set.of(Pattern.compile(".*"));
@@ -52,7 +52,7 @@ public class DefaultImportValidator implements ImportValidator {
       throw new ValidationException("Files may not be imported from %s URLs.".formatted(urlScheme));
     }
 
-    if (getAllowedHostsForImportRequest(importRequest).stream()
+    if (getAllowedHosts(importRequest).stream()
         .noneMatch(allowedHost -> allowedHost.matcher(importUrl.getHost()).matches())) {
       throw new ValidationException(
           "Files may not be imported from %s.".formatted(importUrl.getHost()));
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
@@ -1,13 +1,22 @@
 package org.databiosphere.workspacedataservice.dataimport;
 
+import static java.util.Collections.emptySet;
+
 import com.google.common.collect.Sets;
 import java.net.URI;
+import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
 import org.databiosphere.workspacedataservice.generated.ImportRequestServerModel;
+import org.databiosphere.workspacedataservice.generated.ImportRequestServerModel.TypeEnum;
 import org.databiosphere.workspacedataservice.service.model.exception.ValidationException;
 
 public class DefaultImportValidator implements ImportValidator {
+  private static final Map<TypeEnum, Set<String>> SUPPORTED_URL_SCHEMES_BY_IMPORT_TYPE =
+      Map.of(
+          TypeEnum.PFB, Set.of("https"),
+          TypeEnum.RAWLSJSON, Set.of("gs"),
+          TypeEnum.TDRMANIFEST, Set.of("https"));
   private static final Set<Pattern> ALWAYS_ALLOWED_HOSTS =
       Set.of(
           Pattern.compile("storage\\.googleapis\\.com"),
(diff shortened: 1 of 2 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
```diff
@@ -1,13 +1,22 @@
...
 import org.databiosphere.workspacedataservice.generated.ImportRequestServerModel;
+import org.databiosphere.workspacedataservice.generated.ImportRequestServerModel.TypeEnum;
 import org.databiosphere.workspacedataservice.service.model.exception.ValidationException;
 
 public class DefaultImportValidator implements ImportValidator {
+  private static final Map<TypeEnum, Set<String>> SUPPORTED_URL_SCHEMES_BY_IMPORT_TYPE =
+      Map.of(
+          TypeEnum.PFB, Set.of("https"),
+          TypeEnum.RAWLSJSON, Set.of("gs"),
+          TypeEnum.TDRMANIFEST, Set.of("https"));
```
jladieu, line 19: When scouting some of this out I noticed that TDR allowed gs too: https://github.com/broadinstitute/import-service/blob/a818658d0850818804235c5dc7f239c0de95c8cf/app/translate.py#L30

Should this `Set` include `gs` as well?  If so, this is where having a platform-based intersection might come in handy (eg: only allow `gs` in the control-plane / GCP, exclude it in data-plane / Azure)

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
```diff
@@ -24,15 +33,21 @@ public DefaultImportValidator(Set<Pattern> allowedHosts) {
...
-    if (!urlScheme.equals("https")) {
-      throw new ValidationException(
-          "Files may not be imported from %s URLs.".formatted(importUrl.getScheme()));
+    if (!schemesSupportedForImportType.contains(urlScheme)) {
+      throw new ValidationException("Files may not be imported from %s URLs.".formatted(urlScheme));
     }
 
-    if (allowedHosts.stream()
-        .noneMatch(allowedHost -> allowedHost.matcher(importUrl.getHost()).matches())) {
+    // Imports should be allowed from any GCS bucket.
```
jladieu, line 46: When I was implementing the `ImportValidatorFactory` in the [adjacent change](https://github.com/DataBiosphere/terra-workspace-data-service/pull/750#discussion_r1581557346) I was trying to come up with a way to have behavior different for different types without having to make significant changes the existing Validator logic which is nice and simple.

Maybe instead of `skipHostValidation` you could just instantiate an instance of ImportValidator with a wildcard matcher for the `allowedHosts`?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
nawatts: I wasn't sure if our existing TDR code can actually handle `gs://` URLs, so I was kicking that can down the road.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
nawatts: I'm not really convinced adding a factory layer is any simpler than another condition here...

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
jladieu: Caveat: I recognize my position here is beginning to border on bike-shedding...but I think I'm okay with over-rotating a bit because the code we're fiddling with is a security check.

I agree factory is its own complexity, but I think it's a Java-idiomatic form of complexity that only inflicts itself on the instantiation path rather than conditional/branching complexity in the execution path of the actual security/logic check.  I'm advocating to keep the execution path / security concern as simple as possible.

nawatts: The goal of keeping the execution path simple is to reduce the chances of a developer making a mistake and introducing a vulnerability by making the intent easier to understand, correct? I'd claim that adding a layer of indirection makes the intent and actual behavior _more_ difficult to reason about.

nawatts: > that only inflicts itself on the instantiation path rather than conditional/branching complexity in the execution path of the actual security/logic check

Since all of this is happening in the context of one import request, it seems to me like the factory and instantiation of a validator class are part of the execution path of the security check.

jladieu (approved):

nawatts:
Rebased after merging #749.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
nawatts: Also, I suspect `gs://` URLs for TDR imports is a relic from before TDR switched to signed URLs for imports. When it passed `gs://` URLs to import service, we were plagued by issues with permission propagation delays.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
```diff
@@ -23,15 +32,27 @@ public DefaultImportValidator(Set<Pattern> allowedHosts) {
     this.allowedHosts = Sets.union(ALWAYS_ALLOWED_HOSTS, allowedHosts);
   }
 
+  private Set<Pattern> getAllowedHostsForImportRequest(ImportRequestServerModel importRequest) {
```
jladieu, line 35: Nit: I think it'd be okay to just call this method `getAllowedHosts` as the param type and name make it clear that it's for the request.

jladieu (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/751/commits/a700dc574d5c9fcbc1c1fac26eb4040e91c8611e.

jladieu (approved):

PR #759, opened 2024-04-29
AJ-1748: attribute models

davidangb:
This PR:
* introduces the beginnings of an attribute model hierarchy:
  * `Attribute` base interface
    * `ArrayAttribute` abstract class (not currently used, but will be by the end of AJ-1748)
    * `ScalarAttribute` abstract class
      * `RelationAttribute` class
* refactors the existing standalone `RelationAttribute` class to be part of the attribute hierarchy. This incurs a bunch of renaming in calling classes.

There should be zero functional changes in this PR; it's just renaming/refactoring.


service/src/main/java/org/databiosphere/workspacedataservice/shared/model/RelationAttribute.java
davidangb, line 1: this file was moved into the attribute package; dunno why it's showing as deleted instead of renamed.

jladieu (approved):

yuliadub (approved):

PR #761, opened 2024-04-30
AJ-1748: introduce JsonAttribute

davidangb:
This PR introduces a `JsonAttribute` class into the attribute hierarchy (see #759). TSV uploads now deserialize json columns, and arrays of json, into `JsonAttribute` instead of `Map<String, Object>`.

This is pre-work to supporting arrays-of-arrays. Since `JsonAttribute` can support either a json object or a json array, it is more flexible; the previous `Map<String, Object>` can only support a json object.

However, as prework, there should be no functional changes in this PR; it is only a refactor.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -728,17 +728,20 @@ public class RecordDao {
         return LocalDateTime.parse(sVal, DateTimeFormatter.ISO_LOCAL_DATE_TIME);
       }
     }
+    // TSV-based uploads deserialize json as JsonAttribute.
     if (attVal instanceof JsonAttribute jsonAttribute) {
       return jsonAttribute.sqlValue();
     }
-    //    if (attVal instanceof Map<?, ?>) {
-    //      try {
-    //        return objectMapper.writeValueAsString(attVal);
-    //      } catch (JsonProcessingException e) {
-    //        LOGGER.error("Could not serialize Map to json string", e);
-    //        throw new RuntimeException(e);
-    //      }
-    //    }
+    // json-based APIs deserialize json as LinkedHashMap<String, Object>. Handle those here.
+    // TODO AJ-1748: how to deserialize json-based APIs into JsonAttribute instead?
+    if (attVal instanceof Map<?, ?>) {
+      try {
+        return objectMapper.writeValueAsString(attVal);
+      } catch (JsonProcessingException e) {
+        LOGGER.error("Could not serialize Map to json string", e);
+        throw new RuntimeException(e);
+      }
+    }
     if (typeMapping.isArrayType()) {
       return getArrayValues(attVal, typeMapping);
     }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -48,6 +48,7 @@ import org.databiosphere.workspacedataservice.service.model.exception.BatchDelet
 import org.databiosphere.workspacedataservice.service.model.exception.ConflictException;
 import org.databiosphere.workspacedataservice.service.model.exception.InvalidRelationException;
 import org.databiosphere.workspacedataservice.service.model.exception.MissingObjectException;
+import org.databiosphere.workspacedataservice.service.model.exception.SerializationException;
 import org.databiosphere.workspacedataservice.shared.model.AttributeComparator;
 import org.databiosphere.workspacedataservice.shared.model.Record;
 import org.databiosphere.workspacedataservice.shared.model.RecordAttributes;
@@ -733,8 +734,7 @@ public class RecordDao {
       try {
         return objectMapper.writeValueAsString(jsonAttribute.sqlValue());
       } catch (JsonProcessingException e) {
-        LOGGER.error("Could not serialize JsonAttribute to json string", e);
-        throw new RuntimeException(e);
+        throw new SerializationException("Could not serialize JsonAttribute to json string", e);
       }
     }
     // json-based APIs deserialize json as LinkedHashMap<String, Object>. Handle those here.
@@ -743,8 +743,7 @@ public class RecordDao {
       try {
         return objectMapper.writeValueAsString(attVal);
       } catch (JsonProcessingException e) {
-        LOGGER.error("Could not serialize Map to json string", e);
-        throw new RuntimeException(e);
+        throw new SerializationException("Could not serialize Map to json string", e);
       }
     }
     if (typeMapping.isArrayType()) {
@@ -790,8 +789,8 @@ public class RecordDao {
                 try {
                   return objectMapper.writeValueAsString(el);
                 } catch (JsonProcessingException e) {
-                  LOGGER.error("Could not serialize array element to json string", e);
-                  throw new RuntimeException(e);
+                  throw new SerializationException(
+                      "Could not serialize array element to json string", e);
                 }
               })
           .toList()
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -730,7 +730,12 @@ public class RecordDao {
     }
     // TSV-based uploads deserialize json as JsonAttribute.
     if (attVal instanceof JsonAttribute jsonAttribute) {
-      return jsonAttribute.sqlValue();
+      try {
+        return objectMapper.writeValueAsString(jsonAttribute.sqlValue());
+      } catch (JsonProcessingException e) {
+        LOGGER.error("Could not serialize JsonAttribute to json string", e);
+        throw new RuntimeException(e);
+      }
     }
     // json-based APIs deserialize json as LinkedHashMap<String, Object>. Handle those here.
     // TODO AJ-1748: how to deserialize json-based APIs into JsonAttribute instead?
@@ -1008,7 +1013,7 @@ public class RecordDao {
         return getArrayValue(pgArray.getArray(), typeMapping);
       }
       if (typeMapping == DataTypeMapping.JSON) {
-        return new JsonAttribute(objectMapper.readTree(object.toString()), objectMapper);
+        return new JsonAttribute(objectMapper.readTree(object.toString()));
       }
       return object;
     }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -53,6 +53,7 @@ import org.databiosphere.workspacedataservice.shared.model.Record;
 import org.databiosphere.workspacedataservice.shared.model.RecordAttributes;
 import org.databiosphere.workspacedataservice.shared.model.RecordColumn;
 import org.databiosphere.workspacedataservice.shared.model.RecordType;
+import org.databiosphere.workspacedataservice.shared.model.attributes.JsonAttribute;
 import org.jetbrains.annotations.NotNull;
 import org.postgresql.jdbc.PgArray;
 import org.slf4j.Logger;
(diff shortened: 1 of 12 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -203,9 +206,28 @@ private JsonNode parseToJsonNode(String val) {
...
+    }
+    return null;
+  }
+
+  /**
+   * @deprecated use tryJsonObject instead
+   */
+  @Deprecated(since = "2024-04-30")
+  public boolean isValidJson(String val) {
+    return tryJsonObject(val) != null;
```
davidangb, line 230: This is a rewrite of `isValidJson` that returns either the JsonNode if valid, or null if not valid. It prevents code that parses the json once to see if it's valid, and then a second time to get the JsonNode.

To prevent scope-creeping this PR, I left `isValidJson` in place but delegating to `tryJsonObject`.

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/RecordAttributes.java
```diff
@@ -28,6 +28,7 @@ public class RecordAttributes {
    */
   @JsonCreator
   public RecordAttributes(Map<String, Object> attributes) {
+    // AJ-858: post-process the attributes map to look for JsonAttribute, dates, etc.
     this.attributes = new TreeMap<>(attributes);
   }
```
davidangb, line 33: I haven't explored this route fully: this is (I believe) the method that Jackson calls to instantiate a RecordAttributes. We could conceivably add code here that looks for values of `Map<String, Object>` and transforms them into `JsonAttributes` - doing that kind of post-processing would offer consistency between JSON and TSV paths.

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/attributes/JsonAttribute.java
```diff
@@ -0,0 +1,50 @@
...
+  /**
+   * When serializing the JsonAttribute class, what value should Jackson serialize? We should only
+   * serialize the underlying value, not the JsonAttribute wrapper itself.
+   *
+   * @return the value to use when serializing
+   */
+  @JsonValue
+  public JsonNode jsonValue() {
+    return this.value;
+  }
```
davidangb, line 26: this `jsonValue` method may want to move up to a superclass eventually, depending on how the rest of the attribute hierarchy shakes out when/if we implement it

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/attributes/JsonAttribute.java
```diff
@@ -0,0 +1,50 @@
...
+  @Override
+  public String toString() {
+    // use the supplied ObjectMapper to generate the toString value. This ensures serialization is
+    // consistent with the settings configured on the ObjectMapper.
+    try {
+      return mapper.writeValueAsString(this.value);
+    } catch (JsonProcessingException e) {
+      throw new RuntimeException(e);
+    }
+  }
```
davidangb, line 37: I don't like that this requires an `ObjectMapper` to be supplied in the constructor. The alternative is:
```java
return this.value.toString();
```
which "will produce ... valid JSON using default settings" according to its Javadocs. Thus it would produce different json than what we have set in `JsonConfig` for settings like `WRITE_DATES_AS_TIMESTAMPS` and `WRITE_BIGDECIMAL_AS_PLAIN`.

_Potentially_, we could let this toString() use the defaults, then make sure that anywhere we need to serialize a `JsonAttribute` we use an ObjectMapper in the calling site and never rely on `toString` to serialize.

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/attributes/JsonAttribute.java
```diff
@@ -0,0 +1,50 @@
...
+  public boolean equals(Object obj) {
+    // don't consider the mapper in equals
+    return super.equals(obj);
+  }
+
+  @Override
+  public int hashCode() {
+    // don't consider the mapper in hashcode
+    return super.hashCode();
+  }
```
davidangb, line 49: the `equals` and `hashCode` overrides are only necessary because this class has an `ObjectMapper` member (which should be ignored). If we could get rid of that, we could use the default equals/hashCode

service/src/test/java/org/databiosphere/workspacedataservice/tsv/TsvDeserializerTest.java
```diff
@@ -67,12 +66,14 @@ void cellToAttributeTest(String input, Object expected) {
...
-          actual,
-          "cellToAttribute for input value <%s> should return <%s>".formatted(input, expected));
+      /*
+      } else if (expected instanceof Map) {
+        // map types can differ; don't check instanceof
+        assertEquals(
+            expected,
+            actual,
+            "cellToAttribute for input value <%s> should return <%s>".formatted(input, expected));
+       */
```
davidangb, line 76: we no longer have Maps as expected values, so this block is obsolete and I should just delete it

service/src/test/java/org/databiosphere/workspacedataservice/tsv/TsvJsonArgumentsProvider.java
```diff
@@ -145,12 +146,14 @@ public Stream<? extends Arguments> provideArguments(ExtensionContext context) {
...
 
+        // TODO AJ-1748: where should this test live?
         // array of JSON objects
+        /*
         Arguments.of(
             "[{\"value\":\"foo\"},{\"value\":\"bar\"},{\"value\":\"baz\"}]",
             List.of(Map.of("value", "foo"), Map.of("value", "bar"), Map.of("value", "baz")),
             false),
-
+         */
```
davidangb, line 156: these two commented-out tests are no longer valid, since Json and TSV now deserialize to different objects (Map and JsonAttribute) and therefore are not equivalent.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -192,6 +194,7 @@ public boolean isNumericValue(String sVal) {
     }
   }
 
+  @Nullable
   private JsonNode parseToJsonNode(String val) {
```
jladieu, line 198: Might this be a good spot to use an `Optional<JsonNode>` instead of a `@Nullable`?

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/attributes/JsonAttribute.java
jladieu: >I don't like that this requires an ObjectMapper to be supplied in the constructor.

I think I prefer passing `ObjectMapper` to using the `this.value.toString()` for serialization.

>Potentially, we could let this toString() use the defaults, then make sure that anywhere we need to serialize a JsonAttribute we use an ObjectMapper in the calling site and never rely on toString to serialize.

I like what you're describing here.  My hunch is to avoid using `toString()` for any kind of serialization and instead rely exclusively on Jackson + `@JsonValue`.  That said, this is a nice safeguard against accidental use and okay to leave as is IMO until we take on an endeavor to cut away from `toString()`

jladieu (approved):
I think this all looks fine!

service/src/main/java/org/databiosphere/workspacedataservice/tsv/TsvDeserializer.java
```diff
@@ -130,15 +130,9 @@ public Object cellToAttribute(String val) {
       }
     }
     // JSON objects
-    // TODO: change the "is*" methods in inferer to return their value so we don't parse twice?
```
calypsomatic, line 133: Does this fix the TODO?

davidangb: it fixes the TODO at this location; there is a similar TODO up above at line 124 that is not yet fixed

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
davidangb: Update: I have removed `isValidJson` entirely and switched all callers over to `tryJsonObject`; it wasn't too big.

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/attributes/JsonAttribute.java
davidangb: Update: I have removed `ObjectMapper` from the constructor, and modified `toString()` so callers can't effectively use it for json.

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/attributes/JsonAttribute.java
davidangb: Update: since I removed `ObjectMapper`, these are no longer necessary so I have removed them.

service/src/test/java/org/databiosphere/workspacedataservice/tsv/TsvDeserializerTest.java
davidangb: deleted.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
davidangb: now returns an `Optional`.

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -203,9 +208,19 @@ private JsonNode parseToJsonNode(String val) {
     }
   }
 
-  public boolean isValidJson(String val) {
+  /**
+   * Attempts to parse the input as json. If the input parsed as a valid json object (not array!),
+   * return that ObjectNode; else return null.
```
calypsomatic, line 213: The comments seem out of sync with the behavior; it doesn't return null but an empty Optional

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -727,6 +728,12 @@ private Object getValueForSql(Object attVal, DataTypeMapping typeMapping) {
         return LocalDateTime.parse(sVal, DateTimeFormatter.ISO_LOCAL_DATE_TIME);
       }
     }
+    // TSV-based uploads deserialize json as JsonAttribute.
+    if (attVal instanceof JsonAttribute jsonAttribute) {
+      return jsonAttribute.sqlValue();
+    }
+    // json-based APIs deserialize json as LinkedHashMap<String, Object>. Handle those here.
+    // TODO AJ-1748: how to deserialize json-based APIs into JsonAttribute instead?
```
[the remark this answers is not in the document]
nawatts: > `RecordAttributes` defines a `Map<String, Object>` - thus we are explicitly instructing Jackson to deserialize into a Map

I don't understand... isn't that `Map<String, Object>` the map of attribute names to attribute values, not the values themselves?

service/src/test/java/org/databiosphere/workspacedataservice/tsv/TsvJsonArgumentsProvider.java
nawatts: They no longer pass for `TsvJsonEquivalenceTest`. But they should still be valid cases for `TsvDeserializerTest`, no?

jladieu (approved):
Looks great!

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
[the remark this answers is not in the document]
davidangb: yes, it is the map of attribute names to attribute values … but by specifying the `Object` part of `Map<String, Object>` we lose control over anything else Jackson does. We can't really say "deserialize values as `Object` unless it's a json packet, and then deserialize it into `JsonAttribute`". My eventual goal is to have `RecordAttributes` defined as `Map<String, Attribute>` and then we have control over the whole value hierarchy, but that's a bit off in the future.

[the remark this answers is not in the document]
nawatts: > `RecordAttributes` defines a `Map<String, Object>` - thus we are explicitly instructing Jackson to deserialize into a Map

Oh, I misunderstood that the emphasis of this statement was on the `Object` value type vs the `Map` itself.

service/src/test/java/org/databiosphere/workspacedataservice/tsv/TsvJsonArgumentsProvider.java
davidangb: good call, moved to `TsvDeserializerTest`.

nawatts (approved):

PR #764, opened 2024-04-30
AJ-1810: better handling of ?statuses= param for job listing

davidangb:
For the `jobsInInstanceV1` API, which lists jobs within a given collection:
* If the user specifies an invalid status value, return 400 BadRequest instead of 500 Internal Server Error. This prevents spurious Sentry issues.
* If the user specifies an empty status value - `?status=` - treat it the same as if they specified no param at all and return all jobs

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/JobController.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/JobController.java
@@ -2,7 +2,7 @@ package org.databiosphere.workspacedataservice.controller;
 
 import static org.databiosphere.workspacedataservice.generated.GenericJobServerModel.StatusEnum;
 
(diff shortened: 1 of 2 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/controller/JobController.java
```diff
@@ -2,6 +2,7 @@
 
 import static org.databiosphere.workspacedataservice.generated.GenericJobServerModel.StatusEnum;
 
+import jakarta.annotation.Nullable;
```
nawatts, line 5: Other files use the Spring annotation.

```suggestion
import org.springframework.lang.Nullable
```

nawatts (approved):

calypsomatic (approved):

PR #766, opened 2024-05-01
[AJ-1777] Require a protected workspace for imports from certain sources

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1777

Some data needs to be stored in workspaces with secure monitoring enabled. This allows configuring WDS to enforce requirements on the destination workspace based on the import URL.

`twds.data-import.sources` contains a list of `ImportSourceConfig` objects. Each of those objects contains a list of URL patterns to identify imports from the source and options for imports from those sources. Currently, the only option is to require a workspace with a protected data policy.

`ImportValidator` now looks at the import request and the destination workspace. If the import URL matches a source configured to require a protected data policy, it gets the destination workspace description from WSM and checks that the workspace has a protected data policy.

This approach is more flexible than it needs to be for this. It's intended to support arbitrary configuration for specific data sources. A follow on PR will also add support for an option to require a private workspace.


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
```diff
@@ -57,5 +71,37 @@ public void validateImport(ImportRequestServerModel importRequest) {
...
+  }
+
+  private boolean checkWorkspaceHasProtectedDataPolicy(UUID workspaceId) {
+    try {
+      WorkspaceDescription workspace = wsmDao.getWorkspace(workspaceId);
+      return Optional.ofNullable(workspace.getPolicies()).orElse(emptyList()).stream()
+          .anyMatch(
+              wsmPolicyInput ->
+                  wsmPolicyInput.getNamespace().equals("terra")
+                      && wsmPolicyInput.getName().equals("protected-data"));
```
calypsomatic, line 98: Will we ever be checking for other policies?  Either way, should "protected-data" be a final string at the top so it's not a magic word here?

nawatts: Oh, yeah. Mean to do that. It seems like ideally these constants would exported by one of the shared packages.

But it looks like WSM also hard-codes them. https://github.com/DataBiosphere/terra-workspace-manager/blob/32555d8deb4b5075089d2bbd5a22b7d217289785/service/src/main/java/bio/terra/workspace/service/policy/TpsUtilities.java#L13

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/766/commits/e4226eaddc362b83c022526d8482befa9793550a.

service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerClientFactory.java
```diff
@@ -3,11 +3,15 @@
...
-  ReferencedGcpResourceApi getReferencedGcpResourceApi(String authToken);
+  ReferencedGcpResourceApi getReferencedGcpResourceApi(@Nullable String authToken);
 
-  ResourceApi getResourceApi(String authToken);
+  ResourceApi getResourceApi(@Nullable String authToken);
 
-  ControlledAzureResourceApi getAzureResourceApi(String authToken);
+  ControlledAzureResourceApi getAzureResourceApi(@Nullable String authToken);
+
+  WorkspaceApi getWorkspaceApi(@Nullable String authToken);
```
nawatts, line 16: Sonar complained about passing `null` to this method, but it's applicable to all methods.

nawatts:
Rebased to resolve conflicts after #750 merged.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportRequirementsFactory.java
```diff
@@ -0,0 +1,31 @@
...
+            requiresProtectedDataPolicy = true;
+          }
+          break;
+        }
+      }
+    }
+
+    return new ImportRequirements(requiresProtectedDataPolicy);
+  }
+}
```
davidangb, line 31: I wouldn't block the PR on this, because the data scales involved are so small (one `ImportSourceConfig` with a handful of urls in it). But the ivory-tower programmer in me sees a code smell in the nested loops and nested if conditions, and wants to approach this more from a FP perspective. My suggestion is to do as much work in the constructor as possible: eagerly get the data structures the way you want them once at construction time, and then at runtime when `getRequirementsForImport` is invoked repeatedly, it's less work there. Something like:

```java
public class ImportRequirementsFactory {
  private final List<Pattern> protectedPatterns;

  public ImportRequirementsFactory(List<ImportSourceConfig> sources) {
    // identify a single list of url patterns that require protected workspaces
    this.protectedPatterns =
        sources.stream()
            // only those ImportSourceConfig that require protected data
            .filter(ImportSourceConfig::requireProtectedDataPolicy)
            // extract the urls from those sources
            .map(ImportSourceConfig::urls)
            // flatten
            .flatMap(List::stream)
            .toList();
  }

  public ImportRequirements getRequirementsForImport(URI importUri) {
    boolean requiresProtectedDataPolicy =
        protectedPatterns.stream().anyMatch(x -> x.matcher(importUri.toString()).find());
    return new ImportRequirements(requiresProtectedDataPolicy);
  }
}
```

I am guessing you wrote it this way to allow future flexibility, if/when we add new constraints to sources and if/when we add new sources. I don't have a good sense of when that would happen, or if it will ever happen - my gut says to leave application.yml as-is (it's nicely future-proofed) but let this code be efficient until we need something more flexible.

service/src/main/java/org/databiosphere/workspacedataservice/policy/PolicyUtils.java
```diff
@@ -0,0 +1,10 @@
+package org.databiosphere.workspacedataservice.policy;
+
+public class PolicyUtils {
+  public static final String TERRA_NAMESPACE = "terra";
+  public static final String PROTECTED_DATA_POLICY_NAME = "protected-data";
+
+  private PolicyUtils() {
+    throw new IllegalStateException("Utility class");
+  }
+}
```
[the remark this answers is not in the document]
nawatts: Today I learned. Sonar told me to add the private constructor.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportRequirementsFactory.java
nawatts: Yeah, that was the idea. And there will shortly be another requirement (for private workspaces). With the current set of sources, it will practically be the same as the protected data requirement. However, I wanted to separate them to avoid what we have now where import service has its own definition of "protected data" that doesn't exactly align with the "protected data" policy.

I can definitely rewrite to be more functional, but I suspect any efficiency will be dominated by the matching against the URL patterns.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportValidator.java
```diff
@@ -1,7 +1,8 @@
 package org.databiosphere.workspacedataservice.dataimport;
 
+import java.util.UUID;
 import org.databiosphere.workspacedataservice.generated.ImportRequestServerModel;
 
 public interface ImportValidator {
-  void validateImport(ImportRequestServerModel importRequest);
+  void validateImport(ImportRequestServerModel importRequest, UUID destinationWorkspaceId);
```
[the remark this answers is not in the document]
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/766/commits/5cd2cbfa73d4ae4a81368506d7e89e5ed50d4361.

service/src/main/java/org/databiosphere/workspacedataservice/policy/PolicyUtils.java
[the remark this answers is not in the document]
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/766/commits/7cb0bb5a4347c4fa203468c6f6246c916687c2fa.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportRequirementsFactory.java
nawatts: How does it look after https://github.com/DataBiosphere/terra-workspace-data-service/pull/766/commits/74fd761086fa5b01cf1ff2c11ce36e941b5295bb?

service/src/main/java/org/databiosphere/workspacedataservice/policy/PolicyUtils.java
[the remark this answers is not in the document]
nawatts: Sonar still wants the private constructor...

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportRequirementsFactory.java
davidangb: I definitely think the later commit reads better - thank you. Like I said this isn't a big deal - it's not going to be a bottleneck, and I understand changes are coming. I'm good with it.

service/src/main/java/org/databiosphere/workspacedataservice/policy/PolicyUtils.java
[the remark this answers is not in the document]
davidangb: bah, I think that's silly. The compiler won't let you use the public constructor. But fine, add the private constructor back in to satisfy sonar. You could also make the class `final` at the same time.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportRequirementsFactory.java
nawatts: I suspected there might be some feedback on the for loops as I was writing them... 😁

davidangb (approved):

service/src/main/java/org/databiosphere/workspacedataservice/policy/PolicyUtils.java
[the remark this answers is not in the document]
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/766/commits/fd129145d80e0eb91f302b9823ca553cbc280b89.

PR #767, opened 2024-05-02
AJ-1700: Fix bug while copying `RAWLSJSON` file.

jladieu:
Before this change, the `moveBlob` method would attempt to get the blob based on the original target to return a reference to it.  However, the original `BlobId` used as a reference also contained the original `generation`, and therefore the `getBlob` call would always return `null` and cause an NPE.  Instead, we use the resultant `Blob` from the copy, which has the correct new `generation`, and should actually exist.

Notes about the verifying issue:
- The local mock storage doesn't behave exactly the way the actual/production storage does with regards to existing blobs and how errors are surfaced, so it's not 100% reliable at detecting issues like this.
- In order to reproduce the actual issue I actually temporarily switched over to using the real version of the storage class in the E2E test, hardcoded the bucket name, and used the debugger to step through and understand the problem was related to the wrong `generation` on the `BlobId`.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/rawlsjson/RawlsJsonQuartzJob.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/rawlsjson/RawlsJsonQuartzJob.java
@@ -6,6 +6,7 @@ import static org.databiosphere.workspacedataservice.shared.model.Schedulable.AR
 import static org.databiosphere.workspacedataservice.shared.model.job.JobType.DATA_IMPORT;
 
 import com.google.cloud.storage.Blob;
+import com.google.cloud.storage.BlobId;
 import io.micrometer.observation.Observation;
 import io.micrometer.observation.ObservationRegistry;
 import java.net.URI;
@@ -59,7 +60,7 @@ public class RawlsJsonQuartzJob extends QuartzJob {
   }
 
   private Blob moveBlob(URI sourceUri, String desiredBlobName) {
-    return storage.moveBlob(sourceUri, storage.createBlob(desiredBlobName));
+    return storage.moveBlob(sourceUri, BlobId.of(storage.getBucketName(), desiredBlobName));
   }
 
   private ImportDetails getImportDetails(UUID jobId, JobDataMapReader jobData) {
--- a/service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorage.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorage.java
@@ -9,6 +9,7 @@ import com.google.cloud.spring.storage.GoogleStorageResource;
 import com.google.cloud.storage.Blob;
 import com.google.cloud.storage.BlobId;
 import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.Storage.BlobTargetOption;
 import com.google.cloud.storage.Storage.CopyRequest;
 import com.google.cloud.storage.StorageOptions;
 import com.google.common.annotations.VisibleForTesting;
@@ -17,6 +18,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.URI;
 import java.nio.channels.Channels;
+import java.util.Optional;
 import org.databiosphere.workspacedataservice.config.DataImportProperties;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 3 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorage.java
```diff
@@ -114,13 +116,20 @@ public void deleteBlob(String blobName) {
...
+  public Blob moveBlob(URI sourceUri, BlobId target) {
     BlobId source = BlobId.fromGsUtilUri(sourceUri.toString());
-    storage.copy(CopyRequest.newBuilder().setSource(source).setTarget(target).build());
-    Blob copiedBlob = storage.get(target.getBlobId());
+    BlobTargetOption precondition =
+        Optional.ofNullable(storage.get(target))
+            .map(Blob::getGeneration)
+            .map(BlobTargetOption::generationMatch)
+            .orElseGet(BlobTargetOption::doesNotExist);
+    Blob copiedBlob =
```
jladieu, line 126: This is the key change to fix the bug:

Instead of
```
Blob copiedBlob = storage.get(target.getBlobId());
```

which returns `null` because there no longer exists a `Blob` with `target`'s `generation`, we return the `Blob` of `getResult` which has an updated generation and corresponds with an existing blob.

service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorage.java
```diff
@@ -114,13 +116,20 @@ public void deleteBlob(String blobName) {
    *
    * @return Blob reference to the file's new location
    */
-  public Blob moveBlob(URI sourceUri, Blob target) {
+  public Blob moveBlob(URI sourceUri, BlobId target) {
     BlobId source = BlobId.fromGsUtilUri(sourceUri.toString());
-    storage.copy(CopyRequest.newBuilder().setSource(source).setTarget(target).build());
-    Blob copiedBlob = storage.get(target.getBlobId());
+    BlobTargetOption precondition =
```
jladieu, line 121: This precondition logic is technically optional but is suggested by the GCS docs as a way of preventing race conditions.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/rawlsjson/RawlsJsonQuartzJob.java
```diff
@@ -59,7 +60,7 @@ protected void executeInternal(UUID jobId, JobExecutionContext context) {
   }
 
   private Blob moveBlob(URI sourceUri, String desiredBlobName) {
-    return storage.moveBlob(sourceUri, storage.createBlob(desiredBlobName));
+    return storage.moveBlob(sourceUri, BlobId.of(storage.getBucketName(), desiredBlobName));
```
jladieu, line 63: Instead of creating an empty blob to start, we just pass in the desired `BlobId` which saves a roundtrip to GCS.

service/src/test/java/org/databiosphere/workspacedataservice/storage/GcsStorageTest.java
```diff
@@ -161,8 +162,7 @@ void moveExistingBlob() throws IOException {
 
     // Act
     URI sourceUri = getUri(sourceBlob);
-    Blob targetBlob = storage.createBlob(newBlobName);
-    Blob movedBlob = storage.moveBlob(sourceUri, targetBlob);
+    Blob movedBlob = storage.moveBlob(sourceUri, BlobId.of(storage.getBucketName(), newBlobName));
```
jladieu, line 165: Just to more correctly match what the actual code does, we don't create an empty blob up front.

service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorage.java
calypsomatic: What is a `generation`?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/rawlsjson/RawlsJsonQuartzJob.java
nawatts: I suspect this alone may have fixed the bug as well, since the new `BlobId` wouldn't have the generation.

nawatts (approved):

service/src/main/java/org/databiosphere/workspacedataservice/storage/GcsStorage.java
jladieu: It's like the timestamp / version # on the GCS file.  It gets updated every time something changes about a blob.

davidangb (approved):

PR #768, opened 2024-05-02
AJ-1748: arrays of arrays in TSV

davidangb:
This supports arrays-of-arrays for TSV upload and our json-based REST APIs. It treats arrays-of-arrays as `ARRAY_OF_JSON`. So, we still get a count of the number of first-level elements, but each first-level element is a opaque json blob; we don't provide any special functionality for the nested structures.

`ARRAY_OF_JSON` _can_ support the following use cases:
* **arrays of json objects**: already supported before this PR.
* **arrays of arrays**: added support in this PR.
* **arrays of mixed types**: _partial_ support in this PR.
  * If at least one of the elements of the array is a nested array or object, we treat it as `ARRAY_OF_JSON`. Example: `["one", "two", ["nested", "array"], "three]`
  * If none of the mixed types is an array or object, we maintain the current behavior and treat it as an `ARRAY_OF_STRING`. Example: `[1, 2, "three", 4]`

Changing that last bullet point so that we use `ARRAY_OF_JSON` for _any_ mixed array is trivial and we could certainly add it. It's a behavior change and I want to vet it with the team before making that change.

_**important**_:  this PR only handles arrays-of-arrays for TSV and REST APIs. It does NOT fix it for PFB/TDR. I will handle that in yet another follow-on PR.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -7,7 +7,6 @@ import static org.databiosphere.workspacedataservice.service.model.ReservedNames
 import static org.databiosphere.workspacedataservice.service.model.exception.InvalidNameException.NameType.ATTRIBUTE;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Maps;
@@ -1048,11 +1047,11 @@ public class RecordDao {
     }
 
     private Object[] convertToJson(Object object) throws JsonProcessingException {
+      // json arrays are returned from the db as String[]
       String[] jsonArray = (String[]) object;
-      Object[] result = new Object[jsonArray.length];
+      JsonAttribute[] result = new JsonAttribute[jsonArray.length];
       for (int i = 0; i < jsonArray.length; i++) {
-        result[i] =
-            objectMapper.readValue(jsonArray[i], new TypeReference<Map<String, Object>>() {});
+        result[i] = new JsonAttribute(objectMapper.readTree(jsonArray[i]));
       }
       return result;
     }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
@@ -5,6 +5,7 @@ import static org.databiosphere.workspacedataservice.service.RelationUtils.getTy
 import static org.databiosphere.workspacedataservice.service.RelationUtils.getTypeValueForList;
 import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.ARRAY_OF_DATE;
 import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.ARRAY_OF_DATE_TIME;
+import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.ARRAY_OF_JSON;
 import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.ARRAY_OF_RELATION;
 import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.ARRAY_OF_STRING;
 import static org.databiosphere.workspacedataservice.service.model.DataTypeMapping.BOOLEAN;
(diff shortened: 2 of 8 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -1048,11 +1047,11 @@ private LocalDate[] convertToLocalDate(Object object) {
...
       String[] jsonArray = (String[]) object;
-      Object[] result = new Object[jsonArray.length];
+      JsonAttribute[] result = new JsonAttribute[jsonArray.length];
       for (int i = 0; i < jsonArray.length; i++) {
-        result[i] =
-            objectMapper.readValue(jsonArray[i], new TypeReference<Map<String, Object>>() {});
+        result[i] = new JsonAttribute(objectMapper.readTree(jsonArray[i]));
       }
       return result;
     }
```
davidangb, line 1057: `convertToJson` is only called from `getArrayValue`

service/src/main/java/org/databiosphere/workspacedataservice/service/DataTypeInferer.java
```diff
@@ -242,6 +246,26 @@ private <T> DataTypeMapping findArrayType(List<T> list) {
...
+    //     list.stream().map(Object::getClass).distinct().toList();
+    // and seeing if they are homogenous. If you do this, be careful of multiple
+    // classes which we treat the same, e.g. BigInteger and BigDecimal.
+
+    // is any element of this List itself a List or a Map? This indicates nested
+    // structures, so we should treat this as an array of json.
+    return list.stream()
+        .anyMatch(element -> element instanceof List<?> || element instanceof Map<?, ?>);
+  }
+
```
davidangb, line 268: I don't like that we have essentially duplicate logic in `DataTypeInferer` and `TsvDeserializer` to look at this. But again, until we use a consistent and fully-defined attribute model hierarchy to represent these values, we'll need divergent logic.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerMockMvcTest.java
```diff
@@ -3165,7 +3165,8 @@ void jsonArrayElementsMustBeObjectsOrBeEncodedAsStrings() throws Exception {
             .andReturn();
     RecordResponse record = fromJson(mvcSingleResult, RecordResponse.class);
 
-    List<String> expectedValue = List.of("{value=foo}", "{value=bar}", "baz");
+    // note the mixed values here
+    List<Object> expectedValue = List.of(Map.of("value", "foo"), Map.of("value", "bar"), "baz");
```
davidangb, line 3169: this test's input is an array of [json object, json object, string]. Previously we treated this as ARRAY_OF_STRING but now we treat it as ARRAY_OF_JSON, so the test assertions need to change.

service/src/test/java/org/databiosphere/workspacedataservice/controller/TsvDownloadTest.java
```diff
@@ -49,6 +53,7 @@
 import org.springframework.test.annotation.DirtiesContext;
 import org.springframework.test.context.ActiveProfiles;
 
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
davidangb, line 56: here, and in multiple other tests, we need to use an autowired `ObjectMapper` instance to generate test inputs. This requires the argument providers to not be static, which in turn requires `Lifecycle.PER_CLASS` on the class.

service/src/test/java/org/databiosphere/workspacedataservice/controller/TsvDownloadTest.java
```diff
@@ -207,10 +213,44 @@ private static Stream<Arguments> tsvExemplarData() {
...
+                new JsonAttribute(mapper.readTree("67")),
+                new JsonAttribute(mapper.readTree("{\"some\":\"object\",\"with\":[\"nesting\"]}"))),
+            ARRAY_OF_JSON,
+            "\"[[1,2,3],[\"\"four\"\",\"\"five\"\"],67,{\"\"some\"\":\"\"object\"\",\"\"with\"\":[\"\"nesting\"\"]}]\""));
   }
 
-  @ParameterizedTest(name = "TSV download for a {1} should be correct")
+  // upload a given input value using the JSON APIs, and validate WDS created the expected datatype.
+  // download that column as a TSV and validate the download matches the expected value.
+  @ParameterizedTest(name = "{1} with val {2}")
```
davidangb, line 253: shortened the test names and included more arguments to make them easier to read in IntelliJ

yuliadub (approved):

nawatts (approved):
> - If none of the mixed types is an array or object, we maintain the current behavior and treat it as an ARRAY_OF_STRING. Example: [1, 2, "three", 4]
>
> Changing that last bullet point so that we use ARRAY_OF_JSON for any mixed array is trivial and we could certainly add it. It's a behavior change and I want to vet it with the team before making that change.

I vote yes for that example. As is, it seems like we're ignoring the type information that the user has provided.

However... I could see an argument that the quotes are ambiguous in a TSV. Like is `[1, 2, "three", 4]` 3 numbers and a string or is it 4 strings, one of which includes the quotes?

davidangb:
> I vote yes for that example

Thank you - and I do too. But I will tackle that in a follow-on PR to avoid holding this one up

PR #771, opened 2024-05-03
AJ-1771: id filter for paginated record query

davidangb:
This PR modifies the pre-existing `POST /{instanceid}/search/{v}/{type}` API.

Before this PR, that API accepted a payload of:
```json
{
  "offset": 0,
  "limit": 10,
  "sort": "ASC",
  "sortAttribute": "string"
}
```

After this PR, it will accept a payload of:
```json
{
  "offset": 0,
  "limit": 10,
  "sort": "ASC",
  "sortAttribute": "string",
  "filter": {
    "ids": [
      "string"
    ]
  }
}
```

That is, it will now accept a `filter` object, which can contain an `ids` array of strings.

Both `filter` and `filter.ids` are optional for backwards compatibility. If you omit them, the API will behave as it previously did.

If you DO specify values in `filter.ids`, it will behave like a SQL `IN` clause, and filter the query results to only those records whose ids match one of the supplied values.

Some functional behaviors:
* sorting of results is still controlled by the `sort` and `sortAttribute` params. Sorting does not consider the order of ids in `filter.ids`.
* results still respect the `limit` and `offset` values. If you specify 20 `filter.ids` but a `limit` of 10, you'll get (up to) 10 records back. 

You may want to read through the test cases in `RecordControllerSearchFilterTest` in this PR for detailed examples of behavior.

Anyone who is using a Java/Python client will need to update their client after this PR to get access to the new models.

---

## Open Questions:
1. If the user specifies more `filter.ids` values than `limit`, should we return only `limit` records, or should we throw a 400 Bad Request error?
2. If the user specifies an empty `filter.ids: []`, should we return 0 results always, or should we treat that equivalent as having omitted `filter.ids` and return all results? Or, throw a 400 Bad Request?

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/shared/model/SearchRequest.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/shared/model/SearchRequest.java
@@ -59,7 +59,6 @@ public class SearchRequest {
     this.sortAttribute = sortAttribute;
   }
 
-  @Nullable
   public Optional<SearchFilter> getFilter() {
     return filter;
   }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/shared/model/SearchRequest.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/shared/model/SearchRequest.java
@@ -8,6 +8,7 @@ public class SearchRequest {
   private int offset = 0;
   private SortDirection sort = SortDirection.ASC;
   @Nullable private String sortAttribute = null;
+  @Nullable private SearchFilter filter = null;
 
   public SearchRequest(int limit, int offset, SortDirection sort) {
     this.limit = limit;
@@ -55,5 +56,14 @@ public class SearchRequest {
 
   public void setSortAttribute(String sortAttribute) {
     this.sortAttribute = sortAttribute;
+  }
+
+  @Nullable
+  public SearchFilter getFilter() {
+    return filter;
+  }
+
+  public void setFilter(@Nullable SearchFilter filter) {
+    this.filter = filter;
   }
 }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/shared/model/SearchRequest.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/shared/model/SearchRequest.java
@@ -1,5 +1,6 @@
 package org.databiosphere.workspacedataservice.shared.model;
 
+import java.util.Optional;
 import org.springframework.lang.Nullable;
 
 public class SearchRequest {
@@ -8,7 +9,7 @@ public class SearchRequest {
   private int offset = 0;
   private SortDirection sort = SortDirection.ASC;
   @Nullable private String sortAttribute = null;
-  @Nullable private SearchFilter filter = null;
+  private Optional<SearchFilter> filter = Optional.empty();
 
   public SearchRequest(int limit, int offset, SortDirection sort) {
     this.limit = limit;
@@ -59,11 +60,12 @@ public class SearchRequest {
   }
 
   @Nullable
-  public SearchFilter getFilter() {
+  public Optional<SearchFilter> getFilter() {
     return filter;
   }
 
-  public void setFilter(@Nullable SearchFilter filter) {
+  public void setFilter(
+      @SuppressWarnings("OptionalUsedAsFieldOrParameterType") Optional<SearchFilter> filter) {
     this.filter = filter;
   }
 }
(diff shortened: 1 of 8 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/shared/model/SearchRequest.java
```diff
@@ -56,4 +57,13 @@ public String getSortAttribute() {
   public void setSortAttribute(String sortAttribute) {
     this.sortAttribute = sortAttribute;
   }
+
+  @Nullable
+  public SearchFilter getFilter() {
```
jladieu, line 62: Consider `Optional`?

(I bias toward `Optional` on the way out (to avoid leaking @Nullable concerns to collaborator classes), `@Nullable` on the way in (to avoid Intellij warnings about using `Optional` for parameters & constructor args).

jladieu (approved):
Awesome PR writeup describing the before/after.  I like the approach from an API perspective, and the tests are excellent.

One _optional_ comment about `Optional` :)

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/SearchRequest.java
davidangb: Now using `Optional`.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerSearchFilterTest.java
```diff
@@ -0,0 +1,217 @@
...
+    List<String> expected = List.of("006", "008");
+    executeTest(searchRequestPayload, expected, testCase);
+  }
+
+  @Test
+  void idOrderIsIgnored() {
+    String testCase = "when filter.ids contains more matching ids than the query limit";
+    String searchRequestPayload =
+        """
+{"limit": 3, "sortAttribute": "sortByMe", "sort": "DESC", "filter": {"ids": ["004", "006", "008"]}}
```
calypsomatic, line 172: This is the same as the limit, not more

calypsomatic (approved):

nawatts (approved):
> If the user specifies more filter.ids values than limit, should we return only limit records, or should we throw a 400 Bad Request error?

I would vote for return only `limit` records. This is the difference between "search" and "get"... search suggests we don't know for sure if all those IDs exist.

> If the user specifies an empty filter.ids: [], should we return 0 results always, or should we treat that equivalent as having omitted filter.ids and return all results?

I would vote return 0 results on the basis of the behavior is more explicit. Or, I could see returning 400 for a search query that will always yield 0 results.

service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerSearchFilterTest.java
davidangb: good catch, thank you - I updated the `testCase` description to reflect what the test is actually doing

calypsomatic:
> ## Open Questions:
> 1. If the user specifies more `filter.ids` values than `limit`, should we return only `limit` records, or should we throw a 400 Bad Request error?
> 2. If the user specifies an empty `filter.ids: []`, should we return 0 results always, or should we treat that equivalent as having omitted `filter.ids` and return all results? Or, throw a 400 Bad Request?

1. I think just `limit` records.  A 400 seems too annoying if the user made a mistake.
2. Definitely not 400, for the same reason as above.  I'm torn between the other two choices, though - I lean towards 0 results since it does seem like they're specifically requesting none, but that seems like a dumb thing to do so it seems like it would be a mistake and should just count as omitting it.

nawatts:
> Definitely not 400, for the same reason as above. I'm torn between the other two choices, though - I lean towards 0 results since it does seem like they're specifically requesting none, but that seems like a dumb thing to do so it seems like it would be a mistake and should just count as omitting it.

A 400 response is useful for telling the user that they made a mistake and prompting them to correct it, vs us trying to guess at what they might have intended and maybe getting it wrong.

yuliadub:
> 
> ## Open Questions:
> 1. If the user specifies more `filter.ids` values than `limit`, should we return only `limit` records, or should we throw a 400 Bad Request error?

I still feel strongly that if limit is less then the number of ids, an error should be returned - mostly based on prior experience of people getting confused and thinking something is wrong with the client vs catching their own error. In the end, its an easy thing to resolve if people ask -- but this still means getting to communicate with customers about behavior that is intended (which can be an annoying thing for them). With all that said, this can also be resolved by providing good client documentation and leaving it at that (i.e. that limit will always apply, no matter the filter) 

> 2. If the user specifies an empty `filter.ids: []`, should we return 0 results always, or should we treat that equivalent as having omitted `filter.ids` and return all results? Or, throw a 400 Bad Request?

0 results seems reasonable to me if filter.ids is present. My one worry is that in swagger filter.ids will be there by default and that can be possibly confusing, but again can be resolved by providing good client documentation that states the intended behavior.

yuliadub (approved):

PR #467, opened 2024-01-22
[AJ-1279] Add prepareThreshold=0 to defautl JDBC configuration

nawatts:
While testing attribute data type conversions, I started running into an error after a certain number of tests were run.

```
org.springframework.transaction.TransactionSystemException: JDBC rollback failed
        at app//org.springframework.jdbc.datasource.DataSourceTransactionManager.translateException(DataSourceTransactionManager.java:439)
        at app//org.springframework.jdbc.support.JdbcTransactionManager.translateException(JdbcTransactionManager.java:184)
        at app//org.springframework.jdbc.datasource.DataSourceTransactionManager.doRollback(DataSourceTransactionManager.java:355)
        at app//org.springframework.transaction.support.AbstractPlatformTransactionManager.processRollback(AbstractPlatformTransactionManager.java:896)
        at app//org.springframework.transaction.support.AbstractPlatformTransactionManager.rollback(AbstractPlatformTransactionManager.java:865)
        at app//org.springframework.transaction.interceptor.TransactionAspectSupport.completeTransactionAfterThrowing(TransactionAspectSupport.java:687)
        at app//org.springframework.transaction.interceptor.TransactionAspectSupport.invokeWithinTransaction(TransactionAspectSupport.java:389)
        at app//org.springframework.transaction.interceptor.TransactionInterceptor.invoke(TransactionInterceptor.java:119)
        at app//org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:184)
        at app//org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.proceed(CglibAopProxy.java:765)
        at app//org.springframework.retry.interceptor.RetryOperationsInterceptor$1.doWithRetry(RetryOperationsInterceptor.java:102)
        at app//org.springframework.retry.support.RetryTemplate.doExecute(RetryTemplate.java:335)
        at app//org.springframework.retry.support.RetryTemplate.execute(RetryTemplate.java:211)
        at app//org.springframework.retry.interceptor.RetryOperationsInterceptor.invoke(RetryOperationsInterceptor.java:135)
        at app//org.springframework.retry.annotation.AnnotationAwareRetryOperationsInterceptor.invoke(AnnotationAwareRetryOperationsInterceptor.java:161)
        at app//org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:184)
[...]

PR #773, opened 2024-05-03
AJ-1823: remove @RetryableApi

davidangb:
This PR removes the `@RetryableApi` annotation.

We added that annotation in #194 to address "cached plan must not change result type" errors.

Later, in #467, we realized that the REAL root cause of "cached plan must not change result type" errors was Postgres driver caching. We addressed that more globally by adding `prepareThreshold=0` to our JDBC driver connection string.

With `prepareThreshold=0` in place, I hope there is no need for `@RetryableApi` anymore.

We have a unit test `ConcurrentDataTypeChangesTest` for this issue. That test passes repeatedly with `@RetryableApi` removed but `prepareThreshold=0` in place. That test fails with `@RetryableApi` removed and `prepareThreshold=0` also removed.


service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -86,7 +83,7 @@ public ResponseEntity<TsvUploadResponse> tsvUpload(
     int recordsModified =
         recordOrchestratorService.tsvUpload(instanceId, version, recordType, primaryKey, records);
     return new ResponseEntity<>(
-        new TsvUploadResponse(recordsModified, "Updated " + recordType.toString()), HttpStatus.OK);
+        new TsvUploadResponse(recordsModified, "Updated " + recordType), HttpStatus.OK);
```
davidangb, line 86: huh - IntelliJ or Spotless fixed this for me, I don't even know which. But I agree with it.

nawatts (approved):

jladieu (approved):

PR #441, opened 2024-01-04
AJ-1521 - Remove deprecated Snapshot API

yuliadub:
Did my best to only remove things that were needed to be removed - keep the honest :) 

Reminder:

PRs merged into main will not automatically generate a PR in https://github.com/broadinstitute/terra-helmfile to update the WDS image deployed to kubernetes - this action will need to be triggered manually by running the following github action: https://github.com/DataBiosphere/terra-workspace-data-service/actions/workflows/tag.yml. Dont forget to provide a Jira Id when triggering the manual action, if no Jira ID is provided the action will not fully succeed. 

After you manually trigger the github action (and it completes with no errors), you must go to [the terra-helmfile](https://github.com/broadinstitute/terra-helmfile) repo and verify that this generated a PR that merged successfully.

The terra-helmfile PR merge will then generate a PR in [leonardo](https://github.com/DataBiosphere/leonardo).  This will automerge if all tests pass, but if jenkins tests fail it will not; be sure to watch it to ensure it merges. To trigger jenkins retest simply comment on PR with "jenkins retest".

PR #774, opened 2024-05-03
AJ-2005: delete datarepo dao

davidangb:
Reviving this old PR.

This PR deletes `DataRepoDao` and all of its supporting classes, in an effort to shed code and tests we don't need.

The only method inside `DataRepoDao` is `getSnapshot()`, and that method is never called (except from tests). So, this entire set of classes is unused and just adds to the Spring context.

Yes, WDS supports import from TDR snapshots. But it does this by reading a TDR manifest and Parquet files … it never makes any API calls directly to TDR. This import requires model classes from the TDR client, so we still import the TDR library - we just don't need a DAO.

The potential danger of deleting this code is that we _might_ need it in the future. If we decided, for instance, that we needed to call `getSnapshot()` during TDR or PFB import, we'd want this code back. 

Does deleting `TdrPactTest` have any impact? Do we need to unregister this pact somewhere?


nawatts (approved):
> Does deleting TdrPactTest have any impact? Do we need to unregister this pact somewhere?

Probably doesn't matter, though it may still be enforcing an unnecessary Pact on TDR. The integration can be deleted from the Pact Broker.

calypsomatic:
Did we used to call it from the code and not just a test?

nawatts:
Yes, it was added in https://github.com/DataBiosphere/terra-workspace-data-service/pull/221.

davidangb:
> Did we used to call it from the code and not just a test?

@calypsomatic yes - we called it from the original proof-of-concept snapshot import, which was deleted in #441 - see `DataRepoService` in that PR

calypsomatic:
Ah and now we just use import instead of `DataRepoController`.  Well I'm all for deleting code!

jladieu (approved):

jladieu:
💯

nawatts:
> Alternatively, should we call getSnapshot() during TDR or PFB import to ensure the user has permissions on the snapshots in question?

If we wanted to do that, could we use Sam for that?

calypsomatic:
> Does deleting `TdrPactTest` have any impact? Do we need to unregister this pact somewhere?

My guess is that when the pacts are published that will unregister it, but you might want to ask in #dsp-contract-testing.

kevinmarete (approved):
I see no issues

calypsomatic (approved):

PR #775, opened 2024-05-03
AJ-1812: Require gs:// URIs to match workspaceId

jladieu:
This is a security measure to ensure that a bad actor can't craft a
`RAWLSJSON` URI that hijacks a request for a different workspace.

Paired with orchestration change:
https://github.com/broadinstitute/firecloud-orchestration/commit/16074dd555261e683651c3dfac280912577d49e9

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
@@ -2,6 +2,7 @@ package org.databiosphere.workspacedataservice.dataimport;
 
 import static java.util.Collections.emptyList;
 import static java.util.Collections.emptySet;
+import static java.util.regex.Pattern.compile;
 
 import bio.terra.workspace.model.WorkspaceDescription;
 import bio.terra.workspace.model.WsmPolicyInput;
@@ -13,6 +14,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.databiosphere.workspacedataservice.config.DataImportProperties.ImportSourceConfig;
 import org.databiosphere.workspacedataservice.generated.ImportRequestServerModel;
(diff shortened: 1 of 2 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidator.java
```diff
@@ -88,9 +90,24 @@ public void validateImport(
           "Files may not be imported from %s.".formatted(importUrl.getHost()));
     }
 
+    if (importType == TypeEnum.RAWLSJSON) {
+      validateUriBelongsToWorkspace(importRequest.getUrl(), destinationWorkspaceId);
+    }
     validateDestinationWorkspace(importRequest, destinationWorkspaceId);
   }
 
+  private static final String URI_TEMPLATE = "^gs://test-bucket/to-cwds/%s/.*\\.json$";
```
nawatts, line 99: Should this use `allowedRawlsBucket`?

Or... since we've already validated the host, `validateUriBelongsToWorkspace` could only validate the path?

jladieu: Ahh good call, there's no point in redundantly validating the bucket.  I switched the implementation and method name to reflect it's just doing path validation.

calypsomatic (approved):

jladieu:
Paired with https://github.com/broadinstitute/firecloud-orchestration/pull/1347

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidatorTest.java
```diff
@@ -128,29 +129,57 @@ void rejectsImportsFromOtherSources() {
...
+    assertThatExceptionOfType(ValidationException.class)
+        .isThrownBy(
+            () -> importValidator.validateImport(importRequest, WorkspaceId.of(UUID.randomUUID())))
+        .withMessageContaining("Files may not be imported from rando-bucket.");
+  }
+
+  @Test
+  void rejectsGsUrlsWithMismatchingWorkspaceId() {
+    // Arrange
+    URI importUri = URI.create("gs://test-bucket/to-cwds/does-not-match-workspace-id/random.json");
```
nawatts, line 175: To clarify the intent of the test and also check that validation doesn't fail only because the workspace ID segmetn isn't a UUID, maybe:

```suggestion
    UUID otherWorkspaceId = UUID.randomUUID();
    URI importUri = URI.create("gs://test-bucket/to-cwds/%s/other-workspace-import.json".formatted(otherWorkspaceId));
```

nawatts (approved):

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/DefaultImportValidatorTest.java
jladieu: Cool, applied fix!

PR #779, opened 2024-05-06
[AJ-1787] Add cache control header to responses

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1787

Add a `Cache-Control: no-store` header to all responses.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/ControllerConfig.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/ControllerConfig.java
@@ -1,6 +1,6 @@
 package org.databiosphere.workspacedataservice.controller;
 
-import org.databiosphere.workspacedataservice.service.CacheControlNoStoreFilter;
+import org.databiosphere.workspacedataservice.service.NoCacheFilter;
 import org.springframework.boot.web.servlet.FilterRegistrationBean;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
@@ -52,10 +52,9 @@ public class ControllerConfig {
    * a filter that adds a "Cache-Control: no-store" header to API responses.
    */
   @Bean
-  public FilterRegistrationBean<CacheControlNoStoreFilter> cacheControlNoStoreFilter() {
-    FilterRegistrationBean<CacheControlNoStoreFilter> registrationBean =
-        new FilterRegistrationBean<>();
-    registrationBean.setFilter(new CacheControlNoStoreFilter());
+  public FilterRegistrationBean<NoCacheFilter> cacheControlNoStoreFilter() {
+    FilterRegistrationBean<NoCacheFilter> registrationBean = new FilterRegistrationBean<>();
+    registrationBean.setFilter(new NoCacheFilter());
     registrationBean.addUrlPatterns("/*");
     return registrationBean;
   }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/ControllerConfig.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/ControllerConfig.java
@@ -47,6 +47,10 @@ public class ControllerConfig {
     };
   }
 
+  /**
+   * API responses contain potentially sensitive data and thus should not be cached. This configures
+   * a filter that adds a "Cache-Control: no-store" header to API responses.
+   */
   @Bean
   public FilterRegistrationBean<CacheControlNoStoreFilter> cacheControlNoStoreFilter() {
     FilterRegistrationBean<CacheControlNoStoreFilter> registrationBean =
--- a/service/src/main/java/org/databiosphere/workspacedataservice/controller/ControllerConfig.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/controller/ControllerConfig.java
@@ -1,5 +1,7 @@
 package org.databiosphere.workspacedataservice.controller;
 
+import org.databiosphere.workspacedataservice.service.CacheControlFilter;
+import org.springframework.boot.web.servlet.FilterRegistrationBean;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Profile;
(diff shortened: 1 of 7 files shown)


nawatts:
@davidangb after I opened this, I saw the ticket has a bunch of information about the proxy configuration. Do we _need_ to set this in the proxy? It seems like the application is the source of truth about which responses should or shouldn't be cached and, ideally, those cache settings would apply to _all_ proxies in between the application and user.

yuliadub (approved):

nawatts:
Aligning with BPM, which uses ServletComponentScan to automatically find and load the filter instead of an explicit Component annotation.

calypsomatic (approved):

nawatts:
More updates based on Slack thread: https://broadinstitute.slack.com/archives/C02SJQ0FKHQ/p1715174679743839

yuliadub (approved):

service/src/main/java/org/databiosphere/workspacedataservice/service/CacheControlNoStoreFilter.java
```diff
@@ -0,0 +1,15 @@
...
+
+public class CacheControlNoStoreFilter implements Filter {
+  @Override
+  public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+      throws IOException, ServletException {
+    var servletResponse = (HttpServletResponse) response;
+    servletResponse.setHeader("Cache-Control", "no-store");
+    chain.doFilter(request, response);
+  }
+}
```
davidangb, line 15: the httpd-proxy (for some configs) also sets `Pragma: no-cache`; should we also set that here?

https://github.com/broadinstitute/terra-helmfile/blob/4839caaa650fc9fdf21c29846d9f24ef6c529231/charts/httpd-proxy/templates/_proxy_configmap.yaml#L37-L40

service/src/main/java/org/databiosphere/workspacedataservice/controller/ControllerConfig.java
```diff
@@ -44,4 +46,17 @@ public void configureAsyncSupport(AsyncSupportConfigurer configurer) {
...
+   * a filter that adds a "Cache-Control: no-store" header to API responses.
+   */
+  @Bean
+  public FilterRegistrationBean<CacheControlNoStoreFilter> cacheControlNoStoreFilter() {
+    FilterRegistrationBean<CacheControlNoStoreFilter> registrationBean =
+        new FilterRegistrationBean<>();
+    registrationBean.setFilter(new CacheControlNoStoreFilter());
+    registrationBean.addUrlPatterns("/*");
+    return registrationBean;
+  }
```
davidangb, line 61: `MDCResponseHeaderFilter` simply annotates itself as a `@Component` and thus doesn't need a `FilterRegistrationBean`. Is the difference here that `FilterRegistrationBean` can operate on a subset of url patterns?

service/src/main/java/org/databiosphere/workspacedataservice/service/CacheControlNoStoreFilter.java
```diff
@@ -0,0 +1,15 @@
...
+import jakarta.servlet.*;
+import jakarta.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+public class CacheControlNoStoreFilter implements Filter {
+  @Override
+  public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+      throws IOException, ServletException {
+    var servletResponse = (HttpServletResponse) response;
+    servletResponse.setHeader("Cache-Control", "no-store");
```
davidangb, line 12: for super defensive coding, I suggest avoiding the explicit cast:
```java
    if (response instanceof HttpServletResponse servletResponse) {
      servletResponse.setHeader("Cache-Control", "no-store");  
    }
```

service/src/main/java/org/databiosphere/workspacedataservice/controller/ControllerConfig.java
nawatts: That was what I understood.

Alternatively, the earlier approach using `WebFilter` and `ServletComponentScan` allowed configuring URL patterns without a registration bean.

service/src/main/java/org/databiosphere/workspacedataservice/service/CacheControlNoStoreFilter.java
nawatts: I guess it couldn't hurt, though I'm not sure it's necessary.

> Use Pragma only for backwards compatibility with HTTP/1.0 clients.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Pragma

service/src/main/java/org/databiosphere/workspacedataservice/controller/ControllerConfig.java
nawatts: I wanted the URL patterns for the template repository, and I was hoping to match WDS fairly closely to the template. But since we're filtering _all_ routes (unlike other services, we have no `/api` namespace), maybe I'll just change this to `@Component`.

nawatts: Oh, now I remember... I thought we _could_ exclude Swagger UI static resources from this filter.

nawatts:
Ok, I think I've finally arrived at something that I'm happy with. This feels more "Spring-y" to me and it keeps everything in one class while still allowing configuration based on path (and more flexible configuration than URL patterns).

service/src/main/java/org/databiosphere/workspacedataservice/service/NoCacheFilter.java
```diff
@@ -0,0 +1,31 @@
+package org.databiosphere.workspacedataservice.service;
+
+import jakarta.servlet.*;
```
davidangb, line 3: super nit: get rid of the wildcard import
```suggestion
import jakarta.servlet.FilterChain;
import jakarta.servlet.ServletException;
```

davidangb (approved):
:+1: for the implementation, thanks for sticking with it!

How about a unit test? You could copy from `CorsLocalMockMvcTest` to add a test.

nawatts:
> How about a unit test? 

Added a test in https://github.com/DataBiosphere/terra-workspace-data-service/pull/779/commits/b496bce3f251ee644acf0f83a494ee24fe4ae40c.

> You could copy from CorsLocalMockMvcTest to add a test.

I tried copying from this test, but it wasn't working as expected with `MockMvc`. It did with `TestRestTemplate` though.

PR #783, opened 2024-05-09
AJ-1817: disable logback_event metrics

davidangb:
This disables the "logback_event" Micrometer metrics. These don't seem immediately useful and contribute to overall metrics volume in end users' App Insights (WDS) and in Prometheus/Grafana (cWDS). These metrics simply count how many logback invocations have occurred; the log messages themselves are totally separate.

Here's an example of the metrics, from cWDS dev env:
```
# HELP logback_events_total Number of log events that were enabled by the effective log level
# TYPE logback_events_total counter
logback_events_total{level="info",service="wds",version="0.2.167-SNAPSHOT",} 24.0
logback_events_total{level="debug",service="wds",version="0.2.167-SNAPSHOT",} 0.0
logback_events_total{level="error",service="wds",version="0.2.167-SNAPSHOT",} 0.0
logback_events_total{level="warn",service="wds",version="0.2.167-SNAPSHOT",} 542.0
logback_events_total{level="trace",service="wds",version="0.2.167-SNAPSHOT",} 130252.0
```

In my MRG in the dev env, logback events are the 6th most populous event type in the `AppMetrics` table:
Screenshot 09-05-2024 at 11 31

And since all they do is count logback invocations, I don't see them as super useful. We can get similar metrics by counting/summarizing the actual log messages themselves; do we need to track the metrics separately? Note that it appears logback_event tracks invocations _before_ any loglevel filters in logback-spring.xml; so, logback_event will count TRACE invocations, while we don't actually log any traces and thus wouldn't know how many there are:
Screenshot 09-05-2024 at 11 14

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
@@ -5,6 +5,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.boot.SpringApplication;
+import org.springframework.boot.actuate.autoconfigure.metrics.LogbackMetricsAutoConfiguration;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.cache.annotation.EnableCaching;
 import org.springframework.retry.annotation.EnableRetry;
@@ -16,6 +17,10 @@ import org.springframework.transaction.annotation.EnableTransactionManagement;
       "org.databiosphere.workspacedataservice",
       // terra-common-lib transaction management and DB retry configuration
       "bio.terra.common.retry.transaction"
+    },
+    exclude = {
+      // disable "logback_events" in Micrometer/Prometheus to reduce overall metrics volume
+      LogbackMetricsAutoConfiguration.class
     })
 @EnableRetry
 @EnableTransactionManagement


service/src/main/java/org/databiosphere/workspacedataservice/WorkspaceDataServiceApplication.java
```diff
@@ -16,6 +17,10 @@
       "org.databiosphere.workspacedataservice",
       // terra-common-lib transaction management and DB retry configuration
       "bio.terra.common.retry.transaction"
+    },
+    exclude = {
+      // disable "logback_events" in Micrometer/Prometheus to reduce overall metrics volume
+      LogbackMetricsAutoConfiguration.class
     })
```
davidangb, line 24: I can't figure out a good way to test this. We have a `MetricsConfigTest`, but tests don't inherit the `exclude` setting here - I could add an annotation to that test to exclude `LogbackMetricsAutoConfiguration`, but then I'm simply testing how well I've configured the test, not the runtime app.

jladieu:
Is the stanza you pasted _all_ of the logback metrics, or a subset of a thing with much higher cardinality?  If that's _all_ of them, then IMO the cost of including them in the prometheus scrape is trivial.  Our HTTP metrics will dwarf those 100/1 due to their cardinality.

Having not really noticed these metrics existed, I'd almost advocate to go the other direction and promote these to a dashboard -- it'd be really helpful to know if there were a surge in log metrics over time.

jladieu:
One other thought...from a _prometheus_ perspective, the cost of these metrics seem trivial to me.  They represent a fixed / constant cost in the scrape.

But if the ticking/counting/recording of these events somehow trigger a new row of work for every counting event in appinsights log -- eg: the cost of a metric recording scales with the # of metrics, then I think it makes perfect sense _not_ to send them to appinsights.

davidangb:
closing - we'll let https://github.com/broadinstitute/terra-helmfile/pull/5519 run for a while and assess its impact alone, we can always come back to disabling specific metrics later.

PR #785, opened 2024-05-10
[AJ-1777] Fix check for protected workspace

nawatts:
(C)WDS is configured to require a "protected workspace" for imports from BioData Catalyst. "Protected workspace" means a workspace that has additional security monitoring enabled and would be allowed to import data that has a "protected data" policy.

Currently, CWDS determines if a workspace is protected by checking if the workspace has a protected data policy. That works for Azure workspaces, but not for GCP workspaces. For GCP workspaces, the only indication that the workspace is protected is the name of the GCS bucket. For workspaces with additional security monitoring enabled, the bucket name starts with "fc-secure-" instead of "fc-".

To get the bucket name, CWDS has to get the workspace details from Rawls. However, for WDS, we don't want to involve Rawls and since we're only dealing with Azure workspaces there, we can continue to get workspace policies from WSM. To handle this difference, this adds a `ProtectedDataSupport` interface with Rawls and WSM implementations for the control plane and data plane. `DefaultImportValidator` then uses `ProtectedDataSupport` to check if the import's destination workspace is protected.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/protecteddatasupport/WsmProtectedDataSupport.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/protecteddatasupport/WsmProtectedDataSupport.java
@@ -1,0 +1,30 @@
+package org.databiosphere.workspacedataservice.dataimport.protecteddatasupport;
+
+import static java.util.Collections.emptyList;
+
+import bio.terra.workspace.model.WorkspaceDescription;
+import bio.terra.workspace.model.WsmPolicyInput;
+import java.util.List;
+import java.util.Optional;
+import org.databiosphere.workspacedataservice.annotations.DeploymentMode.DataPlane;
+import org.databiosphere.workspacedataservice.policy.PolicyUtils;
+import org.databiosphere.workspacedataservice.shared.model.WorkspaceId;
+import org.databiosphere.workspacedataservice.workspacemanager.WorkspaceManagerDao;
+import org.springframework.stereotype.Component;
+
+@Component
+@DataPlane
+public class WsmProtectedDataSupport implements ProtectedDataSupport {
+  private final WorkspaceManagerDao wsmDao;
+
+  public WsmProtectedDataSupport(WorkspaceManagerDao wsmDao) {
+    this.wsmDao = wsmDao;
+  }
+
+  public boolean workspaceSupportsProtectedDataPolicy(WorkspaceId workspaceId) {
+    WorkspaceDescription workspace = wsmDao.getWorkspace(workspaceId.id());
+    List<WsmPolicyInput> workspacePolicies =
+        Optional.ofNullable(workspace.getPolicies()).orElse(emptyList());
+    return workspacePolicies.stream().anyMatch(PolicyUtils::isProtectedDataPolicy);
+  }
+}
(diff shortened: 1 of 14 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/rawls/RawlsClient.java
```diff
@@ -17,6 +17,21 @@ public RawlsClient(RawlsApi rawlsApi, RestClientRetry restClientRetry) {
...
   }
 
+  public RawlsWorkspaceDetails getWorkspaceDetails(UUID workspaceId) {
+    try {
+      RestCall<RawlsWorkspaceDetails> restCall =
+          () ->
+              rawlsApi.getWorkspaceDetails(
+                  workspaceId,
+                  /* RawlsWorkspaceDetails only includes the subset of workspace fields needed by CWDS. */
+                  /* fields */ String.join(",", RawlsWorkspaceDetails.SUPPORTED_FIELDS));
```
davidangb, line 27: here's an opportunity for premature optimization! This will perform `String.join()` for every request, even though `RawlsWorkspaceDetails.SUPPORTED_FIELDS` is constant. You could create a private member var like:
```
private final String fieldsParam = String.join(",", RawlsWorkspaceDetails.SUPPORTED_FIELDS);
```
and use that here to avoid the extra work.

I don't think this is a big deal - it tickles the ivory-tower code reviewer in me but I don't think it would matter much in practice - your call.

davidangb (commented):
looking good, one comment about the ProtectedDataSupport interface

service/src/main/java/org/databiosphere/workspacedataservice/rawls/RawlsClient.java
nawatts: I like it. Yeah, at first I had the supported fields defined as a string, but that's hard to read and edit, so I changed it to a List, at the expense of repeatedly joining it. But this sounds like the best of both.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/protecteddatasupport/WsmProtectedDataSupport.java
```diff
@@ -0,0 +1,30 @@
...
+  public WsmProtectedDataSupport(WorkspaceManagerDao wsmDao) {
+    this.wsmDao = wsmDao;
+  }
+
+  public boolean workspaceSupportsProtectedDataPolicy(WorkspaceId workspaceId) {
+    WorkspaceDescription workspace = wsmDao.getWorkspace(workspaceId.id());
+    List<WsmPolicyInput> workspacePolicies =
+        Optional.ofNullable(workspace.getPolicies()).orElse(emptyList());
+    return workspacePolicies.stream().anyMatch(PolicyUtils::isProtectedDataPolicy);
+  }
```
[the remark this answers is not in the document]
nawatts: Moved to `PolicyUtils` in https://github.com/DataBiosphere/terra-workspace-data-service/pull/785/commits/4e8a6b750222488c5f20a2883dba2b856df07066.

service/src/main/java/org/databiosphere/workspacedataservice/rawls/RawlsClient.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/785/commits/ce922318e60e58f2dd175ec292ba121c52419787.

davidangb (approved):
nice, thank you!

jladieu (approved):
This looks great!  Nice work.

PR #787, opened 2024-05-10
[AJ-1822] Remove disk space from status endpoint

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1822

This removes the `diskSpace` section from the `/status` endpoint.

It does so by disabling the default health indicators and explicitly enabling the ones we do want.
https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html#actuator.endpoints.health

While I was here, I noticed the OpenAPI docs for the status endpoint were significantly out of sync with the actual response. So I updated those as well.

### DIFF
--- a/service/src/main/resources/static/swagger/openapi-docs.yaml
+++ b/service/src/main/resources/static/swagger/openapi-docs.yaml
@@ -853,42 +853,6 @@ components:
         time:
           type: string
           format: date-time
-    component:
-      type: object
-      properties:
-        status:
-          type: string
-        details:
-          type: string
-    components:
-      type: object
-      properties:
-        Permissions:
-          $ref: '#/components/schemas/permissionsComponent'
-        db:
-          $ref: '#/components/schemas/dbComponent'
-        livenessState:
-          $ref: '#/components/schemas/component'
-        ping:
-          $ref: '#/components/schemas/component'
-        pubSub:
-          $ref: '#/components/schemas/component'
-        readinessState:
-          $ref: '#/components/schemas/component'
-    dbComponent:
-      type: object
-      properties:
-        status:
-          type: string
-        details:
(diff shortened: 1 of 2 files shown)


service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -863,53 +863,32 @@ components:
     components:
```
davidangb, line 863: how would you feel about simplifying the `StatusResponse` to be really generic, something like:
```yaml
    StatusResponse:
      type: object
      properties:
        status:
          type: string
        components:
          type: object
```
and delete all the strongly-typed `permissionsComponent`, `dbComponent`, etc? I think this would 1) give us freedom to change the components in the future; 2) indicate to end users that the `components` are flexible; 3) reduce complexity in the spec.

We talked about this a long time ago when first putting together the /status API, and this suggestion was downvoted - bringing it up again for reconsideration now!

davidangb (commented):
config change looks good - one question about how to spec it in OpenApi

service/src/main/resources/static/swagger/openapi-docs.yaml
nawatts: Sounds good to me. I was just following the existing pattern. But yeah, there's an inordinate amount of YAML here for the status endpoint, which is unlikely to be used by an end user.

nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/787/commits/5a0cf55f6b05514dc1088682bb13006d77537a2a.

calypsomatic (approved):

davidangb (approved):

PR #788, opened 2024-05-10
AJ-1837: improved logging for snapshot import

davidangb:
AJ-1837

This PR enhances logging for snapshot imports. As I was debugging problems with import, I found the logs lacking in a few places. This beefs them up!

There should be no functional changes at all; only changes to logging.

_reviewer: ignore whitespace!_

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
@@ -26,6 +26,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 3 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJob.java
```diff
@@ -233,29 +242,50 @@ private BatchWriteResult importTables(
       ImportMode importMode,
       RecordSink recordSink) {
     var combinedResult = BatchWriteResult.empty();
+    var numTables = importTables.size();
+    AtomicInteger tableIdx = new AtomicInteger();
+
```
davidangb, line 247: we need to use AtomicInteger here because we're modifying the variable inside the .forEach lambda below, and the lambda will complain about "effectively final" if this is a simple int. There are other workarounds for that but this seems the most readable.

nawatts (approved):

jladieu (approved):

PR #793, opened 2024-05-14
AJ-1704 Error out stalled jobs

calypsomatic:
https://broadworkbench.atlassian.net/browse/AJ-1704
This PR sets up a scheduled task to update any import jobs that haven't completed in 6 hours to `ERROR` status.

I chose not to use a lock or Shedlock to deal with updating multiple replicas at once; all we're doing is changing the status from `RUNNING` to `ERROR`, and as far as I can tell interference would just result in updating a job that's already in `ERROR` to still be in error.  I tried this out on my BEE and although I saw all the replicas update at the same time, I didn't see any other problems.

Reminder:

#### Releasing WDS ####

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
@@ -347,8 +347,6 @@ public class PostgresJobDao implements JobDao {
     OffsetDateTime lastUpdate =
         instantSource.instant().atOffset(ZoneOffset.UTC).minusHours(UPDATE_JOB_FREQUENCY_IN_HOURS);
 
-    // TODO should this be defined in JobService along with terminal job statuses?  Is there a way
-    // to programmatically make it anything that's not in terminal job statuses?
     List<String> nonterminalJobStatuses =
         List.of(
             StatusEnum.CREATED.name(),
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
@@ -13,6 +13,7 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.time.Duration;
 import java.time.InstantSource;
+import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
 import java.util.List;
 import java.util.Optional;
@@ -327,5 +328,19 @@ public class PostgresJobDao implements JobDao {
     }
 
     return tags;
+  }
+
+  // TODO: Genericize to a general get?
+  public List<GenericJobServerModel> getOldRunningJobs() {
+    OffsetDateTime sixHoursAgo = OffsetDateTime.now(ZoneOffset.UTC).minusHours(6);
+
+    return namedTemplate.query(
+        "SELECT id, type, status, created, updated, input, result, error, stacktrace, collection_id "
+            + "FROM sys_wds.job "
+            + "WHERE status = :status AND created < :sixHoursAgo",
+        new MapSqlParameterSource()
+            .addValue("status", StatusEnum.RUNNING.name())
+            .addValue("sixHoursAgo", sixHoursAgo),
+        new AsyncJobRowMapper(mapper));
   }
 }
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
@@ -4,6 +4,7 @@ import static org.databiosphere.workspacedataservice.generated.GenericJobServerM
 import static org.databiosphere.workspacedataservice.generated.GenericJobServerModel.JobTypeEnum.DATA_IMPORT;
 import static org.databiosphere.workspacedataservice.generated.GenericJobServerModel.StatusEnum;
 import static org.databiosphere.workspacedataservice.jobexec.ImportJobUpdater.UPDATE_JOB_FREQUENCY_IN_HOURS;
+import static org.databiosphere.workspacedataservice.service.JobService.NONTERMINAL_JOB_STATUSES;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -175,19 +176,6 @@ public class PostgresJobDao implements JobDao {
     return update(jobId, StatusEnum.ERROR, errorMessage, e.getStackTrace());
   }
 
-  /**
-   * Mark a job as error without logging at the error level
-   *
-   * @param jobId id of the job to update
-   * @param errorMessage a short error message, if the job is in error
-   * @return the updated job
-   */
-  @Override
-  public GenericJobServerModel markError(UUID jobId, String errorMessage) {
-    logger.info("Marking job {} as errored: {}", jobId, errorMessage);
-    return update(jobId, StatusEnum.ERROR, errorMessage, null);
-  }
-
   private GenericJobServerModel update(
       UUID jobId,
       StatusEnum status,
@@ -349,11 +337,7 @@ public class PostgresJobDao implements JobDao {
         instantSource.instant().atOffset(ZoneOffset.UTC).minusHours(UPDATE_JOB_FREQUENCY_IN_HOURS);
 
     List<String> nonterminalJobStatuses =
-        List.of(
-            StatusEnum.CREATED.name(),
-            StatusEnum.QUEUED.name(),
-            StatusEnum.RUNNING.name(),
-            StatusEnum.UNKNOWN.name());
+        NONTERMINAL_JOB_STATUSES.stream().map(status -> status.name()).toList();
 
     return namedTemplate.query(
         "SELECT id, type, status, created, updated, input, result, error, stacktrace, collection_id "
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
@@ -3,6 +3,7 @@ package org.databiosphere.workspacedataservice.dao;
 import static org.databiosphere.workspacedataservice.generated.GenericJobServerModel.JobTypeEnum;
 import static org.databiosphere.workspacedataservice.generated.GenericJobServerModel.JobTypeEnum.DATA_IMPORT;
 import static org.databiosphere.workspacedataservice.generated.GenericJobServerModel.StatusEnum;
+import static org.databiosphere.workspacedataservice.jobexec.ImportJobUpdater.UPDATE_JOB_FREQUENCY_IN_HOURS;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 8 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/jobexec/ImportJobUpdater.java
```diff
@@ -0,0 +1,24 @@
...
+
+  public ImportJobUpdater(JobDao jobDao) {
+    this.jobDao = jobDao;
+  }
+
+  @Scheduled(fixedRate = 21600000) // run every 6 hours
+  public void updateImportJobs() {
+    List<GenericJobServerModel> jobsToUpdate = jobDao.getOldRunningJobs();
+    jobsToUpdate.stream()
+        .forEach(job -> jobDao.fail(job.getJobId(), "Job failed to complete in 6 hours."));
```
davidangb, line 22: I believe `jobDao.fail()` log an exception, which will in turn cause a Sentry error - do we want that?

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/ImportJobUpdater.java
```diff
@@ -0,0 +1,24 @@
...
+@Component
+public class ImportJobUpdater {
+
+  private final JobDao jobDao;
+
+  public ImportJobUpdater(JobDao jobDao) {
+    this.jobDao = jobDao;
+  }
+
+  @Scheduled(fixedRate = 21600000) // run every 6 hours
```
davidangb, line 18: consider using a random `initialDelay` to stagger when multiple pods execute this action. If 2 or 3 pods start at roughly the same time, they'll all run this action at roughly the same time. With a random delay, they're much less likely to run into each other

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
```diff
@@ -328,4 +329,18 @@ private static Tags getTags(GenericJobServerModel job, StatusEnum status) {
 
     return tags;
   }
+
+  // TODO: Genericize to a general get?
+  public List<GenericJobServerModel> getOldRunningJobs() {
+    OffsetDateTime sixHoursAgo = OffsetDateTime.now(ZoneOffset.UTC).minusHours(6);
```
davidangb, line 335: worth breaking out the "6 hour" value here into a config value, and using that both here and in in the `@Scheduled` annotation?

davidangb:
> Should the ImportJobUpdater live where it is or somewhere else? Should it be its own class or does it logically belong in the JobDao or QuartzJob or somewhere?

I think it's fine where it is, architecturally. Implementation-wise, I'd love to see the select statement and subsequent calls to fail the jobs inside a `@WriteTransaction`. I am not sure if a `@WriteTransaction` will work directly on `updateImportJobs` (it could work! I just don't know). If it doesn't work you'll need to insert some other class into the call hierarchy.

> Currently leaning ever so slightly away from using a lock or Shedlock to deal with updating multiple replicas at once; all we're doing is changing the status from RUNNING to ERROR; it seems like interference would just result in updating a job that's already in ERROR, and with this simple task and small number of replicas, perhaps that's the only problem we're likely to face. But! Convince me otherwise!

I think that's fine, as long as we have a way to detect any problems, even if that means searching logs or getting Sentry alerts.

> How generic of a method would be useful for finding the jobs in question? A fully generic search method? One that takes in time and status? Just exactly what I want, as written?

I don't think it needs to be too generic. If we find more use cases later, we can make it more generic later

> Seems like we should updated jobs with a CREATED or QUEUED status as well, if they were created more than 6 hours ago. Or is there a reason to only look for RUNNING?

I'd look for any status that isn't terminal - see JobService.TERMINAL_JOB_STATUSES

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/ImportJobUpdater.java
calypsomatic: Yeah I guess not - kinda annoying that `updateStatus` specifically won't let you do [fail](https://github.com/DataBiosphere/terra-workspace-data-service/blob/7ae1af31125c8856a79d6167ec60bae0e5fef8d3/service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java#L131) but I'll come up with some other way to error a job without logging the error

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
```diff
@@ -143,8 +145,7 @@ public GenericJobServerModel updateStatus(UUID jobId, StatusEnum status) {
    */
   @Override
   public GenericJobServerModel fail(UUID jobId, String errorMessage) {
-    logger.error("Job {} failed: {}", jobId, errorMessage);
-    return update(jobId, StatusEnum.ERROR, errorMessage, null);
+    return fail(jobId, errorMessage, null);
```
calypsomatic, line 148: I just noticed this method was for some reason not using the base `fail` method

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
```diff
@@ -173,6 +174,19 @@ public GenericJobServerModel fail(UUID jobId, String errorMessage, Exception e)
...
 
+  /**
+   * Mark a job as error without logging at the error level
+   *
+   * @param jobId id of the job to update
+   * @param errorMessage a short error message, if the job is in error
+   * @return the updated job
+   */
+  @Override
+  public GenericJobServerModel markError(UUID jobId, String errorMessage) {
```
calypsomatic, line 185: I created a new method that does the same as `fail` except it uses `info` instead of `error` log level to avoid triggering Sentry; potentially I could instead add a `logError` flag to the `fail` method if someone has strong feelings about it.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
```diff
@@ -328,4 +342,27 @@ private static Tags getTags(GenericJobServerModel job, StatusEnum status) {
 
     return tags;
   }
+
+  public List<GenericJobServerModel> getOldNonTerminalJobs() {
+    OffsetDateTime lastUpdate =
+        instantSource.instant().atOffset(ZoneOffset.UTC).minusHours(UPDATE_JOB_FREQUENCY_IN_HOURS);
+
+    // TODO should this be defined in JobService along with terminal job statuses?  Is there a way
```
calypsomatic, line 350: I figure we'll update the list of job statuses rarely if at all, so I shouldn't need to define non terminal statuses programmatically, but it does have an appeal.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
```diff
@@ -328,4 +342,25 @@ private static Tags getTags(GenericJobServerModel job, StatusEnum status) {
 
     return tags;
   }
+
+  public List<GenericJobServerModel> getOldNonTerminalJobs() {
+    OffsetDateTime lastUpdate =
+        instantSource.instant().atOffset(ZoneOffset.UTC).minusHours(UPDATE_JOB_FREQUENCY_IN_HOURS);
+
+    List<String> nonterminalJobStatuses =
```
calypsomatic, line 350: It feels better to define these as anything that's not in terminal job statuses, but since I doubt we'll be updating the list I guess that's not necessary.  Either way, does it make sense for this to live here, or should it be in the `JobService` along with the definition of terminal job statuses?

service/src/main/resources/application.yml
```diff
@@ -112,6 +112,10 @@ spring:
     # always run on the replica where they were created.
     job-store-type: memory
     wait-for-jobs-to-complete-on-shutdown: true
+  task:
+    scheduling:
+      pool:
+        size: 5
```
davidangb, line 118: default is 1, isn't it? Do we actually need to bump this up?

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
nawatts: Generally, I'd vote against anything named with "not", etc. It starts to get hard to understand if you have double negatives like `!nonterminalJobStatues.includes(...)`.

Maybe the list of terminal statuses could be moved to a separate class? JobUtils or something like that? That could then have a method to do as you suggest and return all statuses that aren't terminal.

I think I added them to JobService and wanted to put them in StatusEnum, but couldn't because it's generated.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
davidangb: I'm really waffling on this. I think it could actually be good after all to get alerts when jobs time out like this. If we accept that we're gonna get alerts, then you could do away with the `markError` method, which makes it even more appealing.

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/ImportJobUpdater.java
```diff
@@ -0,0 +1,41 @@
...
+import org.springframework.stereotype.Component;
+
+@Component
+public class ImportJobUpdater {
+
+  private final JobDao jobDao;
+  public static final long UPDATE_JOB_FREQUENCY_IN_HOURS = 6;
+  private static final long UPDATE_FREQUENCY_IN_MILLISECONDS =
+      UPDATE_JOB_FREQUENCY_IN_HOURS * 3600 * 1000;
+  private static final long MAX_INITIAL_DELAY_IN_MILLISECONDS = 60 * 1000; // delay up to 1 minute
```
davidangb, line 18: honestly I'd crank this up, maybe get rid of `MAX_INITIAL_DELAY_IN_MILLISECONDS` entirely and just randomize on `UPDATE_FREQUENCY_IN_MILLISECONDS` as the initial delay so the scheduled tasks on each pod are more widely scattered (though randomization means they could also be close)

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/ImportJobUpdater.java
nawatts: I think I'd lean towards logging the error. Generally, this shouldn't be happening, right?

nawatts (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
calypsomatic: I suppose we could always go back to not alerting if we start getting the alerts and they're not useful.

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/ImportJobUpdater.java
```diff
@@ -0,0 +1,41 @@
...
+  public void updateImportJobs() {
+    List<GenericJobServerModel> jobsToUpdate = jobDao.getOldNonTerminalJobs();
+    logger.info("Updating {} stalled import jobs", jobsToUpdate.size());
+    jobsToUpdate.stream()
+        .forEach(
+            job ->
+                jobDao.markError(
+                    job.getJobId(),
+                    "Job failed to complete in " + UPDATE_JOB_FREQUENCY_IN_HOURS + " hours."));
+  }
```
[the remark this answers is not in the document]
calypsomatic: Ha, I had forgotten about that comment.  But yeah, seems like there's no reason to roll back.

service/src/main/resources/application.yml
calypsomatic: Ah, no, presumably 1 is fine.

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
nawatts: Yeah... most cases this comes up, I'd suggest something like `!isTerminal(...)` vs `isNonTerminal(...)`.

But here we need the list, so 🤷 .

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/ImportJobUpdater.java
davidangb: I bet we'll see a handful of alerts when this launches and we clean up any orphaned jobs since launch - but after that, it shouldn't happen often.

davidangb (approved):
thank you!!

PR #794, opened 2024-05-15
AJ-1844: Prevent pk drop during import.

jladieu:
This prevents primary key from always being dropped during import.

Instead, it only skips the primary key if it matches the name of the table plus an `_id` suffix.  This logic is carried over from import service, see:

https://github.com/broadinstitute/import-service/blob/develop/app/translators/tdr_manifest_to_rawls.py#L242

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -1,5 +1,6 @@
 package org.databiosphere.workspacedataservice.dao;
 
+import static java.util.stream.Collectors.toSet;
 import static org.databiosphere.workspacedataservice.dao.SqlUtils.getQualifiedTableName;
 import static org.databiosphere.workspacedataservice.dao.SqlUtils.quote;
 import static org.databiosphere.workspacedataservice.service.model.ReservedNames.RECORD_ID;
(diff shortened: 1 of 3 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -487,17 +488,26 @@ public void batchUpsert(
   }
 
   private List<RecordColumn> getSchemaWithRowId(
-      Map<String, DataTypeMapping> schema, String recordIdColumn) {
-    // we collect to a set first to handle the case where the user has included their primary key
-    // data in attributes
-    return Stream.concat(
-            Stream.of(new RecordColumn(recordIdColumn, DataTypeMapping.STRING)),
```
jladieu, line 494: [TdrManifestQuartzJobE2ETest.withEntityReferenceListsManifest](https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJobE2ETest.java#L225) actually started failing after I made the fix in the first commit, and it's because WDS was attempting to insert a row with two `id` fields.  The reason for this is the `id` in the Schema had a type of `NUMBER`, while the `id` this logic was using assumed a type of `STRING` -- therefore, both were allowed to appear in the list returned by this method.

nawatts:
I'm kind of skeptical of this approach. Based on the ticket, my understanding is that the WDS behavior is correct and the issue only exists when importing into Rawls? Thus, it sounds like an issue in RawlsRecordSink and I'm not sure why it requires editing the WDS pieces that seem to be working.

Especially based on https://github.com/DataBiosphere/terra-workspace-data-service/pull/794#discussion_r1600874793, this seems like we're changing ParquetRecordConverter to accomodate Rawls, then changing RecordDao to work around that. Is it not possible to have RawlsRecordSink write the primary key / ID value twice, once in the entity ID (where it would be renamed `{tableName}_id` and once in entity attributes (where it would keep its original name)?

PR #796, opened 2024-05-15
AJ-1844: snapshot import preserves primary key column

davidangb:
This PR branches from, and supersedes, #794. This PR builds on the research, findings, and tests from that other PR but attempts to simplify the implementation.

* when translating TDR snapshots, don't exclude the snapshot's primary key column. End users expect that column and its values to exist.
* In WDS (irrelevant for cWDS), when batch-upserting, ensure the primary key column is a string. We have a few places ([example](https://github.com/DataBiosphere/terra-workspace-data-service/blob/11628f599f958f204a376641637ace93a2c46a7d/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java#L422)) that require WDS PKs to be a string, so we need to enforce that.
  * JSON and TSV APIs create PK columns as string. The TDR and PFB paths didn't do that, and the snapshot used in the [TdrManifestQuartzJobE2ETest.withEntityReferenceListsManifest](https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestQuartzJobE2ETest.java#L225) test identified this problem.
  * TDR and PFB paths (and their RecordSinks) exclusively use batchUpsert, no other insert methods, so this is a safe place to add this constraint.

I would love to revisit the assumption that WDS PKs must be a string, but that's pretty well out of scope for this PR.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -1,5 +1,6 @@
 package org.databiosphere.workspacedataservice.dao;
 
+import static java.util.stream.Collectors.toSet;
 import static org.databiosphere.workspacedataservice.dao.SqlUtils.getQualifiedTableName;
 import static org.databiosphere.workspacedataservice.dao.SqlUtils.quote;
 import static org.databiosphere.workspacedataservice.service.model.ReservedNames.RECORD_ID;
@@ -487,15 +488,24 @@ public class RecordDao {
   }
 
   private List<RecordColumn> getSchemaWithRowId(
-      Map<String, DataTypeMapping> schema, String recordIdColumn) {
-    // we collect to a set first to handle the case where the user has included their primary key
-    // data in attributes
-    return Stream.concat(
-            Stream.of(new RecordColumn(recordIdColumn, DataTypeMapping.STRING)),
-            schema.entrySet().stream().map(e -> new RecordColumn(e.getKey(), e.getValue())))
-        .collect(Collectors.toSet())
+      Map<String, DataTypeMapping> schema, String recordIdColumnName) {
+    // make sure the id column is included; first check the schema for it, and if not found, assume
+    // it will be a string
+    RecordColumn idColumn =
+        schema.entrySet().stream()
+            .filter(e -> e.getKey().equals(recordIdColumnName))
+            .findAny()
+            .map(RecordDao::toColumn)
+            .orElse(new RecordColumn(recordIdColumnName, DataTypeMapping.STRING));
+    // ensure the id column occurs exactly once in the result
+    return Stream.concat(Stream.of(idColumn), schema.entrySet().stream().map(RecordDao::toColumn))
+        .collect(toSet())
         .stream()
         .toList();
+  }
+
+  private static RecordColumn toColumn(Map.Entry<String, DataTypeMapping> entry) {
+    return new RecordColumn(entry.getKey(), entry.getValue());
   }
 
   public boolean deleteSingleRecord(UUID collectionId, RecordType recordType, String recordId) {
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -1,6 +1,5 @@
 package org.databiosphere.workspacedataservice.dao;
 
-import static java.util.stream.Collectors.toSet;
 import static org.databiosphere.workspacedataservice.dao.SqlUtils.getQualifiedTableName;
 import static org.databiosphere.workspacedataservice.dao.SqlUtils.quote;
 import static org.databiosphere.workspacedataservice.service.model.ReservedNames.RECORD_ID;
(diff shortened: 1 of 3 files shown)


calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/ParquetRecordConverter.java
```diff
@@ -38,9 +38,7 @@ protected final Record convertBaseAttributes(GenericRecord genericRecord) {
     // for base attributes, skip the id field and all relations
```
nawatts, line 38: This comment is no longer accurate.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -487,14 +487,14 @@ public void batchUpsert(
...
-        .stream()
+      Map<String, DataTypeMapping> schema, String recordIdColumnName) {
+    // Make sure the id column is included, and that it is a string.
+    // This will overwrite any other datatype the id column may have been set to.
+    // WDS requires the id column to be a string (see getPrimaryKeyDef), so we need to enforce that.
+    HashMap<String, DataTypeMapping> workingSchema = new HashMap<>(schema);
+    workingSchema.put(recordIdColumnName, DataTypeMapping.STRING);
+    return workingSchema.entrySet().stream()
+        .map(e -> new RecordColumn(e.getKey(), e.getValue()))
         .toList();
```
davidangb, line 498: The previous logic had a flaw if the `schema` argument contained the `recordIdColumnName` but that column was not a string. Given the `Collectors.toSet()` logic, it would result in the return value containing `recordIdColumnName` twice - once as a string, once as some other data type.

I've update the logic, and hopefully made it easier to read, to ensure we only have `recordIdColumnName` once, and that one instance is a string.

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/ParquetDataTypesTest.java
```diff
@@ -160,6 +163,81 @@ void arrays() throws IOException {
...
+                new ImmutableMap.Builder<String, Object>()
+                    .put("chip_well_barcode", "123")
+                    .put("lab_batch", "456")
+                    .build()));
+  }
+
+  // even though the source TDR manifest contains a column of ${entityType}_id, translation should
+  // include that column. The RawlsAttributePrefixer will later rename this to tdr:${entityType}_id.
+  @Test
+  void primaryKeyIsSkippedWhenMatchingTypeId() {
```
nawatts, line 200: What's the difference between this test and the previous one?

It looks like the column is _not_ skipped here, since it's included in `result.getAttributes()`.

Also the comment is confusing since it says the column should be included but the test name says it should be skipped.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/ParquetRecordConverter.java
```diff
@@ -38,9 +38,7 @@ protected final Record convertBaseAttributes(GenericRecord genericRecord) {
...
     List<String> relationNames =
         relationshipModels.stream().map(r -> r.getFrom().getColumn()).toList();
-    Set<String> allIgnores = new HashSet<>();
-    allIgnores.add(idField);
-    allIgnores.addAll(relationNames);
+    Set<String> allIgnores = new HashSet<>(relationNames);
```
nawatts, line 41: Just trying to understand... so Records created here will contain the primary key in both attributes and ID? But otherwise, Record attributes don't include the primary key?

davidangb: Yes, the primary key value will be in both attributes and id. Consider this record, which was created from a TSV upload for record type `sample` and specifying a pk column of `sample_id`:
```json
{
  "id": "HCC1143",
  "type": "sample",
  "attributes": {
    "sample_id": "HCC1143",
    "platform": "illumina",
    "sample_type": "Tumor",
    "strip_unpaired": true
  }
}
```

davidangb: > But otherwise, Record attributes don't include the primary key?

That's the bug we are fixing! The pk should be in the attributes.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/ParquetRecordConverter.java
davidangb: fixed the code comment, thanks for the catch

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/ParquetRecordConverter.java
nawatts: Ok, if that's the expectation for all cases, then I feel much better about this.

In that case (and where my confusion came from), there's a bunch of tests with unrealistic data (Records without the primary key in attributes).

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/ParquetDataTypesTest.java
nawatts: Still confused about the intent of this test. Otherwise, 👍

nawatts (approved):

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/ParquetDataTypesTest.java
davidangb: I've renamed the test method name, good call (this PR inherited those tests and I didn't read close enough).

The differentiation in this test is that it contains a column whose name is `${entityType}_id`, which is a magic string in Rawls. The test verifies that the Rawls magicness doesn't affect Parquet translation.

PR #798, opened 2024-05-16
AJ-1848: Fix retryable bugs.

jladieu:
* Downgrade error logs to warning.  Eror logs are reported to Sentry via https://docs.sentry.io/platforms/java/guides/logback/; these should only be reported to Sentry when all retry attempts failed.
* Fix bug in test verifying behavior of first attempt, which was looking for an incorrect key, hiding the next bug...
* Fix bug where first attempt is recorded as `retryCount` = 0.  This makes the promQL to detect retries more complicated, so prefer not to count it.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/retry/RestClientRetry.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/retry/RestClientRetry.java
@@ -1,5 +1,7 @@
 package org.databiosphere.workspacedataservice.retry;
 
+import static java.util.Objects.requireNonNull;
+
 import io.micrometer.observation.Observation;
 import io.micrometer.observation.ObservationRegistry;
 import java.util.Objects;
@@ -71,16 +73,16 @@ public class RestClientRetry {
       switch (exceptionHttpCode) {
           // retryable http codes
         case 0:
-          LOGGER.error(loggerHint + " REST request resulted in connection failure", e);
+          LOGGER.warn(loggerHint + " REST request resulted in connection failure", e);
           throw new RestConnectionException();
         case 500:
         case 502:
         case 503:
         case 504:
-          LOGGER.error(
+          LOGGER.warn(
               loggerHint + " REST request resulted in ApiException(" + exceptionHttpCode + ")", e);
           throw new RestServerException(
-              HttpStatus.resolve(exceptionHttpCode), e.getMessage()); // retryable
+              requireNonNull(HttpStatus.resolve(exceptionHttpCode)), e.getMessage()); // retryable
 
           // non-retryable http codes
         case 401:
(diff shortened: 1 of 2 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/retry/RestClientRetry.java
```diff
@@ -71,16 +73,16 @@ public <T> T withRetryAndErrorHandling(RestCall<T> restCall, String loggerHint)
...
         case 500:
         case 502:
         case 503:
         case 504:
-          LOGGER.error(
+          LOGGER.warn(
               loggerHint + " REST request resulted in ApiException(" + exceptionHttpCode + ")", e);
           throw new RestServerException(
-              HttpStatus.resolve(exceptionHttpCode), e.getMessage()); // retryable
+              requireNonNull(HttpStatus.resolve(exceptionHttpCode)), e.getMessage()); // retryable
```
jladieu, line 85: This had an IDE warning about potentially being `null`, but it will never actually be `null` owing to the protection of the prior case statements.

service/src/test/java/org/databiosphere/workspacedataservice/retry/RestClientRetryTest.java
```diff
@@ -362,7 +362,7 @@ void doesNotCountSingleAttemptAsRetry() {
         .hasNumberOfObservationsWithNameEqualTo("wds.outbound", 1)
         .hasObservationWithNameEqualTo("wds.outbound")
         .that()
-        .doesNotHaveLowCardinalityKeyValue("retryAttempt", "0") // don't bother counting 0th attempt
+        .doesNotHaveLowCardinalityKeyValue("retryCount", "0")
```
jladieu, line 365: This bug in the test was hiding the fact that the metric was actually counted.

service/src/main/java/org/databiosphere/workspacedataservice/retry/RestClientRetry.java
```diff
@@ -101,7 +103,9 @@ public <T> T withRetryAndErrorHandling(RestCall<T> restCall, String loggerHint)
           .ifPresent(
               context -> {
                 int retryCount = context.getRetryCount();
-                observation.lowCardinalityKeyValue("retryCount", Integer.toString(retryCount));
+                if (retryCount > 0) {
```
jladieu, line 106: This fixes the bug where the metric was being counted for the 0th attempt.  This was unintended; by counting the 0th attempt, the promQL query to measure retries gets more complicated, and the prom scrapes are unnecessarily bloated.

calypsomatic (approved):

nawatts (approved):

PR #799, opened 2024-05-17
[AJ-1809] Pass ImportJobInput to jobs

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1809

Some import types support options. For example, TDR snapshot imports have an option `tdrSyncPermissions` to sync the snapshot's permissions with the destination workspace's. Rawls JSON imports have an `isUpsert` option.

Currently, `ImportService` adds each of these options individually as an argument to the `Schedulable` job. It also handles default values for each argument.

To generalize options handling, this adds the options to `ImportJobInput` and changes `ImportService` to pass the entire `ImportJobInput` object JSON encoded to the `Schedulable` job. It's then decoded by `ImportDetailsRetriever` and added to `ImportDetails`.

Each type of job has a dedicated `ImportOptions` subclass that parses the options from the import request and handles default values.


service/src/test/java/org/databiosphere/workspacedataservice/common/JsonUtils.java
```diff
@@ -0,0 +1,42 @@
+package org.databiosphere.workspacedataservice.common;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.stereotype.Component;
+
+@Component
+public class JsonUtils implements InitializingBean {
```
nawatts, line 9: Convenience for dealing with JSON in tests. I wanted to be able to serialize some JSON in a static method.

davidangb (commented):
Here's an enhancement for you to consider…

Instead of `ImportJobInput` having a loosely-modeled `Map<String, Object> options`, create some strongly-typed/strongly-modeled classes to represent `PfbOptions`, `TdrManifestOptions`, `RawlsJsonOptions`. Since we can easily read the type (pfb, tdrmanifest) of each import request, we can know which class we should deserialize to.

This would give us nicely-modeled options objects within the Java code. Of course it means we have to update those models every time we need to support a new option, but  supporting a new option necessarily involves new code anyway.

What do you think, is it worth it?

nawatts:
That makes sense. I was thinking we should have some kind of up-front validation on options, so that kind of ties into it.

I have no idea how to convince Java to deserialize a different class for options based on type though. Would that require a separate ImportJobInput subclass for each import type?

nawatts:
It seems like we don't really get any real type improvement without subclasses of `ImportJobInput`. Otherwise, `ImportJobInput` allows mismatches between `importType` and the type of `options`.

nawatts:
> There may be a simpler approach such as [used in PostgresJobDao](https://github.com/DataBiosphere/terra-workspace-data-service/blob/ba9e5818a18f3678344b446e8afe968fe80261b5/service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java#L302-L309): use a switch/case or if/else to determine the target class you want to deserialize to, then pass that target class as argument to readValue

Ok, this has given me an idea...

nawatts:
I'm not sure this can be changed without a migration or work for backwards compatibility with jobs already stored in the database.
https://github.com/DataBiosphere/terra-workspace-data-service/blob/ba9e5818a18f3678344b446e8afe968fe80261b5/service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java#L301-L311

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbTestUtils.java
nawatts, line 1: This was being used for both PFB and TDR tests. This removes TDR specific options from this class and moves them to a new `TdrManifestTestUtils` class.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportJobInput.java
```diff
@@ -1,16 +1,77 @@
...
+    public ImportJobInput deserialize(JsonParser parser, DeserializationContext context)
+        throws IOException {
+
+      JsonNode node = parser.readValueAsTree();
+
+      URI uri = URI.create(node.get("uri").asText());
+      TypeEnum type = TypeEnum.fromValue(node.get("importType").asText());
+
+      JsonNode optionsNode = node.get("options");
+      ObjectMapper mapper = new ObjectMapper();
```
davidangb, line 66: this is _probably_ safe, but it uses an ObjectMapper with different configuration than the rest of WDS - see `JsonConfig` for all the config we set. I think if you do:
```java
ObjectMapper mapper = (ObjectMapper) parser.getCodec();
```
you get the one we configured … at least, that's what `AttributeValueDeserializer` does …

service/src/test/java/org/databiosphere/workspacedataservice/common/JsonUtils.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+  }
+
+  public static <T> T parse(String string, Class<T> valueType) {
+    try {
+      return getInstance().objectMapper.readValue(string, valueType);
+    } catch (JsonProcessingException e) {
+      throw new RuntimeException("Error parsing JSON", e);
+    }
+  }
```
davidangb, line 41: mark `stringify` and `parse` as `@VisibleForTesting`, since they're only used for tests … but, you might be able to get rid of this entirely if you can make `ImportJobInput` `Serializable`

davidangb (commented):
some comments on using `Serializable` and cleaning up test-only methods, great work on getting the input polymorphism working!

service/src/main/java/org/databiosphere/workspacedataservice/service/ImportService.java
```diff
@@ -113,9 +115,7 @@ public GenericJobServerModel createImport(
       arguments.put(ARG_TOKEN, petToken);
       arguments.put(ARG_URL, importRequest.getUrl().toString());
       arguments.put(ARG_COLLECTION, collectionId.toString());
-
-      // maybe pass through import-type specific options
-      maybePassThroughOptions(importRequest, arguments);
+      arguments.put(ARG_IMPORT_JOB_INPUT, objectMapper.writeValueAsString(importJobInput));
```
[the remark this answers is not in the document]
nawatts: I think so, but I didn't understand Serializable enough to be confident in it.

service/src/test/java/org/databiosphere/workspacedataservice/common/JsonUtils.java
nawatts: > mark stringify and parse as @VisibleForTesting, since they're only used for tests

I thought that was implied by this class being in the test package?

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportJobInput.java
```diff
@@ -1,16 +1,77 @@
...
 
 /** User-supplied input arguments for a data import job */
-public record ImportJobInput(URI uri, ImportRequestServerModel.TypeEnum importType)
+@JsonDeserialize(using = ImportJobInput.ImportJobInputDeserializer.class)
+public record ImportJobInput(URI uri, TypeEnum importType, ImportOptions options)
     implements JobInput {
 
+  public ImportJobInput(URI uri, TypeEnum importType) {
+    this(uri, importType, defaultOptionsForType(importType));
+  }
```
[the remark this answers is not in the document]
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/799/commits/73bd015e1d6fdc58d65da39c349793c7c3d596ef.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportJobInput.java
nawatts: Changed in https://github.com/DataBiosphere/terra-workspace-data-service/pull/799/commits/d7c7275d8d3d71c33d5503d265ac2383cc80fdc7.

service/src/test/java/org/databiosphere/workspacedataservice/common/JsonUtils.java
nawatts: > you might be able to get rid of this entirely if you can make `ImportJobInput` `Serializable`

Removed in https://github.com/DataBiosphere/terra-workspace-data-service/pull/799/commits/e90fd032f1e4ca13c728cc59f71e71581697c6c9

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/JobDataMapReader.java
```diff
@@ -84,4 +85,17 @@ public boolean getBoolean(String key) {
...
+   * found/null.
+   */
+  public Object get(String key) {
+    try {
+      return Objects.requireNonNull(jobDataMap.get(key));
+    } catch (Exception e) {
+      throw new JobExecutionException(
+          "Error retrieving key %s from JobDataMap: %s".formatted(key, e.getMessage()), e);
+    }
+  }
```
davidangb, line 100: By adding a type parameter here, you could add validation and prevent call sites from having to do their own unchecked casts. Any `ClassCastException` would get caught and rethrown as `JobExecutionException`:
```suggestion
  /**
   * Retrieve a value from a JobDataMap. Throws a JobExecutionException if the value is not
   * found/null or of the wrong type.
   */
  @SuppressWarnings("unchecked")
  public <T> T get(String key) {
    try {
      return (T) Objects.requireNonNull(jobDataMap.get(key));
    } catch (Exception e) {
      throw new JobExecutionException(
          "Error retrieving key %s from JobDataMap: %s".formatted(key, e.getMessage()), e);
    }
  }
```

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/ImportDetailsRetriever.java
```diff
@@ -35,6 +36,10 @@ public ImportDetails fetch(UUID jobId, JobDataMapReader jobData, PrefixStrategy
     // determine the workspace for the target collection
     CollectionId collectionId = CollectionId.of(targetCollection);
     WorkspaceId workspaceId = collectionService.getWorkspaceId(collectionId);
-    return new ImportDetails(jobId, userEmailSupplier, workspaceId, collectionId, prefixStrategy);
+
+    ImportJobInput importJobInput = (ImportJobInput) jobData.get(ARG_IMPORT_JOB_INPUT);
```
davidangb, line 40: See other comment - you can get rid of the cast here.

davidangb (approved):
One last tweak about casting, otherwise looks good and thank you! giving this a proactive :+1:

service/src/main/java/org/databiosphere/workspacedataservice/jobexec/JobDataMapReader.java
nawatts: 👍 I tried that at first, but Intellij complained about an unchecked cast, which it didn't in ImportDetailsRetriever.

davidangb: `@SuppressWarnings("unchecked")` ftw :)

PR #802, opened 2024-05-22
[AJ-1851] Add import metadata to TDR imports in control plane

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1851

Restoring some of Import Service's behavior... add `import:snapshot_id` and `import:timestamp` fields to all rows imported from TDR snapshots.

- New `MappedRecordSource` wraps another `RecordSource` and transforms the Records that it produces.
- When configured to do so, `TdrManifestQuartzJob` will use `MappedRecordSource` to add `import:snapshot_id` and `import:timestamp` fields to Records from the snapshot.
- Makes `RawlsAttributePrefixer` aware of the `import:` prefix so that it doesn't add the `tdr:` prefix to those fields.
- Enables this feature in the control plane configuration.


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestImportMetadata.java
```diff
@@ -0,0 +1,35 @@
...
+  public static final String IMPORT_METADATA_PREFIX = "import:";
+  private final UUID snapshotId;
+  private final Instant importTime;
+
+  public TdrManifestImportMetadata(UUID snapshotId, Instant importTime) {
+    this.snapshotId = snapshotId;
+    this.importTime = importTime;
+  }
+
+  public MapRecordFunction getAddToRecordFunction() {
```
davidangb, line 22: Does this need its own `MapRecordFunction` functional interface definition? Could this be a `UnaryOperator<Record>` instead? I could go either way - it would mean we could delete one very small file in the WDS codebase but we'd be also have very slightly less strong typing.

Another possibility - could this be a `Supplier<Map<String, Object>>` or a `Supplier<RecordAttributes>`, and just return (a function that returns) the two annotations? Of course then you'd have to add logic into the MappedRecordSource to add those two annotations to the record. I don't have a strong preference between `UnaryOperator` and `Supplier` here; it's your call - I'm just raising the idea.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestImportMetadata.java
```diff
@@ -0,0 +1,35 @@
...
+    String timestamp = ZonedDateTime.ofInstant(importTime, ZoneId.of("UTC")).format(formatter);
+
+    Map<String, Object> attributes =
+        Map.of(
+            "%stimestamp".formatted(IMPORT_METADATA_PREFIX),
+            timestamp,
+            "%ssnapshot_id".formatted(IMPORT_METADATA_PREFIX),
+            snapshotId.toString());
+
+    return record -> record.putAllAttributes(new RecordAttributes(attributes, record.getId()));
```
davidangb, line 33: I am pretty sure that `record.getId()` here is incorrect; that `RecordAttributes` constructor is actually asking for the _name of the primary key column_ instead of the pk value. But it also doesn't matter for this use case; it's only used for ordering the attributes in the json response. You could use the `new RecordAttributes(attributes)` constructor instead.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestImportMetadata.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class TdrManifestImportMetadata {
+  public static final String IMPORT_METADATA_PREFIX = "import:";
+  private final UUID snapshotId;
+  private final Instant importTime;
+
+  public TdrManifestImportMetadata(UUID snapshotId, Instant importTime) {
+    this.snapshotId = snapshotId;
+    this.importTime = importTime;
+  }
```
davidangb, line 20: Style suggestion: consider moving the `Map<String, Object> attributes` to the class level and populating it in the constructor. If you did that, you could eliminate the class-level `snapshotId` and `importTime` variables, and the `getAddToRecordFunction` method becomes a lot smaller.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsAttributePrefixer.java
```diff
@@ -39,13 +41,26 @@ public RawlsAttributePrefixer(PrefixStrategy prefixStrategy) {
...
 
+  /**
+   * Import metadata attributes already have a prefix. Do not add another one.
+   *
+   * @param attributeName the inbound original attribute name
+   * @return whether or not to add a prefix to this attribute name
+   */
+  private boolean shouldPrefix(String attributeName) {
+    return !attributeName.startsWith(IMPORT_METADATA_PREFIX);
+  }
```
davidangb, line 62: you could potentially make this very generic - if the attributeName already contains a colon (indicating it already has a prefix), don't add another prefix.

davidangb (commented):
I like this approach, a handful of comments inline, thank you!

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestImportMetadata.java
nawatts: I did not know `UnaryOperator` existed. That sounds like effectively the same thing.

> Another possibility - could this be a Supplier<Map<String, Object>> or a Supplier<RecordAttributes>, and just return (a function that returns) the two annotations? Of course then you'd have to add logic into the MappedRecordSource to add those two annotations to the record. I don't have a strong preference between UnaryOperator and Supplier here; it's your call - I'm just raising the idea.

`MappedRecordSource` began life as `RecordAnnotator` which did something like this: take a map of attribute values and add them to all Records from a source. But `MappedRecordSource` seemed more generic and possibly broadly useful.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestImportMetadata.java
nawatts: Ah, yeah, I think you're right. Followed the example of some existing tests which are also wrong.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestImportMetadata.java
nawatts: Replaced with `UnaryOperator` in https://github.com/DataBiosphere/terra-workspace-data-service/pull/802/commits/37d86b3f74967d9cfa095280e981a5001c04bb18.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestImportMetadata.java
nawatts: Fixed this and clarified `RecordAttributes`' argument name in https://github.com/DataBiosphere/terra-workspace-data-service/pull/802/commits/c177130f6e8dd4641f72ce6cad21c86a88bd01bb.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/tdr/TdrManifestImportMetadata.java
nawatts: Fixed this and clarified `RecordAttributes`' argument name in https://github.com/DataBiosphere/terra-workspace-data-service/pull/802/commits/c177130f6e8dd4641f72ce6cad21c86a88bd01bb.

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RawlsAttributePrefixer.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/802/commits/195bc192f988f27bf7dc51019f1957fae23bfae0.

davidangb (approved):
nice, thank you!

calypsomatic (approved):

PR #803, opened 2024-05-22
AJ-1809: polymorphic import input

davidangb:
* `ImportJobInput` changes from a record to an abstract class
  * the change in getter names incurs compatibility changes elsewhere in the code
* create three new `ImportJobInput` subclasses: `PfbJobInput`, `RawlsJsonJobInput`, `TdrManifestJobInput`. Each of these hardcodes their correct `TypeEnum` and the correct `ImportOptions` subclass they use for options
  * also incurs a bunch of changes in tests and elsewhere; call sites must now instantiate a subclass instead of the abstract base class
* delete the custom `ImportJobInputDeserializer` json deserialization code in favor of `@JsonTypeInfo` and`@JsonSubTypes` annotations


nawatts (approved):

jladieu (approved):
Looks good to me!

The only rough edge I don't love is how we still have to cast the result of `jobInput.getOptions()` when the subclass of `QuartzJob` needs it, but I don't think this is a biggie.

PR #805, opened 2024-05-29
AJ-1656: Prefer `@SingleTenant` over direct reference to `twds.instance.workspace-id`

jladieu:
- `@SingleTenant WorkspaceId` provides additional guardrails over the direct reference to `twds.instance.workspace-id` so use it where possible
- Also, where possible, stop directly configuring `twds.instance.workspace-id` in test cases, preferring the common one set by `TestBase`
- In the small places where possible, refer to `TestBase.HARDCODED_WORKSPACE_ID` to make coupling more explicit.
- There are still several non-Java files that hardcode the same  `HARDCODED_WORKSPACE_ID`
- There are still several test cases which rely on direct configuration of `twds.instance.workspace-id`, so not _all_ references are eliminated


service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
```diff
@@ -50,7 +49,7 @@
 
 @SpringBootTest
 @DirtiesContext
-@TestInstance(Lifecycle.PER_CLASS)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
jladieu, line 52: Long story, but I inadvertently switched a couple of these over to `@TestInstance(Lifecycle.PER_CLASS)` in touched classes, and decided to revert that change.  However, in doing so, I realized that there were only two spots where we were doing this, so I undid those as well.

calypsomatic (approved):

nawatts (approved):

jladieu:
Pushed fixes for the 3 sonar issues (amended into existing commits), which amounted to removing unused imports.

PR #814, opened 2024-05-31
AJ-1658 Prefer `CollectionId` over `UUID`

jladieu:
Change signature of `CollectionDao#createSchema` to use `CollectionId`
instead of `UUID`.

### DIFF
--- a/service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerSearchFilterTest.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerSearchFilterTest.java
@@ -1,5 +1,6 @@
 package org.databiosphere.workspacedataservice.controller;
 
+import static java.util.UUID.randomUUID;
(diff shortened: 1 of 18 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/controller/RecordControllerSearchFilterTest.java
```diff
@@ -53,7 +56,7 @@ void beforeAll() {
       String leadingZeroes = String.format("%03d", i);
 
       recordOrchestratorService.upsertSingleRecord(
-          COLLECTION_ID,
+          COLLECTION_UUID,
```
jladieu, line 59: Ideally we would be able to drive `COLLECTION_UUID` occurrences to zero while cutting over various APIs to prefer `CollectionId` over `UDID`, so this might be a good search term to act as a proxy for how often `UUID`s are in use.

nawatts (approved):

calypsomatic (approved):

PR #818, opened 2024-06-04
[AJ-1865] Support date/time types in imports from TDR

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1865

Currently, several types of date/time values in GCP TDR imports show up as numbers in data tables.

The cause of the issue is that CWDS only looks at physical type of the data in the Parquet file and ignores the logical type.
This a similar issue to [AJ-1674](https://broadworkbench.atlassian.net/browse/AJ-1674).

The logical types for GCP TDR / BigQuery's date, datetime, time, and timestamp types can be found at https://cloud.google.com/bigquery/docs/reference/storage#arrow_schema_details.

This adds support for the logical types used by those GCP TDR types.
- Date values in GCP TDR are exported as integers representing the number of days since the epoch. Previously, these were stored as a `BigDecimal` attribute on the imported Record. Now, they are stored as a `LocalDate` attribute.
- Datetimes and timestamps in GCP TDR are exported as longs containing the number of microseconds since the epoch. Previously, these were stored as a `BigDecimal` attribute on the imported Record. Now, they are stored as a `LocalDateTime` object.
- Time values in GCP TDR are exported as longs containing the number of microseconds from the beginning of the day. Previously, these were stored as a `BigDecimal` attribute on the imported Record. Now, they are stored as a string formatted `HH:mm:ss` (since WDS has no time type).

We rely on default string conversion to serialize `LocalDate` and `LocalDateTime` attributes in WDS Records to strings for Rawls.

Based on the ticket description, this does result in a change in behavior compared to import service: datetimes are now formatted in ISO 8601 (`yyyy-MM-ddTHH:mm:ss`) vs the old `yyyy-MM-dd HH:mm:ss`. This was done to standardize the date/time format in data tables. For [AJ-1851](https://broadworkbench.atlassian.net/browse/AJ-1851), we restored the `import:timestamp` field, which is (and was also previously) formatted as ISO 8601. It doesn't make sense for imports to use two different date/time formats, so we standardize on the standard one here.

Azure TDR exports only include a logical type for date values. Thus, this also adds support for date values in Azure TDR exports. Other types (datetime, timestamp, and time) remain unsupported because there is no way to distinguish them from a number. This is an issue with Azure TDR, similar to its issues with array types ([AJ-1735](https://broadworkbench.atlassian.net/browse/AJ-1735)).

[AJ-1674]: https://broadworkbench.atlassian.net/browse/AJ-1674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AJ-1851]: https://broadworkbench.atlassian.net/browse/AJ-1851?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/AvroRecordConverter.java
```diff
@@ -97,7 +102,8 @@ protected RecordAttributes extractBaseAttributes(
           objectAttributes.get(fieldName) == null
               ? null
               : convertAttributeType(
-                  destructureElementList(objectAttributes.get(fieldName), field));
+                  destructureElementList(objectAttributes.get(fieldName), field),
+                  getLogicalTypeForValues(field));
```
davidangb, line 106: I think it would be better to pass the entire `field` object into `convertAttributeType`, then call `getLogicalTypeForValues` just-in-time only for those attributes that need it (collections, ints, longs). Else, we're executing `getLogicalTypeForValues` for every single attribute. Since this gets called `(# rows x # attributes)` times, that could be a big reduction. This might incur some refactoring for how to extract the type for collections.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/AvroRecordConverter.java
```diff
@@ -280,6 +316,29 @@ private Object convertFixed(GenericData.Fixed fixedAttr) {
...
+    }
+
+    if (logicalType instanceof LogicalTypes.TimeMicros) {
+      LocalTime time = LocalTime.ofNanoOfDay(longAttr * 1000);
+      return time.format(DateTimeFormatter.ISO_LOCAL_TIME);
+    }
+
+    return BigDecimal.valueOf(longAttr);
+  }
+
```
davidangb, line 341: do you think the `convertFixed` method needs any refactoring, such as to use the logicaltype passed in to `convertAttributeType` instead of calculating that value itself? I'm ok leaving things as-is, just wanna raise the question

davidangb (commented):
It bugged me not knowing _why_ we have the difference in format, so I tracked it down. At

https://github.com/broadinstitute/import-service/blob/a818658d0850818804235c5dc7f239c0de95c8cf/app/translators/tdr_manifest_to_rawls.py#L156

we explicitly use `.isoformat()` to set the `import:timestamp` value.

However, at
 
https://github.com/broadinstitute/import-service/blob/a818658d0850818804235c5dc7f239c0de95c8cf/app/translators/tdr_manifest_to_rawls.py#L222-L225

we use `str(value)` for otherwise-unhandled types such as datetimes … `str()` omits the `T` separator that `.isoformat()` uses:
```python
>>> from datetime import datetime
>>> mydt = datetime.now()
>>> mydt
datetime.datetime(2024, 6, 7, 13, 22, 56, 461508)
>>> str(mydt)
'2024-06-07 13:22:56.461508'
>>> mydt.isoformat()
'2024-06-07T13:22:56.461508'
```
*sigh*

Please make sure to add release notes to the jira ticket so the change is properly announced!

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/AvroRecordConverter.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/818/commits/33b16b76a3822c6e4b26d30ffb9ee0157c1b0d63.

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/AvroRecordConverter.java
nawatts: Done in https://github.com/DataBiosphere/terra-workspace-data-service/pull/818/commits/7832e549ea05734da180035f5f5582784dced4e8.

davidangb (approved):
thank you for the changes!

PR #823, opened 2024-06-07
AJ-1808: don't publish messages to Rawls if import translation fails

davidangb:
Before this PR, cWDS would always publish a "start import" pub/sub message to Rawls, even if the import failed early in the process and there was nothing for Rawls to import.

Now, `RecordSink` has a new `success()` method, which the PFB and TDR jobs call as their last step, and pub/sub publishing is contained in that method. The rawlsjson import job already did the right thing w/r/t publishing.

In previous discussion with Josh, we talked about fancy ways to set callback handlers on the RecordSink, or to automatically detect success vs. failure. After getting into the code, that seemed like overkill, and this approach feels pragmatic.

I considered also adding a `failure(Exception e)` method on `RecordSink` to flesh out its API, but … there's nothing to do there, so I ignored it.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RecordSink.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RecordSink.java
@@ -44,12 +44,20 @@ public interface RecordSink extends AutoCloseable {
   void deleteBatch(RecordType recordType, List<Record> records) throws DataImportException;
 
   /**
-   * Callback invoked at the end of a series of batches operations with the result. Implementers can
-   * commit changes, clean up resources, publish results, etc.
+   * Callback always invoked at the end of a series of batch operations. This should execute any
+   * code to run on both success and failure, such as closing open files.
    *
    * @throws DataImportException if an error occurs while closing the sink
    */
   default void close() throws DataImportException {
     // no-op
   }
+
+  /**
+   * Method manually invoked by callers to indicate success of all batch operations. This should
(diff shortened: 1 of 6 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RecordSink.java
```diff
@@ -44,12 +44,20 @@ void upsertBatch(
...
    */
   default void close() throws DataImportException {
     // no-op
   }
+
+  /**
+   * Method manually invoked by callers to indicate success of all batch operations. This should
+   * execute any code to run only on success, such as sending events to other systems.
+   *
+   * @throws DataImportException if an error occurs while closing the sink
```
nawatts, line 60: Does this still need to throw?

Also, nit on the comment: "closing the sink" looks like it was copied from the `close` docstring.

nawatts (approved):

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/RecordSink.java
davidangb: > Does this still need to throw?

It doesn't need to throw given the current implementation. But, I left the `throws DataImportException` present because every other method in this interface specifies allows throwing that exception … so it was consistent from the perspective of "a RecordSink can always raise a DataImportException error".

I don't feel strongly about this - if you think it would be better to remove, I'll pull it out.

PR #829, opened 2024-06-13
AJ-1854 add concurrency groups to GH workflows

calypsomatic:
https://broadworkbench.atlassian.net/browse/AJ-1854

Adds concurrency groups to 3 workflows that run for pull requests.  
Reviewer, please verify that the [github.head_ref](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) is the appropriate group name.  In the case of `publish_pacts.yaml`, this could be blank (if it's on `push` instead of `pull_request`), in which case I use `github.ref`.  (Or should I remove it from running on `push` as well?) 
Updates `build-and-test.yaml` to only run on `pull_request`, not also `push`. 
Todo:
[ ] Update GHA structure diagram to reflect this change.

Verified that concurrent runs were cancelled - I received emails alerting me to cancelled runs as I was committing to this PR.

Reminder:

#### Releasing WDS ####

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -6,6 +6,10 @@ on:
   pull_request:
     branches: [ '**' ]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -1,8 +1,8 @@
 name: Build and Test
 
 on:
-  push:
-    paths-ignore: [ '*.md' ]
+  # push:
+  #   paths-ignore: [ '*.md' ]
   pull_request:
     branches: [ '**' ]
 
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -1,11 +1,13 @@
 name: Build and Test
 
 on:
-  push:
-    paths-ignore: [ '*.md' ]
```
davidangb, line 5: I don't think we want to stop running this on merge to main. Is there a reason to stop doing so? I think this needs to be:
```yaml
on:
  push:
    branches:
      - main
```

.github/workflows/build-and-test.yml
```diff
@@ -1,11 +1,13 @@
...
   pull_request:
     branches: [ '**' ]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref }}
+  cancel-in-progress: true
```
davidangb, line 9: and, assuming we keep running this workflow on push to main, this would need to be:
```
group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
```

.github/workflows/build-and-test.yml
calypsomatic: Ah yeah I didn't think that through

davidangb (approved):

nawatts (approved):

PR #838, opened 2024-06-21
[AJ-1896] Delete bee from e2e tests

calypsomatic:
The e2e test was only deleting bees if the input said to, but never had any input.  This changes it to always delete the bee and removes reliance on the non-existent input.

Reminder:

#### Releasing WDS ####

### DIFF
--- a/.github/workflows/run-e2e-tests.yaml
+++ b/.github/workflows/run-e2e-tests.yaml
@@ -20,7 +20,7 @@ jobs:
         id: extract-inputs
         run: >
           echo "branch=${{ inputs.branch || 'develop' }}" >> "$GITHUB_OUTPUT"
-          echo "delete-bee=${{ inputs.delete-bee || false }}" >> "$GITHUB_OUTPUT"
+          echo "delete-bee=${{ inputs.delete-bee || true }}" >> "$GITHUB_OUTPUT"
   params-gen:
     runs-on: ubuntu-latest
     outputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/run-e2e-tests.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
         id: extract-inputs
         run: >
           echo "branch=${{ inputs.branch || 'develop' }}" >> "$GITHUB_OUTPUT"
-          echo "delete-bee=${{ inputs.delete-bee || false }}" >> "$GITHUB_OUTPUT"
+          echo "delete-bee=${{ inputs.delete-bee || true }}" >> "$GITHUB_OUTPUT"
```
nawatts, line 23: Won't this always delete the BEE? Even if `inputs.delete-bee` is false?

calypsomatic: Hm, I guess so, unless  `||` only detects whether the variable is present and doesn't evaluate truthiness...  Well, shouldn't we just always delete the bee I guess?  I mean, I don't think there IS any input for this

nawatts: Always deleting the BEE sounds good to me. But if we're going to do that, I'd suggest we remove the `delete-bee` input entirely.

davidangb (approved):
:+1: as-is, I am also curious how other e2e tests handle this

nawatts (approved):

PR #842, opened 2024-06-25
AJ-1893: add total update count to observation for PFB and TDR imports

davidangb:
At a high level, changes in this PR:
* centrally define, as a bean, a [distribution summary](https://docs.micrometer.io/micrometer/reference/concepts/distribution-summaries.html) metric to track the number of upserts per import
* wrap the distribution summary bean in an `ImportMetrics` bean; this is meant to hold a collection of metrics.
    * This level of abstraction is not needed for this PR, but I plan to quickly follow with additional metrics to track snapshot counts, where the abstraction will be useful
* wire the `ImportMetrics` collection-of-metrics into the PFB and TDR import jobs
* persist the appropriate counts from PFB and TDR jobs into the distribution summary

---

Reminder:

#### Releasing WDS ####

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/metrics/MetricsConfig.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/metrics/MetricsConfig.java
@@ -62,6 +62,7 @@ public class MetricsConfig {
                     .build());
   }
 
+  // central definition for the wds.import.upsertCount distribution summary
   @Bean
   RecordUpsertDistributionSummary entityUpsertCountDistributionSummary(
       MeterRegistry meterRegistry) {
--- a/service/src/main/java/org/databiosphere/workspacedataservice/metrics/MetricsConfig.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/metrics/MetricsConfig.java
@@ -5,6 +5,7 @@ import static io.micrometer.core.instrument.config.MeterFilter.deny;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
+import io.micrometer.core.instrument.DistributionSummary;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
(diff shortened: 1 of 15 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/metrics/MetricsConfig.java
```diff
@@ -60,4 +61,18 @@ MeterRegistryCustomizer<MeterRegistry> addWdsVersionTag() {
...
+  // central definition for the wds.import.upsertCount distribution summary
+  @Bean
+  RecordUpsertDistributionSummary entityUpsertCountDistributionSummary(
+      MeterRegistry meterRegistry) {
+    DistributionSummary distributionSummary =
+        DistributionSummary.builder("wds.import.upsertCount")
+            .baseUnit("upserts")
+            .description("Number of upserts in this import job")
+            .publishPercentiles(0.25, 0.5, 0.75, 0.95, 0.99)
+            .register(meterRegistry);
```
davidangb, line 74: This configuration - using `.publishPercentiles(0.3, 0.5, 0.75, 0.95, 0.99)` - generates 7 new Prometheus metrics. These are 5 metrics for the 5 percentiles defined here, plus a `sum` and a `count`. The drawback to this approach is that the percentiles are calculated internally to Micrometer, before they are published to Prometheus/Grafana - which means each k8s replica will be calculating their own percentiles independently of each other. I am of the opinion that this is Good Enough To Start With and we'll be able to get useful info out of it.

The alternative is to use `.publishPercentileHistogram()`. This publishes an auto-generated set of new metrics covering 1 to `+Inf` which can then be calculated into percentiles in Grafana. However, by default this autogenerated 276 new metrics to cover the whole histogram. There's some tweaking we can do by setting an expected maximum value but using reasonable expectations I was only able to get it down to 82 new metrics.

So … I say let's go with predefined percentiles for now.

nawatts (approved):

calypsomatic (approved):

PR #846, opened 2024-07-02
AJ-1909: createOrModifyRecordType should consider primary key

davidangb:
https://broadworkbench.atlassian.net/browse/AJ-1909

The error occurred when importing a table that has > 5000 rows - i.e. across multiple batches. Our batch size is set to 5000.

This fixes the error and adds a unit test to verify behavior across batches.

Related to #796.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/WdsRecordSink.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/recordsink/WdsRecordSink.java
@@ -51,7 +51,7 @@ public class WdsRecordSink implements RecordSink {
           collectionId.id(),
           recordType,
           schema,
-          recordDao.getExistingTableSchemaLessPrimaryKey(collectionId.id(), recordType),
+          recordDao.getExistingTableSchema(collectionId.id(), recordType),
           records);
     }
     return schema;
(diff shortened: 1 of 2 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/recordsink/WdsRecordSink.java
```diff
@@ -51,7 +51,7 @@ public Map<String, DataTypeMapping> createOrModifyRecordType(
           collectionId.id(),
           recordType,
           schema,
-          recordDao.getExistingTableSchemaLessPrimaryKey(collectionId.id(), recordType),
+          recordDao.getExistingTableSchema(collectionId.id(), recordType),
```
davidangb, line 54: this is the root of the error. We previously inspected the schema returned by `getExistingTableSchemaLessPrimaryKey()`, which obviously didn't include the PK column, and therefore thought we had to re-add it.

I am surprised we haven't seen more problems with this.

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/recordsink/WdsRecordSink.java
nawatts: We have seen at least one other problem related to using `getExistingTableSchemaLessPrimaryKey` vs `getExistingTableSchema`: https://github.com/DataBiosphere/terra-workspace-data-service/pull/555#discussion_r1497535873

nawatts (approved):

PR #852, opened 2024-07-10
QA-2320: App versioning golden path

ichengchang:
Jira ticket: [QA-2320](https://broadworkbench.atlassian.net/browse/QA-2320)

Refactor workflows to use new app versioning logic.

**_App Versioning Workflow Diagram (WDS, CWDS)_**
GitHub App Versioning Workflow

**Objective of this PR**
The purpose of this PR is to standardize the app versioning workflow, ensuring that the tag/version file generated by the bumper action is consistently distributed to all workflows triggered by the same commit. We are using WDS / CWDS as an illustration for setting the standard. 

**Problem**
When code is integrated into the mainline (`main` for this repository) from multiple PRs, the GitHub workflow jobs from different commits can run in no specific order. This can introduce inconsistencies in app versioning as an observable edge case. There is extensive discussion on what motivates the current approach of solving this problem [here](https://broadworkbench.atlassian.net/wiki/spaces/IRT/pages/3156246529/App+Versioning+Golden+Path).

**Solution: Standardization**
- To solve this problem, we force the bumper action to run before other workflows. This produces an app context that's uploaded as artifacts, using the commit as a key. Although the time it takes for the bumper to complete is indeterministic, it typically only takes seconds.

- The app context is then distributed to workflow runs that need the app version, based on the commit that triggers these workflows. We use the GitHub REST API to download the app context from artifacts, with retries.

- Reduces workflow complexity: no longer need separate `bump-skip` and `tag.yml` actions.

#### Releasing WDS ####

[QA-2320]: https://broadworkbench.atlassian.net/browse/QA-2320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/build-and-test.yml
```diff
@@ -2,21 +2,64 @@ name: Build and Test
...
   pull_request:
-    branches: [ '**' ]
+    branches:
+      - main
+    paths-ignore: [ '**.md' ]
+    types: [opened, synchronize, reopened]
   push:
-    paths-ignore: [ '*.md' ]
-
-concurrency:
```
calypsomatic, line 9: Did you need to take concurrency groups out for this to work?  We wanted them here

.github/workflows/compute-app-context.yml
```diff
@@ -0,0 +1,107 @@
...
+#     - The part of the version which was bumped.
+#
+# This workflow computes the App Context and uploads the results as a dotenv file
+name: Compute App Context
+on:
+  pull_request:
+    branches:
+      - main
+    paths-ignore: [ '**.md' ]
+    types: [opened, synchronize, reopened]
```
calypsomatic, line 25: Trying to understand what `types` does.  The [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) seem to say that `opened, synchronize, reopened` is the default, so do we need to include it?

calypsomatic (commented):
Working on understanding what's happening here - do you suppose you could update the [GHA diagram](https://github.com/DataBiosphere/terra-workspace-data-service/wiki/GHA-structure-in-WDS) to match your changes?  At least once the PR merges, to keep the diagram up to date, but showing the proposed changes before merge would help reviewers, if you have the time.

.github/workflows/compute-app-context.yml
```diff
@@ -0,0 +1,107 @@
...
+          ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
+          token: ${{ secrets.BROADBOT_TOKEN }}
+
+      - id: ghctx
+        name: Useful GitHub Context
+        run: |
+          echo "Action Type: ${{ github.event.action }}"
+          echo "app_branch=${{ github.event_name == 'pull_request' && github.head_ref || env.MAINLINE }}" >> $GITHUB_OUTPUT
+          app_sha=${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
+          app_short_sha=${app_sha:0:7}
```
calypsomatic, line 63: Do we need double {{}} around these?  Looking at a [run](https://github.com/DataBiosphere/terra-workspace-data-service/actions/runs/9959364298/job/27516041635#step:4:5), it seems like it's not reading the variable

ichengchang:
> Working on understanding what's happening here - do you suppose you could update the [GHA diagram](https://github.com/DataBiosphere/terra-workspace-data-service/wiki/GHA-structure-in-WDS) to match your changes? At least once the PR merges, to keep the diagram up to date, but showing the proposed changes before merge would help reviewers, if you have the time.

Yes, I am also thinking of updating the diagram to explain the new flow. Thanks for reminding me @calypsomatic

.github/workflows/compute-app-context.yml
```diff
@@ -0,0 +1,111 @@
...
+    paths-ignore: [ '**.md' ]
+env:
+  DEFAULT_BUMP: patch
+  MAINLINE: main
+  VERSION_FILE_PATH: build.gradle
+  VERSION_LINE_MATCH: "^\\s*version\\s*=\\s*'.*'"
+
+jobs:
+  bumper-job:
+    if: "!startsWith(github.event.head_commit.message, 'bump ')"
```
davidangb, line 39: see previous comment about this `if` clause

scripts/ci-hooks/get-app-context.sh
```diff
@@ -0,0 +1,79 @@
...
+  local short_sha=${sha:0:7}
+  local retry_count=0
+  local sleep_time=$((${SLEEP_TIME}))
+  local max_retries=$((${MAX_RETRIES}))
+  local list_artifact_url="curl -s -w \"%{http_code}\" \
+  -L \
+  -H 'Accept: application/vnd.github+json' \
+  -H 'X-GitHub-Api-Version: 2022-11-28' \
+  -H 'Authorization: Bearer $GITHUB_TOKEN' \
+  \"https://api.github.com/repos/${GITHUB_REPO}/actions/artifacts?name=$sha\""
```
davidangb, line 24: what's the advantage of using an explicit call to the github API vs. using the `download-artifact` action?

davidangb (commented):
Some syntax-level comments inline, especially wondering about changes that don't seem related to versioning, such as changing the branch targets for pull requests.

At a high level, I'm still confused about the WHY of this PR. Could you provide a concise explanation of the before vs. after and explain why we need to make these changes?

I'm also wondering if you've investigated any alternatives, such as actions already available on the marketplace which might do the same things.

.github/workflows/build-and-test.yml
ichengchang: The concurrency group is only necessary for `compute-app-context.yml` workflow. The new `app-context` job in `build-and-test.yml` workflow is synchronized with `compute-app-context.yml` (through the appropriate commit key in artifacts) so no need for another concurrency group.

.github/workflows/compute-app-context.yml
ichengchang: You're right. I updated this with comment.

.github/workflows/compute-app-context.yml
ichengchang: Yes, this still worked, even it showed up as ${...}, it will correctly expand.

.github/workflows/build-and-test.yml
```diff
@@ -2,21 +2,65 @@ name: Build and Test
 
 on:
   pull_request:
-    branches: [ '**' ]
+    branches:
+      - main
```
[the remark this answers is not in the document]
ichengchang: @davidangb Thanks for leaving your feedback here. I updated the beginning of this PR with context info, hope that makes sense. Yes, about the change in `branches` config. I have a question, how often do we have PRs that don't target `main`. What's the purpose of opening a PR that targets another branch not the mainline? I'm trying to understand is this something done very often?

.github/workflows/build-and-test.yml
```diff
@@ -2,21 +2,65 @@ name: Build and Test
...
 
 jobs:
-  build:
+  # The app-context job retrieves the App Context artifact that was uploaded
+  # through the compute-app-context.yml workflow. If the last commit was
+  # a version file bump, this workflow is skipped. Additionally, this workflow
+  # runs when a PR targeting mainline is opened, updated (with merge commit or rebase),
+  # or when there is a push to the mainline branch.
+  app-context:
+    if: "!startsWith(github.event.head_commit.message, 'bump ')"
```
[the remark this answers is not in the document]
ichengchang: Thinking along the same line, let me see if there is a way to make this more specific.

.github/workflows/compute-app-context.yml
ichengchang: Replied to same above.

.github/workflows/build-and-test.yml
davidangb: That's not the only reason we implemented concurrency. We are also trying to reduce concurrent test runs. If an engineer commits to a PR and kicks off tests, then quickly issues another commit, we don't want two simultaneous test runs. We want to cancel the first and only run tests on the second commit.

scripts/ci-hooks/get-app-context.sh
davidangb: `download-artifact` has retries built into it; they're always on. Search for "retry" in their issues, for instance ([link](https://github.com/actions/download-artifact/issues?q=is%3Aissue+retry+is%3Aclosed)) to see some discussion of that feature

davidangb: are you running into stability issues where custom retry handling is necessary?

ichengchang: The `actions/download-artifact` has a built-in [retry](https://github.com/actions/download-artifact/issues/72) policy, but it doesn't appear to be user-configurable. The custom retry implementation allows the download script to wait for a period exceeding the built-in retry provided by `actions/download-artifact`, ensuring that the artifact becomes available even if there are temporary delays.

.github/workflows/build-and-test.yml
```diff
@@ -48,6 +92,8 @@ jobs:
       - uses: actions/checkout@v4
         # Needed by sonar to get the git history for the branch the PR will be merged into.
         with:
+          ref: ${{ needs.app-context.outputs.app_sha }}
+          token: ${{ secrets.BROADBOT_TOKEN }}
```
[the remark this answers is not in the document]
ichengchang: Copy-paste from another workflow. We can ignore token here, removed.

.github/workflows/build-and-test.yml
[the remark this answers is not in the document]
ichengchang: I think all we need is this condition `if: "github.event.head_commit.committer.name != 'bumptabot'"`. I pushed the fix.

.github/workflows/compute-app-context.yml
ichengchang: Addressed.

.github/workflows/compute-app-context.yml
```diff
@@ -0,0 +1,148 @@
...
+  VERSION_FILE_PATH: build.gradle
+  VERSION_LINE_MATCH: "^[ \\t]*version[ \\t]*=[ \\t]*'(.*)'"
+
+jobs:
+  bumper-job:
+    if: "github.event.head_commit.committer.name != 'bumptagbot'"
+    runs-on: ubuntu-latest
+    concurrency:
+      group: ${{ github.repository_owner }}-${{ github.event.repository.name }}-${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || 'main' }}-bumper
+      cancel-in-progress: false
```
calypsomatic, line 59: We had `cancel-in-progress: true` for when two commits are made in quick succession.  Is there a reason to switch it to `false`?

ichengchang: Here's the logic on why `false` was chosen based on the documentation below. We want to keep things very simple when it comes to _concurrency group_ so it is only applied to the bumper job in `compute-app-context.yml`. The bumper job runs once and won't repeat for the same PR commit or push to main. We use action `get-app-context/action.yml` to distribute the App Context data for all other workflow jobs within the same GitHub Context to maintain version consistency. 

According to the [concurrency group](https://docs.github.com/en/enterprise-cloud@latest/actions/using-jobs/using-concurrency#using-concurrency-in-different-scenarios) documentation, there can be at most one running and one pending job in a concurrency group at any time. Setting `cancel-in-progress: true` cancels a running job in the concurrency group when a new job is triggered.

Here we have chosen an option that will prevent versioning inconsistencies. Consider the following scenarios:

Scenario 1: `cancel-in-progress: true`
The `compute-app-context.yml` workflow is triggered by a merge, followed by another merge.
The workflow from the first merge (instance A) is running, and the workflow from the second merge (instance B) is pending.
Instance A gets canceled when instance B starts running, potentially leading to inconsistent versioning due to the premature termination of the bumper job.

Scenario 2: `cancel-in-progress: false`
The `compute-app-context.yml` workflow is triggered by a merge, followed by another merge.
The workflow from the first merge (instance A) is running, and the workflow from the second merge (instance B) is pending.
Instance A completes the bumper job before instance B starts running, ensuring version consistency.

To avoid inconsistency, we have chosen to set `cancel-in-progress: false`.

The above points are important details so please let me know if you would like to follow up with further discussion.

calypsomatic: Hm, in that case can we have `cancel-in-progress: false` for `compute-app-context` but leave it `true` for running the tests?

ichengchang: Assuming you are referring to the concurrency group in the `build-and-integration-test.yml` workflow: This is an automated workflow and currently does not associate the app version in reporting test results. You can leave the concurrency group setting as it is.

When the time comes to use app versioning in integration tests, the concurrency group will no longer be needed for `build-and-integration-test.yml` because it will be replaced by `get-app-context`, which will fetch the version like a promise.

calypsomatic: But can we add the concurrency group back into `build-and-test.yml`?

ichengchang: The `build-and-test.yml` job does not need a concurrency group. Here's why: the `app-context` job in `build-and-test.yml` uses the shared action `get-app-context/actions.yml`, which will automatically block until the App Context for versioning becomes available. Are there specific scenarios you have in mind that would require the concurrency group? If so, could you elaborate on those scenarios?

calypsomatic: See David's comment [here](https://github.com/DataBiosphere/terra-workspace-data-service/pull/852/files#r1681565463).  I suppose it's not the end of the world if we're running the same tests four times in a few minutes, but it's something we'd like to avoid if possible.

ichengchang: Gotcha! This makes sense. Let me take a look and reinstate the concurrency group. Thanks for pointing out David's comment, somehow I missed that.

ichengchang: I put the concurrency group back but made a slight modification to the name of the group. Also added a few lines of comments to explain why concurrency group is being used.

calypsomatic: Thank you!

calypsomatic (approved):

PR #854, opened 2024-07-11
[AJ-1934] add Troubleshooting section to README

mspector:
Does a README update need a Jira ticket for compliance? or is this fine?


calypsomatic (approved):
Logically, it shouldn't require a ticket, but I think it might anyway?

davidangb (approved):

PR #856, opened 2024-07-15
AJ-1933: handle NaNs in Parquet exports

davidangb:
Fixes a `Character N is neither a decimal digit number, decimal point, nor "e" notation exponential mark.` error when importing a TDR snapshot, if that snapshot contained `NaN` values.

TDR, BigQuery, and Parquet allow NaNs in floating-point columns. However, WDS did not handle this. If WDS saw a float/double, and the value was not `null`, we tried to feed it into the `BigDecimal` constructor. Now, we check for both `null` and `NaN` before calling `BigDecimal`.

WDS now treats `NaN` as `null`. At some point in the future, we could consider supporting `NaN` as a first-class value, since [Postgres also supports it](https://www.postgresql.org/docs/8.2/datatype-numeric.html#DATATYPE-NUMERIC-DECIMAL). But, I suspect that would have ripple effects elsewhere in our code.

---

_reviewer: I am prone to ignore the Sonar complexity error. Should I refactor `AvroRecordConverter.convertAttributeType()` instead, to reduce complexity?_


mspector (approved):

calypsomatic (approved):
I think we can ignore the sonar warning.

PR #857, opened 2024-07-15
AJ-1923: proposal signatures for v1 collection APIs

davidangb:
OpenAPI specs for the multi-tenant v1 collection APIs, plus the code classes they generate.

These definitions live ONLY in `apis-v1.yaml`. I have not added any `$ref`s for them into `openapi-docs.yaml`; this means they won't show up when running WDS. I suggest we add the `$ref`s as part of the follow-on work in AJ-1924 to avoid confusion.

If you want to visualize these APIs, copy-paste the contents of `apis-v1.yaml` into https://editor.swagger.io/:
Screenshot 15-07-2024 at 17 17

---

_Note: Sonar has a lot to say about this PR, but it's all about the generated code. I wonder how we can tell Sonar to ignore that._


mspector (approved):
I have some questions about the fact that the term `instance` still appears in parts of the spec, but I think we covered it in refinement, and I'm just not completely remembering the conversation.

Are `instance` and `collection` synonyms? or are they actually separate concepts? If they are synonyms, was the reasoning that replacing `instance` everywhere would be too big of a chance to do at one time?

davidangb:
@mspector "instance" and "collection" are synonyms. We started by calling this thing "instance", but that terminology was too confusing, and we renamed to "collection". We have updated a lot of places to use "collection" instead of "instance", but API signatures are the long tail due to issues with backwards compatibility. As we naturally replace existing APIs with newer versions, we'll use "collection" instead.

calypsomatic (approved):

PR #858, opened 2024-07-16
AJ-1924: use Spring Data JDBC to power v1 Collection APIs

davidangb:
Implement 3 of the 5 v1 collection APIs: create, delete, list. I am saving the get and update APIs for a separate PR, to keep PR size down. Also saving name validation for later.

This PR introduces Spring Data into WDS and uses Spring Data JDBC to power the database CRUD for the collections table.

See my comments inline to explain the PR.

I reviewed the 10 Sonar warnings; they're either about TODOs which I am saving for a future PR or deprecation warnings which … are for a future PR.


service/src/main/java/org/databiosphere/workspacedataservice/controller/CollectionController.java
```diff
@@ -0,0 +1,101 @@
...
+import org.databiosphere.workspacedataservice.shared.model.WorkspaceId;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.RestController;
+
+/**
+ * Implementations for collection API routes. This implements CollectionApi, which is auto-generated
+ * from our OpenAPI spec.
+ */
+@DataPlane
```
mspector, line 19: I see it isn't new, but this seems interesting! would love to learn more about it sometime

service/src/main/java/org/databiosphere/workspacedataservice/controller/CollectionController.java
```diff
@@ -0,0 +1,101 @@
...
+   *
+   * @param workspaceId Workspace id (required)
+   * @param collectionId WdsCollection id (required)
+   * @return The collection object. (status code 200)
+   */
+  @Override
+  public ResponseEntity<CollectionServerModel> getCollectionV1(
+      UUID workspaceId, UUID collectionId) {
+    // TODO: implement
+    return CollectionApi.super.getCollectionV1(workspaceId, collectionId);
```
mspector, line 70: testing my own understanding: when it comes time to implement this, we'll implement some kind of "get collection" method in collectionServiceV1? if so, does `CollectionApi.super.getCollectionV1` go unused?

service/src/main/java/org/databiosphere/workspacedataservice/dao/CollectionRepository.java
```diff
@@ -0,0 +1,26 @@
...
+ * Spring Data repository for managing WdsCollection objects in the sys_wds.collection table.
+ *
+ * <p>By extending CrudRepository, Spring will autogenerate 10+ methods to save, find, count and
+ * delete rows in the sys_wds.collection table. See CrudRepository for those method signatures.
+ *
+ * <p>When you inject a CollectionRepository bean into some other class, such as CollectionService,
+ * you can call those save/find/delete methods and they will just work. Spring generates a proxy
+ * class which implements those methods, and that proxy is what gets injected.
+ *
+ * <p>This means you won't see any code that actually implements those methods. They just work.
```
mspector, line 19: this comment is great. I think it satisfies the need for the kind of documentation I mentioned in our meeting.

going off on that tangent a bit further: I wonder if there's an easy way to make this kind of commentary more discoverable. I like the proximity of the comment to the relevant code; is it worth trying to also expose this more at the point of introduction of the repo to a new developer?

(I understand there's some effort being made to standardize READMEs and introductions to our repositories more generally)

service/src/main/java/org/databiosphere/workspacedataservice/controller/CollectionController.java
```diff
@@ -0,0 +1,101 @@
...
+  /**
+   * GET /collections/v1/{workspaceId} : List all collections in this workspace.
+   *
+   * @param workspaceId Workspace id (required)
+   * @return List of collections in this workspace. (status code 200)
+   */
+  @Override
+  public ResponseEntity<List<CollectionServerModel>> listCollectionsV1(UUID workspaceId) {
+    List<CollectionServerModel> collections = collectionServiceV1.list(WorkspaceId.of(workspaceId));
+    return new ResponseEntity<>(collections, HttpStatus.OK);
```
mspector, line 82: generally speaking, how do we decide what gets handled by the controller, and what gets handled by the service? is the controller a place where potentially multiple services meet?

mspector:
No further questions or feedback - it's a nice demonstration!

service/src/main/java/org/databiosphere/workspacedataservice/controller/CollectionController.java
davidangb: we use `@DataPlane` and `@ControlPlane` annotations to control which beans are instantiated in WDS vs. cWDS. In this case, we only want these APIs to be turned on for WDS in the data plane.

service/src/main/java/org/databiosphere/workspacedataservice/controller/CollectionController.java
davidangb: That's correct. At a high level:
* we define API and payload signatures in `apis-v1.yaml`
* we use openapi-generator to autogenerate server stubs for those APIs. In this case that's `CollectionApi`. If you look at CollectionApi you can see that is just a stub - it hardcodes a response of `NOT_IMPLEMENTED`
* we then hand-write controllers which inherit from/override the autogenerated stubs, with real implementations.

This approach ensures that the actual API signatures and payloads are true to the OpenApi spec, and that our spec is true to our implementation.

service/src/main/java/org/databiosphere/workspacedataservice/controller/CollectionController.java
davidangb: this is pretty philosophical, but general guidelines:
* controllers should be pretty light, focused on routing
* services are where we implement business logic, which could include talking to other services or to multiple DAOs
* DAOs encapsulate interaction with a single external target, be it some other service (Sam, TDR) or resource (Postgres)

service/src/main/java/org/databiosphere/workspacedataservice/controller/CollectionController.java
```diff
@@ -0,0 +1,101 @@
...
+   * @param workspaceId Workspace id (required)
+   * @param collectionId WdsCollection id (required)
+   * @return The collection object. (status code 200)
+   */
+  @Override
+  public ResponseEntity<CollectionServerModel> getCollectionV1(
+      UUID workspaceId, UUID collectionId) {
+    // TODO: implement
+    return CollectionApi.super.getCollectionV1(workspaceId, collectionId);
+  }
```
davidangb, line 71: `getCollectionV1` and `updateCollectionV1` are for a future PR, to keep PR size down

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
```diff
@@ -33,24 +48,186 @@ public class CollectionService {
...
+    if (!canCreateCollection(workspaceId)) {
+      // the user doesn't have permission to create a new collection.
+      // do they have permission to read the workspace at all?
+      if (canReadWorkspace(workspaceId)) {
+        throw new AuthorizationException("Caller does not have permission to create collection.");
+      } else {
+        throw new AuthenticationMaskableException("Workspace");
+      }
+    }
+    // TODO: validate name against the CollectionServerModel.getName() pattern
```
davidangb, line 98: saving this for a follow-on PR. There's some Spring validation magic we can use here, but it would enlarge the PR; I'm saving it for later

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
```diff
@@ -33,24 +48,186 @@ public class CollectionService {
...
+      throw dbActionExecutionException;
+    }
+    if (cause instanceof DuplicateKeyException duplicateKeyException) {
+      // kinda ugly: we need to determine if the DuplicateKeyException is due to an id conflict
+      // or a name conflict.
+      String msg = duplicateKeyException.getMessage();
+      if (msg.contains("duplicate key value violates unique constraint")
+          && msg.contains("instance_pkey")) {
+        // TODO: this allows phishing for collection ids.
+        throw new ConflictException("Collection with this id already exists");
```
davidangb, line 219: `// TODO: this allows phishing for collection ids.`:

A malicious user could determine the existence of collections in other workspaces. By specifying a uuid when creating a collection (in a workspace they do have permissions for), they could check for id conflicts. If there's a conflict, they will know that uuid exists somewhere else. I don't like this. One alternative is to disallow specifying the collection id. However, that would cause problems with our approach of "by convention, there exists a collection with the same id as the containing workspace".

Any good ideas here?

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
```diff
@@ -131,20 +308,32 @@ public void validateCollection(UUID collectionId) {
...
-    return getSamAuthorizationDao(getWorkspaceId(collectionId)).hasWriteWorkspacePermission();
+    return canWriteWorkspace(getWorkspaceId(collectionId));
+  }
+
+  private boolean canReadWorkspace(WorkspaceId workspaceId) {
+    return getSamAuthorizationDao(workspaceId).hasReadWorkspacePermission();
+  }
+
+  private boolean canWriteWorkspace(WorkspaceId workspaceId) {
+    return getSamAuthorizationDao(workspaceId).hasWriteWorkspacePermission();
```
davidangb, line 336: I've added a `canListCollections` helper. Beyond that, it's just refactoring: extracting `canReadWorkspace` and `canWriteWorkspace` helpers.

mspector (approved):
preemptive 👍 ... are those two SQL sonarcloud issues worth troubling ourselves over? is that an actual security risk?

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
calypsomatic: Aren't there just too dang many UUIDs for such phishing to be practical though?  Even programmatically running through possible UUIDs should be untenable.

calypsomatic (approved):

PR #859, opened 2024-07-18
[Needs ticket] Fix bee-retry name in e2e test

calypsomatic:
Recently when we had create-bee failures in our e2e test, I noticed that the retry failed with a [different error](https://github.com/broadinstitute/terra-github-workflows/actions/runs/9989978977/job/27609700090) than the original failure:  "duplicate key value violates unique constraint \"environments_name_unique\"".  It looks like the retry uses the same bee-name, which I'm guessing is the cause of the failure, if the first attempt got far enough to create something with the same name.  
Incidentally, it seems like we never use `RUN_NAME_SUFFIX`; its value is just a subset of `ATTACH_BP_TO_LZ_RUN_NAME`, so I also deleted that line.

Question for reviewers:
Will the addition of these two characters make the bee name too long or otherwise incorrect?  I seem to recall some failures with bee names not conforming to some expected format or length.

Reminder:

#### Releasing WDS ####

### DIFF
--- a/.github/workflows/run-e2e-tests.yaml
+++ b/.github/workflows/run-e2e-tests.yaml
@@ -8,7 +8,6 @@ env:
   BEE_NAME: 'wds-${{ github.run_id }}-${{ github.run_attempt}}-dev'
   TOKEN: '${{ secrets.BROADBOT_TOKEN }}'
   ATTACH_BP_TO_LZ_RUN_NAME: 'attach-billing-project-to-landing-zone-${{ github.event.repository.name }}-${{ github.run_id }}-${{ github.run_attempt }}'
-  RUN_NAME_SUFFIX: ${{ github.event.repository.name }}-${{ github.run_id }}-${{github.run_attempt }}
 jobs:
   init-github-context:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/run-e2e-tests.yaml
```diff
@@ -52,7 +51,7 @@ jobs:
         if: steps.FirstAttemptCreateBee.outcome == 'failure'
         uses: ./.github/actions/create-bee
         with:
-          bee_name: ${{ env.BEE_NAME }}
+          bee_name: ${{ env.BEE_NAME }}-2
```
mspector, line 54: can we use a more descriptive suffix for the retry? maybe something like `-retry`, or `-2nd-attempt`, or similar?

mspector (approved):

.github/workflows/run-e2e-tests.yaml
calypsomatic: Done

mspector (approved):
🎉 for deleting code!

davidangb (approved):
> Should we also remove the retry for attach-landing-zone?

Do you have any idea how frequently `attach-landing-zone` fails but is rescued by a retry? If infrequent, I am fine deleting that.

calypsomatic:
> > Should we also remove the retry for attach-landing-zone?
> 
> Do you have any idea how frequently `attach-landing-zone` fails but is rescued by a retry? If infrequent, I am fine deleting that.

No idea - you can't really search for it and I don't remember ever seeing it happen.

PR #861, opened 2024-07-24
AJ-1924: collection name validation

davidangb:
Follow-on PR to #858. Introduces validation for collection names.

We are quite likely to use collection names in URLs in the future. While url-encoding is certainly possible, it feels ok right now to limit collection names to alphanumeric/dash/underscore only.

Instead of writing a custom validation to do this, I'm relying on `spring-boot-starter-validation` and `openapi-generator`. The workflow here is:
1. Define the validation constraint in the OpenAPI spec, [here](https://github.com/DataBiosphere/terra-workspace-data-service/blob/26fda06444c1eb0c421c538f4c84eb8b43f00194/service/src/main/resources/static/swagger/apis-v1.yaml#L309)
2. `openapi-generator` uses this constraint to generate a `@Pattern` annotation on its model, as seen [here](https://github.com/DataBiosphere/terra-workspace-data-service/blob/26fda06444c1eb0c421c538f4c84eb8b43f00194/service/src/main/java/org/databiosphere/workspacedataservice/generated/CollectionServerModel.java#L74)
3. `openapi-generator` marks its API stub as `@Validated` [here](https://github.com/DataBiosphere/terra-workspace-data-service/blob/26fda06444c1eb0c421c538f4c84eb8b43f00194/service/src/main/java/org/databiosphere/workspacedataservice/generated/CollectionApi.java#L37), and marks the request body with `@Valid` [here](https://github.com/DataBiosphere/terra-workspace-data-service/blob/26fda06444c1eb0c421c538f4c84eb8b43f00194/service/src/main/java/org/databiosphere/workspacedataservice/generated/CollectionApi.java#L76)
4. The presence of `spring-boot-starter-validation` in the classpath automatically invokes validation for the API, as configured by `@Pattern`, `@Validated`, and `@Valid`; Spring throws a `MethodArgumentNotValidException` if the user inputs bad data
5. The `handleMethodArgumentNotValid` override in our pre-existing `@ControllerAdvice` `GlobalExceptionHandler` massages the `MethodArgumentNotValidException` into a somewhat-nicer error message to the end user

The resultant error message is still somewhat engineer-oriented, in that it exposes the regex:
Screenshot 24-07-2024 at 09 07
```json
{
  "timestamp": "2024-07-24T12:57:34.045+00:00",
  "status": 400,
  "error": "Invalid request content.",
  "message": "name: must match \"[a-zA-Z0-9-_]{1,128}\"",
  "path": "/collections/v1/b95017ef-d0c9-4baf-9a1b-a9e692c37bce"
}
```

But the flip side is that by including `spring-boot-starter-validation` and having generic handling for `MethodArgumentNotValidException`, we gain built-in validation for all other APIs and can easily add more validation in the future


service/src/main/java/org/databiosphere/workspacedataservice/controller/GlobalExceptionHandler.java
```diff
@@ -115,19 +152,31 @@ public ResponseEntity<Map<String, Object>> maskAuthenticationExceptions(
...
+    Map<String, Object> errorBody = new LinkedHashMap<>();
+
+    errorBody.put(TIMESTAMP, new Date());
+    errorBody.put(STATUS, statusCode);
+    errorBody.put(ERROR, error);
+    errorBody.put(MESSAGE, message);
+    errorBody.put(PATH, path);
+
+    return errorBody;
+  }
```
davidangb, line 178: I extracted this so I could reuse it from `handleMethodArgumentNotValid`

davidangb:
This PR got messy merge/rebase stuff - closing in favor of #862

PR #863, opened 2024-07-24
[AJ-1782] Address dnsjava vulnerability

calypsomatic:
Addresses https://github.com/DataBiosphere/terra-workspace-data-service/security/dependabot/95

Reminder:

#### Releasing WDS ####

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -149,6 +149,10 @@ dependencies {
         implementation('org.apache.commons:commons-configuration2:2.11.0') {
             because("CVE-2024-29131, CVE-2024-29133")
         }
+        implementation('dnsjava:dnsjava:3.6.0') {
+            because("CVE-2024-25638")
+        }
+
     }
 }
 
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -116,6 +116,7 @@ dependencies {
         exclude(group: 'org.apache.kerby')
         exclude(group: 'com.google.protobuf')
         exclude(group: 'org.bouncycastle')
+        exclude(group: 'dnsjava')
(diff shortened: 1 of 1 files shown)


davidangb (approved):
:+1: for syntax

This could use AJ-1782, probably doesn't need its own ticket

I am curious if you can exclude `dnsjava` via `withoutHadoopExcludes` [here](https://github.com/DataBiosphere/terra-workspace-data-service/blob/09445abb939834d5a39589c5f0f33087386eba0f/service/build.gradle#L105-L119) instead of updating it, or if it's really necessary for functionality

calypsomatic:
> I am curious if you can exclude `dnsjava` via `withoutHadoopExcludes` [here](https://github.com/DataBiosphere/terra-workspace-data-service/blob/09445abb939834d5a39589c5f0f33087386eba0f/service/build.gradle#L105-L119) instead of updating it, or if it's really necessary for functionality

The tests passed and bootRun started up when I excluded it; what should I explicitly test to verify we don't need is?

davidangb:
> what should I explicitly test to verify we don't need it?

@calypsomatic try running a real import into a locally-running WDS? If tests pass I'm pretty sure we're good, but a manual smoke test would be great

calypsomatic:
> > what should I explicitly test to verify we don't need it?
> 
> @calypsomatic try running a real import into a locally-running WDS? If tests pass I'm pretty sure we're good, but a manual smoke test would be great

Seems to work!

davidangb (approved):
reiterating :+1:

mspector (approved):

PR #867, opened 2024-07-24
AJ-1924: update-collection v1 API

davidangb:
In this PR:
* Implementation and test coverage for the update-collection v1 collection API. This is the last unimplemented v1 collection API!
* Update summary/description for the deprecated v0.2 instance APIs in the OpenAPI spec
* add `@WriteTransaction` annotation for save/delete methods

Runtime code changes are relatively small, but there's a lot of lines of test coverage.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
@@ -244,6 +244,15 @@ public class CollectionService {
             .find(workspaceId, collectionId)
             .orElseThrow(() -> new MissingObjectException(COLLECTION));
 
+    // optimization: if the request doesn't change anything, don't bother writing to the db
+    if (found.name().equals(collectionServerModel.getName())
+        && found.description().equals(collectionServerModel.getDescription())) {
+      // make sure this response has an id
+      CollectionServerModel response = new CollectionServerModel(found.name(), found.description());
+      response.id(found.collectionId().id());
+      return response;
+    }
+
     // update the found object with the supplied name and description
     WdsCollection updateRequest =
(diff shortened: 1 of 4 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
```diff
@@ -86,6 +87,7 @@ void setWorkspaceId(@Nullable @SingleTenant WorkspaceId workspaceId) {
    * @param collectionServerModel the collection definition
    * @return the created collection
    */
+  @WriteTransaction
```
davidangb, line 90: `save()` and `delete()` need `@WriteTransaction` because they execute multiple SQL queries which should be committed/rolled back atomically. `list()`, `get()`, and `update()` each execute only one SQL query, so they don't need an explicit transaction annotation.

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
```diff
@@ -121,24 +123,15 @@ public CollectionServerModel save(
...
 
-    // save
-    WdsCollection actual = null;
-    try {
-      actual = collectionRepository.save(wdsCollectionRequest);
-    } catch (DbActionExecutionException dbActionExecutionException) {
-      handleDbException(dbActionExecutionException);
-    }
+    // save, handle exceptions, and translate to the response model
+    CollectionServerModel response = saveAndHandleExceptions(wdsCollectionRequest);
```
davidangb, line 127: I extracted logic into a method that can be shared by `save()` and `update()`

calypsomatic (approved):

mspector (approved):

PR #868, opened 2024-07-25
AJ-1238: filter-by-column initial implementation

davidangb:
Initial spike for filter-by-column.

Limitations of this PR:
1. It only works for string attribute. It does not work for array-of-string or any other datatype.
2. It is exact-match only, not substring match or natural language search.
3. It is case-sensitive.
4. It works on a single column only; you can't search for e.g. "column1:foo AND column2:bar".
5. Users cannot search for null
6. The `totalRecords` value in the response of the `queryRecords` API may be considered incorrect. It returns the count of records for the specified record type _ignoring_ any filtering. So, if your filter only finds 1 record out of 100 samples, `totalRecords` will still return 100. 

In my opinion, we can incrementally roll out all of those enhancements. Some of them will require decisions on what we _want_ the business logic to be. Purpose of this PR is to get an initial spike merged so we have something to build on.

This adds a `filter.query` field to the `queryRecords` API request payload. The syntax for that `query` field is a small subset of [Lucene Query Parser syntax](https://lucene.apache.org/core/2_9_4/queryparsersyntax.html) - we only support the form `columnName:filterTerm` for now, with an intent to build out more features later.

### Example payloads for the `queryRecords` API

All defaults, no filtering:
```
{}
```

Filter the `sample` column for a value `HCC1143_10`:
```
{
  "filter": {
    "query": "sample:HCC1143_10"
  }
}
```

Filter the `sample_type` column for a value `Tumor`, within the three records with ids "HCC1143_1000", "HCC1143_1001", and "HCC1143_1002". Return the second page of 10 results, sorted by "sample_id" descending:
_note that filter-by-ids is not part of this PR; this example just shows how the filters interact_
```
{
  "offset": 10,
  "limit": 10,
  "sortAttribute": "sample_id",
  "sort": "DESC",
  "filter": {
    "ids": ["HCC1143_1000", "HCC1143_1001", "HCC1143_1002"],
    "query": "sample_type:Tumor"
  }
}
```


service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -270,19 +271,38 @@ public List<Record> queryForRecords(
...
+        // the filter index
+        var paramName = "filter" + idx;
+        whereClause.append(" and ");
+        whereClause.append(quote(filter.getColumn()));
+        whereClause.append(" = :");
+        whereClause.append(paramName);
+        sqlParams.addValue(paramName, filter.getFind());
+      }
+    }
+
```
davidangb, line 305: todo: we could extract the where-clause generation into a standalone method; that would make it nicely unit-testable

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -200,16 +202,40 @@ public RecordQueryResponse queryForRecords(
               + MAX_RECORDS
               + ", and offset must be positive.");
     }
+    // retrieve schema to use in validations
+    // TODO: why doesn't this include the primary key column?
+    Map<String, DataTypeMapping> schema =
+        recordDao.getExistingTableSchemaLessPrimaryKey(collectionId, recordType);
```
davidangb, line 208: I'm maintaining the logic from line 205 below, but I question the logic. It also means we can't sort on the primary key column.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -270,19 +271,38 @@ public List<Record> queryForRecords(
...
+      whereClause.append(quote(primaryKeyDao.getPrimaryKeyColumn(recordType, collectionId)));
+      whereClause.append(" in (:filterIds) ");
       sqlParams.addValue("filterIds", filterIds.get());
     }
 
+    // if this query has specified filter.filters, populate the where clause and bind params
+    Optional<List<FilterColumn>> filterColumns = searchFilter.flatMap(SearchFilter::filters);
+    if (filterColumns.isPresent() && !filterColumns.get().isEmpty()) {
+      if (whereClause.isEmpty()) {
+        whereClause.append(" where 1=1");
```
calypsomatic, line 290: This is just to turn it into a return all?

davidangb: ah, I bet I can find a way to do this more elegantly. This is a noop SQL clause to allow the subsequent code to append "and foo=?", "and bar=?" without causing SQL syntax errors.

davidangb: now refactored, with additional test coverage

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
davidangb: done!

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -891,6 +891,17 @@ components:
...
+    FilterColumn:
+      type: object
+      required:
+        - column
+        - find
+      properties:
+        column:
+          type: string
+        find:
+          type: string
```
mspector, line 903: is `find` basically the search term?

are we limiting ourselves by making this a string? I may be pushing into premature optimization territory, but just thinking out loud: if we do want to eventually support other kinds of search, will we regret not making this more general?

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
davidangb: Update: I'm fixing this in #869

service/src/main/resources/static/swagger/openapi-docs.yaml
davidangb: yes, this is the search term. "find" makes sense in the context of "find and replace" but admittedly isn't a great name outside of that context.

And good call on the datatype here. It definitely should be more flexible.

davidangb: I have completely overhauled this; marking as resolved.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -297,10 +294,57 @@ public List<Record> queryForRecords(
...
+    // init an empty list of clauses and empty map of bind params
+    List<String> clauses = new ArrayList<>();
+    MapSqlParameterSource sqlParams = new MapSqlParameterSource();
+
+    // extract potential record ids from the `filter` param
+    Optional<List<String>> filterIds = searchFilter.flatMap(SearchFilter::ids);
+
+    // if this query has specified filter.ids, populate the where clause and bind params
+    if (filterIds.isPresent()) {
+      // we know ids is non-empty due to the check in queryForRecords above
```
calypsomatic, line 314: I know it's just a comment but this is really throwing me off - don't we know it's non-empty because of `filterIds.isPresent` RIGHT there, and not due to anything in `queryRecords`?  Do we need this comment anymore?

davidangb: I updated the comment to be (hopefully) more explanatory

calypsomatic: Thanks!

calypsomatic (approved):
Good as is - since the UI isn't equipped to use it I think it's fine to roll it out incrementally

mspector (approved):

PR #869, opened 2024-07-25
AJ-1945: allow explicit sorting on primary key column

davidangb:
Before this PR:
* you could sort by the primary key column by omitting the `sortAttribute` field entirely from the API request payload. Omitting the field caused WDS to impose a default sort attribute, which was the primary key
* sorting by primary key worked in Terra UI, because Terra UI would omit the `sortAttribute` field
* you got an error if you included the `sortAttribute` field and its value was the primary key column.

After this PR:
* you can set the `sortAttribute` field to the primary key column
* omitting the `sortAttribute` field works as before
* Terra UI works as before

This PR is related to #868 - I found this issue while working on that other PR, and the change in this PR helps that other one.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
@@ -202,7 +202,7 @@ public class RecordOrchestratorService { // TODO give me a better name
     }
     if (searchRequest.getSortAttribute() != null
         && !recordDao
-            .getExistingTableSchemaLessPrimaryKey(collectionId, recordType)
+            .getExistingTableSchema(collectionId, recordType)
             .containsKey(searchRequest.getSortAttribute())) {
       throw new MissingObjectException("Requested sort attribute");
     }
--- a/service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
@@ -33,6 +33,7 @@ import org.databiosphere.workspacedataservice.shared.model.RecordRequest;
 import org.databiosphere.workspacedataservice.shared.model.RecordResponse;
 import org.databiosphere.workspacedataservice.shared.model.RecordType;
 import org.databiosphere.workspacedataservice.shared.model.SearchRequest;
+import org.databiosphere.workspacedataservice.shared.model.SortDirection;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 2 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -202,7 +202,7 @@ public RecordQueryResponse queryForRecords(
     }
     if (searchRequest.getSortAttribute() != null
         && !recordDao
-            .getExistingTableSchemaLessPrimaryKey(collectionId, recordType)
+            .getExistingTableSchema(collectionId, recordType)
```
davidangb, line 205: this is the only runtime change. We were explicitly omitting the primary key column from the list of allowed sort columns. I don't know why.

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
```diff
@@ -116,6 +117,94 @@ void queryForRecords() {
...
+    RecordQueryResponse resp =
+        recordOrchestratorService.queryForRecords(
+            COLLECTION_UUID, TEST_TYPE, VERSION, searchRequest);
+
+    assertEquals(3, resp.totalRecords());
+
+    // extract actual ids, in order, from the response
+    List<String> actualIds = resp.records().stream().map(RecordResponse::recordId).toList();
+    assertEquals(List.of("two", "three", "one"), actualIds); // descending alpha sort on pk
+  }
```
davidangb, line 206: this test fails without the runtime changes in `RecordOrchestratorService`

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
calypsomatic: I'm mildly worried there was a good reason for it, but if all the tests pass and we can't see any issues... 🤷

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
```diff
@@ -116,6 +117,94 @@ void queryForRecords() {
...
+
+  @Test
+  void sortPrimaryKeyExplicit() {
+    // insert records out of any order to ensure native db order doesn't give false positives
```
calypsomatic, line 188: It looks like you are inserting them in the order you're expecting them though

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
```diff
@@ -116,6 +117,94 @@ void queryForRecords() {
...
+
+    // extract actual ids, in order, from the response
+    List<String> actualIds = resp.records().stream().map(RecordResponse::recordId).toList();
+    assertEquals(List.of("three", "two", "one"), actualIds);
+  }
+
+  @Test
+  void sortPrimaryKeyImplicit() {
+    // insert records out of any order to ensure native db order doesn't give false positives
+    testCreateRecord("two", TEST_KEY, "value2");
```
calypsomatic, line 167: Here too - they're inserted in the order expected

calypsomatic (approved):

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceTest.java
davidangb: ahhhh great catch! I only had them out-of-order for the values, not the pks. Fixed across the board.

mspector (approved):

PR #871, opened 2024-07-26
AJ-1660: standardized, controller-layer auth via PermissionService

davidangb:
# Description

This is a relatively intrusive refactor. Summary of this PR:
* introduce a new `PermissionService` to centralize and standardize read/write permission checking and error throwing in case of permission failures.
* move permission enforcement out of the service layer and into the controller layer, utilizing `PermissionService`

The goals of this change are:
* add permission checks to all APIs that did not previously have them
* standardize the error messages and exception classes thrown by permission errors
* simplify the service layer
* make the service layer easier to test, because it will no longer require mocking Sam (except in corner cases where we do something with Sam above and beyond checking user permissions)
* allow deleting any service-layer permissions tests

# Review Notes

at 2000+ lines of code changed, I confess this is a big PR. When reviewing, I think you can break it down into a few categories to try to ease the review:

### PermissionService
PermissionService, PermissionServiceTest, and AllControllersPermissionsTest. This is the core implementation and test coverage to support it. AllControllersPermissionsTest is a bit of a new test pattern in that it tests multiple controllers at once; it provides an overview of how permissions are set up across all APIs.

### src/main/.../controller/*
Auth has moved to the controller layer. Controllers now invoke PermissionService directly.

### src/main/.../service/*
Auth has moved to the controller layer. Service classes now do NOT need any auth, and thus I've removed code.

### src/test/**/*
Except for PermissionServiceTest and AllControllersPermissionsTest, which you've already reviewed, all test changes involve removing obsolete Sam mocks, removing obsolete tests, and simplifying.

# Out of Scope

I'm saving this work for follow-on PRs, to _try_ to keep this current PR small:
* renaming RecordOrchestratorService.validateAndPermissions()
* assessing all test usages of the `mock-sam` profile, and removing more mocks where possible
* simplifying SamAuthorizationDao and SamDao

I have reviewed the Sonar warnings and do not believe the remaining warnings are worth fixing.


service/src/main/java/org/databiosphere/workspacedataservice/service/PermissionService.java
```diff
@@ -0,0 +1,140 @@
...
+    throw new AuthenticationMaskableException(objectType);
+  }
+
+  /**
+   * Check read permission on the single-tenant workspace id set in the environment.
+   *
+   * @deprecated Use {@link #requireReadPermission(WorkspaceId)} or {@link
+   *     #requireReadPermission(CollectionId)} instead.
+   */
+  @Deprecated(since = "v0.14.0")
```
calypsomatic, line 101: What's going on with the deprecated methods?  Are they in use somewhere until we do some more refactoring?

davidangb: yeah - once the instance APIs and cloning APIs are retired, we can retire these methods. Those APIs do not accept a workspaceId and therefore need to use defaults.

calypsomatic (approved):

mspector (approved):

PR #879, opened 2024-08-07
AJ-1238: filter by column enhancements

davidangb:
Followup to #868 

* report more detailed capabilities so the UI can distinguish individual query features
* filter-by-column is now case-insensitive
* support filtering on array_of_string, number, and array_of_number columns

Still todo:
1. Support boolean, file, relation, date, etc columns
2. It is exact-match only, not substring match or natural language search.
4. It works on a single column only; you can't search for e.g. "column1:foo AND column2:bar".
5. Users cannot search for null
6. The `totalRecords` value in the response of the `queryRecords` API may be considered incorrect. It returns the count of records for the specified record type _ignoring_ any filtering. So, if your filter only finds 1 record out of 100 samples, `totalRecords` will still return 100.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
@@ -51,7 +51,6 @@ import org.databiosphere.workspacedataservice.service.model.exception.ConflictEx
 import org.databiosphere.workspacedataservice.service.model.exception.InvalidRelationException;
 import org.databiosphere.workspacedataservice.service.model.exception.MissingObjectException;
 import org.databiosphere.workspacedataservice.service.model.exception.SerializationException;
-import org.databiosphere.workspacedataservice.service.model.exception.ValidationException;
 import org.databiosphere.workspacedataservice.shared.model.AttributeComparator;
 import org.databiosphere.workspacedataservice.shared.model.Record;
 import org.databiosphere.workspacedataservice.shared.model.RecordAttributes;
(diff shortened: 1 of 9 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -320,18 +319,7 @@ static WhereClause generateQueryWhereClause(
...
-          throw new ValidationException(
-              "Column specified in query does not exist in this record type");
-        }
-        var datatype = schema.get(column);
-        if (!DataTypeMapping.STRING.equals(datatype)) {
-          throw new ValidationException("Column specified in query must be a string type");
-        }
-      }
-
+      WhereClausePart queryPart = new QueryParser(schema).parse(filterQuery.get());
```
davidangb, line 322: validation has moved inside `QueryParser`

service/src/test/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorServiceFilterQueryTest.java
davidangb, line 1: I've rewritten this test pretty thoroughly. Before, it was tied tightly to STRING columns and needed a revamp to support the other datatypes.

service/src/main/java/org/databiosphere/workspacedataservice/search/QueryParser.java
```diff
@@ -33,29 +41,67 @@ public WhereClausePart parse(String query) {
...
-      clauses.add(quote(column) + " = :" + paramName);
-      values.put(paramName, value);
-      columns.add(column);
 
-      return new WhereClausePart(clauses, values, columns);
+      // based on the datatype of the column, build relevant SQL
+      switch (datatype) {
+        case STRING:
+          // LOWER("mycolumn") = 'mysearchterm'
+          clauses.add("LOWER(" + quote(column) + ") = :" + paramName);
```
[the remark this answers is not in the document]
davidangb: the search term is lowercased on the next line:
```
values.put(paramName, value.toLowerCase());
```

what this line does is generate a SQL statement with a bind parameter: `LOWER("mycolumn") = : filterquery0`. The next line populates the parameter map: `values.put("filterquery0", "mylowercasedvalue");`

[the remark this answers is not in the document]
calypsomatic: Gotcha

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/search/QueryParser.java
```diff
@@ -33,29 +41,67 @@ public WhereClausePart parse(String query) {
...
       var paramName = "filterquery0";
-      clauses.add(quote(column) + " = :" + paramName);
-      values.put(paramName, value);
-      columns.add(column);
 
-      return new WhereClausePart(clauses, values, columns);
+      // based on the datatype of the column, build relevant SQL
+      switch (datatype) {
+        case STRING:
+          // LOWER("mycolumn") = 'mysearchterm'
```
mspector, line 66: is this comment here as a visualization aid for what the query will look like, since the `clauses.add ...` statement below is hard to parse?

mspector (approved):

PR #881, opened 2024-08-09
AJ-1682: Remove `enforce-collections-match-workspace-id` test overrides

davidangb:
See https://broadworkbench.atlassian.net/browse/AJ-1682. The A/C for that ticket is to remove `enforce-collections-match-workspace-id` entirely. I don't think that's the right thing to do; I think it's good and fine to leave that config flag in place.

That config flag is only used, via `TenancyProperties`, in two places:
1. When saving a new collection, [here](https://github.com/DataBiosphere/terra-workspace-data-service/blob/76d9479887c5f10b670e4aeb39afdfb238f99f36/service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java#L87-L91).
2. When retrieving the workspaceId for a collection, [here](https://github.com/DataBiosphere/terra-workspace-data-service/blob/76d9479887c5f10b670e4aeb39afdfb238f99f36/service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java#L390-L402).

Those seem like minimally-invasive usages, and useful as long as we think we may want to support single-tenant deployments.

I _have_ removed all but one override of that config flag from tests. We still have one override in `CollectionControllerMockMvcTest`; I've added a code comment to that class to explain.

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
@@ -384,9 +384,8 @@ public class CollectionService {
       }
     }
 
-    // safety check: if we found a workspace id in the db, it indicates we are in a data-plane
-    // single-tenant WDS. Verify the workspace matches the $WORKSPACE_ID env var.
-    // we must remove this check in a future multi-tenant WDS.
+    // safety check: if we are running as a single-tenant WDS, verify the workspace matches the
+    // $WORKSPACE_ID env var.
     if (tenancyProperties.getEnforceCollectionsMatchWorkspaceId()
         && rowWorkspaceId != null
         && !rowWorkspaceId.equals(workspaceId)) {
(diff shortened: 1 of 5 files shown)


davidangb:
> I'm wondering... in single-tenant mode, is it actually useful to continue to enforce workspaceId == collectionId?

@mspector this property enforces a subtly different constraint. It checks that the workspaceId associated with any collection is equal to the `$WORKSPACE_ID` environment variable specified at WDS startup. In other words, it enforces that collections - regardless of their collectionId - belong to the single-tenant workspace; it disallows creating collections in any workspace outside the single-tenant one.

There's no enforcement that `workspaceId == collectionId`. However, there may be lingering _assumptions_ hidden in the code somewhere; that's what we're actively trying to rid ourselves of!

kevinmarete (approved):

mspector:
Ah, I see, thanks for clarifying!

PR #882, opened 2024-08-09
AJ-1951: API proposal for initWorkspaceV1

davidangb:
This is a signature definition - and minimal Java scaffolding - for the new `initWorkspaceV1` API. There's no implementation yet; calling this API will return a 501 error.

Hoping to get any feedback or suggestions on the API signature before diving in to implementation stuff!

## Current State
1. User creates or clones a workspace
2. Rawls calls Leo to deploy a WDS app into a single workspace in the data plane
3. On startup, WDS initializes that workspace, either with empty data tables or by cloning

## Proposed State
1. User creates or clones a workspace
2. Rawls calls this new init-workspace API in cWDS, passing a workspaceId _(this could eventually move to a Workspace.onCreate event, but for the time being will use a direct API call)_
3. cWDS initializes that workspace, either with empty data tables or by cloning

Screenshots below, or view it live via the [swagger viewer](https://petstore.swagger.io/?url=https://raw.githubusercontent.com/DataBiosphere/terra-workspace-data-service/da_AJ-1951_initWorkspace/service/src/main/resources/static/swagger/openapi-docs.yaml#/Workspace/initWorkspaceV1).

Note this API returns the same response payload as the import API; that is, it will return an async job id. This is on the assumption that init/cloning operations will be async.

Screenshot 09-08-2024 at 17 04
Screenshot 09-08-2024 at 17 03


service/build.gradle
```diff
@@ -224,7 +224,7 @@ sonar {
         property "sonar.organization", "broad-databiosphere"
         property "sonar.host.url", "https://sonarcloud.io"
         // exclude generated files from Sonar scans
-        property "sonar.exclusions", "service/src/main/java/org/databiosphere/workspacedataservice/generated/**/*"
+        property "sonar.exclusions", "**/generated/*"
```
davidangb, line 227: bonus fix: at some point this config stopped working, and Sonar started analyzing the `openapi-generator`-generated files again. I tweaked the config so Sonar ignores these once more.

mspector (approved):
This seems good to me. I wonder, though, why not make this part of the `collections` API? are there likely to be future operations that callers will want to make on `workspaces` by way of WDS?

kevinmarete (approved):
No concerns!

dvoet (approved):
The shape of the API LGTM. I am curious about access control checks on the workspaces as well as accessing the resulting job.

davidangb:
> The shape of the API LGTM. I am curious about access control checks on the workspaces as well as accessing the resulting job.

@dvoet thank you for the review! And I will happily include you as a reviewer on the implementation PR, once that is ready :)

PR #887, opened 2024-08-12
AJ-1956: Remove Prometheus RSocket Spring from WDS

kevinmarete:
**Description:**
This PR removes the prometheus-rsocket-spring library from WDS. The library was added for data plane observability but was never fully implemented or used. This approach is no longer needed as we're shifting to a centralized observability service.

**Key Changes:**
- Removed the prometheus-rsocket-spring library and related configurations.
- Ensured that metrics at /prometheus remained unaffected.

This change helps simplify our observability setup by removing unused components.


davidangb (approved):
lgtm, thank you!

mspector (approved):

PR #888, opened 2024-08-12
AJ-1951: init-workspace API implementation (minus cloning)

davidangb:
Implements the new init-workspace API (minus cloning). See #882 for some explanation of that API.

When called, the API will create the workspace's default collection, assuming that collection does not already exist.

Executing the API requires the `write` action on the workspace; this is implemented in this PR in `WorkspaceController`. Retrieving the job details after the fact requires the `read` action on that workspace; that has already been in place for a while.

The API is synchronous, but returns a job reference just like our async import API. This allows the API to move to an asynchronous implementation later without requiring a new API version. It is out of scope to make the API synchronous now; that would entail a nontrivial refactor to allow async jobs to be associated with a workspace instead of with a collection.

I've added TODO comments in places where we'll implement cloning as part of AJ-1952.


service/build.gradle
```diff
@@ -135,6 +135,7 @@ dependencies {
     testImplementation 'au.com.dius.pact.provider:spring:4.6.13'
     testImplementation 'au.com.dius.pact.consumer:junit5:4.6.13'
     testImplementation 'org.junit-pioneer:junit-pioneer:2.2.0'
+    testImplementation 'org.awaitility:awaitility'
```
davidangb, line 138: provides test utilities for working with async operations. No version specified because it is managed by `io.spring.dependency-management`

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
```diff
@@ -300,7 +301,11 @@ private StatusEnum getStatus(ResultSet rs) throws SQLException {
 
     private JobInput getJobInput(JobTypeEnum jobType, ResultSet rs) throws SQLException {
       Class<? extends JobInput> targetClass =
-          DATA_IMPORT.equals(jobType) ? ImportJobInput.class : JobInput.class;
+          switch (jobType) {
+            case DATA_IMPORT -> ImportJobInput.class;
+            case WORKSPACE_INIT -> WorkspaceInitJobInput.class;
+            default -> JobInput.class;
+          };
```
davidangb, line 308: the `WORKSPACE_INIT` type is new in this PR

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
```diff
@@ -275,6 +291,15 @@ private void handleDbException(DbActionExecutionException dbActionExecutionExcep
...
 
+  public boolean exists(WorkspaceId workspaceId, CollectionId collectionId) {
+    // if this deployment allows virtual collections, there is nothing to check
+    if (tenancyProperties.getAllowVirtualCollections()) {
+      return true;
+    }
+    // else, check if this collection has a row in the collections table
+    return collectionRepository.find(workspaceId, collectionId).isPresent();
+  }
+
```
davidangb, line 302: this `exists()` method is new in this PR, and used for idempotency of the init-workspace API

service/src/main/java/org/databiosphere/workspacedataservice/service/WorkspaceService.java
```diff
@@ -0,0 +1,109 @@
...
+      }
+    }
+
+    /* initialize the default collection:
+       - name and description of "default"
+       - collectionId equal to workspaceId
+    */
+    CollectionRequestServerModel collectionRequestServerModel =
+        new CollectionRequestServerModel(NAME_DEFAULT, NAME_DEFAULT);
+    collectionService.save(workspaceId, collectionId, collectionRequestServerModel);
```
davidangb, line 88: there does exist a (pretty fast) race condition between the call to `collectionService.exists()` above and this call to `collectionService.save()`. Some other thread could create the collection between these two calls. In that case, the call to `collectionService.save()` will throw a `ConflictException`, which will bubble up. I think it's a fair tradeoff to not prevent this race condition and instead simply expose the `ConflictException` in those rare cases. Thoughts?

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
mspector: this `PostgresJobDao` is new to me - it sounds like it's related to async jobs, but I know we're keeping this implementation synchronous. No feedback, just looking for an intro to what this is doing.

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
```diff
@@ -83,16 +110,31 @@ void setWorkspaceId(@Nullable @SingleTenant WorkspaceId workspaceId) {
...
+    return save(workspaceId, collectionId, collectionRequestServerModel);
+  }
+
+  /**
+   * Insert a new collection, specifying the collection id. This should only be called internally;
+   * users should not be granted the ability to specify the collection id.
+   *
+   * @param workspaceId the workspace to contain this collection
+   * @param collectionId the id of the collection to create
+   * @param collectionRequestServerModel the collection definition
```
mspector, line 124: can you help me understand how the name of this class maps onto its meaning?

is it: "the model of a collection, as understood by the server, as delivered by a collection request"?

service/src/main/java/org/databiosphere/workspacedataservice/dao/PostgresJobDao.java
mspector: actually, I think I see it now (but keeping my previous comment in case I'm wrong): this *is* for async jobs, we're just writing this API with all the async machinery around it so that it can be changed easily later.

mspector (approved):

service/src/main/java/org/databiosphere/workspacedataservice/service/WorkspaceService.java
davidangb: this comment is now outdated, I have reduced probability of the race condition even more by wrapping calls inside a transaction. Resolving this conversation.

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
davidangb: Update: `exists()` is no longer used at runtime. However, since I wrote it and it seems useful, I'm leaving it in case we need it in the future. Alternately … should I just delete it?

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
davidangb: There exist in WDS two categories of model classes:
* classes we hand-write, which typically live under o.d.w.service.model or o.d.w.shared.model<sup>1</sup>
* classes automatically generated by `openapi-generator` from our OpenAPI spec, which live under o.d.w.generated

We wanted a standard way - beyond the package location - to differentiate between the two, and to avoid name conflicts. So, we tweaked the `openapi-generator` code generation to automatically append "ServerModel" to all model classes it generates. That setting is [here](https://github.com/DataBiosphere/terra-workspace-data-service/blob/8db552af537ce6a94cf4c11cb6525e6dc92ce1b2/service/build.gradle#L167).

So, `CollectionRequestServerModel` should be read as "a CollectionRequest, using generated code".


1. there's no good distinction between o.d.w.service.model or o.d.w.shared.model, it's messy

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
mspector: thinking out loud... preserving unused methods seems like a double edged sword. the code contains information: namely that any collection with a given workspaceId and collectionId can be said to exist if WDS allows "virtual collections" (N.B. I don't actually know what that means).

If we think that fact will remain true forever / for a while, then I think keeping it is a good idea. if that fact might change, I would lean towards deleting it

davidangb: ty - I'm going to delete it - it's simple enough to add back in later.

davidangb: removed in d59c19a6663cb3bd8bdf2576adcd3b8adaedb723; resolving this thread

marctalbott (approved):
LGTM although I have limited WDS knowledge

dvoet (approved):

PR #889, opened 2024-08-13
[AJ-1955] add wds.searchRequest.includesFilter observation

mspector:
Reminder:

#### Releasing WDS ####

### DIFF
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
@@ -211,7 +211,9 @@ public class RecordOrchestratorService { // TODO give me a better name
 
     Observation observation =
         Observation.start("wds.RecordOrchestratorService.queryForRecords", observations)
-            .lowCardinalityKeyValue("wds.searchRequest.includesFilter", String.valueOf(searchRequest.getFilter().isPresent()));
+            .lowCardinalityKeyValue(
+                "wds.searchRequest.includesFilter",
+                String.valueOf(searchRequest.getFilter().isPresent()));
 
     LOGGER.info("queryForEntities: {}", recordType.getName());
     List<Record> records =
--- a/service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
+++ b/service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
@@ -4,10 +4,13 @@ import static org.databiosphere.workspacedataservice.service.RecordUtils.validat
 import static org.databiosphere.workspacedataservice.service.model.ReservedNames.RECORD_ID;
 
 import bio.terra.common.db.ReadTransaction;
+import io.micrometer.common.KeyValue;
+import io.micrometer.common.KeyValues;
 import io.micrometer.observation.Observation;
 import io.micrometer.observation.ObservationRegistry;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -204,6 +209,12 @@ public RecordQueryResponse queryForRecords(
       return new RecordQueryResponse(searchRequest, Collections.emptyList(), totalRecords);
     }
 
+    Observation observation =
+        Observation.start("wds.RecordOrchestratorService.queryForRecords", observations)
```
davidangb, line 213: nit: take `RecordOrchestratorService` out of the observation name; no need to include specifics of our Java class structure

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -204,6 +209,12 @@ public RecordQueryResponse queryForRecords(
...
+            .lowCardinalityKeyValue(
+                "wds.searchRequest.includesFilter",
+                String.valueOf(searchRequest.getFilter().isPresent()));
```
davidangb, line 216: could you add a key-value that is specific to filter-by-column? Simply checking `searchRequest.getFilter().isPresent()` only tells you that _some_ filter exists. It could be a filter-by-ids or filter-by-column (and I wonder if it even returns true if the user submits a payload of `{ "filter": {} }`. To be specific to filter-by-column you'll need to check for `filter.query`, e.g. a payload of `{ "filter": { "query": "col:term" } }`.

Stretch goal is to have separate KVs for filter-by-ids and filter-by-column; that would cover all possible filters!

davidangb (commented):
Can you write a unit test for this? Check out `RestClientRetryTest`'s `outboundHttpMeasurementOnSuccess` and `outboundHttpMeasurementOnFailure` and especially note the `WithTestObservationRegistry` annotation.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
mspector: is `wds.queryForRecords` good? I note that other observations don't seem to include the method name, but I couldn't discern a naming convention

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -209,11 +210,17 @@ public RecordQueryResponse queryForRecords(
...
+
+    if (searchRequest.getFilter().isPresent()) {
+      SearchFilter filter = searchRequest.getFilter().get();
+      if (filter.ids().isPresent() && !filter.ids().get().isEmpty()) {
+        observation.lowCardinalityKeyValue("queryForRecords.includesFilterById", "true");
+      }
+      if (filter.query().isPresent()) {
+        observation.lowCardinalityKeyValue("queryForRecords.includesFilterByQuery", "true");
+      }
+    }
```
mspector, line 223: @davidangb , this feels clunky to me. given these optionals, is this how you would do it (nested if statements)? feedback is welcome!

davidangb: I don't think you can be much less verbose, but how about breaking the logic out into a helper method? So, here you'd just have a one-liner:
```java
observation.lowCardinalityKeyValues(generateLowCardinalityKeyValues(searchRequest));
```

and then elsewhere in the class you can have the helper:
```java
  private KeyValues generateLowCardinalityKeyValues(SearchRequest searchRequest) {
    List<KeyValue> kvs = new ArrayList<>();

    searchRequest
        .getFilter()
        .ifPresent(
            filter -> {
              // check for non-empty ids
              if (filter.ids().isPresent() && !filter.ids().get().isEmpty()) {
                kvs.add(KeyValue.of("queryForRecords.includesFilterById", "true"));
              }
              // check for non-blank query
              if (filter.query().isPresent() && !filter.query().get().isBlank()) {
                kvs.add(KeyValue.of("queryForRecords.includesFilterByQuery", "true"));
              }
            });

    return KeyValues.of(kvs);
  }
```

I have not tested this code to see if it works, hopefully it illustrates what I'm thinking though

mspector: makes sense, I'll do that -- thanks!

davidangb (approved):
nice, thank you!

cahrens (approved):

PR #891, opened 2024-08-15
AJ-1954: more datatypes for filter-by-column

davidangb:
This PR completes datatype support for filter-by-column; it adds support for `relation`, `array_of_relation`, `json`, and `array_of_json`.

It also prevents errors if the column is of datatype `null` or `empty_array`; we expect these datatypes to be rare. Filters on these columns will always return an empty result set, but at least they won't throw errors.


service/src/main/java/org/databiosphere/workspacedataservice/search/QueryParser.java
```diff
@@ -65,65 +68,94 @@ public WhereClausePart parse(String query) {
...
-          break;
-
-          /* TODO AJ-1954: support
-              NULL, EMPTY_ARRAY, (uncommon)
-              RELATION, ARRAY_OF_RELATION,
-              JSON, ARRAY_OF_JSON
+        }
+        case NULL, EMPTY_ARRAY ->
+        // results in a `where false` clause. These columns are nonsensical to filter on, as
+        // they cannot contain anything. Would it be better to throw InvalidQueryException?
```
mspector, line 123: does NULL or EMPTY_ARRAY refer to the data type of the whole column? I can imagine why this would be rare, but trying to understand in what kind of situation this might happen... what is the purpose of a column that can only contain null?

mspector (approved):

service/src/main/java/org/databiosphere/workspacedataservice/search/QueryParser.java
davidangb: I don't believe we have use cases for these columns, when thinking from the user journey side. However, when thinking from the data side … if somebody uploaded a TSV and had a column of entirely `[]`, we create the column as `EMPTY_ARRAY`. If that TSV has a column with entirely empty values, we create the column as `NULL`. We have no way to infer if they meant for it to be a numeric column but forgot the values, or if they intended it to be a string column but forgot the values … so, using `NULL` and `EMPTY_ARRAY` here allows us to be precise about what happened, and the user can later modify the datatype of the column to what they intended.

calypsomatic (approved):

PR #896, opened 2024-08-19
AJ-1957: multi-cloud RecordSinkFactory, SnapshotSupportFactory

davidangb:
https://broadworkbench.atlassian.net/browse/AJ-1957

High-level goal is to allow the centralized cWDS to handle imports for _both_ Entity Service workspaces (by sending JSON to Rawls) and WDS workspaces (by writing to cWDS' own database).

Previously, we branched control-plane vs. data-plane by having two separate implementations of `RecordSinkFactory`, each annotated with either `@ControlPlane` or `@DataPlane`. In this PR, I delete those two implementations and replace them with a new `MultiCloudRecordSinkFactcory`. This new implementation asks `WorkspaceService` about the workspace in question (see #892 ), then returns the appropriate `RecordSink` instance.

Similarly, we had control-plane vs. data-plane implementations of `SnapshotSupportFactory`. Those are deleted and replaced with a new `MultiCloudSnapshotSupportFactory`. As part of that change, I made the `RawlsClient` available in both the data plane and the control plane to satisfy bean dependencies; the rawls client is not actually invoked in the data plane, though.

IMPORTANT: this PR does not complete AJ-1957. We'll need the same kind of branching for `ProtectedDataSupport` and at least; there may be more. I am hoping to get reviews/opinions on this approach before tackling those other beans.


service/src/main/java/org/databiosphere/workspacedataservice/recordsink/MultiCloudRecordSinkFactory.java
```diff
@@ -0,0 +1,79 @@
...
+  @Override
+  public RecordSink buildRecordSink(CollectionId collectionId) {
+    WorkspaceId workspaceId = collectionService.getWorkspaceId(collectionId);
+    return switch (workspaceService.getDataTableType(workspaceId)) {
+      case RAWLS -> throw new NotImplementedException(
+          "MultiCloudRecordSinkFactory does not support building a RawlsRecordSink from a CollectionId");
+      case WDS -> new WdsRecordSink(recordService, recordDao, dataTypeInferer, collectionId);
+    };
+  }
+}
```
davidangb, line 79: these two `buildRecordSink` methods are really close, but have just enough differences that I don't see a good way to consolidate them further

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/MultiCloudSnapshotSupportFactory.java
```diff
@@ -0,0 +1,44 @@
...
+import org.databiosphere.workspacedataservice.activitylog.ActivityLogger;
+import org.databiosphere.workspacedataservice.rawls.RawlsClient;
+import org.databiosphere.workspacedataservice.service.WorkspaceService;
+import org.databiosphere.workspacedataservice.shared.model.WorkspaceId;
+import org.databiosphere.workspacedataservice.workspacemanager.WorkspaceManagerDao;
+import org.springframework.stereotype.Component;
+
+/** Factory to generate the appropriate {@link SnapshotSupport} for any given workspace. */
+@Component
+public class MultiCloudSnapshotSupportFactory implements SnapshotSupportFactory {
```
mspector, line 12: do we need to distinguish MultiCloudSnapshotSupportFactory from SnapshotSupportFactory?

mspector (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/MultiCloudSnapshotSupportFactory.java
davidangb: do we still need both the `SnapshotSupportFactory` interface and the `MultiCloudSnapshotSupportFactory` implementing class? We don't _need_ them - but think that developing to interfaces, especially in Spring, helps to avoid problems. It's much easier to swap in a different implementation class later if everything uses the interface!

PR #898, opened 2024-08-19
AJ-1957: use local db cache for workspace type

davidangb:
Reviewer: is this premature optimization?

This PR adds a `sys_wds.workspace` table to the WDS database, with a column to describe whether the workspace should use Rawls Entity Service data tables or WDS data tables.

The `RawlsDataTableTypeInspector` first looks in the db to see if a row for a given workspace exists. If a row exists, we use its value. If no row exists, we make a REST call out to Rawls, inspect the result, and then write the result to the db for subsequent usage.

In the short term, this will reduce the number of REST calls cWDS needs to make to Rawls. In the long run, it will allow us to easily toggle a workspace from Rawls-powered tables to WDS-powered tables at will. This will be necessary eventually if we want to incrementally migrate existing GCP workspaces to WDS. But admittedly, we don't really need it now.


service/src/main/java/org/databiosphere/workspacedataservice/workspace/WorkspaceRecord.java
```diff
@@ -0,0 +1,83 @@
...
+    this.workspaceId = workspaceId;
+    this.dataTableType = dataTableType;
+    this.newFlag = false;
+  }
+
+  public WorkspaceRecord(
+      WorkspaceId workspaceId, WorkspaceDataTableType dataTableType, boolean newFlag) {
+    this.workspaceId = workspaceId;
+    this.dataTableType = dataTableType;
+    this.newFlag = newFlag;
```
calypsomatic, line 31: What is `newFlag`?

calypsomatic (approved):

PR #900, opened 2024-08-22
suggested fixes

davidangb:
@kevinmarete A PR into your branch, not into main. `GeneratedClientTests` is passing for me after these changes. See comments inline for explanation. Feel free to merge into your branch at will.

### DIFF
--- a/service/src/test/java/org/databiosphere/workspacedataservice/GeneratedClientTests.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/GeneratedClientTests.java
@@ -11,12 +11,14 @@ import java.util.List;
 import java.util.UUID;
 import java.util.stream.Stream;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.RandomStringUtils;
 import org.databiosphere.workspacedata.api.CollectionApi;
 import org.databiosphere.workspacedata.api.GeneralWdsInformationApi;
 import org.databiosphere.workspacedata.api.RecordsApi;
 import org.databiosphere.workspacedata.api.SchemaApi;
 import org.databiosphere.workspacedata.client.ApiClient;
 import org.databiosphere.workspacedata.client.ApiException;
+import org.databiosphere.workspacedata.model.Collection;
 import org.databiosphere.workspacedata.model.CollectionRequest;
 import org.databiosphere.workspacedata.model.RecordAttributes;
 import org.databiosphere.workspacedata.model.RecordQueryResponse;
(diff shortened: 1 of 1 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/GeneratedClientTests.java
```diff
@@ -51,22 +53,22 @@ class GeneratedClientTests extends TestBase {
   @LocalServerPort int port;
 
   @Autowired TwdsProperties twdsProperties;
-  private final UUID workspaceId = twdsProperties.workspaceId().id();
-  private final UUID collectionId = UUID.randomUUID();
+  private UUID collectionId;
```
davidangb, line 56: collection creation no longer allows the end user to specify the collection id. Therefore, setting the collection id had no effect. In this PR, the `createNewCollection` helper method now returns the actual id it created, and the `init` method saves that value back into this variable before each test.

service/src/test/java/org/databiosphere/workspacedataservice/GeneratedClientTests.java
```diff
@@ -51,22 +53,22 @@ class GeneratedClientTests extends TestBase {
   @LocalServerPort int port;
 
   @Autowired TwdsProperties twdsProperties;
-  private final UUID workspaceId = twdsProperties.workspaceId().id();
```
davidangb, line 54: autowired beans aren't available at the phase of object instantiation where `private final` variables are initialized. So, I've eliminated the class-level `workspaceId` variable. Wherever that variable was used, I've replaced it with a reference to `twdsProperties.workspaceId().id()`. By the time those references are needed, the autowired bean is populated.

service/src/test/java/org/databiosphere/workspacedataservice/GeneratedClientTests.java
```diff
@@ -51,22 +53,22 @@ class GeneratedClientTests extends TestBase {
...
   @BeforeEach
   void init() throws ApiException {
     apiClient = new ApiClient();
     apiClient.setBasePath("http://localhost:" + port);
     CollectionRequest collectionRequest = new CollectionRequest();
-    collectionRequest.setName("default");
-    createNewCollection(collectionRequest, workspaceId);
+    collectionRequest.setName(RandomStringUtils.randomAlphabetic(16));
+    collectionRequest.setDescription("description");
+    collectionId = createNewCollection(collectionRequest, twdsProperties.workspaceId().id());
```
davidangb, line 66: * description is required, so I added it.
* using a random name is NOT required if everything works perfectly. However, if any single test cleanup fails - including if the test is aborted midway through and you then try to run the test again - you'll hit a conflict with the old code because a collection named "default" already exists. Using a random name eliminates that corner case.
* modified this code to save the `createNewCollection` result back to the class-level `collectionId` variable

kevinmarete (approved):

PR #908, opened 2024-08-26
[AJ-1277] WIP apis-v1

mspector:
Reminder:

#### Releasing WDS ####


service/src/main/resources/static/swagger/apis-v1.yaml
```diff
@@ -407,4 +434,31 @@ components:
...
+    IncludedRecordIDs:
+      type: object
+      description: included record IDs array
+      properties:
+        include: # Is this a clear enough term for this property?
+          type: array
+          description: an array of record IDs that should be deleted
+          items:
+            type: string
+            # should this be UUID format? format: uuid
```
calypsomatic, line 453: No, the record ids would be text not UUIDs

service/src/main/resources/static/swagger/apis-v1.yaml
```diff
@@ -249,6 +251,31 @@ paths:
...
+  /records/v1/{workspaceId}/delete:
+    post:
+      summary: Submit a request for record deletion in this workspace.
+      operationId: deleteRecordV1
+      description: |
+        deletes records by ID. TODO finish this description!
+      tags:
+        - Record
+      parameters:
+        - $ref: '#/components/parameters/workspaceIdPathParam'
```
calypsomatic, line 266: I'm still wrapping my head around the collection/workspace structure so I genuinely don't know - should this include a collection Id in addition to/instead of workspaceId?

mspector: yeah good question, maybe @davidangb can chime in

davidangb: yes, this would need to be a collection id. A workspace can contain multiple collections; collections contain tables.

However … thinking into the future … what if we used `{workspaceId}/{collectionName}` instead of `{collectionId}`? That is, we start identifying collections by their _name_ and workspaceId, instead of by an id?

Advantages of using `{workspaceId}/{collectionName}`:
* requiring a workspaceId in the url will make permission implementation easier at the code level
* using a collection name makes things human-readable, especially when/if we support reference data; urls will be something like `{workspaceId}/default` or `{workspaceId}/reference`

Disadvantages:
* requires callers to know the workspaceId; hopefully they do already
* a UUID and a name is longer than just a UUID; urls get longer
* url signature is inconsistent with older APIs

service/src/main/resources/static/swagger/apis-v1.yaml
```diff
@@ -249,6 +251,31 @@ paths:
...
+      requestBody:
+        description: Record deletion specification
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/DeleteRecords'
+      responses:
+        202:
+          description: Deletion request accepted
```
davidangb, line 276: Do you envision this as a synchronous or asynchronous API? 202 Accepted suggests it would be asynchronous. If async, this should probably return a jobId so users can check up on what happened.

If synchronous - consider returning 200 OK and some payload indicating what happened?

service/src/main/resources/static/swagger/apis-v1.yaml
calypsomatic: Is it overkill to have two APIs, one with `collectionId` and one with `collectionName` that do the same thing?  Then we can use the name for the cases where it's available or makes sense, and the id otherwise.

davidangb: I don't think that's overkill from the user perspective - it's very user-friendly, giving flexibility to callers. It's more code for us to maintain but I don't think it's that much more.

For identifying a collection, are you suggesting `{collectionId}` or `{workspaceId}/{collectionId}`? The latter is a bit redundant but makes the user's intent very explicit.

service/src/main/resources/static/swagger/apis-v1.yaml
mspector: thanks, I intended synchronous, do you think that's fine?

davidangb: I think synchronous is fine - just consider changing the response status code and define a payload

service/src/main/resources/static/swagger/apis-v1.yaml
```diff
@@ -249,6 +251,63 @@ paths:
...
+              $ref: '#/components/schemas/DeleteRecordsRequest'
+      responses:
+        200:
+          description: Deletion successful
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/DeleteRecordsResponse'
+
+  /records/v1/{collectionId}/delete:
```
davidangb, line 284: Hm - while in theory I like having both APIs, in practice I worry about conflicts.

The two urls:
```
/records/v1/{workspaceId}/{collectionName}/delete
/records/v1/{collectionId}/delete
```

are very close from a Spring route-parsing perspective and might potentially lead to confusion. I know it was my suggestion to identify a collection via workspaceId+collectionName but given the url confusion, I am backtracking. I'd suggest for this PR to only implement the `/records/v1/{collectionId}/delete` route, and maybe we'll move to supporting workspaceId+collectionName  for a new set of v2 routes

mspector:
The sonarcloud issues are related to dynamically generated SQL, but I used the same patterns I saw elsewhere in `RecordDao`... can anyone spot differences I'm not seeing?

calypsomatic:
> The sonarcloud issues are related to dynamically generated SQL, but I used the same patterns I saw elsewhere in `RecordDao`... can anyone spot differences I'm not seeing?

For the other methods you can see we [add](https://github.com/DataBiosphere/terra-workspace-data-service/blob/3d2134de78249eb64d6ef42456afd5ba7a4caa2d/service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java#L219) `@SuppressWarnings("squid:S2077")` which is supposed to stop those sonar warnings - otherwise you can just tell sonar to shut up on these.

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -240,4 +248,57 @@ public ResponseEntity<BatchResponse> streamingWrite(
...
+  @Override
+  public ResponseEntity<DeleteRecordsResponseServerModel> deleteRecordsByCollectionV1(
+      UUID collectionId,
+      String recordType,
+      DeleteRecordsRequestServerModel deleteRecordsRequestServerModel) {
+    return deleteRecords(
+        collectionId, RecordType.valueOf(recordType), deleteRecordsRequestServerModel);
+  }
+
+  private ResponseEntity<DeleteRecordsResponseServerModel> deleteRecords(
```
calypsomatic, line 261: This method seems a bit business-logic heavy for the controller - is it possible to move it to the service layer?  Not sure if that means it'd be `RecordService` or `RecordOrchestratorService` (again, can't remember the difference between them)

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -240,4 +248,57 @@ public ResponseEntity<BatchResponse> streamingWrite(
...
+          "delete_all cannot be set to true if record_ids or excluded_record_ids are nonempty");
+    }
+
+    int deletionCount;
+    if (hasRecordIds) {
+      deletionCount =
+          recordOrchestratorService.deleteRecords(
+              collectionId, recordType, deleteRecordsRequestServerModel.getRecordIds());
+    } else if (hasExcludedRecordIds) {
+      System.out.println("hasExcludedRecordIds: " + hasExcludedRecordIds);
```
calypsomatic, line 290: Should this be a logging statement or just removed?

mspector: oh, thank you, just removed

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
mspector: I don't mind moving it - do you know if I can throw `ResponseStatusException`s in the service layer? am I overthinking that...?

calypsomatic: I think so?

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -240,4 +243,13 @@ public ResponseEntity<BatchResponse> streamingWrite(
...
   }
+
+  @Override
+  public ResponseEntity<DeleteRecordsResponseServerModel> deleteRecordsByCollectionV1(
+      UUID collectionId,
+      String recordType,
+      DeleteRecordsRequestServerModel deleteRecordsRequestServerModel) {
+    return recordOrchestratorService.deleteRecords(
+        collectionId, RecordType.valueOf(recordType), deleteRecordsRequestServerModel);
+  }
```
davidangb, line 254: hmmmm … I could go either way on this. Should we make a brand new `RecordV1Controller` (or whatever name, naming is hard) for this API, or keep these methods here? I kinda like having the separation of the v0.2 APIs and the v1 API … but it would be another controller to keep track of.

service/src/main/java/org/databiosphere/workspacedataservice/dao/CollectionRepository.java
```diff
@@ -29,4 +29,9 @@ public interface CollectionRepository extends CrudRepository<WdsCollection, Coll
   @Query(
       "SELECT * FROM sys_wds.collection WHERE workspace_id = :workspaceId and id = :collectionId")
   Optional<WdsCollection> find(WorkspaceId workspaceId, CollectionId collectionId);
+
+  // custom method to get a collection by workspaceId and collectionName
+  @Query(
+      "SELECT * FROM sys_wds.collection WHERE workspace_id = :workspaceId and name = :collectionName")
+  Optional<WdsCollection> find(WorkspaceId workspaceId, String collectionName);
```
davidangb, line 36: with the removal of the workspaceId+collectionName API, this and its parent methods are now unused. Remove them?

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -550,6 +551,53 @@ public boolean deleteSingleRecord(UUID collectionId, RecordType recordType, Stri
...
+              + quote(recordTypePrimaryKey)
+              + " in (:recordIds)",
+          new MapSqlParameterSource(RECORD_IDS_PARAM, recordIds));
+    } catch (DataIntegrityViolationException e) {
+      if (e.getRootCause() instanceof SQLException sqlEx) {
+        checkForTableRelation(sqlEx);
+      }
+      throw e;
+    }
+  }
```
davidangb, line 571: Sonar complains about this because of the table name and primary key column name - those are added to the SQL statement through plain string concatenation and Sonar notices the string concatenation. We are aware that Sonar will continue to ping us on this; bind parameters are not supported for table names/column names.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -550,6 +551,53 @@ public boolean deleteSingleRecord(UUID collectionId, RecordType recordType, Stri
     }
   }
 
+  @SuppressWarnings("squid:S2077")
+  public int deleteRecords(UUID collectionId, RecordType recordType, List<String> recordIds) {
```
davidangb, line 555: consider using `CollectionId collectionId` instead of `UUID collectionId`. With lots of UUIDs being thrown around (workspace ids, collection ids, job ids), we're trying to avoid using raw UUIDs to avoid any confusion

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -550,6 +551,53 @@ public boolean deleteSingleRecord(UUID collectionId, RecordType recordType, Stri
...
+            new MapSqlParameterSource(RECORD_IDS_PARAM, excludedRecordIds));
+      }
+
+    } catch (DataIntegrityViolationException e) {
+      if (e.getRootCause() instanceof SQLException sqlEx) {
+        checkForTableRelation(sqlEx);
+      }
+      throw e;
+    }
+  }
```
davidangb, line 599: I could envision `deleteRecords` and `deleteAllRecords` being combined, or both delegating to a common private method; the only differences are the `where` clauses. Or, there could be three methods, for "delete inclusions", "delete all but exclusions", and "delete all".

Your call; I wouldn't hold the PR for this and one could argue that keeping the methods separate is more readable.

service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
```diff
@@ -255,6 +275,19 @@ public CollectionServerModel get(WorkspaceId workspaceId, CollectionId collectio
...
+   *
+   * @param workspaceId the workspace containing the collection to be retrieved
+   * @param collectionName name of the collection to be retrieved
+   * @return the collection
+   */
+  public CollectionServerModel get(WorkspaceId workspaceId, String collectionName) {
+    Optional<CollectionServerModel> found = find(workspaceId, collectionName);
+    return found.orElseThrow(() -> new MissingObjectException(COLLECTION));
+  }
+
```
davidangb, line 290: these are both unused now

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -284,6 +289,49 @@ public void deleteRecordType(UUID collectionId, String version, RecordType recor
         user -> user.deleted().table().ofQuantity(1).withRecordType(recordType));
   }
 
+  public ResponseEntity<DeleteRecordsResponseServerModel> deleteRecords(
```
davidangb, line 292: Somewhere in the call stack, we need a `@WriteTransaction` annotation. You could add it right here, or move some of this implementation to `RecordService` and annotate it there

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -284,6 +289,49 @@ public void deleteRecordType(UUID collectionId, String version, RecordType recor
...
+      throw new ResponseStatusException(
+          HttpStatus.BAD_REQUEST, "record_ids and excluded_record_ids are mutually exclusive");
+    }
+
+    Boolean deleteAll = deleteRecordsRequestServerModel.getDeleteAll();
+    if (Boolean.TRUE.equals(deleteAll) && (hasRecordIds || hasExcludedRecordIds)) {
+      throw new ResponseStatusException(
+          HttpStatus.BAD_REQUEST,
+          "delete_all cannot be set to true if record_ids or excluded_record_ids are nonempty");
+    }
```
davidangb, line 313: what's the difference between deleting all records of a given record type vs. deleting the record type (see `deleteRecordType` API)?

I realize that deleting all records of a given record type will keep the record type definition - it will result in a data table with zero records; deleting the record type itself will remove the table entirely.

For the purposes of keeping complexity down - do we need both features? If it isn't too complex to keep both, keeping both seems good.

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -284,6 +289,49 @@ public void deleteRecordType(UUID collectionId, String version, RecordType recor
...
+    Boolean hasExcludedRecordIds =
+        !deleteRecordsRequestServerModel.getExcludedRecordIds().isEmpty();
+
+    if (hasRecordIds && hasExcludedRecordIds) {
+      throw new ResponseStatusException(
+          HttpStatus.BAD_REQUEST, "record_ids and excluded_record_ids are mutually exclusive");
+    }
+
+    Boolean deleteAll = deleteRecordsRequestServerModel.getDeleteAll();
+    if (Boolean.TRUE.equals(deleteAll) && (hasRecordIds || hasExcludedRecordIds)) {
```
davidangb, line 309: throughout this method, consider using `boolean` instead of `Boolean`. That should allow you to simplify here:
```suggestion
    boolean deleteAll = deleteRecordsRequestServerModel.getDeleteAll();
    if (deleteAll && (hasRecordIds || hasExcludedRecordIds)) {
```

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -284,6 +289,49 @@ public void deleteRecordType(UUID collectionId, String version, RecordType recor
...
+          recordDao.deleteRecords(
+              collectionId, recordType, deleteRecordsRequestServerModel.getRecordIds());
+    } else if (hasExcludedRecordIds) {
+      deletionCount =
+          recordDao.deleteAllRecords(
+              collectionId, recordType, deleteRecordsRequestServerModel.getExcludedRecordIds());
+    } else if (deleteAll) {
+      deletionCount = recordDao.deleteAllRecords(collectionId, recordType, Collections.emptyList());
+    } else {
+      deletionCount = 0;
```
davidangb, line 327: if the user specified deleteAll=false, also specified no records to include or exclude - should it be a 400 Bad Request?

service/src/main/resources/static/swagger/apis-v1.yaml
```diff
@@ -302,6 +335,13 @@ components:
       schema:
         type: string
         format: uuid
+    recordTypePathParam:
+      name: recordType
+      in: path
+      description: Record type
+      required: true
+      schema:
+        type: string
```
davidangb, line 344: consider adding a `pattern` here for initial validation.

Here's an example of using `pattern`:
https://github.com/DataBiosphere/terra-workspace-data-service/blob/3367d0cc877d12fcf9179b7bef0a1a81fd82ec90/service/src/main/resources/static/swagger/apis-v1.yaml#L322

And, record type inputs are validated here:
https://github.com/DataBiosphere/terra-workspace-data-service/blob/3367d0cc877d12fcf9179b7bef0a1a81fd82ec90/service/src/main/java/org/databiosphere/workspacedataservice/shared/model/RecordType.java#L17
if you follow that down you can find the regex it uses!

service/src/main/resources/static/swagger/apis-v1.yaml
```diff
@@ -407,4 +447,30 @@ components:
...
+          description: an array of record IDs that should NOT be deleted.
+          items:
+            type: string
+        delete_all:
+          type: boolean
+          description: if true, deletes all records
+          default: false
+    DeleteRecordsResponse:
+      type: object
+      description: a required deletion specification
```
davidangb, line 469: um … what does this description mean?

service/src/main/resources/static/swagger/apis-v1.yaml
```diff
@@ -407,4 +447,30 @@ components:
       required:
         - sourceWorkspaceId
       additionalProperties: false
+    DeleteRecordsRequest:
+      type: object
+      description: request body to delete records by ID
```
davidangb, line 452: I'd love to see some more detail here on how `record_ids`, `excluded_record_ids`, and `delete_all` interact - for instance, `record_ids` and `excluded_record_ids` are mutually exclusive.

When first looking at this options, I assumed that `delete_all=true` is required for `excluded_record_ids` to take effect. I think that's not true, though?

davidangb (commented):
comments inline, thank you!

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
mspector: I'm ok with a new V1 controller. The hard part (to me) about keeping track of them is remembering what they're for. Calling it `RecordV1Controller` makes it obvious what it is

service/src/main/java/org/databiosphere/workspacedataservice/dao/CollectionRepository.java
mspector: whoops - yes, good catch, I'll find other unused methods from that removal

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
mspector: yes, I think that's a good idea

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
mspector: I did keep them separate for the sake of readability... but let me think a little more about your suggestions

service/src/main/resources/static/swagger/apis-v1.yaml
mspector: oops, that should have been a "TODO: write something that makes sense here"

service/src/main/resources/static/swagger/apis-v1.yaml
mspector: > When first looking at this options, I assumed that delete_all=true is required for excluded_record_ids to take effect. I think that's not true, though?

Correct, that's not what I intended, though I'm interested to know what reviewers think is likely to be the most intuitive behavior

mspector: actually, I think you're right that `delete_all=true` should be required for `excluded_record_ids` to take effect

service/src/main/java/org/databiosphere/workspacedataservice/service/RecordOrchestratorService.java
```diff
@@ -284,6 +290,56 @@ public void deleteRecordType(UUID collectionId, String version, RecordType recor
...
+              collectionId, recordType, deleteRecordsRequestServerModel.getExcludedRecordIds());
+    } else if (deleteAll) {
+      deletionCount = recordDao.deleteAllRecords(collectionId, recordType, Collections.emptyList());
+    } else {
+      throw new ResponseStatusException(
+          HttpStatus.BAD_REQUEST,
+          "No records were specified for deletion. Set delete_all=true (and optionally use excluded_record_ids) or use record_ids to delete records.");
+    }
+
+    response.setCount(deletionCount);
```
davidangb, line 338: ah, I should have thought of this earlier - could you please add activity logging here? We need to log on any data change. See the activity logger example in `deleteSingleRecord()` - I think you can mostly copy that, just change the `.withId()` to `.ofQuantity()`

service/src/main/resources/static/swagger/apis-v1.yaml
```diff
@@ -302,6 +335,14 @@ components:
...
         type: string
         format: uuid
+    recordTypePathParam:
+      name: recordType
+      in: path
+      description: Record type
+      required: true
+      schema:
+        type: string
+        pattern: '[a-zA-Z0-9-_]{1,128}'
```
davidangb, line 345: nit: I don't think we enforce a 128-char limit here (though, uh … we probably should enforce a limit, I think Postgres will fail at > 63 chars)

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/RecordType.java
```diff
@@ -42,6 +42,9 @@ public void validate() {
     if (name.startsWith(RESERVED_NAME_PREFIX)) {
       throw new InvalidNameException(InvalidNameException.NameType.RECORD_TYPE);
     }
+    if (name.length() > 63) {
+      throw new InvalidNameException(InvalidNameException.NameType.RECORD_TYPE);
```
mspector, line 46: @davidangb , your comment didn't explicitly ask for this, but let me know if this seems right

service/src/main/resources/static/swagger/apis-v1.yaml
```diff
@@ -249,6 +251,37 @@ paths:
...
+  ##############################
+  # Records APIs
+  ##############################
+  /records/v1/{collectionId}/{recordType}/delete:
+    post:
+      summary: Bulk delete records
+      operationId: deleteRecords
+      description: |
+        Deletes records by record ID, using collection ID.
+        To delete by workspace ID and collection name instead, see deleteRecordsByWorkspaceV1
```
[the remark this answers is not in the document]
mspector: whoops, no we do not. thank you

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/shared/model/RecordType.java
davidangb: If you're gonna add this, we should also explain the limitation in the error message of `InvalidNameException`. I'm ok updating the error message in a follow-on PR, or removing the length validation from this PR and making a follow-on PR for validation, unit tests, and error messaging … probably the latter is the easiest, since we should check how Postgres actually behaves in practice

davidangb (approved):
Giving a proactive :+1:, I suggest removing the entity type length validation and saving for follow-on work

PR #913, opened 2024-08-27
AJ-1967: Enable APIs in cWDS [dev]

kevinmarete:
## Description
This PR enables the necessary APIs in the cWDS environment to prepare for upcoming pen-testing and security review processes. Currently, cWDS has most APIs disabled, with only the import APIs enabled. To allow workspaces to start storing their data tables in cWDS, this PR enables all necessary APIs but only in the `dev` environment.

### Key Changes:

- Enabled all required APIs in the cWDS `dev` environment.
- Ensured that the single-tenant APIs are not enabled in any environment:
- Updated Swagger UI in cWDS to display all enabled APIs in the `dev` environment.

Screenshot 2024-08-26 at 9 56 57 PM


calypsomatic (approved):

.github/workflows/release-python-client.yml
```diff
@@ -90,7 +90,7 @@ jobs:
           export SPRING_PROFILES_ACTIVE=data-plane,local
           ./gradlew --build-cache build -x test
           # Allow http imports from localhost to test with files from resources served by the mock API server
-          ./gradlew bootRun --args='--twds.data-import.require-validation=false' &
+          ./gradlew bootRun --args='--twds.data-import.require-validation=false --controlPlanePreview=on' &
```
davidangb, line 93: is this change necessary?

davidangb (commented):
instead of annotating controllers with:
```java
@DataPlane
@ControlPlane
@ConditionalOnProperty(name = "controlPlanePreview", havingValue = "on")
```

could you just remove any mention of dataplane/controlplane for those controllers and simply have this?
```java
@ConditionalOnProperty(name = "controlPlanePreview", havingValue = "on")
```

Then, if you add a `controlPlanePreview: "on"` property into `application-data-plane.yml`, I believe the end result will be that these APIs will be enabled in:
* data plane, all environments, via `application-data-plane.yml`
* control plane, dev env only, via `application-dev.yml`

This should also eliminate the need to specify `"mock-cwds-preview"` in all/most of the tests. It looks like all/most of the tests where you added that are running under the data plane profile.

.github/workflows/release-python-client.yml
kevinmarete: Removed

davidangb (approved):
I did not test it myself, code looks good, thank you!

PR #916, opened 2024-08-28
AJ-1971: getWorkspaceId respects Rawls vs. WDS workspace types

davidangb:
## Description

This PR modifies the logic in `CollectionService.getWorkspaceId()` such that virtual collections are only allowed in Rawls-powered workspaces. The logic is:

If workspace is WDS, return the value from sys_wds.collection or throw an error if no row exists. The virtual collections setting is ignored.

If workspace is Rawls, return the value from sys_wds.collection if one exists. Else, return workspaceId == collectionId if virtual collections are allowed. Else, throw an error.

`CollectionService.getWorkspaceId()` continues to respect the `twds.tenancy.enforce-collections-match-workspace-id` config setting. If that setting is true, the method enforces that any workspaceIds returned must match the `$WORKSPACE_ID` env var. This logic hasn't changed; it exists to enforce single-tenancy.

## Questions and Notes for Reviewers

1. `CollectionService.getWorkspaceId()` is a superset of `CollectionService.validateCollection()`. I am not touching `validateCollection()` in this PR, even though there's overlap - in yet one more follow-on PR I will tackle that.
2. I've taken two approaches to fixing unit tests - one "real" approach of inserting into the db and another with mocks. Which do you prefer? See comments inline for details.


service/src/main/java/org/databiosphere/workspacedataservice/service/CollectionService.java
```diff
@@ -25,38 +26,43 @@
...
-  private final ActivityLogger activityLogger;
-  private final TenancyProperties tenancyProperties;
-  private final TwdsProperties twdsProperties;
 
+  private final ActivityLogger activityLogger;
   private final CollectionRepository collectionRepository;
+  private final DataTableTypeInspector dataTableTypeInspector;
   private final NamedParameterJdbcTemplate namedTemplate;
+  private final TenancyProperties tenancyProperties;
+  private final TwdsProperties twdsProperties;
```
davidangb, line 47: I alphabetized these here and in the constructor

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobControlPlaneE2ETest.java
```diff
@@ -144,6 +146,11 @@ class PfbQuartzJobControlPlaneE2ETest {
   void setup() {
     collectionId = UUID.randomUUID();
 
+    // do not create a collection; we want to test virtual collections here.
+    // mock out the DataTableTypeInspector to show that this workspace is Rawls-powered.
+    when(dataTableTypeInspector.getWorkspaceDataTableType(WorkspaceId.of(collectionId)))
+        .thenReturn(WorkspaceDataTableType.RAWLS);
+
```
davidangb, line 153: This approach uses a mock for DataTableTypeInspector to indicate the workspace is Rawls-powered and thus supports virtual collections. Do you like this mocked approach better, as compared to `RawlsJsonQuartzJobControlPlaneE2ETest` just below?

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/rawlsjson/RawlsJsonQuartzJobTest.java
```diff
@@ -108,7 +118,11 @@ private void assertPubSubMessage(Map<String, String> expectedMessage) {
   private void assertSingleBlobWritten(String expectedBlobName) {
     List<Blob> blobsWritten = listBlobs(storage);
     assertThat(blobsWritten).hasSize(1);
-    assertThat(blobsWritten.get(0).getName()).isEqualTo(expectedBlobName);
+    assertThat(blobsWritten.get(0).getName())
+        .withFailMessage(
+            "Expected blob name is incorrect; expected %s got %s",
+            expectedBlobName, blobsWritten.get(0).getName())
+        .isEqualTo(expectedBlobName);
```
davidangb, line 125: no functional change here; I added a custom failure message to help me when I was debugging failures in this test

service/src/test/java/org/databiosphere/workspacedataservice/service/CollectionServiceNoWorkspaceTest.java
davidangb, line 1: this test was obsolete

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/rawlsjson/RawlsJsonQuartzJobControlPlaneE2ETest.java
```diff
@@ -67,15 +76,23 @@ class RawlsJsonQuartzJobControlPlaneE2ETest {
...
 
   @BeforeEach
   void setup() {
+    TestUtils.cleanAllWorkspaces(namedTemplate);
     collectionId = UUID.randomUUID();
+    workspaceId = WorkspaceId.of(collectionId);
+    // do not create a collection; we want to test virtual collections here.
+    // however, create a record of the corresponding workspace indicating the workspace is
+    // Rawls-powered, so we know this workspace is valid for virtual collections.
+    workspaceRepository.save(new WorkspaceRecord(workspaceId, WorkspaceDataTableType.RAWLS, true));
```
davidangb, line 89: This approach uses no mocks; instead it relies on setting up the appropriate data in the db to indicate the workspace is Rawls-powered and thus supports virtual collections. Do you like this mocked approach better, as compared to `PfbQuartzJobControlPlaneE2ETest` above?

service/src/test/java/org/databiosphere/workspacedataservice/service/CollectionServiceGetWorkspaceIdTest.java
```diff
@@ -1,6 +1,7 @@
 package org.databiosphere.workspacedataservice.service;
```
davidangb, line 1: I added a few more test cases to this file

service/src/test/java/org/databiosphere/workspacedataservice/service/CollectionServiceNoWorkspaceTest.java
```diff
@@ -1,50 +0,0 @@
-package org.databiosphere.workspacedataservice.service;
```
davidangb, line 1: this test was obsolete

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobControlPlaneE2ETest.java
calypsomatic: I think the mocked approach is good, assuming we've tested the code that checks whether the workspaceId is in the DB somewhere else.

kevinmarete (approved):
Looks good

service/src/test/java/org/databiosphere/workspacedataservice/dataimport/pfb/PfbQuartzJobControlPlaneE2ETest.java
davidangb: cool, thank you - moved tests over to the mock approach in c45fda46a99d0f2c378130ffeb9f2799e440ecf5

calypsomatic (approved):

PR #917, opened 2024-08-28
AJ-1971: remove CollectionService.validateCollection()

davidangb:
This PR removes `CollectionService.validateCollection()`. Why?

All controllers<sup>1</sup> include a call to `PermissionService.requireWritePermission(CollectionId collectionId)` or `PermissionService.requireReadPermission(CollectionId collectionId)`. An example is [here](https://github.com/DataBiosphere/terra-workspace-data-service/blob/8ee3fa18488bc01a7046a31870c401cba9f3e897/service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java#L87).

The implementation of `PermissionService.requireWritePermission(CollectionId collectionId)` and `PermissionService.requireReadPermission(CollectionId collectionId)` includes a call to `CollectionService.getWorkspaceId()`. You can see that implementation [here](https://github.com/DataBiosphere/terra-workspace-data-service/blob/8ee3fa18488bc01a7046a31870c401cba9f3e897/service/src/main/java/org/databiosphere/workspacedataservice/service/PermissionService.java#L50).

As of #916, `CollectionService.getWorkspaceId()` includes fully multi-tenant validation that the collection in question exists.

Thus, all collection-specific controllers include, as part of their permission check, validation that the target collection exists. Since that is true, we can _remove validation elsewhere_. That's what this PR does - it removes `CollectionService.validateCollection()`.

On the test side, we have extensive test coverage of `PermissionService` in `PermissionServiceTest`, and extensive test coverage of `CollectionService.getWorkspaceId()` in `CollectionServiceGetWorkspaceIdTest`. With that coverage in place, I don't feel the need to add any more test coverage for controllers.

---

1. to be precise, all controllers which operate against a collection. Some, like `initWorkspaceV1` and `listCollectionsV1`, operate at a workspace level; other controllers are fully public, like `capabilities`.


service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceDataPlaneTest.java
```diff
@@ -1,126 +0,0 @@
-package org.databiosphere.workspacedataservice.service;
```
davidangb, line 1: I consolidated `JobServiceDataPlaneTest ` and `JobServiceControlPlaneTest ` into `JobServiceTest`

service/src/test/java/org/databiosphere/workspacedataservice/service/JobServiceTest.java
```diff
@@ -16,7 +16,6 @@
 import java.util.List;
```
davidangb, line 16: I consolidated `JobServiceDataPlaneTest ` and `JobServiceControlPlaneTest ` into `JobServiceTest`

kevinmarete (approved):

calypsomatic (approved):

PR #928, opened 2024-09-05
AJ-1983 Enhance initWorkspaceV1 API to Insert Workspaces Proactively

kevinmarete:
#### Summary

This PR updates the `initWorkspaceV1` API to automatically insert a workspace into the `sys_wds.workspace` table when a new workspace is created. This change improves efficiency by avoiding the need to query Rawls for missing workspaces.

#### Key Changes:

- Automatic Insertion: The initWorkspaceV1 API now inserts a workspace into the table when it's created, whether it's new or a clone.
- Idempotency: The API remains idempotent, preventing conflicts if called multiple times for the same workspace.


service/src/main/java/org/databiosphere/workspacedataservice/service/WorkspaceService.java
```diff
@@ -121,4 +130,20 @@ private GenericJobServerModel initClone(WorkspaceInitJobInput jobInput) {
...
+  public void initSystemWorkspace(WorkspaceId workspaceId) {
+    Optional<WorkspaceRecord> maybeWorkspaceRecord = workspaceRepository.findById(workspaceId);
+    if (maybeWorkspaceRecord.isEmpty()) {
+      WorkspaceDataTableType dataTableType =
+          dataTableTypeInspector.getWorkspaceDataTableType(workspaceId);
+      WorkspaceRecord newWorkspaceRecord =
+          new WorkspaceRecord(workspaceId, dataTableType, /* newFlag= */ true);
+      workspaceRepository.save(newWorkspaceRecord);
+    }
+  }
```
davidangb, line 148: instead of querying for a row's existence, then conditionally inserting a row, how about writing a new SQL query to `insert into sys_wds.workspace … on conflict do nothing`, and always executing it?

The query-then-conditional-insert pattern is vulnerable to race conditions; if the API is called more than once at just the right time, multiple threads can conflict:
* thread 1: findById() returns false
* thread 2: findById() returns false
* thread 1: save() inserts the row
* thread 2: save() encounters an error because the row already exists

Alternate approaches:
* don't write a new insert method; always call `save()` but trap for/handle errors on primary key conflict; see `CollectionService.handleDbException`
* wrap all of this in a `@WriteTransaction`, which means [moving it to another class](https://medium.com/javarevisited/spring-transactional-mistakes-everyone-did-31418e5a6d6b)

service/src/main/java/org/databiosphere/workspacedataservice/service/WorkspaceService.java
```diff
@@ -121,4 +130,20 @@ private GenericJobServerModel initClone(WorkspaceInitJobInput jobInput) {
...
+  /**
+   * Initialize the workspace record in the database (sys_wds.workspace) if it does not exist.
+   *
+   * @param workspaceId the workspace to initialize
+   */
+  public void initSystemWorkspace(WorkspaceId workspaceId) {
+    Optional<WorkspaceRecord> maybeWorkspaceRecord = workspaceRepository.findById(workspaceId);
+    if (maybeWorkspaceRecord.isEmpty()) {
+      WorkspaceDataTableType dataTableType =
+          dataTableTypeInspector.getWorkspaceDataTableType(workspaceId);
```
davidangb, line 143: Hm - do we actually need to call `getWorkspaceDataTableType`? Perhaps we could assume that any workspace for which this API is called will be a WDS-powered workspace.

service/src/main/java/org/databiosphere/workspacedataservice/workspace/RawlsDataTableTypeInspector.java
```diff
@@ -37,17 +37,9 @@ public WorkspaceDataTableType getWorkspaceDataTableType(WorkspaceId workspaceId)
...
-    // persist the Rawls result to the local db for future use
-    WorkspaceRecord newWorkspaceRecord =
-        new WorkspaceRecord(workspaceId, dataTableType, /* newFlag= */ true);
-    workspaceRepository.save(newWorkspaceRecord);
-
-    return dataTableType;
+    return switch (details.workspace().workspaceType()) {
+      case MC -> WorkspaceDataTableType.WDS;
+      case RAWLS -> WorkspaceDataTableType.RAWLS;
+    };
```
davidangb, line 43: Why is this logic being deleted? I think this logic is still needed to handle pre-existing workspaces (especially Rawls-powered workspaces). Without this, those workspaces will never get a row in `sys_wds.workspace` and thus we'll always be making a REST call to Rawls.

If you put this logic back, the corresponding unit test will need to revert, too

service/src/test/java/org/databiosphere/workspacedataservice/service/WorkspaceServiceControlPlaneTest.java
```diff
@@ -52,4 +63,64 @@ void dataTableType(
...
+    CollectionService mockCollectionService = Mockito.mock(CollectionService.class);
+
+    WorkspaceId workspaceId = WorkspaceId.of(UUID.randomUUID());
+    RawlsWorkspaceDetails.RawlsWorkspace rawlsWorkspace =
+        new RawlsWorkspaceDetails.RawlsWorkspace("bucketName", workspaceType);
+    RawlsWorkspaceDetails rawlsWorkspaceDetails =
+        new RawlsWorkspaceDetails(rawlsWorkspace, List.of());
+    WorkspaceRecord workspaceRecord = new WorkspaceRecord(workspaceId, dataTableType, true);
+
+    when(rawlsClient.getWorkspaceDetails(workspaceId.id())).thenReturn(rawlsWorkspaceDetails);
```
davidangb, line 83: I think this `getWorkspaceDetails` is unnecessary, which also makes the `RawlsWorkspace` and `RawlsWorkspaceDetails` models above unnecessary

service/src/test/java/org/databiosphere/workspacedataservice/service/WorkspaceServiceControlPlaneTest.java
```diff
@@ -52,4 +63,64 @@ void dataTableType(
...
+    when(workspaceRepository.findById(workspaceId)).thenReturn(Optional.empty());
+    when(dataTableTypeInspector.getWorkspaceDataTableType(workspaceId)).thenReturn(dataTableType);
+
+    WorkspaceService mockWorkspaceService =
+        new WorkspaceService(
+            mockJobDao, mockCollectionService, dataTableTypeInspector, workspaceRepository);
+
+    mockWorkspaceService.initSystemWorkspace(workspaceId);
+
+    verify(workspaceRepository).save(workspaceRecord); // verify that workspaceRecord is saved
```
davidangb, line 93: for a slightly stricter test, use `verify(workspaceRepository, times(1))`

service/src/main/java/org/databiosphere/workspacedataservice/workspace/RawlsDataTableTypeInspector.java
```diff
@@ -37,17 +37,9 @@ public WorkspaceDataTableType getWorkspaceDataTableType(WorkspaceId workspaceId)
...
     RawlsWorkspaceDetails details = rawlsClient.getWorkspaceDetails(workspaceId.id());
 
-    WorkspaceDataTableType dataTableType =
-        switch (details.workspace().workspaceType()) {
-          case MC -> WorkspaceDataTableType.WDS;
-          case RAWLS -> WorkspaceDataTableType.RAWLS;
-        };
-
-    // persist the Rawls result to the local db for future use
-    WorkspaceRecord newWorkspaceRecord =
```
calypsomatic, line 47: Are we sure we're never going to need to save the workspace when getting its type?  What about old workspaces that weren't initialized through cwds?

service/src/test/java/org/databiosphere/workspacedataservice/service/WorkspaceServiceControlPlaneTest.java
```diff
@@ -52,4 +63,64 @@ void dataTableType(
...
   }
+
+  @ParameterizedTest(name = "Initialize `{0}` system workspace for `{1}-backed` data tables")
+  @MethodSource("workspaceTypeArguments")
+  void testInitSystemWorkspace_WhenWorkspaceNotExists(
+      RawlsWorkspaceDetails.RawlsWorkspace.WorkspaceType workspaceType,
+      WorkspaceDataTableType dataTableType) {
+
+    JobDao mockJobDao = Mockito.mock(JobDao.class);
+    CollectionService mockCollectionService = Mockito.mock(CollectionService.class);
```
davidangb, line 74: no functional difference, but the code pattern here is slightly inconsistent. `workspaceRepository` and `dataTableTypeInspector` are declared as class-level variables using `@Mock`, but `mockJobDao` and `mockCollectionService` are declared as method-level variables in the new unit tests. Could you use the same approach for all of these?

davidangb (commented):
comments inline, please let me know if you have any questions!

service/src/main/java/org/databiosphere/workspacedataservice/service/WorkspaceService.java
kevinmarete: New SQL Query: https://github.com/DataBiosphere/terra-workspace-data-service/pull/928/commits/0f0a198bf1c560bded1a425aa51e287ebf643e6f

service/src/main/java/org/databiosphere/workspacedataservice/workspace/RawlsDataTableTypeInspector.java
kevinmarete: I have put it back: https://github.com/DataBiosphere/terra-workspace-data-service/pull/928/commits/9f324f12c6677dd8c56bffcda443c2468b19e420

service/src/main/java/org/databiosphere/workspacedataservice/service/WorkspaceService.java
kevinmarete: Yes this was unnecessary since all workspaces will be WDS powered. Updated in L69

service/src/main/java/org/databiosphere/workspacedataservice/workspace/RawlsDataTableTypeInspector.java
kevinmarete: I have put this back: https://github.com/DataBiosphere/terra-workspace-data-service/pull/928/commits/9f324f12c6677dd8c56bffcda443c2468b19e420

service/src/test/java/org/databiosphere/workspacedataservice/service/WorkspaceServiceControlPlaneTest.java
kevinmarete: Removed opted for the new test based on repository implementation:  https://github.com/DataBiosphere/terra-workspace-data-service/pull/928/commits/2a7eac9d5b4d6c202506310c6dcc535c485ae1ce

davidangb (approved):
looks good to me, thank you!

We should use this new `on conflict do nothing` insert method in `RawlsDataTableTypeInspector` too - I realize it has the same race-condition problem (and I wrote that one, my bad!). Would you mind making that change too? I realize it's out of the original scope for this ticket: https://github.com/DataBiosphere/terra-workspace-data-service/blob/689530c9b3f525bd37a5ff99b6cee5369edd3ac7/service/src/main/java/org/databiosphere/workspacedataservice/workspace/RawlsDataTableTypeInspector.java#L49

calypsomatic (approved):

PR #929, opened 2024-09-05
AJ-1981: continue moving tests to control-plane profile

davidangb:
Followup to #927.

This moves another 20 unit tests from the data-plane profile to the control-plane profile. There are still ~25 tests in the data-plane profile; this is incremental progress.

Some tests required relatively minor setup/teardown changes; some required no changes at all other than the profile switch.

The remaining tests will require the most attention, either because:
* they are actually testing data plane functionality and therefore we need to decide on keeping them or not
* they rely on controllers which are not yet enabled in the control plane

### DIFF
--- a/service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/ParquetDataTypesTest.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/ParquetDataTypesTest.java
@@ -25,23 +25,20 @@ import org.apache.hadoop.fs.Path;
 import org.apache.parquet.hadoop.ParquetReader;
 import org.apache.parquet.hadoop.util.HadoopInputFile;
 import org.apache.parquet.io.InputFile;
-import org.databiosphere.workspacedataservice.common.DataPlaneTestBase;
+import org.databiosphere.workspacedataservice.common.ControlPlaneTestBase;
 import org.databiosphere.workspacedataservice.recordsource.RecordSource;
 import org.databiosphere.workspacedataservice.service.model.TdrManifestImportTable;
 import org.databiosphere.workspacedataservice.shared.model.Record;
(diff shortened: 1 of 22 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/dataimport/tdr/ParquetDataTypesTest.java
```diff
@@ -25,23 +25,20 @@
...
-import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.core.io.Resource;
 
 @SpringBootTest
-class ParquetDataTypesTest extends DataPlaneTestBase {
-  private final Logger logger = LoggerFactory.getLogger(getClass());
+class ParquetDataTypesTest extends ControlPlaneTestBase {
```
davidangb, line 41: logger was unused

service/src/test/java/org/databiosphere/workspacedataservice/startup/CollectionInitializerBeanTest.java
```diff
@@ -64,6 +64,8 @@ class CollectionInitializerBeanTest extends DataPlaneTestBase {
   void setUp() throws InterruptedException {
     when(mockLock.tryLock(anyLong(), any())).thenReturn(true);
     when(registry.obtain(anyString())).thenReturn(mockLock);
+    // start fresh
+    TestUtils.cleanAllCollections(collectionService, namedTemplate);
```
davidangb, line 68: I can't find what changed, but this test started being flaky; giving it a fresh start seems to help

mspector (approved):

kevinmarete (approved):

calypsomatic (approved):

PR #931, opened 2024-09-10
AJ-1959 Expression evaluation in WDS

dvoet:
https://broadworkbench.atlassian.net/browse/AJ-1959


service/src/main/java/org/databiosphere/workspacedataservice/expressions/ExpressionService.java
```diff
@@ -0,0 +1,530 @@
...
+            // plus one to see if there are more records
+            pageSize + 1,
+            offset);
+    // the query results may have an extra record to see if there are more records
+    // if there are more records, remove the extra record
+    var hasNext =
+        resultsByQuery.values().stream().map(LinkedHashMap::size).anyMatch(s -> s > pageSize);
+    if (hasNext) {
+      removeExtraRecords(pageSize, resultsByQuery);
+    }
```
dvoet, line 138: I am curious what people thing about this `hasNext` implementation. It seems a little shifty because of the `removeExtraRecords` things and reliance on `LinkedHashMap`. OTOH, it avoids another db call and associated code to get a full count that either needs to happen on each page or in a different api call.

service/src/main/antlr/org/databiosphere/workspacedataservice/expressions/parser/antlr/TerraExpression.g4
```diff
@@ -0,0 +1,60 @@
+/*
+ * This is a modified version of the official JSON grammar, supporting the proposed extensions. All
+ * other TerraExpression* files are generated by antlr from this one. Antlr supports both JS and Java
+ * as target languages.
+ *
+ * The rule `root` is needed to not allow expressions with unbalanced quotes. So something like
+ * ""str" is invalid.
+ */
+
+grammar TerraExpression;
```
dvoet, line 10: copied from https://github.com/broadinstitute/rawls/blob/develop/core/src/main/antlr4/org/broadinstitute/dsde/rawls/expressions/parser/antlr/TerraExpression.g4

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -117,6 +126,39 @@ public RecordQueryResponse queryForRecords(
         instanceId, recordType, version, searchRequest);
   }
 
+  @PostMapping("/{instanceid}/records/{version}/{recordType}/{recordId}/evaluateExpressions")
+  public EvaluateExpressionsResponse evaluateExpressions(
```
dvoet, line 130: response will look like
```
{
  "evaluations": {
    "<expression name>": <expression value>
  }
}
```

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -117,6 +126,39 @@ public RecordQueryResponse queryForRecords(
...
+      @RequestBody EvaluateExpressionsRequest request) {
+    permissionService.requireReadPermission(CollectionId.of(instanceId));
+    return new EvaluateExpressionsResponse(
+        expressionService.evaluateExpressions(
+            instanceId, version, recordType, recordId, request.toMap()));
+  }
+
+  @PostMapping(
+      "/{instanceid}/records/{version}/{recordType}/{recordId}/evaluateExpressionsWithArray")
+  public EvaluateExpressionsWithArrayResponse evaluateExpressionsWithArray(
```
dvoet, line 144: response will look like
```
{
  "hasNext": boolean,
  "results":  [
    {"recordId": string, "evaluations": {"<expression name>": <expression value>, ... }
  ]
}

File: service/src/main/java/org/databiosphere/workspacedataservice/expressions/ExpressionService.java (hunk 1 above)
Comment (mspector) [reply service/src/main/java/org/databiosphere/workspacedataservice/expressions/ExpressionService.java:138]: I don't feel strongly about it, and so would like to see what other AJ people think, but it seems that getting a full count is arguably a more intuitive approach, and that might be worth something from a code readability point of view. Whether it's worth another db call / more code, I don't know.

**Review by mspector — APPROVED** [review]
There's a lot going on here -- it might be worth a walkthrough to go over the new ANTLR-related classes, or more commentary about how those new classes work for those that aren't already familiar with it from the Rawls implementation, but I wouldn't insist on it for my sake alone. I'll leave that up to you and the other members of AJ who may be more familiar.

File: service/src/main/java/org/databiosphere/workspacedataservice/expressions/ExpressionService.java [hunk 5]
```diff
@@ -0,0 +1,530 @@
...
+            recordDao.getRelationArrayCols(collectionId, priorRecordType).stream())
+        .filter(r -> r.relationColName().equals(relationName))
+        .findFirst()
+        .orElseThrow(
+            () ->
+                new ResponseStatusException(
+                    HttpStatus.BAD_REQUEST,
+                    "The relation %s in expression %s does not exist"
+                        .formatted(relationName, arrayRelationExpression)));
+  }
```
davidangb, line 226: this probably doesn't matter in practice, but making two SQL queries in `getRelationCols()` and `getRelationArrayCols()` which could return a few rows each, then filtering to just the one desired row … we could optimize this. But it's premature optimization.

service/src/main/java/org/databiosphere/workspacedataservice/expressions/ExpressionService.java
```diff
@@ -0,0 +1,530 @@
...
+   * records.
+   */
+  @VisibleForTesting
+  List<Relation> getArrayRelations(
+      UUID collectionId, RecordType arrayRecordType, String arrayRelationExpression) {
+    var arrayRelationAttributeLookups = extractRecordAttributeLookups(arrayRelationExpression);
+    if (arrayRelationAttributeLookups.size() != 1) {
+      throw new ResponseStatusException(
+          HttpStatus.BAD_REQUEST,
+          "Expected a single expression in %s".formatted(arrayRelationExpression));
```
davidangb, line 182: use `ExpressionParsingException` instead of `ResponseStatusException`? (there are a few instances of `ResponseStatusException` in this class)

service/src/main/java/org/databiosphere/workspacedataservice/expressions/ExpressionService.java
```diff
@@ -0,0 +1,530 @@
...
+                                  .toList();
+                          if (queryInfoAndResult.getKey().isArray()) {
+                            return Map.entry(
+                                lookup.lookupText(),
+                                objectMapper.createArrayNode().addAll(attributeValues));
+                          } else if (attributeValues.size() > 1) {
+                            throw new ResponseStatusException(
+                                HttpStatus.BAD_REQUEST,
+                                "Expected a single value for attribute %s but got %s"
+                                    .formatted(lookup.attribute(), attributeValues));
```
davidangb, line 375: consider returning the count of values found instead of the values themselves:
```suggestion
                                "Expected a single value for attribute %s but got %s"
                                    .formatted(lookup.attribute(), attributeValues.size()));
```

service/src/main/java/org/databiosphere/workspacedataservice/expressions/ExpressionService.java
```diff
@@ -0,0 +1,530 @@
...
+        : record.getAttributeValue(lookup.attribute());
+  }
+
+  /** Get the name of the id attribute for the record type in the form [recordType]_id. */
+  private String getIdName(RecordType recordType, List<Relation> relations) {
+    return (relations.isEmpty()
+            ? recordType.getName()
+            : relations.get(relations.size() - 1).relationRecordType().getName())
+        + "_id";
+  }
```
davidangb, line 402: ah, we probably need this for backwards compatibility, huh? We'd been trying to do away with magic naming like this. I wonder if it's actually needed … if we migrated existing Rawls-powered workspaces to WDS, those tables would end up with a physical column named `${recordType}_id`, so we wouldn't need special handling. See also `PrimaryKeyDao.getPrimaryKeyColumn()`.

service/src/main/java/org/databiosphere/workspacedataservice/expressions/ExpressionService.java
```diff
@@ -0,0 +1,530 @@
...
+
+  private JsonNode toJsonNode(Object attributeValue) {
+    if (attributeValue == null) {
+      return objectMapper.nullNode();
+    } else if (attributeValue instanceof JsonNode) {
+      return (JsonNode) attributeValue;
```
davidangb, line 408:
```suggestion
    } else if (attributeValue instanceof JsonNode jsonNode) {
      return jsonNode;
```

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
calypsomatic: We're moving away from the use of `instance` to `collection` - it's not reflected in these other APIs because they're the last to be updated since that requires releasing a new API version, but for new APIs we should avoid use of the term "instance".  Just renaming variables

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
```diff
@@ -299,6 +301,218 @@ public List<Record> queryForRecords(
...
+      String arrayRecordId,
+      List<Relation> arrayRelations,
+      List<Relation> relations,
+      int pageSize,
+      int offset) {
+    if (arrayRelations.isEmpty()) {
+      throw new IllegalArgumentException("Array relations must not be empty");
+    }
+
+    var rootRecordType = arrayRelations.get(arrayRelations.size() - 1).relationRecordType();
```
davidangb, line 332: see comment in ExpressionService about validating that all relations have the same root Type

davidangb (commented):
See comment details inline. Thank you so much for all the contribution. My main points for review are:
* move to v1 APIs, which use generated models and controller interfaces
* do we need both APIs or would one do?

For follow-on work after this PR:
* add Micrometer `Observation`s (which will inherit tracing when we tackle that)

calypsomatic:
> There's a lot going on here -- it might be worth a walkthrough to go over the new ANTLR-related classes, or more commentary about how those new classes work for those that aren't already familiar with it from the Rawls implementation, but I wouldn't insist on it for my sake alone. I'll leave that up to you and the other members of AJ who may be more familiar.

I would also love a walkthrough!

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -229,6 +229,74 @@ paths:
             'application/json':
               schema:
                 $ref: '#/components/schemas/ErrorResponse'
+  /{instanceid}/records/{v}/{type}/{id}/evaluateExpressions:
```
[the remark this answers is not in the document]
calypsomatic: V1

service/src/main/resources/static/swagger/openapi-docs.yaml
```diff
@@ -229,6 +229,74 @@ paths:
...
+                $ref: '#/components/schemas/ErrorResponse'
+        400:
+          description: Bad request
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+  /{instanceid}/records/{v}/{type}/{id}/evaluateExpressionsWithArray:
+    post:
+      summary: Evaluate expressions on array of records
```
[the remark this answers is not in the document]
dvoet: I think that the benefit is in clarity of the API. If we collapse them then `recordType` and `recordId` are handled differently based on if `arrayExpression` is present. Also `pageSize` and `offset` are unused if `arrayExpression` is absent.

The benefit of collapsing them for the WDS code is minimal because the code paths diverge early on.

service/src/main/java/org/databiosphere/workspacedataservice/expressions/ExpressionService.java
```diff
@@ -0,0 +1,530 @@
...
+      String version,
+      RecordType arrayRecordType,
+      String arrayRecordId,
+      String arrayRelationExpression,
+      Map<String, String> expressionsByName,
+      int pageSize,
+      int offset) {
+    validateVersion(version);
+    var arrayRelations = getArrayRelations(collectionId, arrayRecordType, arrayRelationExpression);
+    var queryRecordType = arrayRelations.get(arrayRelations.size() - 1).relationRecordType();
```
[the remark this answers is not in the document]
dvoet: No, `arrayRelations` is the list of relations to follow. For example, `arrayRelationExpression` could be `this.foo.bar.samples` and `arrayRelations` would contain a relation for `foo`, `bar` and `samples`. Then `queryRecordType` is `Sample`.

service/src/main/java/org/databiosphere/workspacedataservice/dao/RecordDao.java
dvoet: they will not all have the same type and here we care only about the type of the last one because we are only selecting columns for that type

service/src/main/java/org/databiosphere/workspacedataservice/expressions/ExpressionService.java
dvoet: I agree it does not matter in practice and perhaps caching will be a solution. However, I have a fix just because I wanted to figure out how to do it with streams

service/src/main/java/org/databiosphere/workspacedataservice/expressions/ExpressionService.java
dvoet: I am going to leave this in. I think the main thing to optimize for here is code burden and I think this is less code and not hard to follow. I don't think performance will be much of a factor either way (although I am sure the current code has the best performance).

service/src/main/java/org/databiosphere/workspacedataservice/controller/RecordController.java
```diff
@@ -43,11 +44,15 @@ public class RecordController {
...
+  private final ExpressionService expressionService;
 
   public RecordController(
-      RecordOrchestratorService recordOrchestratorService, PermissionService permissionService) {
+      RecordOrchestratorService recordOrchestratorService,
+      PermissionService permissionService,
+      ExpressionService expressionService) {
     this.recordOrchestratorService = recordOrchestratorService;
     this.permissionService = permissionService;
+    this.expressionService = expressionService;
```
davidangb, line 55: these changes are now obsolete - `RecordController` doesn't need `ExpressionService` any more

service/src/main/java/org/databiosphere/workspacedataservice/expressions/ExpressionService.java
davidangb: nice, I like this!

service/src/main/resources/static/swagger/apis-v1.yaml
```diff
@@ -279,7 +279,66 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/DeleteRecordsResponse'
-
+  /records/v1/{collectionId}/{recordType}/{recordId}/evaluateExpressions:
```
davidangb, line 282: to expose these APIs in the actual live swagger UIs, you'll need to add `$ref`s to the main YAML file, cwds-api-docs-preview.yaml. See [here](https://github.com/DataBiosphere/terra-workspace-data-service/blob/b67b001b50c357b68b7742d05a9afdfc918f43c3/service/src/main/resources/static/swagger/cwds-api-docs-preview.yaml#L58-L59) for an example.

Our OpenAPI yaml is currently pretty complicated. We have:
* apis-v1.yaml: source for autogeneration, containing new-style APIs, but never displayed to users directly
* openapi-docs.yaml: APIs for the single-tenant data plane, includes $refs to apis-v1.yaml
* cwds-api-docs.yaml: APIs for the multi-tenant control plane IN PROD, includes $refs to apis-v1.yaml
* cwds-api-docs-preview.yaml: APIs for the multi-tenant control plane IN DEV, includes $refs to apis-v1.yaml

cwds-api-docs.yaml and cwds-api-docs-preview.yaml both exist for a short time while we're in appsec review and have some APIs turned on in dev but not in prod. The -preview will go away soon after we get approvals.

openapi-docs.yaml would go away once we migrate out of the data plane and into the multi-tenant control plane

so, if you want these APIs to appear in the dev env, but not in prod until we get approved, cwds-api-docs-preview.yaml is the right place.

davidangb (approved):
thank you! Two hopefully not-big comments inline; I'm giving this a proactive :+1: assuming you can tackle those two changes

PR #932, opened 2024-09-10
AJ-2005 remove mention of datarepo url from readme

calypsomatic:
After https://github.com/DataBiosphere/terra-workspace-data-service/pull/774 we don't need the data repo url anymore; this PR removes its mention from the README and its use from `.yml` files.


davidangb (approved):
thank you!!!

mspector (approved):

PR #938, opened 2024-09-25
[AJ-2023] cWDS sends cors headers

mspector:
Reminder:

#### Releasing WDS ####


service/src/test/java/org/databiosphere/workspacedataservice/controller/CorsLiveMockMvcTest.java
```diff
@@ -1,57 +0,0 @@
...
-import java.util.UUID;
-import org.junit.jupiter.params.ParameterizedTest;
-import org.junit.jupiter.params.provider.ValueSource;
-import org.springframework.test.web.servlet.MvcResult;
-
-/**
- * This class tests CORS behavior for live deployments. WDS in live deployments sits behind Azure
- * Relay, and Relay handles CORS headers. Therefore, WDS should *not* reply with CORS headers, as it
- * will cause a conflict.
- *
```
mspector, line 16: I'm advocating for removing this test, since the underlying assumptions are changing for the V1 APIs.

service/src/main/resources/application-bee.yml
```diff
@@ -5,3 +5,5 @@ workspacemanagerurl: https://workspace.${beename}.bee.envs-terra.bio
 leoUrl: https://leonardo.${beename}.bee.envs-terra.bio
 rawlsUrl: https://rawls.${beename}.bee.envs-terra.bio
 
+# Enable preview of control plane apis
+controlPlanePreview: "on"
```
mspector, line 9: if controlPlanePreview is enabled in dev by default, it feels like it should be enabled in BEEs by default

service/src/test/java/org/databiosphere/workspacedataservice/controller/CorsLiveMockMvcTest.java
davidangb: If you keep this test, but add an annotation of:
```
@ActiveProfiles(
    value = {"data-plane"},
    inheritProfiles = false)
```
does it pass?

This annotation would change the purpose of the test to be: "when running in the data plane behind Azure Relay, ensure we don't send CORS headers"

service/src/main/java/org/databiosphere/workspacedataservice/controller/ControllerConfig.java
```diff
@@ -21,7 +21,7 @@ public class ControllerConfig {
    */
   @SuppressWarnings("java:S5122") // we explicitly want to allow * for origins
   @Bean
-  @Profile("local-cors")
+  @Profile({"local-cors", "control-plane"})
```
davidangb, line 24: I wonder if the `local-cors` profile has any use after this change. We might be able to do away with it entirely. I don't think we need to block the PR on that decision.

service/src/test/java/org/databiosphere/workspacedataservice/controller/CorsLiveMockMvcTest.java
mspector: You know what - I tried this initially, and couldn't make it work, but I'm realizing that it's because I hadn't set up my environment properly in IntelliJ 🤦 

this is a better solution - thanks

service/src/main/java/org/databiosphere/workspacedataservice/controller/ControllerConfig.java
mspector: isn't it still useful if we want to run WDS in data plane mode, locally?

service/src/test/java/org/databiosphere/workspacedataservice/controller/CorsLiveMockMvcTest.java
mspector: Actually... something about this isn't agreeing with our github actions environment. It may take some more fiddling to make this work as intended

mspector: solved 👍

calypsomatic: Just curious - what was the issue?

calypsomatic (approved):

service/src/main/java/org/databiosphere/workspacedataservice/controller/ControllerConfig.java
davidangb: yes, agreed - good call!

davidangb (approved):

service/src/test/java/org/databiosphere/workspacedataservice/controller/CorsLiveMockMvcTest.java
mspector: in my local environment, the test was using the workspace ID that I'd set with an environment variable. with the "data-plane" active profile, the test (somehow) injects that workspace ID as a bean from that environment variable, but it's unset in the github actions environment.

the solution is to mock the WorkspaceId bean (crucially, using the `@SingleTenant` annotation)

PR #941, opened 2024-10-02
AJ-2022: add Content-Security-Policy meta header for swagger-ui

davidangb:
Basing these changes on https://github.com/DataBiosphere/terra-landing-zone-service/pull/502

This PR adds a `Content-Security-Policy` to swagger-ui via `<meta http-equiv …` tags.

### DIFF
--- a/service/src/main/resources/swagger-ui-control-plane-preview.html
+++ b/service/src/main/resources/swagger-ui-control-plane-preview.html
@@ -4,6 +4,8 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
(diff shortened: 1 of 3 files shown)


kevinmarete (changes requested):
You [cannot set `frame-ancestors 'none';` on a meta tag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors)

All else looks good.

kevinmarete (approved):

calypsomatic (approved):

PR #956, opened 2024-12-05
CORE-69: Spring Boot 3.4, part 1

davidangb:
Upgrade WDS to Spring Boot 3.4. Upgrade documentation is [here](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.4-Release-Notes)

In this PR:
* upgrade to the 3.4 libraries
* address deprecated configs related to Actuator
* fix okhttp versions no longer being managed by Spring Boot
* switch to webjars-locator-lite as recommended by the upgrade doc
* fix compatibility issues

NOT in this PR:
* remove deprecated usages of `@MockBean` and `@SpyBean`. These annotations are deprecated but still work, and addressing them will cause a LOT of files to change, so I'd rather do it in a separate PR.


client/swagger.gradle
```diff
@@ -2,8 +2,8 @@ dependencies {
     //Need to include libraries for generated code to work
     api 'com.google.code.gson:gson'
     api 'io.gsonfire:gson-fire:1.9.0'
-    api 'com.squareup.okhttp3:okhttp'
-    api 'com.squareup.okhttp3:logging-interceptor'
+    api 'com.squareup.okhttp3:okhttp:4.12.0'
+    api 'com.squareup.okhttp3:logging-interceptor:4.12.0'
```
davidangb, line 6: okhttp is no longer managed by Spring Boot, so we need to add a version.

service/src/main/java/org/databiosphere/workspacedataservice/controller/SwaggerUiConfig.java
```diff
@@ -36,6 +37,7 @@ public String controlPlaneSwaggerHtml(
   }
 
   @DataPlane
+  @ConditionalOnMissingBean(name = "swaggerHtml") // for tests that use both dat & control plane
```
davidangb, line 40: we have one test - `AnnotatedApisTest` - which specifies both `control-plane` and `data-plane` profiles. That test started to fail on Spring Boot 3.4 on the "swaggerHtml" bean. Adding this annotation fixes it.

service/src/main/resources/application.yml
```diff
@@ -14,19 +14,21 @@ env:
...
         liveness:
           include: livenessState,db
     prometheus:
-      enabled: true
+      access: read_only
   endpoints:
-    enabled-by-default: false
+    access:
+      default: none
+      max-permitted: read_only
```
davidangb, line 31: Most of these changes recommended by the [release notes](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.4-Release-Notes#controlling-access-to-actuator-endpoints). I added the new `max-permitted` setting for enhanced security.

samanehsan (approved):

trholdridge (approved):

PR #958, opened 2024-12-06
CORE-69: remove deprecated MockBean/SpyBean

davidangb:
In #956, we upgraded to Spring Boot 3.4. 

Spring Boot 3.4 deprecated `@MockBean` and `@SpyBean`. This PR removes all usages of those annotations and replaces them with the newly-recommended `@MockitoBean` and `@MockitoSpyBean`.

This is almost completely a find/replace. There is one single test - `QuartzJobTest` - which needed a slight syntax change; I'll add a comment inline to call that out.

### DIFF
--- a/service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
+++ b/service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
@@ -33,7 +33,7 @@ import org.databiosphere.workspacedataservice.config.DataImportProperties;
 import org.databiosphere.workspacedataservice.dao.JobDao;
 import org.databiosphere.workspacedataservice.generated.GenericJobServerModel;
 import org.databiosphere.workspacedataservice.sam.TokenContextUtil;
-import org.junit.jupiter.api.BeforeAll;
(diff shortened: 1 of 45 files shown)


service/src/test/java/org/databiosphere/workspacedataservice/jobexec/QuartzJobTest.java
```diff
@@ -44,22 +44,22 @@
...
-  @MockBean DataImportProperties dataImportProperties;
+  @MockitoBean JobDao jobDao;
+  @MockitoBean DataImportProperties dataImportProperties;
   @Autowired MeterRegistry meterRegistry;
   @Autowired TestObservationRegistry observationRegistry;
 
-  @BeforeAll
-  void beforeAll() {
+  @BeforeEach
+  void beforeEach() {
```
davidangb, line 62: this change from `@BeforeAll` to `@BeforeEach` is the only extra change in this PR.

I note that elsewhere in this test, we run the `jobCompletion()` test, which changes the return value from the mock. This used to work ok in Spring Boot 3.3 but caused failures here … so now I'm simply resetting the mock's return value before each test.

dvoet (approved):

marctalbott (approved):

PR #982, opened 2025-02-07
CORE-302: remove backup/restore, collection initialization, and Leo/WDS DAOs

davidangb:
Remove a bunch of Azure-specific functionality:
* backup/restore
* cloning
* DAO for connecting to another WDS, which was only used by cloning
* DAO for connecting to Leo, which was only used by cloning
* at-startup collection initialization


service/src/main/java/org/databiosphere/workspacedataservice/retry/RestClientRetry.java
```diff
@@ -178,14 +178,6 @@ private int extractResponseCode(Exception e) {
...
       return tdrException.getCode();
     }
-    // Leo
-    if (e instanceof org.broadinstitute.dsde.workbench.client.leonardo.ApiException leoException) {
-      return leoException.getCode();
-    }
-    // WDS
-    if (e instanceof org.databiosphere.workspacedata.client.ApiException wdsException) {
-      return wdsException.getCode();
-    }
```
davidangb, line 188: since this PR removes the Leo and WDS clients, we no longer need to retry the exceptions from those clients.

( the WDS client was used internally during cloning; this PR doesn't affect our publishing of the WDS client for use by other services )

service/src/test/python/test.py
```diff
@@ -100,7 +100,9 @@ class WdsTests(TestCase):
     schema_client = wds_client.SchemaApi(api_client)
     import_client = wds_client.ImportApi(api_client)
     job_client = wds_client.JobApi(api_client)
+    collections_client = wds_client.CollectionApi(api_client)
     current_workspaceId = os.environ['WORKSPACE_ID']
+    current_collectionId = '' # will be overwritten by setup_method
```
davidangb, line 105: We needed changes in this Python test because this PR changes at-startup behavior. We no longer create the default collection at startup. This only happened in the data plane, but this test relied on that behavior. So I've added a `setup_class` method to execute before the tests; that method creates the collection.

calypsomatic (approved):

PR #989, opened 2025-02-11
CORE-302: remove WsmProtectedDataSupport data-plane-only bean

davidangb:
Deletes the `WsmProtectedDataSupport` bean, which was only used in the data plane.


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/protecteddatasupport/RawlsProtectedDataSupport.java
```diff
@@ -9,7 +8,6 @@
 import org.springframework.stereotype.Component;
 
 @Component
-@ControlPlane
 public class RawlsProtectedDataSupport implements ProtectedDataSupport {
   private final RawlsClient rawlsClient;
```
davidangb, line 12: RawlsProtectedDataSupport is now used everywhere, instead of RawlsProtectedDataSupport for control plane and WsmProtectedDataSupport for the data plane. Since we are moving the codebase to control-plane-only, this is ok

calypsomatic (approved):

marctalbott (approved):

PR #990, opened 2025-02-11
CORE-302: remove WdsDataTableTypeInspector

davidangb:
Removes the `WdsDataTableTypeInspector`, which was only used for the data plane. Includes fixes for unit tests and python tests.


service/src/test/python/test.py
```diff
@@ -120,13 +121,13 @@ class WdsTests(TestCase):
...
+    ### setup: initialize the workspace for use with data tables
+    ### This needs to happen first, before other tests that rely on the initialization
     @classmethod
     def setup_class(cls):
-      collection_request = wds_client.CollectionRequest(name='default', description='default')
-      collection_created = cls.collections_client.create_collection_v1(cls.current_workspaceId, collection_request)
-      cls.current_collectionId = collection_created.id
+      workspace_init_request = wds_client.WorkspaceInit()
+      init_response = cls.workspace_client.init_workspace_v1(cls.current_workspaceId, workspace_init_request)
+      cls.current_collectionId = cls.current_workspaceId
```
davidangb, line 130: calling the init-workspace API serves dual purpose:
* creates a collection
* makes the DataTableTypeInspector say that it's ok to use data tables in this workspace

So this call can replace the previous call to `create_collection_v1`

kevinmarete (approved):
Very nice!

kevinpalis (approved):

PR #1015, opened 2025-05-03
[CORE-480] Automate Releases on Every Commit to Main

kevinmarete:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-480

### PR Changes:
- Updates `build-and-test` to `build-test-tag-and-publish`
- Removes `tag-and-publish` and `release-drafter` workflows

---


calypsomatic (approved):
LGTM

davidangb (approved):
:+1: with a question: is there a point to keeping the release drafter functionality at all? Since we will only have one commit for each release, it feels kinda pointless. I'm fine leaving it as-is and seeing if we still find value in the release drafter.

kevinmarete:
> 👍 with a question: is there a point to keeping the release drafter functionality at all? Since we will only have one commit for each release, it feels kinda pointless. I'm fine leaving it as-is and seeing if we still find value in the release drafter.

Makes sense, I have removed it here: https://github.com/DataBiosphere/terra-workspace-data-service/pull/1015/commits/cde595ac1f0bc167e68c656f8ec2f34c4cd66331

calypsomatic (approved):

davidangb (commented):
Looking at this a bit more critically … I wonder if this will cause race conditions upon merging a PR to `main`. Both `tag-and-publish.yml` and `build-and-test.yml` are configured to run on push to main. Both perform version bumps and both call `publish-docker.yml`. Do you think this will be a problem? Perhaps it makes more sense to consolidate these two workflows into a single one to avoid any contention? Rawls handles this by checking if the current event is a push to `develop` vs. a PR by checking the ref name; see [here](https://github.com/broadinstitute/rawls/blob/8b0baabc17279feb54ef51b05975883485e3f8b5/.github/workflows/rawls-build-tag-publish-and-run-tests.yaml#L117).

kevinmarete:
> Looking at this a bit more critically … I wonder if this will cause race conditions upon merging a PR to `main`. Both `tag-and-publish.yml` and `build-and-test.yml` are configured to run on push to main. Both perform version bumps and both call `publish-docker.yml`. Do you think this will be a problem? Perhaps it makes more sense to consolidate these two workflows into a single one to avoid any contention? Rawls handles this by checking if the current event is a push to `develop` vs. a PR by checking the ref name; see [here](https://github.com/broadinstitute/rawls/blob/8b0baabc17279feb54ef51b05975883485e3f8b5/.github/workflows/rawls-build-tag-publish-and-run-tests.yaml#L117).

This has been updated: https://github.com/DataBiosphere/terra-workspace-data-service/pull/1015/commits/973827be37b26708ad6637bed3f3349a922af9af

davidangb (commented):
this is very close, one comment inline.

With this commit, we can probably do away with the `tag-and-publish.yml` workflow too - what do you think?

davidangb (approved):
Thank you! Could you keep an eye on the next handful of commits to make sure this is working as expected?

PR #1016, opened 2025-05-12
[CORE-488] use snapshot v3

calypsomatic:
https://broadworkbench.atlassian.net/browse/CORE-488

---
Updates cWDS to use Rawls' v3 snapshot API.  This API already checks for existing references to the snapshot in the workspace, so the methods and logic that did that in cWDS could be removed.  This left `RawlsSnapshotSupport` to be pretty dinky; given that we've already removed any other type of `SnapshotSupport`, I've simplified the architecture a bit by moving `RawlsSnapshotSupport`'s remaining code up into `SnapshotSupport` and making it no longer an abstract class.  I've also removed the `SnapshotSupportFactory` interface and simply use `MultiCloudSnapshotSupportFactory` directly; I think it makes sense to keep the factory since a new one is made with the needed workspace id for each import, but I'm open to arguments to get rid of it as well. 


service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/SnapshotSupport.java
```diff
@@ -85,76 +92,23 @@ protected UUID safeGetSnapshotId(ResourceDescription resourceDescription) {
...
-        .map(ResourceMetadata::getProperties)
-        .map(
-            properties ->
-                properties.stream()
-                    .anyMatch(
-                        property ->
-                            PROP_PURPOSE.equals(property.getKey())
-                                && PURPOSE_POLICY.equals(property.getValue())))
-        .orElse(false);
+    // TODO do we need to figure out how many succeeded vs failed?
```
calypsomatic, line 110: Now that Rawls handles finding previously linked snapshots and takes all snapshots in one batch, we don't have the details on how many were already present and how many failed.  I'm not sure what the response is if some snapshots are successfully linked and some fail, or if it's all or nothing.  Should the `SnapshotLinkResult` be updated to be a single number, or should it just be a boolean success/fail or what?

dvoet: I think success/fail

dvoet (approved):

service/src/main/java/org/databiosphere/workspacedataservice/dataimport/snapshotsupport/SnapshotSupport.java
marctalbott: It depends on where it fails in the process, but if one of the snapshots fails to link after Rawls has already started linking then we'll have partially linked the snapshots -- Rawls doesn't go back and unlink the already linked snapshots nor does it undo the auth domain changes. The auth domain will have been fully updated, it just won't reflect in TPS.

The best thing for the user to do would be to retry linking to finish linking the PAOs in TPS since Rawls can handle repeated requests.

calypsomatic: Hm, so even a failure doesn't mean 0 snapshots were linked?  That does mess up the metrics

marctalbott (approved):

PR #1031, opened 2025-07-15
[Core-593] Switch from jfrog to gar

calypsomatic:
https://broadworkbench.atlassian.net/browse/CORE-593

Switches from publishing the java client to jfrog to publishing it to GAR. 

---


client/swagger.gradle
```diff
@@ -23,6 +23,8 @@ openApiGenerate {
         disallowAdditionalPropertiesIfNotPresent: "false",
         useJakartaEe                            : "true",
         prependFormOrBodyParameters             : "true", // orders arguments in the Java client same as previous swagger-codegen
+        hideGenerationTimestamp                 : "true", // prevent unnecessary diff noise
+        serializableModel                       : "true", // make models serializable
```
davidangb, line 27: does `serializableModel=true` change the models in the client in a breaking way? If a consumer upgrades to the latest version of this client will they encounter problems?

calypsomatic: I'm not planning on keeping this; this is just one of the many things I've tried to trick it into publishing

calypsomatic: I'm not planning on keeping this; this is just one of the many things I've tried to trick it into publishing

PR #1058, opened 2025-09-18
[CORE-661] allow test fixture bucket on staging

calypsomatic:
---
https://broadworkbench.atlassian.net/browse/CORE-661 
https://github.com/DataBiosphere/terra-workspace-data-service/pull/1052 restricted which URLS can be imported, but failed to allow-list the buckets that are used for files in ui integration tests.  This PR adds that bucket to all environments


calypsomatic:
> I wonder if any of our external partners, or even internal folks, use staging to test PFB etc. integrations - and if this would cause those integrations to fail

Yeah, they would've started failing once the original PR went out.  This PR will make anything using that HCA bucket fail as well.  Hopefully they'll scream if that happens?  Should we put out an announcement?

kevinmarete (approved):

